PR #3, opened 2016-10-26
Update 2-state-of-the-art.md

pjz:
BGP is the core routing protocol of the Internet; it's actually p2p as well, with the only central authority being ARIN (for ASNs and IP blocks)


lgierth:
You're right, thanks!

PR #5, opened 2017-01-24
Change 'multiplex' multistream id to 'mplex'

whyrusleeping:
Its less confusing, there are other multiplexers with different names, shortening it to mplex both distinguishes itself as its own protocol, and saves bytes.

Plus, this is already in go-ipfs as /mplex


diasdavid:
We can change it.

diasdavid:
//cc @VictorBjelkholm

lgierth:
Thanks :+1: this was bothering me ;)

PR #6, opened 2017-02-08
Add and link circuit relay spec

lgierth:
Review of the spec itself concluded in libp2p/go-libp2p-relay#1 -- the additional changes in this PR are purely linking it into place.

I chose to categorize the relay primarily as a transport, since that is what it looks like to the layers using it -- you use it by means of `/p2p-circuit` multiaddrs, just like other transports. The fact that it's also a swarm protocol is an implementation detail from this point of view.

### DIFF
--- a/3-requirements.md
+++ b/3-requirements.md
@@ -13,6 +13,10 @@ It is recommended that implementations use one of the many NAT traversal librari
 
 Unfortunately, due to symmetric NATs, container and VM NATs, and other impossible-to-bypass NATs, `libp2p` MUST fallback to relaying communication to establish a full connectivity graph. To be complete, implementations MUST support relay, though it SHOULD be optional and able to be turned off by end users.
 
+Connection relaying SHOULD be implemented as a transport, in order to be transparent to upper layers.
+
+For an instantiation of relaying, see the [/ipfs/relay/circuit transport](transports/circuit-relay.md).
+
 ## 3.3 Encryption
 
 Communications on `libp2p` may be:
--- a/4-architecture.md
+++ b/4-architecture.md
@@ -71,6 +71,8 @@ Protocol multiplexing is done through [`multistream`](https://github.com/jbenet/
 
 ### 4.2.6 Relay
 
+See [/ipfs/relay/circuit transport](transports/circuit-relay.md).
+
 ## 4.3 Distributed Record Store
 
 ### 4.3.1 Record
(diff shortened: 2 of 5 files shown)


3-requirements.md
```diff
@@ -13,6 +13,10 @@ It is recommended that implementations use one of the many NAT traversal librari
 
 Unfortunately, due to symmetric NATs, container and VM NATs, and other impossible-to-bypass NATs, `libp2p` MUST fallback to relaying communication to establish a full connectivity graph. To be complete, implementations MUST support relay, though it SHOULD be optional and able to be turned off by end users.
 
+Connection relaying SHOULD be implemented as a transport, in order to be transparent to upper layers.
+
+For an instantiation of relaying, see the [/ipfs/relay/circuit transport](transports/circuit-relay.md).
```
diasdavid: `/ipfs/` or `/libp2p/`?

4-architecture.md
```diff
@@ -71,6 +71,8 @@ Protocol multiplexing is done through [`multistream`](https://github.com/jbenet/
 
 ### 4.2.6 Relay
 
+See [/ipfs/relay/circuit transport](transports/circuit-relay.md).
```
diasdavid: same as above?

7-properties.md
```diff
@@ -195,17 +195,18 @@ The wire protocol is -- of course -- wrapped with encryption. We use cyphersuite
...
+:---- | :---- | :----
+secio | /secio/1.0.0 |
+TLS | /tls/1.3.0 | not implemented
+plaintext | /plaintext/1.0.0 |
+spdy | /spdy/3.1.0 |
+yamux | /yamux/1.0.0 |
+multiplex | /mplex/6.7.0 |
+identify | /ipfs/id/1.0.0 |
+ping | /ipfs/ping/1.0.0 |
+relay | /ipfs/relay/line/0.1.0 | discontinued
```
diasdavid: just delete it, as it 'never actually started' :)

diasdavid:
Still a great fan of the dramatization :) I would say that we should take this opportunity to make sure that the protocol strings are very clear and well scoped (e.g relay is a libp2p thing and not an ipfs one)

diasdavid:
@lgierth I've updated the links and the multicodec in the multicodecs table. I understand that his spec is still a WIP, but let's make it a WIP with a placeholder in master, so that we have an entry point for it and then update the spec as the implemention develops. (Trying to avoid having 50 spec fragments everywhere)

lgierth:
Yes totally agree that we should merge this. While the spec is WIP and will be for a while longer, it's definitely good enough to be merged as a first draft.

PR #7, opened 2017-02-13
Feat/add pubsub

diasdavid:


diasdavid:
spec was written by @whyrusleeping on https://github.com/ipfs/specs/pull/148

PR #8, opened 2017-02-16
Clarify the difference between a tunnel and a relay

hsanjuan:
It took me a while to realize that a relay it's not just
tunneling a stream to a destination through a proxy peer
but rather it is making a destination peer
aware that the source peer is reachable through the relaying
node. Hopefully this makes that more explicit.

License: MIT
Signed-off-by: Hector Sanjuan <code@hector.link>

### DIFF
--- a/relay/README.md
+++ b/relay/README.md
@@ -26,18 +26,24 @@ libp2p already has modules for NAT ([go-libp2p-nat](https://github.com/libp2p/go
 but these don't always do the job, just because NAT traversal is complicated.
 That's why it's useful to have a simple relay protocol.
 
-One node asks a relay node to connect to another node on its behalf.
-The relay node shortcircuits its streams to the two nodes,
+Unlike a transparent **tunnel**, where a libp2p peer would just proxy a
+communication stream to a destination (the destination being unaware of the
+original source), a circuit-relay makes the destination aware of the original
+source and the circuit followed to establish communication between the two.
+This provides the destination side with full knowledge of the circuit which,
+if needed, could be rebuilt in the opposite direction.
+
+Apart from that, this relayed connection behaves just like a regular
+connection would, but over an existing swarm stream with another peer
+(instead of e.g. TCP.): One node asks a relay node to connect to another node
+on its behalf. The relay node shortcircuits its streams to the two nodes,
 and they are then connected through the relay.
-This relayed connection behaves just like a regular connection would,
-because it is in fact just that,
-but over an existing swarm stream with another peer, instead of e.g. TCP.
 
 Relayed connections are end-to-end encrypted just like regular connections.
 
 The circuit relay is both a tunneled transport and a mounted swarm protocol.
-The transport is the means of *establishing* and *accepting* connections,
-and the swarm protocol is the means to *relaying* connections.
+The transport is the means of ***establishing*** and ***accepting*** connections,
+and the swarm protocol is the means to ***relaying*** connections.
 
 ```
 +-------+    /ip4/.../tcp/.../ws/p2p/QmRelay     +---------+     /ip4/.../tcp/.../p2p/QmTwo        +-------+


relay/README.md
```diff
@@ -24,18 +24,24 @@ libp2p already has modules for NAT ([go-libp2p-nat](https://github.com/libp2p/go
 but these don't always do the job, just because NAT traversal is complicated.
 That's why it's useful to have a simple relay protocol.
 
-One node asks a relay node to connect to another node on its behalf.
-The relay node shortcircuits its streams to the two nodes,
+Unlike a transparent **tunnel**, where a libp2p peer would just proxy a communication
+stream to a destination (the destination being unaware of the original source),
+a circuit-relay makes the destination aware of the original source and
+the circuit followed to stablish communication between the two. This provides
```
lgierth: s/stablish/establish/

relay/README.md
```diff
@@ -24,18 +24,24 @@ libp2p already has modules for NAT ([go-libp2p-nat](https://github.com/libp2p/go
...
+the destination side with the capacity of re-contacting the original source
+by rebuilding the circuit in the oposite direction, using the same relay.
```
lgierth: s/oposite/opposite/

lgierth:
Awesome, thanks! Two tiny typos, LGTM :+1:

relay/README.md
```diff
@@ -24,18 +24,24 @@ libp2p already has modules for NAT ([go-libp2p-nat](https://github.com/libp2p/go
...
+
+Apart from that, this relayed connection behaves just like a regular
+connection would, but over an existing swarm stream with another peer
+(instead of e.g. TCP.): One node asks a relay node to connect to another node
+n its behalf. The relay node shortcircuits its streams to the two nodes,
```
dryajov: s/n/on/

relay/README.md
dryajov: @hsanjuan @lgierth can this part be clarified a little further, because the wording is explicit in saying that the connection is rebuilt using the same relay, which makes me ask why can't it be rebuilt using an alternative relay, in case the original one is unavailable. _I guess this is just wording, and in case the original is unavailable a new one will be selected, or not?_

dryajov: @hsanjuan @lgierth can this part be clarified a little further, because the wording is explicit in saying that the connection is rebuilt using the same relay, which makes me ask why can't it be rebuilt using an alternative relay, in case the original one is unavailable. _I guess this is just wording, and in case the original is unavailable a new one will be selected, or not?_

hsanjuan:
@lgierth Thanks, fixed the typos. I need to pay more attention.

@dryajov I have changed the wording, does it sound better to you?

relay/README.md
```diff
@@ -24,18 +24,25 @@ libp2p already has modules for NAT ([go-libp2p-nat](https://github.com/libp2p/go
...
 That's why it's useful to have a simple relay protocol.
 
-One node asks a relay node to connect to another node on its behalf.
-The relay node shortcircuits its streams to the two nodes,
+Unlike a transparent **tunnel**, where a libp2p peer would just proxy a
+communication stream to a destination (the destination being unaware of the
+original source), a circuit-relay makes the destination aware of the original
+source and the circuit followed to establish communication between the two.
+This provides the destination side with full knowledge of the circuit which,
+if needed, could be rebuilt in the opposite direction using
```
dryajov: I think the key point that confused me and something worth highlighting here is that, the relay connection can be rebuilt using the same `relay` node or by asking the swarm to route it over some other relay peer. So I would perhaps word it as `if needed, could be rebuilt either through the same relay node, or if unavailable by asking the swarm to route the connection over any other known relay`.

hsanjuan: But @dryajov, the spec does not really specify how to "ask the swarm" or do relay discovery. Therefore I find your wording more confusing (it opens more questions). I just want to highlight that in a tunnel the receiving node would not have information about the circuit, while in a relay like this it does. Perhaps I will just remove the sentence?

dryajov: @hsanjuan Fair enough. Maybe leaving the mention that the connection could be rebuilt and leaving out the details could make it more readable. Perhaps, something like this:

`This provides the destination side with full knowledge of the circuit which, if needed, could be rebuilt in the opposite direction.`

hsanjuan: @dryajov we have a deal! I have pushed it in the last form.

dryajov: Thanks @hsanjuan, that looks much better!

PR #13, opened 2017-03-21
Make heading levels match

malcolmredheron:
This heading was at a higher level than the others, making the later headings look like subheadings.


lgierth:
Thanks :+1:

Issue #12, opened 2017-03-01
Clarify/confirm `<src><dst><data>` multistream flow in relay


dryajov:
@lgierth @diasdavid @dignifiedquire  

The current spec describes the multistream interaction as `<src><dst><data>`, further defining it as:

>After getting a stream to the relay node from its libp2p swarm, the dialing transport writes the header to the stream. The relaying node reads the header, gets a stream to the destination node, then writes the header to the destination stream and shortcircuits the two streams.

In order to make sure that this is unambiguous an example might be required, such as:

* `Peer A` connects to relay (`Relay`) and writes the destination address as a header
    * The header is in the form of `/ipfs/relay/circuit/1.0.0\n/<valid multiaddr of dest node>`
* The `Relay` reads the header and the address of `Peer B` from the stream
     * `Relay` then connects to `Peer B` and writes a header in the form of `/ipfs/relay/circuit/1.0.0\n/<valid multiaddr of source node>`
     * `Peer B` stores the origin/source address to be able to later dial back (is this the case?)
* Once connected to both peers, the `Relay` short circuits the two streams

This is what I currently have implemented as well, but I still feel that I might be reading the spec incorrectly, so just making sure this is the intended flow. Adding an example like this to the spec might also help future implementors. Please feel free to correct were things are off.

lgierth:
Mh yes, thanks for catching this. The semantics of the header are pretty under-specified...

It should be exactly as you say above. There's an A-to-R header (dest multiaddr), and an R-to-B header (src multiaddr). (Note that the multicodec now starts in /libp2p though.)

```
/libp2p/relay/circuit/1.0.0\n
/p2p/QmPeerB\n
<data-stream>
```

R needs the dest multiaddr to know what to do. It MUST be at least a `/p2p` multiaddr, but *later* the relay SHOULD also accept other multiaddrs which would bypass R's peer routing. Being able to run R without the overhead of peer routing, as a thin stream-in-stream-out service, will be very useful in the future. (This is called "forced addresses" in the current spec.)

B doesn't *need* the src multiaddr, but it's pretty helpful in that it allows to establish the reverse connection.

lgierth:
I think it's good to go :+1: 

One more explanation addition:

> B doesn't need the src multiaddr, but it's pretty helpful in that it allows to establish the reverse connection.

[...]

PR #15, opened 2017-03-28
The Circuit Relay Specification: The first iteration

diasdavid:
In an attempt to converge all the relay spec notes and discussions, I'm listing them here, so that they can all be considered for the revamp of the relay spec. (Order represents the timeline of how theses appear, last should be considered as last 'discussed')

- first iteration of the spec - https://github.com/libp2p/specs/blob/master/relay/README.md
- question: wire protocol message format - https://github.com/libp2p/specs/issues/12
- question: relay vs tunnel - https://github.com/libp2p/specs/pull/8
- implementation notes - https://github.com/libp2p/js-libp2p-circuit/blob/master/implementation-notes.md
- Why's concerns about current implementation - https://github.com/libp2p/js-libp2p-circuit/issues/8
- Relay use cases - https://github.com/libp2p/js-libp2p-circuit/issues/6
- moar notes - https://github.com/libp2p/js-libp2p-circuit/issues/2
- moar discussion - https://github.com/libp2p/js-libp2p-circuit/issues/4
- errors table: https://gist.github.com/whyrusleeping/7665164d04e43171e6cb7b8aa6205645#file-errors-md
- wire protocol: https://gist.github.com/whyrusleeping/7665164d04e43171e6cb7b8aa6205645#file-relay-md

### DIFF
--- a/relay/README.md
+++ b/relay/README.md
@@ -68,47 +68,66 @@ Scene 2:
 
 ## Addressing
 
-`/p2p-circuit` multiaddrs don't carry any meaning of their own.
-They need to encapsulate a `/p2p` address, or
-be encapsulated in a `/p2p` address, or both.
+`/p2p-circuit` multiaddrs don't carry any meaning of their own. They need to encapsulate a `/p2p` address, or be encapsulated in a `/p2p` address, or both.
 
-As with all other multiaddrs, encapsulation of different protocols
-determines which metaphorical tubes to connect to each other.
+As with all other multiaddrs, encapsulation of different protocols determines which metaphorical tubes to connect to each other.
+
+A `/p2p-circuit` circuit address, is formated following:
+
+`/p2p-circuit[<relay peer multiaddr>]<destination peer multiaddr>`
+
+This opens the room for multiple hop relay, where the first relay is encapsulated in the seconf relay multiaddr, such as:
+
+`/p2p-circuit/p2p-circuit/<first relay>/<first hop multiaddr>/<destination peer multiaddr>`
 
 A few examples:
 
+Using any relay available:
+
 - `/p2p-circuit/p2p/QmTwo`
-  - Dial QmTwo, through any available relay node.
-  - The relay node will use peer routing to find an address for QmTwo.
-- `/p2p/QmRelay/p2p-circuit/p2p/QmTwo`
-  - Dial QmTwo, through QmRelay.
-  - Use peer routing to find an address for QmRelay.
-  - The relay node will also use peer routing, to find an address for QmTwo.
+  - Dial QmTwo, through any available relay node (or find one node that can relay).
+  - The relay node will use peer routing to find an address for QmTwo if it doesn't have a direct connection.
 - `/p2p-circuit/ip4/../tcp/../p2p/QmTwo`
   - Dial QmTwo, through any available relay node,
     but force the relay node to use the encapsulated `/ip4` multiaddr for connecting to QmTwo.
   - We'll probably not support forced addresses for now, just because it's complicated.
+
+Specify a relay:
+
+- `/p2p-circuit/p2p/QmRelay/p2p/QmTwo`
+  - Dial QmTwo, through QmRelay.
+  - Use peer routing to find an address for QmRelay.
+  - The relay node will also use peer routing, to find an address for QmTwo.
+- `/p2p-circuit/ip4/../tcp/../p2p/QmRelay/p2p/QmTwo`
+  - Dial QmTwo, through QmRelay.
+  - Includes info for connecting to QmRelay.
+  - The relay node will use peer routing to find an address for QmTwo.
+
+Double relay:
+
+- `/p2p-circuit/p2p/QmTwo/p2p-circuit/p2p/QmThree`
+  - Dial QmThree, through a relayed connection to QmTwo.
+  - The relay nodes will use peer routing to find an address for QmTwo and QmThree.
+  - We'll probably not support nested relayed connections for now, there are edge cases to think of.
+
+--
+
+?? I don't understand the usage of the following:
+
 - `/ip4/../tcp/../p2p/QmRelay/p2p-circuit`
   - Listen for connections relayed through QmRelay.
   - Includes info for connecting to QmRelay.
   - Also makes QmRelay available for relayed dialing, based on how listeners currently relate to dialers.
 - `/p2p/QmRelay/p2p-circuit`
   - Same as previous example, but use peer routing to find an address for QmRelay.
-- `/p2p-circuit/p2p/QmTwo/p2p-circuit/p2p/QmThree`
-  - Dial QmThree, through a relayed connection to QmTwo.
-  - The relay nodes will use peer routing to find an address for QmTwo and QmThree.
-  - We'll probably not support nested relayed connections for now, there are edge cases to think of.
-- `/ip4/../tcp/../p2p/QmRelay/p2p-circuit/p2p/QmTwo`
-  - Dial QmTwo, through QmRelay.
-  - Includes info for connecting to QmRelay.
-  - The relay node will use peer routing to find an address for QmTwo.
+
+?? I believe we don't need this one:
 - `/p2p-circuit`
   - Use relay discovery to find a suitable relay node. (Neither specified nor implemented.)
   - Listen for relayed connections.
   - Dial through the discovered relay node for any `/p2p-circuit` multiaddr.
 
-TODO: figure out forced addresses.
-TODO: figure out nested relayed connections.
+TODO: figure out forced addresses. -> what is forced addresses?
 
 ## Wire format
 
--- a/relay/README.md
+++ b/relay/README.md
@@ -12,32 +12,18 @@ Table of Contents
 - [Addressing](#addressing)
 - [Wire protocol](#wire-protocol)
 - [Interfaces](#interfaces)
-- [Removing existing relay protocol](#removing-existing-relay-protocol)
+- [Implementation Details](#implementation-details)
 
 
 ## Overview
 
-The circuit relay is a means of establishing connectivity between
-libp2p nodes (such as IPFS) that wouldn't otherwise be able to connect to each other.
+The circuit relay is a means of establishing connectivity between libp2p nodes (such as IPFS) that wouldn't otherwise be able to connect to each other.
 
-This helps in situations where nodes are behind NAT or reverse proxies,
-or simply don't support the same transports (e.g. go-ipfs vs. browser-ipfs).
-libp2p already has modules for NAT ([go-libp2p-nat](https://github.com/libp2p/go-libp2p-nat)),
-but these don't always do the job, just because NAT traversal is complicated.
-That's why it's useful to have a simple relay protocol.
+This helps in situations where nodes are behind NAT or reverse proxies, or simply don't support the same transports (e.g. go-ipfs vs. browser-ipfs). libp2p already has modules for NAT ([go-libp2p-nat](https://github.com/libp2p/go-libp2p-nat)), but these don't always do the job, just because NAT traversal is complicated. That's why it's useful to have a simple relay protocol.
 
-Unlike a transparent **tunnel**, where a libp2p peer would just proxy a
-communication stream to a destination (the destination being unaware of the
-original source), a circuit-relay makes the destination aware of the original
-source and the circuit followed to establish communication between the two.
-This provides the destination side with full knowledge of the circuit which,
-if needed, could be rebuilt in the opposite direction.
+Unlike a transparent **tunnel**, where a libp2p peer would just proxy a communication stream to a destination (the destination being unaware of the original source), a circuit-relay makes the destination aware of the original source and the circuit followed to establish communication between the two. This provides the destination side with full knowledge of the circuit which, if needed, could be rebuilt in the opposite direction.
 
-Apart from that, this relayed connection behaves just like a regular
-connection would, but over an existing swarm stream with another peer
-(instead of e.g. TCP.): One node asks a relay node to connect to another node
-on its behalf. The relay node shortcircuits its streams to the two nodes,
-and they are then connected through the relay.
+Apart from that, this relayed connection behaves just like a regular connection would, but over an existing swarm stream with another peer (instead of e.g. TCP.): One node asks a relay node to connect to another node on its behalf. The relay node shortcircuits its streams to the two nodes, and they are then connected through the relay.
 
 Relayed connections are end-to-end encrypted just like regular connections.
 
@@ -50,18 +36,13 @@ and the swarm protocol is the means to ***relaying*** connections.
 | QmOne | <------------------------------------> | QmRelay | <-----------------------------------> | QmTwo |
 +-------+   (/libp2p/relay/circuit multistream)  +---------+  (/libp2p/relay/circuit multistream)  +-------+
       ^                                            +-----+                                           ^
-      |                                            |     |                                           |
       |            /p2p-circuit/QmTwo              |     |                                           |
       +--------------------------------------------+     +-------------------------------------------+
 ```
 
-Note: we're using the `/p2p` multiaddr protocol instead of `/ipfs` in this document.
-`/ipfs` is currently the canonical way of addressing a libp2p or IPFS node,
-but given the growing non-IPFS usage of libp2p, we'll migrate to using `/p2p`.
+Note: we're using the `/p2p` multiaddr protocol instead of `/ipfs` in this document. `/ipfs` is currently the canonical way of addressing a libp2p or IPFS node, but given the growing non-IPFS usage of libp2p, we'll migrate to using `/p2p`.
 
-Note: at the moment we're not including a mechanism for discovering relay nodes.
-For the time being, they should be configured statically.
-
+Note: at the moment we're not including a mechanism for discovering relay nodes. For the time being, they should be configured statically.
 
 ## Dramatization
 
@@ -131,10 +112,9 @@ TODO: figure out nested relayed connections.
 
 ## Wire format
 
-The wire format (or codec) is named `/ipfs/relay/circuit` and is simple.
-A variable-length header consisting of two length-prefixed multiaddrs
-is followed by a bidirectional stream of arbitrary data,
-and the eventual closing of the stream.
+The multicodec for the circuit relay protocol is: `/libp2p/relay/circuit`.
+
+A variable-length header consisting of two length-prefixed multiaddrs is followed by a bidirectional stream of arbitrary data, and the eventual closing of the stream.
 
 ```
 <src><dst><data>
@@ -149,13 +129,11 @@ and the eventual closing of the stream.
  +------------ multiaddr of the dialing node
 ```
 
-After getting a stream to the relay node from its libp2p swarm,
-the dialing transport writes the header to the stream.
-The relaying node reads the header, gets a stream to the destination node,
-then writes the header to the destination stream and shortcircuits the two streams.
+After getting a stream to the relay node from its libp2p swarm, the dialing transport writes the header to the stream.
 
-Each relayed connection corresponds to two multistreams,
-one between QmOne and QmRelay, the other between QmRelay and QmTwo.
+The relaying node reads the header, gets a stream to the destination node, then writes the header to the destination stream and shortcircuits the two streams.
+
+Each relayed connection corresponds to two multistreams, one between QmOne and QmRelay, the other between QmRelay and QmTwo.
 
 Implementation details:
 - The relay node has the `Swarm.EnableRelaying` config option enabled
@@ -166,9 +144,7 @@ Implementation details:
 
 ## Interfaces
 
-As explained above, the relay is both a transport (`tpt.Transport`)
-and a mounted stream protocol (`p2pnet.StreamHandler`).
-In addition it provides a means of specifying relay nodes to listen/dial through.
+As explained above, the relay is both a transport (`tpt.Transport`) and a mounted stream protocol (`p2pnet.StreamHandler`). In addition it provides a means of specifying relay nodes to listen/dial through.
 
 TODO: the usage of p2pnet.StreamHandler is a little bit off, but it gets the point across.
 
@@ -193,7 +169,9 @@ fund NewCircuitRelay(h p2phost.Host)
 ```
 
 
-### Removing existing relay protocol
+## Implementation details
+
+### Removing existing relay protocol in Go
 
 Note that there is an existing swarm protocol colloqiually called relay.
 It lives in the go-libp2p package and is named `/ipfs/relay/line/0.1.0`.
@@ -208,5 +186,4 @@ It lives in the go-libp2p package and is named `/ipfs/relay/line/0.1.0`.
 - Capable of *accepting* connections, and *relaying* connections.
 - Not capable of *connecting* via relaying.
 
-Since the existing protocol is incomplete, insecure, and certainly not used,
-we can safely remove it.
+Since the existing protocol is incomplete, insecure, and certainly not used, we can safely remove it.
--- a/relay/README.md
+++ b/relay/README.md
@@ -1,6 +1,6 @@
 # Circuit Relay
 
-> Circuit Switching in libp2p
+> Circuit Switching for libp2p, also known as TURN or Relay in Networking literature.
 
 ## Implementations
 
@@ -18,9 +18,9 @@
 
 ## Overview
 
-The circuit relay is a means of establishing connectivity between libp2p nodes (such as IPFS) that wouldn't otherwise be able to connect to each other.
+The circuit relay is a means of establishing connectivity between libp2p nodes (such as IPFS) that wouldn't otherwise be able to establish a direct connection to each other.
 
-This helps in situations where nodes are behind NAT or reverse proxies, or simply don't support the same transports (e.g. go-ipfs vs. browser-ipfs). libp2p already has modules for NAT ([go-libp2p-nat](https://github.com/libp2p/go-libp2p-nat)), but these don't always do the job, just because NAT traversal is complicated. That's why it's useful to have a simple relay protocol.
+This helps in situations where nodes are behind NAT or reverse proxies, or simply don't support the same transports (e.g. go-ipfs vs. browser-ipfs). libp2p already has modules for NAT ([go-libp2p-nat](https://github.com/libp2p/go-libp2p-nat)), however piercing through NATs is not always a option due to their implementation differences. The circuit relay protocol exist to overcome those scenarios.
 
 Unlike a transparent **tunnel**, where a libp2p peer would just proxy a communication stream to a destination (the destination being unaware of the original source), a circuit-relay makes the destination aware of the original source and the circuit followed to establish communication between the two. This provides the destination side with full knowledge of the circuit which, if needed, could be rebuilt in the opposite direction.
 
@@ -40,9 +40,9 @@ The transport is the means of ***establishing*** and ***accepting*** connections
      +-----------------------------------------+     +-----------------------------------------+
 ```
 
-Note: we're using the `/p2p` multiaddr protocol instead of `/ipfs` in this document. `/ipfs` is currently the canonical way of addressing a libp2p or IPFS node, but given the growing non-IPFS usage of libp2p, we'll migrate to using `/p2p`.
+## Notes for the reader
 
-Note: at the moment we're not including a mechanism for discovering relay nodes. For the time being, they should be configured statically.
+1. We're using the `/p2p` multiaddr protocol instead of `/ipfs` in this document. `/ipfs` is currently the canonical way of addressing a libp2p or IPFS node, but given the growing non-IPFS usage of libp2p, we'll migrate to using `/p2p`.
 
 ## Dramatization
 
@@ -73,11 +73,16 @@ As with all other multiaddrs, encapsulation of different protocols determines wh
 
 A `/p2p-circuit` circuit address, is formated following:
 
-`/p2p-circuit[<relay peer multiaddr>]<destination peer multiaddr>`
+`[<relay peer multiaddr>]/p2p-circuit/<destination peer multiaddr>`
 
-This opens the room for multiple hop relay, where the first relay is encapsulated in the seconf relay multiaddr, such as:
+Examples:
 
-`/p2p-circuit/p2p-circuit/<first relay>/<first hop multiaddr>/<destination peer multiaddr>`
+- `/p2p-circuit/p2p/QmVT6GYwjeeAF5TR485Yc58S3xRF5EFsZ5YAF4VcP3URHt` - No known relay
+- `/ip4/127.0.0.1/tcp/5002/ipfs/QmdPU7PfRyKehdrP5A3WqmjyD6bhVpU1mLGKppa2FjGDjZ/p2p-circuit/p2p/QmVT6GYwjeeAF5TR485Yc58S3xRF5EFsZ5YAF4VcP3URHt` - Known relay
+
+This opens the room for multiple hop relay, where the first relay is encapsulated in the second relay multiaddr, such as:
+
+`<1st relay>/p2p-circuit/<2nd relay>/p2p-circuit/<dst multiaddr>`
 
 A few examples:
 
@@ -87,17 +92,15 @@ Using any relay available:
   - Dial QmTwo, through any available relay node (or find one node that can relay).
   - The relay node will use peer routing to find an address for QmTwo if it doesn't have a direct connection.
 - `/p2p-circuit/ip4/../tcp/../p2p/QmTwo`
-  - Dial QmTwo, through any available relay node,
-    but force the relay node to use the encapsulated `/ip4` multiaddr for connecting to QmTwo.
-  - We'll probably not support forced addresses for now, just because it's complicated.
+  - Dial QmTwo, through any available relay node, but force the relay node to use the encapsulated `/ip4` multiaddr for connecting to QmTwo.
 
 Specify a relay:
 
-- `/p2p-circuit/p2p/QmRelay/p2p/QmTwo`
+- `/p2p/QmRelay/p2p-circuit/p2p/QmTwo`
   - Dial QmTwo, through QmRelay.
   - Use peer routing to find an address for QmRelay.
   - The relay node will also use peer routing, to find an address for QmTwo.
-- `/p2p-circuit/ip4/../tcp/../p2p/QmRelay/p2p/QmTwo`
+- `/ip4/../tcp/../p2p/QmRelay/p2p-circuit/p2p/QmTwo`
   - Dial QmTwo, through QmRelay.
   - Includes info for connecting to QmRelay.
   - The relay node will use peer routing to find an address for QmTwo.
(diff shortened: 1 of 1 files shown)


relay/README.md
```diff
@@ -87,54 +68,72 @@ Scene 2:
...
+
+?? I believe we don't need this one:
 - `/p2p-circuit`
   - Use relay discovery to find a suitable relay node. (Neither specified nor implemented.)
   - Listen for relayed connections.
   - Dial through the discovered relay node for any `/p2p-circuit` multiaddr.
 
-TODO: figure out forced addresses.
-TODO: figure out nested relayed connections.
+TODO: figure out forced addresses. -> what is forced addresses?
```
diasdavid: @lgierth Can you elaborate on what "forced addresses" are?

relay/README.md
```diff
@@ -87,54 +68,72 @@ Scene 2:
...
-  - The relay node will use peer routing to find an address for QmTwo.
+  - Also makes QmRelay available for relayed dialing, based on how listeners currently relate to dialers.
+- `/p2p/QmRelay/p2p-circuit`
+  - Same as previous example, but use peer routing to find an address for QmRelay.
+
+?? I believe we don't need this one:
 - `/p2p-circuit`
   - Use relay discovery to find a suitable relay node. (Neither specified nor implemented.)
   - Listen for relayed connections.
   - Dial through the discovered relay node for any `/p2p-circuit` multiaddr.
```
diasdavid: To me, the act of 'enabling relayed connections' is a config value and not a multiaddr. Do you agree @lgierth ?

relay/README.md
```diff
@@ -87,54 +68,72 @@ Scene 2:
...
+
+?? I don't understand the usage of the following:
+
+- `/ip4/../tcp/../p2p/QmRelay/p2p-circuit`
+  - Listen for connections relayed through QmRelay.
   - Includes info for connecting to QmRelay.
-  - The relay node will use peer routing to find an address for QmTwo.
+  - Also makes QmRelay available for relayed dialing, based on how listeners currently relate to dialers.
+- `/p2p/QmRelay/p2p-circuit`
+  - Same as previous example, but use peer routing to find an address for QmRelay.
```
diasdavid: Just to confirm, these addresses exist to have a way to make the node always dial to the relay peer (punch out), correct?

diasdavid:
From now on, all comments, reviews, edits and so on, should be applied directly to the document so that we keep everything in place :)

relay/README.md
```diff
@@ -149,26 +148,42 @@ and the eventual closing of the stream.
...
+| ----- |:-----------------:|:----------:|
+| 100   | OK                      | Relay was setup correctly    |
+| 220   | "src address too long"  | |
+| 221   | "dst address too long"  | |
+| 250   | "failed to parse src addr: no such protocol ipfs" | The `<src>` multiaddr in the header was invalid |
+| 251   | "failed to parse dst addr: no such protocol ipfs" | The `<dst>` multiaddr in the header was invalid |
+| 260   | "passive relay has no connection to dst" | |
+| 261   | "active relay could not connect to dst: connection refused" | relay could not form new connection to target peer |
+| 262   | "could not open new stream to dst: BAD ERROR" | relay has connection to dst, but failed to open a new stream |
+| 270   | "<dst> does not support relay" | |
```
diasdavid: Do we really want to use 2XX+3XX for errors? Can we follow the pattern laid by HTTP:

```
- 1×× Informational
- 2×× Success
- 3×× Redirection
- 4×× Client Error
- 5×× Server Error
```

relay/README.md
```diff
@@ -149,26 +148,42 @@ and the eventual closing of the stream.
...
+| 320   | "src address too long" | |
```
diasdavid: From IRC:

```
14:13 <@daviddias> I don't understand why the 'send the src' addr keeps appearing as a thing for relay
16:05 <+dryajov> diasdavid re srs addr: I dont think we need it anymore, you are right. We can always do getPeerInfo on the circuited conn and should be able to get the src addrs. It would only make sense, if the full relay addr is sent so that the relay can be rebuilt from the dst to the src, if that ever makes sense?
16:23 <@daviddias> " It would only make sense, if the full relay addr is sent so that the relay can be rebuilt from the dst to the src, if that ever makes sense?" even in that case, we can always verify who dialed us with the getPeerInfo call
16:26 <+dryajov> True, that would be a swarm addr and should come back with the getPeerInfo. We don't need it.
```

relay/README.md
dryajov: 👍HTTP codes are well known, which makes them more intuitive and expected.

relay/README.md
```diff
@@ -84,61 +65,79 @@ Scene 2:
...
 
-As with all other multiaddrs, encapsulation of different protocols
-determines which metaphorical tubes to connect to each other.
+A `/p2p-circuit` circuit address, is formated following:
+
+`/p2p-circuit[<relay peer multiaddr>]<destination peer multiaddr>`
+
+This opens the room for multiple hop relay, where the first relay is encapsulated in the seconf relay multiaddr, such as:
+
+`/p2p-circuit/p2p-circuit/<first relay>/<first hop multiaddr>/<destination peer multiaddr>`
```
dryajov: To me it makes a lot more sense to do something like - `/p2p-circuit<first relay>/<first hop multiaddr>/p2p-circuit<second relay>/<second hop multiaddr>/<destination peer multiaddr>`

relay/README.md
```diff
@@ -84,91 +65,189 @@ Scene 2:
...
-then writes the header to the destination stream and shortcircuits the two streams.
+It then waits for a response in the form of:
+```
+<uvarint error code><uvarint msglength><message>
+```
+
+Once it receives that, it checks if the status code is `OK`. If it is, it passes the new connection to its `NewConnHandler`.
+Otherwise, it returns the error message to the caller.
+
+### 'hop' protocol handler
```
diasdavid: By `hop`, you mean `/libp2p/circuit/relay` ?

relay/README.md
```diff
@@ -84,91 +65,189 @@ Scene 2:
...
+
+B now verifies that `<dst>` matches its peer ID. It then also checks that `<src>` is valid. It uses src as the
+'remote addr' of the new 'incoming relay connection' it will create.
+
+Peer B now writes back a message of the form:
+```
+<uvarint 'OK'><uvarint len(msg)><string "OK">
+```
+
+And passes the relayed connection into its `NewConnHandler`.
```
diasdavid: I'm confused by this `stop` handler. What case is it enabling exactly?

dryajov: Thats the end point in the relay flow, `hop` the relay listener, `stop` listens for relayed connections coming from `hop`.

diasdavid: Understood.

diasdavid:
One thing that the current spec is missing is versioning. Right now we have:

- `/libp2p/circuit/relay/hop`
- `/libp2p/circuit/relay/stop`

We the possibility of having a `/libp2p/circuit/relay/ls`, still on discussion.

If we were to add versioning, it makes it a tad complicated because it would create a version for each of these protocols (and we already know that assuming that something will never change is an antipattern™), ending up on:

- `/libp2p/circuit/relay/hop/1.0.0`
- `/libp2p/circuit/relay/stop/1.0.0`

This can be ok (and maybe a question of preference), but I would rather have a single `/libp2p/circuit/relay/1.0.0` to maintain a single versioning over the circuit relay protocol, having the selection of {`hop`, `stop`, `ls`} as part of the RPC. 

```
// cbor encoded json
{
  type: `hop`
  dstAddr: <some multiaddr>
}
```

It is not that we are trying to avoid RPC messages, we are already sending the multiaddr after first dial and we have the status codes going on.

relay/README.md
```diff
@@ -84,91 +65,189 @@ Scene 2:
...
-`/p2p-circuit` multiaddrs don't carry any meaning of their own.
-They need to encapsulate a `/p2p` address, or
-be encapsulated in a `/p2p` address, or both.
+`/p2p-circuit` multiaddrs don't carry any meaning of their own. They need to encapsulate a `/p2p` address, or be encapsulated in a `/p2p` address, or both.
+
+As with all other multiaddrs, encapsulation of different protocols determines which metaphorical tubes to connect to each other.
+
+A `/p2p-circuit` circuit address, is formated following:
+
+`/p2p-circuit[<relay peer multiaddr>]<destination peer multiaddr>`
```
whyrusleeping: maybe put slashes in here?

relay/README.md
whyrusleeping: No, i mean `/libp2p/circuit/relay/1.0.0/hop`

relay/README.md
whyrusleeping: It doesnt map exactly though, i'm fine moving the 100 OK to 200 OK. 

I guess we could map all the 200 errors to 400 'client' errors, and map all the 300 errors to 500 'server' errors

dryajov:
@diasdavid re versioning:

<•lgierth>  /libp2p/relay/circuit/1.0.0/{hop,stop}
15:37 W<•whyrusleeping> because theyre sub-endpoints of the same protocol
15:38 L<•lgierth> yeah and they're different semantics on the same wire format
15:38 D<•dryajov> ah, sure… I had the impression that the version would always come at the end
15:39 that makes sense tho
15:39 L<•lgierth> well i guess until now it always has
15:39 "it's always been like this" is usually a good reason to do it differently eventually :D
15:39 D<•dryajov> problem is that now its non deterministic… if its always at the end, you can easily extract it, if its not, you need to define syntax around it
15:40 L<•lgierth> no you just go through all the mounted protocols you have, and pick the one which supports /libp2p/relay/circuit/$versionRange
15:40 W— •whyrusleeping checks his parsing code
15:40 L<•lgierth> hah yeah better check
15:40 W— •whyrusleeping is sure that this will break his parsing code
15:40 W<•whyrusleeping> but no worries
15:41 we have to update that package anyways
15:41 can fix it along the way
15:41 or just use explicit protocol matching for now
15:41 D<•dryajov> yeah… but… how do you know whats a version and whats not :D?
15:41 L<•lgierth> your mount protocols know
15:41 W<•whyrusleeping> Yeah, that does get hard
15:41 L<•lgierth> transports work similarly, just without a version
15:41 D<•dryajov> ah… I see what you mean
15:42 L<•lgierth> each transport has a Matches() func
15:42 D<•dryajov> yeah, you don’t really need to check anything, just do an exact match
15:42 :D

whyrusleeping:
what should the 'remote multiaddr' look like on a relay connection?

My proposal is:
```
<src maddr, from header>/p2p-circuit/QmRelayPeerID
```

diasdavid:
@whyrusleeping can you enlighten me on the `src` being something that is transferred here? See https://github.com/libp2p/specs/pull/15#discussion_r108462775

whyrusleeping:
Irc logs for context: http://lets.git.sexy/relaynotes

dryajov:
@whyrusleeping  can you clarify what the 

- src maddr
- from header

in this - `<src maddr, from header>`

whyrusleeping:
@dryajov I mean the source multiaddr that is sent by 'A' in their initial relay request

whyrusleeping:
Another question, is it `/p2p-relay/PEERID` or `/p2p-relay/<multiaddr>` ?

dryajov:
@whyrusleeping - both. In the first case it's peer routing, the second is over an explicit transport.

diasdavid:
```
2017-03-27 17:48:29	+dryajov	hsanhuan brought some valid points there, basically, the idea is that you can rebuild the circuit from dst to src
2017-03-27 17:48:47	+dryajov	thats why src address is written to dst,
2017-03-27 17:48:54	+dryajov	otherwise it would work
2017-03-27 17:49:27	@whyrusleeping	Then we should go back to lgierth's original spec with both a src and a dst
```

I see you want to the full initial multiaddr, including any IPs, ports and so on. I question the usefulness of this, as you would only really need the peer Id (which you get from secio) and the relay addr (which you know, because you also have a conn established with it). But I guess it doesn't hurt.

dryajov:
@whyrusleeping or do you mean, 

- `/p2p-circuit/QmPeer` (not a valid mr as it is right now?) 
- `/p2p-circuit/ipfs/QmPeer` - peer routing ma
- `/p2p-circuit/ip4/127.0.0.1/tcp/80/ipfs/QmPeer`

?

whyrusleeping:
> as you would only really need the peer Id (which you get from secio)

No, we reallllllly have to stop assuming "we will just get it from secio". Thats an assumption that hurts us

I think its important to know where the connection is coming from. Though I think with the two separate protocols we could now get away with a single address, with different meaning per protocol (hop vs stop)

whyrusleeping:
@dryajov 

> - `/p2p-circuit/QmPeer` (not a valid mr as it is right now?) 
> - `/p2p-circuit/ipfs/QmPeer` - peer routing ma
> - `/p2p-circuit/ip4/127.0.0.1/tcp/80/ipfs/QmPeer`

The issue here is that `<anything>/ipfs/QmPeerID` implies that this multiaddress is connecting to that peer. I think doing `<anything>/p2p-circuit/QmPeerID` makes the most sense, as its saying we are relaying through peer `QmPeerID`.

`/p2p-circuit/ip4/127.0.0.1/tcp/80/ipfs/QmPeer` is very confusing. Consider the following:
``/p2p-circuit/ip4/127.0.0.1/tcp/80/ipfs/QmPeer/p2p-circuit/ip4/127.0.0.1/tcp/123/ipfs/QmOtherPeer`

Am I relaying through `QmPeer` to `QmOtherPeer`? Or is this a relay whose exit point is `QmOtherPeer`?

In a diagram so i can more clearly explain myself:
```
A -> [QmPeer] -> [QmOtherPeer] -> ?
or
A -> [QmPeer] -> QmOtherPeer
or
A -> [QmPeer -> [QmOtherPeer]] -> ?
```

relay/README.md
```diff
@@ -84,91 +65,189 @@ Scene 2:
...
+
+This opens the room for multiple hop relay, where the first relay is encapsulated in the seconf relay multiaddr, such as:
```
whyrusleeping: typo: s/seconf/second/

whyrusleeping:
The main distinction i'm worried about is this:

Multihop relay, A wants to connect to B, A only has a connection to X and B only has a connection to Y.
X and Y are connected.

How do i distinguish between Relay(Relay(AX, XY), YB) and Relay(AX, Relay(XY, YB))

dryajov:
@whyrusleeping 

IMO a peer id without /ipfs looses context and overloads /p2p-circuit even further, which is already pretty bloated in terms of contextual meaning 😄 

Here is what make sense to me:

-` /p2p-circuit/ < optional explicit transport > /ipfs/QmRelay` means this is a relay addr
   - When used as a swarm addr, you get connections relayed over this address
   - if a peer is queried for its addresses (.getPeerInfo) you get `/p2p-circuit/ < optional explicit transport > /ipfs/QmRelay`

From the libp2p-swarm perspective, this is how you currently dial

- Dial over a PeerInfo. That should have the relay as part of its multiaddr and you know the dest Peer id from it as well        
- Dial over an ma string, this requires embedding the dest peer id into the maddr
    - This could look something like this `/p2p-circuit/ipfs/QmRelay/ipfs/QmDst`, note the nested `/ipfs/QmDst` not sure thats supported, but it seems like a valid case for this scenario
    - Multi hop dialing works pretty much the same, but nests /p2p-circuit addresses, i.e. `/p2p-circuit/ipfs/QmRelay1/p2p-circuit/ipfs/QmRelay2/ipfs/QmDest`, QmDest is reachable over QmRelay1, over QmRelay2


I think the above takes care of pretty much all scenarios? 

So we have:

- /p2p-circtuit/ipfs/QmRelay - a relay address
- /p2p-circuit/ipfs/QmRelay/ipfs/QmDest - QmDest is reachable over QmRealy
- /p2p-circuit/ipfs/QmRelay1/p2p-circuit/ipfs/QmRela2/ipfs/QmDest a multi hop relay address

I don't see a use case for /p2p-circuit/p2p-circuit/ipfs/QmSomething mentioned above.

I know this is slimmed down a lot from the initial set of examples, but if you think about it, it covers all the uses cases we want.

dryajov:
@whyrusleeping 

>How do i distinguish between Relay(Relay(AX, XY), YB) and Relay(AX, Relay(XY, YB))

Why would you need to distinguish them? I'm not following this particular scenario? To me a multi hop add is explicit about the path you want to follow to reach the destination - its a flat list of hops?

What am I missing here?

whyrusleeping:
@dryajov could you write out the multiaddrs for both those scenarios then?

diasdavid:
Posted this on IRC, but just for log purposes:

Is there anyone else thinking that the multihop stuff is becoming a rabbit hole of complexity? It is like trying to implement/support onion routing but without all the security properties.

Most users of relay should only have to do `nHops = 1` and for most enterprise networks will only increase that number by one.

I say this because I feel that `/p2p-circuit/p2p-circuit/<first relay>/<first hop multiaddr>/<destination peer multiaddr>` is elegant for these cases, which were the ones we were targeting with Relay in the first place.

diasdavid:
On the question, how to denote `Relay(Relay(AX, XY), YB) and Relay(AX, Relay(XY, YB))`

My answer is:

- Relay(Relay(AX, XY), YB) would be: `/p2p-circuit/p2p-circuit/X/Y/B`
- Relay(AX, Relay(XY, YB)) would be: `/p2p-circuit/X/p2p-circuit/Y/B`

dryajov:
@diasdavid I keep wondering why we need this format /p2p-circuit/p2p-circuit/<first relay>/<first hop multiaddr>/<destination peer multiaddr>  vs /p2p-circuit/ < ma >? What's he difference?

whyrusleeping:
@diasdavid In more concrete terms:
```
/p2p-circuit/p2p-circuit/ipfs/QmX/ipfs/QmY/ipfs/QmB
and
/p2p-circuit/ipfs/QmX/p2p-circuit/ipfs/QmY/ipfs/QmB
```

so the algorithm for a dialer trying to figure out what to do would be:

- split the address after the first `p2p-circuit`
- take everything until the last `/ipfs/` segment to be 'relay addr'
- connect to 'relay addr', ask to be relayed to the portion of the multiaddr after 'relay addr'

Running through this:
```
addr = /p2p-circuit/p2p-circuit/ipfs/QmX/ipfs/QmY/ipfs/QmB
take = /p2p-circuit/ipfs/QmX/ipfs/QmY/ipfs/QmB
relayaddr = /p2p-circuit/ipfs/QmX/ipfs/QmY
dest = /ipfs/QmB

And

addr = /p2p-circuit/ipfs/QmX/p2p-circuit/ipfs/QmY/ipfs/QmB
take = /ipfs/QmX/p2p-circuit/ipfs/QmY/ipfs/QmB
relayAddr = /ipfs/QmX/p2p-circuit/ipfs/QmY
dest = /ipfs/QmB
```

I think that makes sense. Someone double check the algorithm?

diasdavid:
@dryajov if we did `/p2p-circuit/ < ma >/p2p-circuit/ < ma >`, we couldn't fit @whyrusleeping first scenario `Relay(Relay(AX, XY), YB)`, because it describes a case where the relay also needs to establish a relayed connection. With `/p2p-circuit/ < ma >/p2p-circuit/ < ma >`, we can only specify "first do this relay and then do the other relay".

diasdavid:
@whyrusleeping in the case of `Relay(Relay(AX, XY), YB)` (aka `/p2p-circuit/p2p-circuit/ipfs/QmX/ipfs/QmY/ipfs/QmB`), the tricky bit is that the first relay needs to be a proactive relay that accepts hints (taking in a suggested relay route) of how to dial to the dest node. This would have to be factored in on the `hop` protocol.

whyrusleeping:
@diasdavid thats fine, I just want to make sure we have the notation correct

diasdavid:
In this one:

```
addr = /p2p-circuit/ipfs/QmX/p2p-circuit/ipfs/QmY/ipfs/QmB
take = /ipfs/QmX/p2p-circuit/ipfs/QmY/ipfs/QmB
relayAddr = /ipfs/QmX/p2p-circuit/ipfs/QmY
dest = /ipfs/QmB
```

The relayAddr should be just `/ipfs/QmX` and then we have to hint QmX to do a Relay dance with `/p2p-circuit/ipfs/QmY/ipfs/QmB`

dryajov:
@whyrusleeping @diasdavid 

Lets break this up one more time

- /p2p-circuit/p2p-circuit/ipfs/QmX/ipfs/QmY/ipfs/QmB
   - This implies asking a (any) relay to create another relay for you from from `/p2p-circuit/ipfs/QmX/ipfs/QmY`
       - dial any available relay and send `/p2p-circuit/ipfs/QmX` as dest
            - use the connection from above to dial QmB for you

- /p2p-circuit/ipfs/QmX/p2p-circuit/ipfs/QmY/ipfs/QmB
   - This is a chained addr
       - Dial into /ipfs/QmX send /ipfsQmY as the dest
       - ask the QmX<->QmY to dial QmB for you

dryajov:
@whyrusleeping @diasdavid 

One more thing. I think its good if we define some terminology around this, I think chaining vs nesting are good candidates.

* Chaining: A flat list of relay hops, 
   * the dialer calls them in sequence
      * /p2p-circuit/ipfs/QmA/p2p-circuit/QmB/ipfs/QmDst
* Nesting: A list with nested /p2p-circuit segments - i.e. /p2p-circuit/p2p-circuit/ipfs/QmA/ipfs/QmDst
    * The dialer will send the nested circuit address to a relay for further routing and will use the resulting connection for relaying to /ipfs/QmDst

dryajov:
@whyrusleeping @diasdavid 

```
whyrusleeping> dryajov: your examples don't quite make sense
14:40 you never just ask 'any' relay to relay a connection, thats the point of these addresses
14:41 we're explicitly specifying which node will do the relaying
```

We do. From the perspective of the dialer, if I can't reach a peer over any of its existing transports I will use one of the relays I know of to dial that peer.

***************

But here is what I'm trying to understand and so far its not been clear for the examples above:

Whats the difference between 

`/p2p-circuit/p2p-circuit/ipfs/QmA/ipfs/QmDst` 

and 

 `/p2p-circuit/ipfs/QmA/p2p-circuit/ipfs/QmB/ipfs/QmDst`


More importantly - `/p2p-circuit/p2p-circuit` as a prefix vs one `/p2p-circuit`?

lgierth:
A multiaddr represents a stack of protocols encapsulated in each other.
This is the "relativity" part of multiaddrs.
You can add and remove encapsulation layers based on context.
For example an `/ipfs` stream is the same no matter if it's within `/tcp` or `/ws`,
and a `/tcp` stream is the same no matter if it's `/ipfs/QmFoo` or `/mc/QmMediachain` within.

Let's stick with the following protocols here:

- `/ip4/<ipaddr>`
- `/tcp/<portno>`
- `/ipfs/<peerid>`
- `/p2p-circuit`

And lets also use spacing to make the multiaddr encapsulations more visible.

`/p2p-circuit` is just another layer of this encapsulation:

    /ip4/10.0.0.1/tcp/4001 /ipfs/QmRelay /p2p-circuit /ip4/1.2.3.4/tcp/4001 /ipfs/QmDest

This example is the most explicit `/p2p-circuit` multiaddr.
The topology resembles a local network gateway (10.0.0.1, QmRelay)
that we use to connect to a node on the public network (1.2.3.4, QmDest).
We're not only explicit about the relay node, but also about its network address.
We're also explicit about the destination node's network address,
which the relay node will dial on our behalf.

In other words, our libp2p connection (`/ipfs`) to QmDest
is encapsulated in a `/p2p-circuit` stream, in a libp2p connection to QmRelay.

QmRelay might not allow us to specify QmDest's network address,
or we might not even know one.
We can be less explicit
by relying on our peer routing, or on QmRelay's peer routing, or both.
Here, QmRelay uses its peer routing to pick a network address for QmDest,
and we use our own peer routing to pick one for QmRelay.

    /ipfs/QmRelay /p2p-circuit /ipfs/QmDest

We can be even less explicit
by letting our local `/p2p-circuit` dialer pick a relay node for us:

    /p2p-circuit /ipfs/QmDest

We can construct multi-hop relays too, specifying the whole route:

    /ipfs/QmRelay /p2p-circuit /ipfs/QmOtherRelay /p2p-circuit /ipfs/QmDest

Or multi-hop relays with an arbitrary route,
where the first relay node picks the second, and so on.

    /p2p-circuit /p2p-circuit /p2p-circuit /ipfs/QmDest

XXX: Big question, is this `Relay(Relay(Relay(Foo, Bar), Baz), Dest)`
or `Relay(Foo, Relay(Bar, Relay(Baz, Dest)))`?

- It depends on how the local swarm splits the address.
- The former would make a relay stream with Foo,
  and through that a relay stream to Bar, then to Baz, and then to Dest.
  - It'd basically unwrap the multiaddr for every occasion of `/p2p-circuit`.
- The latter would do it the other way round.
  - It'd unwrap the multiaddr only on the first `/p2p-circuit`,
    and send the rest to Foo, as the destination address.
- => The former only ever tells relay nodes the next hop.
- => The latter is less resource intensive locally the more hops there are,
  because it shakes hands only with the first hop and the destination
- => Both aren't mutually exclusive.
  You could switch between the two "modes" with a config setting,
  or they could be separate multiaddr protocols,
  or you could switch on a per-peerid basis.

Lastly, a `/p2p-circuit` stream doesn't neccessarily encapsulate an `/ipfs` or `/p2p` stream.
For all I care, it could be anything in there.
The relay node will tell you if it's something it doesn't support or allow.

whyrusleeping:
Note: dropping the header down to a single length prefixed multiaddr.
In the hop protocol this will be the destination addr.
In the stop protocol, this will be the source addr (set by the hop node).

lgierth:
### Listening

For the listener side, the same explicitness semantics apply.

Generally accept relayed streams:

    /p2p-circuit

Accept relayed streams only from QmRelay:

    /ipfs/QmRelay /p2p-circuit

Accept relayed streams only over 3 arbitrary hops:

    /p2p-circuit /p2p-circuit /p2p-circuit

The respective announced addresses could be the same, but with your `/ipfs/<peerid>` added.

lgierth:
Specifying a `/p2p-circuit` listener would also add a `/libp2p/relay/circuit/0.0.0/stop` stream handler.

Enabling being a relay would add a `/libp2p/relay/circuit/0.0.0/hop` stream handler.

dryajov:
@lgierth thanks for the proto stack visualization, it all makes sense now 👍 

Here is my summary:

* From that visualization, `p2p-circuit` encapsulates all other protos (tho it doesn't really care what they are). In plain english, `p2p-circuit` comes at the end of the multiaddr sequence/stream.

However:

* In practice `p2p-circuit` could come at the beginning, in between or at the end of a multiaddr sequence. (This makes it delightfully recursive 😺  but I think I've got it mostly covered here - https://github.com/whyrusleeping/js-mafmt/pull/14)

***************

> XXX: Big question, is this Relay(Relay(Relay(Foo, Bar), Baz), Dest)
or Relay(Foo, Relay(Bar, Relay(Baz, Dest)))?

* Two things regarding the above two cases (onion vs telescoping?)
   1. It be good to have the maddress convey which way the dialing happen, tho after thinking about it for a while I'm drawing a blank. We can do a flag as @lgierth suggested
   2. Should we make a decision on which mode of dialing multihops we want implemented first? So far, @diasdavid and I have been talking about having the client peer dial in sequence (onion?). But telescoping should actually be even easier, you just simply send the rest of the multiaddr as the dest to the first hop in the addr.

*******************

* `/p2p-circuit /p2p-circuit /p2p-circuit` and `/p2p-circuit /p2p-circuit /p2p-circuit /ipfs/QmDest` what are the real world use cases for this? I can see a few issues if we enable this addresses.

   * What happens if my peer only knows of one relay? 
       * Does it dial on that relay three times or does it skip the other two relays in the multiaddr? 
       * Similarly, when listening on a relay with several peer routed `hops`, if the dialer only knows of one relay, whats happens?
    * IMO, unless this has some real life applications, we should leave them out of the implementation.

diasdavid:
Interesting perspective, @lgierth, that does it more readable and parsing it is also easier. I like it :)

If everyone agrees with that (:+1: this comment) and I'll update this PR.

dryajov:
@diasdavid A few tings worth adding (maybe even as placeholders), while you're at it:

- Active/Passive dialing

- Relay selection
   - Look for relays that are already connected to my dst peer
       - if none found, look for an active relay that can dial for me
           - if none found, do a random/roundrobin selection of known relays
               - if none found, ask the dht for a relay

- How listening on a relayed connection works?
   - According to @lgierth, listening on a relayed address disables being reached over any other relay/transport (thats at least how I understood it from our IRC convo), which differs from what we discussed previously

- Relay discovery

dryajov:
@lgierth @diasdavid @whyrusleeping 

> How listening on a relayed connection works?
According to @lgierth, listening on a relayed address disables being reached over any other relay/transport (thats at least how I understood it from our IRC convo), which differs from what we discussed previously.

I think this should be our next topic of discussion. 

Here is how I understand listening in general:

* By default all peers accept relayed connections
* Peers can listen on specific relay addresses
   * It could be several, just like any other transport

(NOTE: relay is only enabled if muxing is present)

Now, to disable listening on other transports or not do disable, that is the question: 😄

* Q: what are we trying to accomplish by disabling other transports and only listening on the relay address?
* A: Force traffic through that address, because of:
    * e.i. you're in a slow network and don't want to talk to other peers directly cause they are slow
    * (list other reasons)

I think this is requirement is outside of the concerns of relay and, should be handled with a different mechanism that would allow ranking enabling/disabling transports based on conditions and should probably work for other transports as well.

lgierth:
> I think this is requirement is outside of the concerns of relay and, should be handled with a different mechanism that would allow ranking enabling/disabling transports based on conditions and should probably work for other transports as well.

Yeah I agree this might not have to be decided now

relay/README.md
dryajov: I think it depends on how you look at relay. To me, relay is just another transport, hence making it behave like any other transport makes sense - here, I'm specifically talking about the dialer/listener (/stop) parts of the relay. The circuit-relay itself, can/should be controlled by config options.

relay/README.md
dryajov: Also, after @lgierth clarifications this should be `<relay peer multiaddr>/p2p-circuit/<destination peer multiaddr>`

relay/README.md
```diff
@@ -84,91 +65,189 @@ Scene 2:
...
 
-As with all other multiaddrs, encapsulation of different protocols
-determines which metaphorical tubes to connect to each other.
+`/p2p-circuit/p2p-circuit/<first relay>/<first hop multiaddr>/<destination peer multiaddr>`
```
dryajov: Trying to reformat this example to match @lgierth clarifications `/p2p-circuit/p2p-circuit/<first relay>/<first hop multiaddr>/<destination peer multiaddr>`

I think the above is intended to read as follows:

* `<first relay>/p2-pcircuit/<second relay>/p2p-circuit/ipfs/QmDst`
    * use the first hop `<first relay>/p2-pcircuit`
         * either peer routed or using its explicit transport if specified
     * use the second hop `<second relay>/p2p-circuit`
          * either peer routed or using its explicit transport if specified
      * circuit `/ipfs/QmDst` over the <first relay> <-> <second relay> pipe

relay/README.md
```diff
@@ -84,91 +65,189 @@ Scene 2:
...
 
 A few examples:
 
+Using any relay available:
+
 - `/p2p-circuit/p2p/QmTwo`
```
dryajov: Should be:

/p2p/QmTwo/p2p-circuit

Dial QmTwo, through any available relay node (or find one node that can relay).
The relay node will use peer routing to find an address for QmTwo if it doesn't have a direct connection.

relay/README.md
```diff
@@ -84,91 +65,189 @@ Scene 2:
...
-  - Dial QmTwo, through any available relay node.
-  - The relay node will use peer routing to find an address for QmTwo.
-- `/p2p/QmRelay/p2p-circuit/p2p/QmTwo`
-  - Dial QmTwo, through QmRelay.
-  - Use peer routing to find an address for QmRelay.
-  - The relay node will also use peer routing, to find an address for QmTwo.
+  - Dial QmTwo, through any available relay node (or find one node that can relay).
+  - The relay node will use peer routing to find an address for QmTwo if it doesn't have a direct connection.
 - `/p2p-circuit/ip4/../tcp/../p2p/QmTwo`
```
dryajov: Should be : 

`/ip4/../tcp/../p2p-circuit/p2p/QmTwo`

relay/README.md
```diff
@@ -84,91 +65,189 @@ Scene 2:
...
   - Dial QmTwo, through any available relay node,
     but force the relay node to use the encapsulated `/ip4` multiaddr for connecting to QmTwo.
   - We'll probably not support forced addresses for now, just because it's complicated.
-- `/ip4/../tcp/../p2p/QmRelay/p2p-circuit`
-  - Listen for connections relayed through QmRelay.
+
+Specify a relay:
+
+- `/p2p-circuit/p2p/QmRelay/p2p/QmTwo`
```
dryajov: Should be:

`/p2p/QmRelay/p2p-circuit/p2p/QmTwo`

relay/README.md
```diff
@@ -84,91 +65,189 @@ Scene 2:
...
+  - Dial QmTwo, through QmRelay.
+  - Use peer routing to find an address for QmRelay.
+  - The relay node will also use peer routing, to find an address for QmTwo.
+- `/p2p-circuit/ip4/../tcp/../p2p/QmRelay/p2p/QmTwo`
```
dryajov: Should be:

`/ip4/../tcp/../p2p/QmRelay/p2p-circuit/p2p/QmTwo`

relay/README.md
```diff
@@ -84,91 +65,189 @@ Scene 2:
...
+#### Process
+
+- A opens new stream `sAR` to R using protocol RELAY
+- A writes Bs multiaddr `/ipfs/QmB` on `sAR`
+- R receives stream `sAR` and reads `/ipfs/QmB` from it.
+- R opens a new stream `sRB` to B using protocol RELAY
+- R writes `/ipfs/QmB` on `sRB`
+- B receives stream `sRB` and reads `/ipfs/QmB` from it.
+- B sees that the multiaddr it read is its own and chooses to handle this stream as an endpoint instead of attempting to relay further
+- TODO: step for R to send back status code to A
```
dryajov: Thoughts on status codes:

* R doesn't send code if the circuit is successful only if there was a failure
* A would get a success or failure code from B once the circuit is established

relay/README.md
dryajov: @diasdavid @whyrusleeping 

After, re-reading the table... They don't really map to HTTP at all... 200s are relay codes and 300 are dst errors. So, I think 1xx, 2xx, 3xx make sense here.

diasdavid:
I've reviewed: 

- [x] Overview
- [x] Dramatization
- [x] Addressing

Missing:

- [ ] Wire Protocol
- [ ] Interfaces
- [ ] Implementation details

I'll review these last 3 in separate PRs to this one to make sure I don't merge anything to the PR without having everyone in the same page.

lgierth:
Three comments, then LGTM :+1:

diasdavid:
Applied @lgierth review, now onto the remaining bits.

diasdavid:
https://github.com/libp2p/specs/pull/18 up for review

diasdavid:
**Not yet**, https://github.com/libp2p/specs/pull/15#issuecomment-299949023 needs to happen

lgierth:
@diasdavid there's an inaccurate spec already merged though -- this PR is a great improvement over the existing document, and isn't technically blocked by the remaining tasks. Anyhow I added a note about this work-in-progress in #20 as an alternative.

diasdavid:
A new chapter here:
==================

Following up the discussion on IRC (#ipfs-dev@freenode), I'll be working on finishing this spec to my best understanding and contemplating all the improvements discussed. Once this revamp is done, I'll open a new PR for formal review.

diasdavid:
It is here! https://github.com/libp2p/specs/pull/22

Guess what, using a protobuf simplifies so much the 'under the microscope section' :)

PR #18, opened 2017-05-09
review Wire protocol section from the relay spec

diasdavid:
Reviewed the wire protocol section. There were some incoherences in the high-level order of events and some question marks everywhere, I fixed them based on commons and our conversations.

### DIFF
--- a/relay/README.md
+++ b/relay/README.md
@@ -114,54 +114,46 @@ Double relay:
 
 ## Wire format
 
-### Overview
+We start the description of the Wire format by illustrating a possible flow scenario and then describing them in detail by phases.
 
-#### Setup
+### High level overview of establishing a relayed connection
 
-Peers involved:
-- A, B, R
+Setup:
+- Peers involved, A, B, R
 - A wants to connect to B, but needs to relay through R
 
-#### Assumptions
+Assumptions:
+- A has connection to R, R has connection to B
 
-A has connection to R, R has connection to B
-
-#### Sequence of events
-
-- A opens new stream `sAR` to R using protocol RELAY
-- A writes Bs multiaddr `/p2p/QmB` on `sAR`
-- R receives stream `sAR` and reads `/p2p/QmB` from it.
-- R opens a new stream `sRB` to B using protocol RELAY
-- R writes `/p2p/QmB` on `sRB`
-- B receives stream `sRB` and reads `/p2p/QmB` from it.
-- B sees that the multiaddr it read is its own and chooses to handle this stream as an endpoint instead of attempting to relay further
-- TODO: step for R to send back status code to A
-- R now pipes `sAR` and `sRB` together
-- TODO: step for B to send back status code to A
-- B passes stream to `NewConnHandler` to be handled like any other new incoming connection
+Events:
+- phase I: Open a request for a relayed stream (A to R)
+  - A dials a new stream `sAR` to R using protocol `/libp2p/relay/hop` 
+  - A writes Bs multiaddr `/p2p/QmB` on `sAR`
+  - R receives stream `sAR` and reads `/p2p/QmB` from it.
+- phase II: Open a stream to be relayed (R to B)
+  - R opens a new stream `sRB` to B using protocol `/libp2p/relay/stop`
+  - R writes `/p2p/QmA` on `sRB`
+  - R writes OK to `sAR`
+- phase III: Streams are piped together, establishing a circuit
+  - B receives stream `sRB` and reads `/p2p/QmA` from it.
+  - B writes OK to `sRB`
+  - B passes stream to `NewConnHandler` to be handled like any other new incoming connection
 
 ### Under the microscope
 
-Peer A wants to connect to peer B through peer R.
+Notes for the reader:
+- `maddrA` is peer A's multiaddr
+- `maddrR` is peer R's multiaddr
+- `maddrB` is peer B's multiaddr
+- `maxAddrLen` is arbitrarily 1024
+- `readLpMaddr` is a function that does the following:
 
-`maddrA` is peer A's multiaddr
-`maddrR` is peer R's multiaddr
-`maddrB` is peer B's multiaddr
-`maxAddrLen` is arbitrarily 1024
+Read a Uvarint `V` from the stream. If the value is higher than `maxAddrLen`, writes an error code back, closes the stream and halts the relay attempt. If not, then reads `V` bytes from the stream and checks if its a valid multiaddr. Check for multiaddr validity and error back if not a valid formated multiaddr.
 
-#### Function definitions
-##### Process for reading a multiaddr
-We define `readLpMaddr` to be the following:
+#### Phase I
 
-Read a Uvarint `V` from the stream. If the value is higher
-than `maxAddrLen`, (write an error message back?) close the 
-stream and halt the relay attempt.
+Peer A opens a new stream to R on the '/libp2p/relay/hop' protocol and writes:
 
-Then read `V` bytes from the stream and checks if its a valid multiaddr.
-If it is not a valid multiaddr (write an error back?) close the stream and return.
-
-#### Opening a relay
-Peer A opens a new stream to R on the 'hop' protocol and writes:
 ```
 <uvarint len(maddrA)><madddrA><uvarint len(maddrB)><madddrB>
 ```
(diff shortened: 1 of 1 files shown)


relay/README.md
```diff
@@ -114,54 +114,46 @@ Double relay:
...
-- TODO: step for R to send back status code to A
-- R now pipes `sAR` and `sRB` together
-- TODO: step for B to send back status code to A
-- B passes stream to `NewConnHandler` to be handled like any other new incoming connection
+Assumptions:
+- A has connection to R, R has connection to B
+
+Events:
+- phase I: Open a request for a relayed stream (A to R)
+  - A dials a new stream `sAR` to R using protocol `/libp2p/relay/hop`
```
lgierth: It's `/libp2p/relay/circuit/1.0.0/hop`

relay/README.md
```diff
@@ -114,54 +114,46 @@ Double relay:
...
+Assumptions:
+- A has connection to R, R has connection to B
+
+Events:
+- phase I: Open a request for a relayed stream (A to R)
+  - A dials a new stream `sAR` to R using protocol `/libp2p/relay/hop` 
+  - A writes Bs multiaddr `/p2p/QmB` on `sAR`
+  - R receives stream `sAR` and reads `/p2p/QmB` from it.
+- phase II: Open a stream to be relayed (R to B)
+  - R opens a new stream `sRB` to B using protocol `/libp2p/relay/stop`
```
lgierth: Same here `/libp2p/relay/circuit/1.0.0/stop`

relay/README.md
```diff
@@ -114,54 +114,46 @@ Double relay:
...
+- `maddrA` is peer A's multiaddr
+- `maddrR` is peer R's multiaddr
+- `maddrB` is peer B's multiaddr
+- `maxAddrLen` is arbitrarily 1024
+- `readLpMaddr` is a function that does the following:
 
-#### Function definitions
-##### Process for reading a multiaddr
-We define `readLpMaddr` to be the following:
+Read a Uvarint `V` from the stream. If the value is higher than `maxAddrLen`, writes an error code back, closes the stream and halts the relay attempt. If not, then reads `V` bytes from the stream and checks if its a valid multiaddr. Check for multiaddr validity and error back if not a valid formated multiaddr.
```
lgierth: This line should be indented as a sub-item of the previous list item

relay/README.md
lgierth: And a couple more of the same throughout the rest of the doc

relay/README.md
```diff
@@ -171,31 +163,38 @@ It then waits for a response in the form of:
...
 
-### 'hop' protocol handler
+Peer R receives a new stream on the '/libp2p/relay/hop' protocol.
 
-Peer R receives a new stream on the 'hop' protocol.
 It then calls `readLpMaddr` twice. The first value is `<src>` and the second is `<dst>`.
-Peer R checks to make sure that `<src>` matches the remote peer of the stream its reading
-from. If it does not match, it (writes an error back?) closes the stream and halts the relay attempt.
+
+Peer R checks to make sure that `<src>` matches the remote peer of the stream its reading from. If it does not match, it writes an Error message, closes the stream and halts the relay attempt.
```
lgierth: typo its/**it's**

relay/README.md
```diff
@@ -171,31 +163,38 @@ It then waits for a response in the form of:
...
+#### Phase II
+
+R now opens a new stream to B with the '/libp2p/relay/stop' relay protocol multicodec and writes:
+
 ```
 <uvarint len(maddrA)><madddrA><uvarint len(maddrB)><madddrB>
 ```
 
-After this, R simply pipes the stream from A and the stream it opened to B together. R's job is complete.
+After this, R pipes the stream from A and the stream it opened to B together. R also writes OK back to A. R's job is complete.
```
lgierth: Maybe worth noting that R's job actually continues: it now needs to keep the connections alive

relay/README.md
```diff
@@ -171,31 +163,38 @@ It then waits for a response in the form of:
...
+
+#### Phase III
 
-### 'stop' protocol handler
+Peer B receives a new stream on the '/libp2p/relay/stop' protocol. It then calls `readLpMaddr` twice on this stream.
```
dryajov: Writing both the `<src>` and `<dst>` seems a bit redundant here. The current js implementation only cares about the `<src>` addr. What is the use case for sending the `<dst>` as well?

dryajov: Other than that LGTM! 👍

relay/README.md
diasdavid: This is the definition of that function. I can change how that is communicated.

diasdavid:
Thanks for the review @lgierth @dryajov <3

There is just one thing that is itching me, the fact that we are mixing a protocol codec with an RPC method name on having `/libp2p/relay/circuit/1.0.0/hop` and `/libp2p/relay/circuit/1.0.0/stop`. I'm afraid that this will encourage people to start designing RPCs using multistream-select as their mechanism to hit the right function and as far libp2p should be a concern, it should just do the Protocol Muxing, not RPC select.

If we want to say that we have the `hop` protocol and the `stop` protocol within relay/circuit scope, that is '''fine''', but that would mean that semver would appear at the end.

I also don't understand the resistant to use a serializable data format like protobuf or cbor this time, we use it everywhere already (see Identify, Bitswap, DHT, and others) and it gives a nice feature that we can add more information later without having to do complicated retro compatibility tests to make sure all the bytes are read in the same order and what not. Using a protobuf this time would give us the ability to just specify the type of message (hop, stop) on the message itself.

relay/README.md
lgierth: Just trying to save space :)

dryajov:
> There is just one thing that is itching me, the fact that we are mixing a protocol codec with an RPC method name on having /libp2p/relay/circuit/1.0.0/hop and /libp2p/relay/circuit/1.0.0/stop. I'm afraid that this will encourage people to start designing RPCs using multistream-select as their mechanism to hit the right function and as far libp2p should be a concern, it should just do the Protocol Muxing, not RPC select.

For the sake of playing the devils advocate, what is the danger of using multistream select as an RPC mechanism (note, I'm neither proposing nor condemning that use case). Also, how placing the semver at the end would prevent that - the current syntax maps nicely to a _sub protocol_ in my mind, which is accurate in this case.

> I also don't understand the resistant to use a serializable data format like protobuf or cbor this time, we use it everywhere already (see Identify, Bitswap, DHT, and others) and it gives a nice feature that we can add more information later without having to do complicated retro compatibility tests to make sure all the bytes are read in the same order and what not. Using a protobuf this time would give us the ability to just specify the type of message (hop, stop) on the message itself.

I would love to hear the reasoning behind this also, the only thing that comes to mind is that the amount of structured data exchanged between peers is minimal, and going with a text based protocol saves the additional dependencies, complexity and overhead that cbor/protobuf introduce?

diasdavid:
> what is the danger of using multistream select as an RPC mechanism

multistream-select has a cost:
- it adds to the list of protocols supported, which means that there is one more element for the linear search that happens every time a protocol mux happens.
- it makes it weird with semver, because as you said 'subprotocol', what is the criteria that the subprotocol doesn't have a version as well? When is a thing a protocol and when is a thing a subprotocol? It just adds complexity to the parsing
- if we decide to argue about bytes (which has been a discussion in the past even when the argument is just to save 1 byte), `hop` and `stop` will be more bytes than just a number on a protobuf message, since multistream messages are not binary packed
- it encourages to others do the same and you should not have to inspect multistream-select logs to check behaviors of your own protocol that is mounted on top

> the only thing that comes to mind is that the amount of structured data exchanged between peers is minimal, and going with a text based protocol saves the additional dependencies, complexity and overhead that cbor/protobuf introduce?

protobuf RPC exist to reduce complexity and we already know that things change over time really well, that is why we have a whole project just for self-describable data formats :)

lgierth:
`/hop` and `/stop` were introduced to differentiate semantics (opening vs. receiving a relayed connection). The second multiaddr in the header was removed because it wasn't needed anymore once this differentiation was there. I agree that byte-savings are a weak argument here :)

I'd be down for us reworking this part -- I kinda agree that the changes to multistream-select would be non-trivial and likely risky, while the neccessity for them isn't very strong.

I'd also be down to at the very least put a multicodec header in front of the header. (Good call.)

dryajov:
@lgierth 
What you mean by this:

> I'd also be down to at the very least put a multicodec header in front of the header. (Good call.)

lgierth:
>  What you mean by this:

Similar to how there's a multicodec-packed header in a CID to describe the following hash, we'd put one in front of the relay header to describe what follows. For now this would just be the multicodec code for multiaddr (0x32), but being able to switch this over to e.g. the cbor code, would make the header more future-proof.

See https://github.com/multiformats/multicodec

dryajov:
@lgierth ah, makes sense. Thanks!

dryajov:
Re: RPC

I think we need to make a decision to move circuit forward. Lets time box this, how about we make a decision by next `Friday, Jun. 23rd` the latest?

@lgierth what are your thoughts on this, I gather that you tend to agree with what @diasdavid is proposing.

I'm on board with reworking the RPC calls with a proper, binary packed format instead.

Next question is, what format should we use - `cbor` or `protobuf` (or something else)? I see that `libp2p` uses `protobuf` heavily, is that the preferred route? I don't have strong opinion on neither one of them as I haven't used either.

dryajov:
Vote CBOR: 👍 or 👎

dryajov:
Vote protobufs 👍 or 👎

whyrusleeping:
i prefer protobufs here, pretty much just because they will be much smaller

diasdavid:
I'm also in favour to use protobufs. It has suited well our needs and as @whyrusleeping points out, they are smaller.

dryajov:
Looks like we're rewriting the protocol with protobufs - If there are no objections I'll start working on this ASAP.

@whyrusleeping @diasdavid @lgierth @dignifiedquire

dryajov:
@lgierth @diasdavid @whyrusleeping 

I though I put something together for the actual protobuf message to get the ball rolling. 

```
message Circuit {
  required string version = 1;
  required RelayMessage message = 2;
  
  message Peer {
    required string id = 1; // peer id
    repeated string address = 2; // peer's dialable addresses      
  }
  
  message RelayMessage {
    required Peer source = 1;
    required Peer dest = 2;
  }
}
```

PR #20, opened 2017-05-22
relay: add note about in-progress review and update

lgierth:
@diasdavid this is just a temporary note while the review and update is in progress -- to prevent confusion and wasted effort on the part of others.


diasdavid:
Thank you @lgierth, this will help a ton of people :)

PR #22, opened 2017-07-08
The Circuit Relay Specification

diasdavid:
As promised on Friday, July 8, I've cleaned up, coalesced and added what was missing to the Relay spec, taking the bits from PRs #18 and #15 and all the comments. 

Hope you enjoy :)

### DIFF
--- a/relay/README.md
+++ b/relay/README.md
@@ -1,4 +1,4 @@
-# Circuit Relay
+# Circuit Relay v0.1.0
 
 > Circuit Switching for libp2p, also known as TURN or Relay in Networking literature.
 
@@ -19,7 +19,7 @@
 
 ## Overview
 
-The circuit relay is a mean to establish connectivity between libp2p nodes (e.g IPFS) that wouldn't otherwise be able to establish a direct connection to each other.
+The circuit relay is a means to establish connectivity between libp2p nodes (e.g IPFS) that wouldn't otherwise be able to establish a direct connection to each other.
 
 Relay is needed in situations where nodes are behind NAT, reverse proxies, firewalls and/or simply don't support the same transports (e.g. go-ipfs vs. browser-ipfs). Even though libp2p has modules for NAT traversal ([go-libp2p-nat](https://github.com/libp2p/go-libp2p-nat)), piercing through NATs isn't always an option. The circuit relay protocol exists to overcome those scenarios.
 
@@ -60,7 +60,7 @@ Scene 2:
 - All three nodes have learned to speak the `/ipfs/relay/circuit` protocol.
 - QmRelay is configured to allow relaying connections between other nodes.
 - QmOne is configured to use QmRelay for relaying.
-- QmOne automatically added `/p2p-circuit/ipfs/QmTwo` to its set of QmTwo addresses.
+- QmOne automatically added `/p2p-circuit/p2p/QmTwo` to its set of QmTwo addresses.
 - QmOne tries to connect via relaying, because it shares this transport with QmTwo.
 - A lively and prolonged dialogue ensues.
 
@@ -77,7 +77,7 @@ A `/p2p-circuit` circuit address, is formated following:
 Examples:
 
 - `/p2p-circuit/p2p/QmVT6GYwjeeAF5TR485Yc58S3xRF5EFsZ5YAF4VcP3URHt` - Arbitrary relay node
-- `/ip4/127.0.0.1/tcp/5002/ipfs/QmdPU7PfRyKehdrP5A3WqmjyD6bhVpU1mLGKppa2FjGDjZ/p2p-circuit/p2p/QmVT6GYwjeeAF5TR485Yc58S3xRF5EFsZ5YAF4VcP3URHt` - Specific relay node
+- `/ip4/127.0.0.1/tcp/5002/p2p/QmdPU7PfRyKehdrP5A3WqmjyD6bhVpU1mLGKppa2FjGDjZ/p2p-circuit/p2p/QmVT6GYwjeeAF5TR485Yc58S3xRF5EFsZ5YAF4VcP3URHt` - Specific relay node
 
 This opens the room for multiple hop relay, where the second relay is encapsulated in the first relay multiaddr, such as:
 
@@ -184,8 +184,8 @@ This is a table of status codes and sample messages that may occur during a rela
 | 250   | "failed to parse src addr: no such protocol ipfs" | The `<src>` multiaddr in the header was invalid |
 | 251   | "failed to parse dst addr: no such protocol ipfs" | The `<dst>` multiaddr in the header was invalid |
 | 260   | "passive relay has no connection to dst"          | |
-| 261   | "active relay could'nt dial to dst: conn refused" | relay could not form new connection to target peer |
-| 262   | "could'nd dial to dst: BAD ERROR"                 | relay has conn to dst, but failed to open a stream |
+| 261   | "active relay couldn't dial to dst: conn refused" | relay could not form new connection to target peer |
+| 262   | "couldn't' dial to dst"                           | relay has conn to dst, but failed to open a stream |
 | 270   | "dst does not support relay"                      | |
 | 280   | "can't relay to itself"                           | The relay got its own address as destination |
 | 320   | "src address too long"                            | |
--- a/relay/README.md
+++ b/relay/README.md
@@ -19,13 +19,13 @@
 
 ## Overview
 
-The circuit relay is a means to establish connectivity between libp2p nodes (e.g IPFS) that wouldn't otherwise be able to establish a direct connection to each other.
+The circuit relay is a means to establish connectivity between libp2p nodes (e.g. IPFS nodes) that wouldn't otherwise be able to establish a direct connection to each other.
 
 Relay is needed in situations where nodes are behind NAT, reverse proxies, firewalls and/or simply don't support the same transports (e.g. go-ipfs vs. browser-ipfs). Even though libp2p has modules for NAT traversal ([go-libp2p-nat](https://github.com/libp2p/go-libp2p-nat)), piercing through NATs isn't always an option. The circuit relay protocol exists to overcome those scenarios.
 
 Unlike a transparent **tunnel**, where a libp2p peer would just proxy a communication stream to a destination (the destination being unaware of the original source), a circuit relay makes the destination aware of the original source and the circuit followed to establish communication between the two. This provides the destination side with full knowledge of the circuit which, if needed, could be rebuilt in the opposite direction.
 
-Apart from that, this relayed connection behaves just like a regular connection would, but over an existing swarm stream with another peer (instead of e.g. TCP.). A node asks a relay node to connect to another node on its behalf. The relay node short-circuits streams between the two nodes, enabling them to reach each other.
+Apart from that, this relayed connection behaves just like a regular connection would, but over an existing swarm stream with another peer (instead of e.g. TCP). A node asks a relay node to connect to another node on its behalf. The relay node short-circuits streams between the two nodes, enabling them to reach each other.
 
 Relayed connections are end-to-end encrypted just like regular connections.
 
--- a/relay/README.md
+++ b/relay/README.md
@@ -1,69 +1,48 @@
 # Circuit Relay
 
-> Circuit Switching in libp2p
+> Circuit Switching for libp2p, also known as TURN or Relay in Networking literature.
 
-Implementations
-- [js-libp2p-circuit](https://github.com/libp2p/js-libp2p-circuit) -- status: started
-- [go-libp2p-circuit](https://github.com/libp2p/go-libp2p-circuit) -- status: ready-to-start
+## Implementations
 
-Table of Contents
+- [js-libp2p-circuit](https://github.com/libp2p/js-libp2p-circuit)
+- [go-libp2p-circuit](https://github.com/libp2p/go-libp2p-circuit)
+
+## Table of Contents
+
 - [Overview](#overview)
 - [Dramatization](#dramatization)
 - [Addressing](#addressing)
 - [Wire protocol](#wire-protocol)
 - [Interfaces](#interfaces)
+- [Implementation Details](#implementation-details)
 - [Removing existing relay protocol](#removing-existing-relay-protocol)
-
-Note: this document is currently in the process of being double-checked and updated. Please see [libp2p/specs#15 (comment)](https://github.com/libp2p/specs/pull/15#issuecomment-299949023) before using it as the basis for an implementation.
-
 
 ## Overview
 
-The circuit relay is a means of establishing connectivity between
-libp2p nodes (such as IPFS) that wouldn't otherwise be able to connect to each other.
+The circuit relay is a mean to establish connectivity between libp2p nodes (e.g IPFS) that wouldn't otherwise be able to establish a direct connection to each other.
 
-This helps in situations where nodes are behind NAT or reverse proxies,
-or simply don't support the same transports (e.g. go-ipfs vs. browser-ipfs).
-libp2p already has modules for NAT ([go-libp2p-nat](https://github.com/libp2p/go-libp2p-nat)),
-but these don't always do the job, just because NAT traversal is complicated.
-That's why it's useful to have a simple relay protocol.
+Relay is needed in situations where nodes are behind NAT, reverse proxies, firewalls and/or simply don't support the same transports (e.g. go-ipfs vs. browser-ipfs). Even though libp2p has modules for NAT traversal ([go-libp2p-nat](https://github.com/libp2p/go-libp2p-nat)), piercing through NATs isn't always an option. The circuit relay protocol exists to overcome those scenarios.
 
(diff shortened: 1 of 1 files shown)


relay/README.md
```diff
@@ -1,69 +1,48 @@
...
 
-Unlike a transparent **tunnel**, where a libp2p peer would just proxy a
-communication stream to a destination (the destination being unaware of the
-original source), a circuit-relay makes the destination aware of the original
-source and the circuit followed to establish communication between the two.
-This provides the destination side with full knowledge of the circuit which,
-if needed, could be rebuilt in the opposite direction.
+Unlike a transparent **tunnel**, where a libp2p peer would just proxy a communication stream to a destination (the destination being unaware of the original source), a circuit relay makes the destination aware of the original source and the circuit followed to establish communication between the two. This provides the destination side with full knowledge of the circuit which, if needed, could be rebuilt in the opposite direction.
```
dryajov: Do we really need this? It seems like there could be a lot of complexity involved in making this work properly, as opposed to leaving the job of re-establishing a dropped connection to the `source`. So far I have left this out in the js implementation.

relay/README.md
```diff
@@ -86,93 +64,144 @@ Scene 2:
...
-
- ^    ^    ^
- |    |    |
- |    |    +-- bidirectional data stream
- |    |         (usually /multistream-select in the case of /p2p multiaddrs)
- |    |
- |    +------- multiaddr of the listening node
- |
- +------------ multiaddr of the dialing node
+message CircuitRelay {
```
dryajov: Looks good, but should we have a version as part of the message as well? I know its communicated as part of the multistream-select rpc end point, but having a version in the message might help preventing stupid mistakes like `we made changes to the protobuff msg, but forgot to bump the multistream version`.

relay/README.md
```diff
@@ -86,93 +64,144 @@ Scene 2:
...
+  enum Type { // RPC identifier, either HOP, STOP or STATUS
+    HOP = 1; 
+    STOP = 2; 
+    STATUS = 3;
```
dryajov: If we're returning the status as part of the message, we should probably make the return codes protobuf enums as well.

relay/README.md
diasdavid: - protobufs have a very clean way to be updated already
- "we made changes to the protobuf msg, but forgot to bump the multistream version" - this is not a valid technical reason, the truth is that adding a version won't stop you from making any of the mistakes you suggest, in fact, it opens the door to even more.

relay/README.md
diasdavid: > if needed, could be rebuilt

It is not handled by the module. It is "if needed, the dst has the information to rebuild the conn back"

relay/README.md
diasdavid: I like that 👍

relay/README.md
```diff
@@ -86,93 +64,144 @@ Scene 2:
...
+- Peers involved, A, B, R
+- A wants to connect to B, but needs to relay through R
+
+**Assumptions:**
+- A has connection to R, R has connection to B
+
+**Events:**
+- phase I: Open a request for a relayed stream (A to R).
+  - A dials a new stream `sAR` to R using protocol `/libp2p/circuit/relay/0.1.0`.
+  - A sends a CircuitRelay message with `{ type: 'HOP', srcPeer: '/p2p/QmA', dstPeer: '/p2p/QmB' }` to R through `sAR`.
```
JustinDrake: Should this `HOP` message be immediately acknowledged by `R` with a message? Opening a new stream `sRB` can take some time, and it would be good for `A` to be able to distinguish timeouts where `R` is slow, or where `B` is slow.

relay/README.md
```diff
@@ -1,69 +1,48 @@
...
 - [Removing existing relay protocol](#removing-existing-relay-protocol)
 
-Note: this document is currently in the process of being double-checked and updated. Please see [libp2p/specs#15 (comment)](https://github.com/libp2p/specs/pull/15#issuecomment-299949023) before using it as the basis for an implementation.
-
-
 ## Overview
 
-The circuit relay is a means of establishing connectivity between
-libp2p nodes (such as IPFS) that wouldn't otherwise be able to connect to each other.
+The circuit relay is a mean to establish connectivity between libp2p nodes (e.g IPFS) that wouldn't otherwise be able to establish a direct connection to each other.
```
JustinDrake: Typo "mean" should be "means". (The singular of "means" is "means".)

relay/README.md
```diff
@@ -86,93 +64,144 @@ Scene 2:
...
+| Code  | Message                                           | Meaning    |
+| ----- |:--------------------------------------------------|:----------:|
+| 100   | OK                                                | Relay was setup correctly |
+| 220   | "src address too long"                            | |
+| 221   | "dst address too long"                            | |
+| 250   | "failed to parse src addr: no such protocol ipfs" | The `<src>` multiaddr in the header was invalid |
+| 251   | "failed to parse dst addr: no such protocol ipfs" | The `<dst>` multiaddr in the header was invalid |
+| 260   | "passive relay has no connection to dst"          | |
+| 261   | "active relay could'nt dial to dst: conn refused" | relay could not form new connection to target peer |
+| 262   | "could'nd dial to dst: BAD ERROR"                 | relay has conn to dst, but failed to open a stream |
```
JustinDrake: Typo "could'nd" -> "couldn't"

relay/README.md
```diff
@@ -86,93 +64,144 @@ Scene 2:
...
+**Events:**
+- phase I: Open a request for a relayed stream (A to R).
+  - A dials a new stream `sAR` to R using protocol `/libp2p/circuit/relay/0.1.0`.
+  - A sends a CircuitRelay message with `{ type: 'HOP', srcPeer: '/p2p/QmA', dstPeer: '/p2p/QmB' }` to R through `sAR`.
+  - R receives stream `sAR` and reads the message from it.
+- phase II: Open a stream to be relayed (R to B).
+  - R opens a new stream `sRB` to B using protocol `/libp2p/circuit/relay/0.1.0`.
+  - R sends a CircuitRelay message with `{ type: 'STOP', srcPeer: '/p2p/QmA', dstPeer: '/p2p/QmB' }` on `sRB`.
+  - R sends a CircuitRelay message with `{ type: 'STATUS', code: 'OK' }` on `sAR`.
+- phase III: Streams are piped together, establishing a circuit
```
JustinDrake: Where exactly does the piping happen? Does it happen after R sends `{ type: 'STATUS', code: 'OK' }` on `sAR`? If so, is there a race condition here? Should it happen before R sends that message?

relay/README.md
JustinDrake: What is a BAD ERROR? The name "BAD" isn't very descriptive.

relay/README.md
```diff
@@ -1,69 +1,48 @@
 # Circuit Relay
```
JustinDrake: Maybe add the version number (0.1.0)

relay/README.md
```diff
@@ -86,93 +64,144 @@ Scene 2:
...
+As with all other multiaddrs, encapsulation of different protocols determines which metaphorical tubes to connect to each other.
+
+A `/p2p-circuit` circuit address, is formated following:
+
+`[<relay peer multiaddr>]/p2p-circuit/<destination peer multiaddr>`
+
+Examples:
+
+- `/p2p-circuit/p2p/QmVT6GYwjeeAF5TR485Yc58S3xRF5EFsZ5YAF4VcP3URHt` - Arbitrary relay node
+- `/ip4/127.0.0.1/tcp/5002/ipfs/QmdPU7PfRyKehdrP5A3WqmjyD6bhVpU1mLGKppa2FjGDjZ/p2p-circuit/p2p/QmVT6GYwjeeAF5TR485Yc58S3xRF5EFsZ5YAF4VcP3URHt` - Specific relay node
```
JustinDrake: `/ipfs` should be `/p2p`. Same with `/ipfs/QmTwo` => `/p2p/QmTwo`.

relay/README.md
```diff
@@ -86,93 +64,144 @@ Scene 2:
...
-As explained above, the relay is both a transport (`tpt.Transport`)
-and a mounted stream protocol (`p2pnet.StreamHandler`).
-In addition it provides a means of specifying relay nodes to listen/dial through.
+This is a table of status codes and sample messages that may occur during a relay setup. Codes in the 200 range are returned by the relay node. Codes in the 300 range are returned by the destination node.
 
-TODO: the usage of p2pnet.StreamHandler is a little bit off, but it gets the point across.
+
+| Code  | Message                                           | Meaning    |
+| ----- |:--------------------------------------------------|:----------:|
+| 100   | OK                                                | Relay was setup correctly |
```
JustinDrake: Should this be "One leg of the relay was setup correctly", instead of the full relay?

relay/README.md
diasdavid: It's after R sends that `{ type: 'STATUS', code: 'OK' }`. The order in which the items are listed is relevant.

Where is the race condition here?

JustinDrake: The race condition is as follows. `A` may try to send a message for `B` on `sAR` after receiving `{ type: 'STATUS', code: 'OK' }`, but before piping has happened, i.e. before a circuit to `B` is properly formed. In practice, network latency will likely be high enough that piping will happen in time. Still, piping should maybe happen before sending `{ type: 'STATUS', code: 'OK' }` to avoid the (theoretical) race condition.

dryajov: @diasdavid agree with @JustinDrake in the current js implementation `OK` is sent after the pipe has been established otherwise there is no way to tell if the pipe is ready without an additional message that would indicate that.

relay/README.md
dryajov: @diasdavid you're correct, reading a bit more into protobufs clarified this.

relay/README.md
diasdavid: If A writes on sAR and sAR hasn't been piped to sRB then the message is buffered (backpressure).

relay/README.md
diasdavid: This is a very good question, I need it needs to be clarified. 

Dialing from A to R self-confirms because of all the libp2p dance (multistream, secio, protocol muxing), we really just need to signal that we managed to reach the other end (B).

This brings be back to => https://github.com/libp2p/specs/pull/22#discussion_r126312261

relay/README.md
diasdavid: Now a valid point is that the status should be sent before the dial from R to B happens. Going to fix that.

relay/README.md
diasdavid: @dryajov wanna PR that addition?

relay/README.md
diasdavid: Ah! Now that I read the spec again, I've already accounted for this. See the order:

```
  - R sends a CircuitRelay message with `{ type: 'STOP', srcPeer: '/p2p/QmA', dstPeer: '/p2p/QmB' }` on `sRB`.
  - R sends a CircuitRelay message with `{ type: 'STATUS', code: 'OK' }` on `sAR`.=
```

relay/README.md
diasdavid: ✅

relay/README.md
diasdavid: ✅

relay/README.md
diasdavid: ✅

relay/README.md
```diff
@@ -73,106 +52,156 @@ Cast:
...
 - QmOne tries to connect via relaying, because it shares this transport with QmTwo.
 - A lively and prolonged dialogue ensues.
 
-
 ## Addressing
 
-`/p2p-circuit` multiaddrs don't carry any meaning of their own.
-They need to encapsulate a `/p2p` address, or
-be encapsulated in a `/p2p` address, or both.
+`/p2p-circuit` multiaddrs don't carry any meaning of their own. They need to encapsulate a `/p2p` address, or be encapsulated in a `/p2p` address, or both.
```
JustinDrake: How can a `/p2p-circuit` multiaddr be encapsulated in a `/p2p` address without encapsulating a `/p2p` address? In other words, shouldn't the destination `/p2p` address be mandatory?

relay/README.md
dryajov: @diasdavid whats the benefit of sending `OK` before the pipe is established?

JustinDrake: > If A writes on sAR and sAR hasn't been piped to sRB then the message is buffered (backpressure).

Backpressure may not completely address the race condition. The reason is that the recipient of messages sent to the `sAR` stream could be either `R` or `B`. As it stands the protocol is unambiguous because after `A` sends the initial `HOP` message to `R` it is not expected to send further messages to `R`. But in theory the protocol could be extended to allow for `A` to send further messages to `R` before the circuit is established. Doing things the other way round may be more futureproof.

relay/README.md
diasdavid: The wording might not be perfect. Could you illustrate the case that doesn't make sense to you?

relay/README.md
dryajov: @diasdavid sure thing.

relay/README.md
dryajov: Looking at this closer, I see another issue that might come up with multihop dialing. In shallow/matreshka dialing mode, there is no way for the `R` to tell weather the circuit has been completed or not until the last portion of the multihop address is dialed, for that reason I have the `STOP` handler signaling the `OK` rather than `HOP`.

relay/README.md
```diff
@@ -210,5 +237,11 @@ It lives in the go-libp2p package and is named `/ipfs/relay/line/0.1.0`.
...
 
-Since the existing protocol is incomplete, insecure, and certainly not used,
-we can safely remove it.
+Since the existing protocol is incomplete, insecure, and certainly not used, we can safely remove it.
+
+## Future work
+
+We have considered more features but won't be adding them on the first iteration of Circuit Relay, the features are:
+
+- Multihop relay - With this specification, we are only enabling single hop relays to exist. Multihop relay will come at a later stage as Packet Switching.
```
dryajov: The js implementation does implement multihop dialing in the form of `shallot/matreshka` dialing. This was based on the conversations we had at the time of me starting the implementation, what has changed that we now have scrapped this part?

relay/README.md
JustinDrake: The case that is not allowed is `<relay peer multiaddr>/p2p-circuit/` (destination omitted). What about the following wording?

> A `/p2p-circuit` multiaddress needs to encapsulate a `/p2p` destination address, and optionally be encapsulated by a `/p2p` relay address.

relay/README.md
dryajov: @diasdavid #23 - here it is

relay/README.md
diasdavid: There is nothing wrong with experimentation. It was part of Future Work spec wise from the beginning.

dryajov: Makes sense. :)

relay/README.md
dryajov: @lgierth @whyrusleeping @diasdavid wanna chime in on this one? Seems like the only contention point right now? I agree with the points that @JustinDrake is bringing up.

relay/README.md
dryajov: I think this can be added as additional bullet points to the main sentense:

> `/p2p-circuit` multiaddrs don't carry any meaning of their own. They need to encapsulate a `/p2p` address, or be encapsulated in a `/p2p` address, or both.
> - /p2p-circuit multiaddress encapsulates a /p2p destination address - e.g. `/p2p-circuit/p2p/QmHash`
> - can optionally be encapsulated by a /p2p relay address - e.g. `/p2p/ipfs/QmRelay/p2p-circuit/p2p/QmDest`

relay/README.md
whyrusleeping: i'm not sure i see an issue here, If `R` sends the message to `A`, then pipes the connection, any message sent from A to R will just be buffered (or A will not be able to finish the write) until things are piped. It is assumed we are operating over a reliable ordered stream (not something like UDP where packets could get dropped if we don't read them)

dryajov: @whyrusleeping I guess you're right. I remember dealing with some issues initially, which made me experiment a bit with we're the ok is sent, but most likely it was unrelated to the ordering of the messages. I'll give it another go sinse I'll have to rewrite message handling anyways.

diasdavid:
From IRC:

```
06:06 <@daviddias> Seems that other than some miner typos, there are no outstanding issues in the relay spec
06:07 <@daviddias> Shall we merge it today?
06:07 <@whyrusleeping> sounds good to me
```

# 📣 If no blockers appear, I shall merge the spec today 📯

JustinDrake:
@diasdavid Are you going to address https://github.com/libp2p/specs/pull/22#discussion_r126303619 before merging?

relay/README.md
diasdavid: It's acknowledged in the sense that:

- a connection is open
- a multistream + secio + spdy + multistream again to /libp2p/circuit/relay is done

The next ack is when R finishes sRB

diasdavid:
@JustinDrake addressed last Wednesday

diasdavid:
All right, this is it, no one has raised any other blockers, we can say that this spec is **COMPLETE**

Issue #27, opened 2017-07-16
Proposal: Add a `CAN_HOP` message Type to circuit-relay to allow passive relay discovery


dryajov:
Having separate multistream endpoints for `hop` and `stop` allowed to quickly determine if a peer had relay capabilities, with the removal of the separate multistream endpoints this isn't possible anymore. 

@vyzo and myself discussed the possibility of adding a new message type that would allow determining if a host is a relay or isn't. This is crucial functionality as it would otherwise require to resort to ugly hacks to know which hosts are relays and which aren't. 

A couple I can think of:

- Send a `HOP` message with an invalid peer id (or somesuch) and check for an explicit `CANT_SPEAK_RELAY` response code, if its something else then assume it is a relay. 
- A (arguably) worst way would be to simply dial every peer in the peerbook asking to relay to the desired destination, this is both slow and generates unnecessary load on the network. 

In the current js implementation, I was _trying_ over the `hop` endpoint on every incoming connection and adding it to an internal relay table that I would use when dialing over a relay address. This is IMO more efficient and cleaner.

This proposal tries to address this shortcomings by adding an explicit `CAN_HOP` message that would allow quickly asking a peer if it is a relay. 

The `CAN_HOP` message either succeeds or fails and doesn't provide any additional information, leaving further capability discovery functionality to the already proposed `ls` protocol - https://github.com/ipfs/notes/issues/237. 

@diasdavid @whyrusleeping @lgierth

dryajov:
This PR adds the `CAN_HOP` message - https://github.com/libp2p/specs/pull/28

vyzo:
I think it's very important to have this capability, relay functionality will be seriously hampered without the means to discover hop relays. And I don't like having to resort to hacks either, so I'll add it to the Go implementation.

dryajov:
@vyzo I've also added this to the js implementation.

diasdavid:
Merged https://github.com/libp2p/specs/pull/28#pullrequestreview-71027537 

Thank you for coming up with the proposal and explaining the rationale :)

PR #28, opened 2017-07-16
feat: adding an explicit CAN_HOP message type to allow passive relay …

dryajov:
feat: adding an explicit CAN_HOP message type to allow passive relay discovery

PR addressing proposal - https://github.com/libp2p/specs/issues/27

### DIFF
--- a/relay/README.md
+++ b/relay/README.md
@@ -143,6 +143,7 @@ message CircuitRelay {
     HOP = 1;
     STOP = 2;
     STATUS = 3;
+    CAN_HOP = 4; // is peer a relay?
   }
 
   message Peer {


dryajov:
@diasdavid @Stebalien @vyzo can we get this merged? This is already part of the implementation, but we're missing it from the spec :(

relay/README.md
```diff
@@ -143,6 +143,7 @@ message CircuitRelay {
     HOP = 1;
     STOP = 2;
     STATUS = 3;
+    CAN_HOP = 4; // is peer a relay?
```
diasdavid: When is this used? Won't the multistream handshake fail on the relay protocol eitherway?

Stebalien: IIRC, this allows one to distinguish between a peer that simply accepts relayed connections and a peer that is willing to act as a relay.

dryajov: Here is the background of this change - https://github.com/libp2p/specs/issues/27. This was something that both @vyzo and myself thought is _important_ to have, that would otherwise complicate the code base.

diasdavid: Sounds good. I understand where the need comes from (since we stopped using multicodecs to signal if nodes are hops or stops 🙏🏽).

PR #35, opened 2017-10-18
Update README.md

kaotisk-hund:
Removed tl;tr typo in first quote


Stebalien:
Not sure it was a typo but I agree it's better without it.

Stebalien:
Thanks!

PR #48, opened 2018-04-18
Update old addresses

camelmasa:
I just updated old links.


lgierth:
Thank you, much appreciated!

PR #54, opened 2018-05-14
Add details on PubSub

jamesray1:


jamesray1:
You're welcome! By way of introduction, I'm interested in implementing rust-libp2p for developing a sharding network protocol for Ethereum, with [Drops of Diamond](https://github.com/Drops-of-Diamond/diamond_drops).

PR #56, opened 2018-05-21
RFC: rendezvous

victorb:
Only thing I can see missing is adding the `E_UNAVAILABLE` which will be used when registrations are happening too fast or it's generally overloaded with the already registered clients.

Previous: https://github.com/libp2p/specs/pull/44 (mostly doesn't load)

### DIFF
--- a/rendezvous/README.md
+++ b/rendezvous/README.md
@@ -157,6 +157,7 @@ message Message {
     E_INVALID_COOKIE    = 103;
     E_NOT_AUTHORIZED    = 200;
     E_INTERNAL_ERROR    = 300;
+    E_UNAVAILABLE       = 400;
   }
 
   message PeerInfo {
--- a/rendezvous/README.md
+++ b/rendezvous/README.md
@@ -144,6 +144,16 @@ R -> D: {[REGISTER{my-app, {QmE, AddrE}}],
          c3}
 ```
 
+### Proof of Work
+
+The protocol as described so far is susceptible to spam attacks from
+adversarial actors who generate a large number of peer identities and
+register under a namespace of interest (eg: the relay namespace). This
+can be mitigated by requiring a Proof of Work scheme for client
+registrations.
+
+This is TBD before finalizing the spec.
+
 ### Protobuf
 
 ```protobuf
--- a/rendezvous/README.md
+++ b/rendezvous/README.md
@@ -1,5 +1,9 @@
 # Rendezvous Protocol
 
+Author: vyzo
+
+Revision: DRAFT; 2019-01-18
+
 The protocol described in this specification is intended to provide a
 lightweight mechanism for generalized peer discovery. It can be used
 for bootstrap purposes, real time peer discovery, application specific
--- a/rendezvous/README.md
+++ b/rendezvous/README.md
@@ -87,12 +87,15 @@ Clients `A` and `B` connect to the rendezvous point `R` and register for namespa
 
 ```
 A -> R: REGISTER{my-app, {QmA, AddrA}}
+R -> A: {OK}
 B -> R: REGISTER{my-app, {QmB, AddrB}}
+R -> B: {OK}
 ```
 
 Client `C` connects and registers for namespace `another-app`:
 ```
 C -> R: REGISTER{another-app, {QmC, AddrC}}
+R -> C: {OK}
 ```
 
 Another client `D` can discover peers in `my-app` by sending a `DISCOVER` message; the
@@ -123,6 +126,7 @@ query that discovers just that peer by including the timestamp:
 
 ```
 E -> R: REGISTER{my-app, {QmE, AddrE}}
+R -> E: {OK}
 D -> R: DISCOVER{ns: my-app, since: t1}
 R -> D: {[REGISTER{my-app, {QmE, AddrE}}],
          t3}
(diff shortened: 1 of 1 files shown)


dryajov:
We should make sure that rendezvous doesn't allow registering arbitrary peers under a namespace. Basically, the peer id in the register request should match the peer id of the peer initiating the request, otherwise this is a potential DDoS vector. I know we're communicating the peer id in our register requests, so this might be already covered, but its worth pointing out.

whyrusleeping:
Great point, care to suggest wording in the spec to clarify this?

rendezvous/README.md
```diff
@@ -0,0 +1,201 @@
...
+real-time distribution, while still providing a simple interface to
+clients.
+
+
+## The Protocol
+
+The rendezvous protocol provides facilities for real-time peer
+discovery within application specific namespaces. Peers connect to the
+rendezvous point and register their presence in one or more
+namespaces.
```
dryajov: As suggested in https://github.com/libp2p/specs/pull/56#issuecomment-393022342, here is some possibly wording to spell this out:

No arbitrary peers are allowed to be registered under a namespace, only the peer initiating the register request is allowed to register itself.

dirkmc:
I would suggest using the word `cursor` instead of `cookie`

jacobheun:
With the addition of an `E_UNAVAILABLE ` response and the recommended language changes mentioned above I think this looks good and would like to get this merged in. 

I've added a couple more core members of libp2p for review so we can move this forward.

vyzo:
Another point that was discussed in Glasgow was the addition of a Proof of Work function for registration.
We want to deter trivial attacks where an attack launches 1M nodes and registers them in (say) the relay namespace, intending to censor or mutilate client performance with junk nodes.

A PoW function (like equihash) would add some minimum work/time before the registration is accepted, significantly limiting the damage a single attacker can do without committing considerable resources.

vyzo:
Also discussed in Glasgow -- the registration function, apart from PoW, should return the actual TTL of the registration so that a background process can re-register when appropriate.

jacobheun:
For the PoW function, we would apply that to every registration? So if an app wanted to register to multiple namespaces they would require a PoW for each and each subsequent re-registration on TTL expiration?

Would it be valuable to add some concept of limited multi registration in combination with the PoW function to balance this out?

vasco-santos:
Except from @dirkmc comment regarding the cookie keyword, the current state of the spec looks good to me. 

Moreover, totally agree with the addition of the PoW function. And subscribe @jacobheun questions about finding a balance to this function.

rendezvous/README.md
```diff
@@ -0,0 +1,201 @@
...
+  rendezvous can be used to discover pubsub peers within a topic.
+- In a real time setting, applications can poll rendezvous points in
+  order to discover new peers in a timely fashion.
+- In an application specific routing setting, rendezvous points can be
+  used to progressively discover peers that can answer specific queries
+  or host shards of content.
+
+### Replacing ws-star-rendezvous
+
+We intend to replace ws-star-rendezvous with a few rendezvous daemons
```
bigs: first few sentences here are a bit redundant. furthermore, what mechanisms allow this to replace `ws-star-rendezvous`? good to have that up here

rendezvous/README.md
```diff
@@ -0,0 +1,201 @@
...
+time peer discovery by registering and polling rendezvous points.
+This allows us to replace a fragile centralized component with a
+horizontally scalable ensemble of daemons.
+
+### Rendezvous and pubsub
+
+Rendezvous can be naturally combined with pubsub for effective
+real-time discovery.  At a basic level, rendezvous can be used to
+bootstrap pubsub: nodes can utilize rendezvous in order to discover
+their peers within a topic.  Contrariwise, pubsub can also be used as
```
bigs: -Contrariwise +Alternatively

aarshkshah1992:
Hey guys,

This looks like a great feature to have as there is no 'standard' mechanism for discovery when building a P2P networking using libp2p. Does anyone know when this will be merged ?

daviddias:
Where can we find impl status?

chr15m:
With regards to the proof-of-work function, there are questions above about whether to make the difficulty configurable and what the default proof-of-work required for registration should be.

You might consider the "hashcash auction" technique to allow the clients to determine the difficulty of the proof-of-work amongst themselves in a bottom-up, decentralized way:

https://chr15m.github.io/hashcash-auctions-for-decentralized-resource-allocation.html

You don't need to read the post to understand the basic idea which is as follows:

 * Each rendezvous server would set an upper limit on the amount of messages & connections per unit time it can process (i.e. `R` registrations per hour). This can be estimated based on CPU, memory, bandwidth resources available or set at some sensible easily serviceable limit.
 * When a client wants to connect and register it must perform a proof-of-work.
 * The server publishes the current highest, lowest, and median hash power achieved for each registration by other clients that it currently holds active.
 * A registration slot will be granted based on the amount of proof-of-work the client supplies. This is done by sorting all registrations by the hash power each achieved and slicing the array of registrations so that only the top (`R`) are granted the ability to register.

There are three beneficial outcomes from this scheme:

 * As with traditional Hashcash there is basic spam prevention as DoS attacks now incur a literal CPU (and therefore energy) cost. The energy cost is proportional to the size of the DoS effort so it is somewhat self defeating - the harder they DoS the more expensive they make it.
 * Legitimate users now have a mechanism to out perform a DoS attack and continue to use the service rather than the normal failure mode where everybody is locked out.
 * Because a client will know of multiple registration servers and their relative proof-of-work requirements (which will be proportional to their load) clients will naturally head for registration servers with lower proof-of-work and therefore lower load. This creates a bottom-up load balancing mechanism without requiring any centralized coordination between registration servers.

Stebalien:
There are two things to keep in mind when considering proof of work:

1. When it comes to clients, hardware matters. An PoW that's expensive on one device can be virtually free on a another. This can be a serious problem for mobile users.
2. When it comes to servers, we're basically asking people to run these servers as a public service. Given that we have to be careful about not disincentivizing participation by making the service costly to run (in terms of power).

Really, I'm more worried about the first issue. The second issue will always come down to the fact that honest nodes will need to spend enough to overcome malicious nodes.

---

@chr15m my worry with your proposal is 1. It would be pretty easy to completely lock-out mobile users by driving the PoW cost up.

chr15m:
@Stebalien yes those are important concerns.

Under the scheme I outlined the server does not perform any proof-of-work. It is specifically design to protect the people you are worried about: those running servers as a public service. The idea is to allow such resource volunteers certainty that their offering can't be abused by anonymous clients. The clients are the ones who must perform proof-of-work.

With regards to 1, @vyzo mentioned Equihash which is memory-hard rather than CPU-hard and possibly therefore friendlier to mobile phones. Scrypt is another alternative with good pure-JS presence in browsers.

For this particular use-case where most registration servers will have spare capacity it should not be a big issue since the PoW will be negligible for all clients anyway. Only in the situation of massive DoS of multiple registration servers will mobile clients find a harder time registering than higher powered CPU machines and that situation would hopefully be extraordinary in any case.

Stebalien:
I'm not really a fan of memory hard problems (at least for use cases like this):

1. It's easier to share CPU time than it is to share memory. That is, if I frequently need a large chunk of memory but don't *always* need it, it still needs to be *available* most of the time. Basically, if we use a memory hard problem, the libp2p instance will need to hoard the memory it needs.
2. I'm not convinced of the overall security of multiple memory hard problems that can be computed in parallel. The hard part about memory is usually *latency* not *throughput*. However, when you just need to compute a bunch of memory hard problems in parallel and care more about how many problems you finish, not how long it takes to finish each one, it's throughput that matters.

chr15m:
@Stebalien yes good points. Of course, using plain PoW under a fixed/configurable difficulty setting also suffers from the issues you have mentioned.

Of course the real question is whether using PoW provides better protection and features than no PoW at all, and further, whether using the "hashcash auction" scheme outlined provides better protection and features than plain fixed difficulty PoW.

rendezvous/README.md
vyzo: the very act of separating discovery from relay, so that we can replace `ws-star-rendezvous` with a few rendezvous daemons and a fleet of circuit relays...
It's already there.

rendezvous/README.md
vyzo: done

rendezvous/README.md
vyzo: done.

vyzo:
Changes:
- addressed @dryajov's request for explicit wording prohibiting arbitrary peer registration
- added `E_UNAVAILABLE` error for temporary availability errors
- added the actual ttl in the registration response
- added section (TBD) for proof of work

PR #68, opened 2018-06-06
Add implementations and definitions of floodsub and gossipsub

jamesray1:


bigs:
I think these blurbs are likely to go stale quite quickly, but I do think it's important that we get some specs for Floodsub and Gossipsub in this repo and reference them from this doc. Individual docs could elaborate on specifics of each protocol and point to implementations.

jamesray1:
Agreed, but it's better to update now (and update again when necessary, or at least it will be outdated but still provide more information than otherwise).

jamesray1:
I still think the technical details on floodsub and pubsub will not go out of date, even if they are replaced by another later on.

lgierth:
Is gossipsub what's called epidemic-pubsub in libp2p/research-pubsub#17? Anyway, there's another impl of gossipsub in gerbil: https://github.com/vyzo/gerbil-simsub

jamesray1:
Thanks, I added a link for the gerbil impl. See also https://github.com/libp2p/go-floodsub/issues/77#issuecomment-387404293 and https://en.wikipedia.org/wiki/Gossip_protocol.

> A gossip protocol[1] is a procedure or process of computer-computer communication that is based on the way social networks disseminate information or how epidemics spread.

So based on that, but not having read https://github.com/libp2p/research-pubsub/pull/17 yet, I would say that gossipsub can also be referred to as epidemic-pubsub.

lgierth:
> So based on that

I wasn't asking based on definitions ;) We have two implementations which we call gossipsub, and we have a spec that we call epidemic-pubsub, and I was wondering whether they're the same -- they are.

lgierth:
> I was wondering whether they're the same -- they are.

I'm gonna merge that spec by tomorrow, and sprinkle in a couple of mentions of "gossipsub".

lgierth:
@jamesray1 hey the gossipsub spec is now merged, and I updated links wherever it seemed appropriate. Does that look good to you now?

PR #73, opened 2018-06-28
gossipsub: work draft 1 into the specs 

lgierth:
This is taken from libp2p/research-pubsub#17, with a few logistical changes that are all contained in the last commit.

It's fine that this isn't a perfect spec document yet -- not letting important information rot in issues and PRs is one step towards better documentation :)


lgierth:
cc @vasco-santos

PR #76, opened 2018-07-05
eagler -> eagerer

jamesray1:


jamesray1:
Whoops, should've noted that more changes were coming!

PR #77, opened 2018-07-05
Grammar

jamesray1:
Remove unnecessary line breaks;
eg => e.g., ttl -> TTL, followin - > following + more
on the same time -> at the same time
for all by the first source -> for all but the first source
C_rand `JOIN` message => `C_rand` `JOIN` messages
in world full => in a world full


pubsub/gossipsub/README.md
```diff
@@ -424,7 +423,7 @@ GETNODES {}
 
 NODES {
  peers []peer.ID
- ttl int
+ TTL int
```
Stebalien: These should probably be lower case for consistency.

jamesray1: Ah sorry, that was from a Find and replace, I forgot about those.

PR #80, opened 2018-07-23
[RFC] peer discovery with mDNS

richardschneider:
Background discussion at [Future of mDNS discovery](https://github.com/libp2p/libp2p/issues/28)

### DIFF
--- a/discovery/mdns.md
+++ b/discovery/mdns.md
@@ -92,4 +92,36 @@ peer's discovery details. To accomodate this, a peer should respond to the follo
 
 ## Worked Example
 
-**TODO**
+Asumming that `peer-id` is `QmQusTXc1Z9C1mzxsqC9ZTFXCgSkpBRGgW4Jk2QYHxKE22`.  Then the `peer-name` is `ciqcmoputolsfsigvm7nx5fwkko2eq26h46qhbj6o4co7uyn2f2srdy`.
+
+To make the examples more readable `id` and `name` are used.
+
+
+### Meta Query
+
+Find all services on the local network.
+
+#### Question
+
+`_services._dns-sd._udp.local PTR`
+
+#### Answer
+
+_services._dns-sd._udp.local IN PTR _p2p._udp.local
+
+### Find All Peers
+
+Find all peers on the local network
+
+#### Question
+
+_p2p._udp.local PTR
+
+#### Answer
+
+_p2p._udp.local IN PTR `name`._p2p._udp.local
+
+#### Additional Records
+
+- `name`._p2p._udp.local IN TXT dnsaddr=/ip6/fe80::7573:b0a8:46b0:bfea/tcp/4001/ipfs/`id`
+- `name`._p2p._udp.local IN TXT dnsaddr=/ip4/192.168.178.21/tcp/4001/ipfs/'id'
--- a/discovery/mdns.md
+++ b/discovery/mdns.md
@@ -1,55 +1,51 @@
-# Multicast DNS
+# Multicast DNS (mDNS)
 
 Author: Richard Schneider (makaretu@gmail.com)
 
 ## Overview
 
-The goal is to allow peers to discover each other when on the same local network with zero configuration. MDNS uses a multicast system of DNS records; this allows all peers on the local network to see all query responses.
+The goal is to allow peers to discover each other when on the same local network with zero configuration. mDNS uses a multicast system of DNS records; this allows all peers on the local network to see all query responses.
 
-Conceptually, it is very simple.  When a peer starts (or detects a network change), it sends a query for all peers. As responses come in, the peer adds the other peers information into is local database of peers.
+Conceptually, it is very simple. When a peer starts (or detects a network change), it sends a query for all peers. As responses come in, the peer adds the other peers' information into its local database of peers.
 
 ## Definitions
 
-- `service-name` is the DNS-SD service name for all peers. It is defined as `_p2p._udp.local`.
-- `host-name` is the fully qualified name of the peer.  It is derived from the peer's name and `p2p.local`.
-- `peer-name` is the case-insenstive unique identifier of the peer and less than 64 characters.  
-
-It is normally the base-32 encoding of peer's ID. 
-
-If the encoding of the peer's ID exceeds 63 characters, then the [Split at 63rd character](https://github.com/ipfs/in-web-browsers/issues/89#issue-341357014) workaround can be used.
+- `service-name` is the DNS Service Discovery (DNS-SD) service name for all peers. It is defined as `_p2p._udp.local`.
+- `host-name` is the fully qualified name of the peer. It is derived from the peer's name and `p2p.local`.
+- `peer-name` is the case-insensitive unique identifier of the peer, and is less than 64 characters. It is normally the base-32 encoding of peer's ID.
+  - If the encoding of the peer's ID exceeds 63 characters, then the [Split at 63rd character](https://github.com/ipfs/in-web-browsers/issues/89#issue-341357014) workaround can be used.
 
 ## Peer Discovery
 
 ### Request
 
-To find all peers, a DNS message is sent with the question `_p2p._udp.local PTR`. Peers will then start responding with their details.  
+To find all peers, a DNS message is sent with the question `_p2p._udp.local PTR`. Peers will then start responding with their details.
 
-Note that a peer must respond to it's own query.  This allows other peers to passively discover it.
+Note that a peer must respond to its own query. This allows other peers to passively discover it.
 
 ### Response
 
-On receipt of a `find all peers` query, a peer sends a DNS response message (QR = 1) that contains
-the **answer**
+On receipt of a `find all peers` query, a peer sends a DNS response message (QR = 1) that contains the **answer**
 
 ```
 <service-name> PTR <peer-name>.<service-name>
 ```
 
-The **additional records** of the response contain the peer's discovery details
+The **additional records** of the response contain the peer's discovery details:
 
 ```
 <peer-name>.<service-name> TXT "dnsaddr=..."
 ```
 
-The TXT record contains the multiaddresses that the peer is listening on.  Each multiaddress is a TXT attribute with the form `dnsaddr=.../p2p/QmId`.  Multiple `dnsaddr` attributes and/or TXT records are allowed.
+The TXT record contains the multiaddresses that the peer is listening on. Each multiaddress is a TXT attribute with the form `dnsaddr=/.../p2p/QmId`. Multiple `dnsaddr` attributes and/or TXT records are allowed.
 
 ## DNS Service Discovery
 
-DNS-SD support is not needed for peers to discover each other.  However, it is extremely usefull for network administrators to discover what is running on the network.
+DNS-SD support is not needed for peers to discover each other. However, it is extremely useful for network administrators to discover what is running on the network.
 
 ### Meta Query
 
-This allows discovery of all services.  The question is `_services._dns-sd._udp.local PTR`.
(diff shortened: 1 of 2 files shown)


discovery/mdns.md
```diff
@@ -0,0 +1,94 @@
...
+
+`peer-id` is the ID of the peer.  It normally is the base-58 enconding of the hash of the peer's public key.
+
+`port` is a port that the peer listens on. Normally 4001.
+
+## Peer Discovery
+
+### Request
+
+To find all peers, a DNS message is sent with the question `_ipfs._udp.local PTR`.
```
Stebalien: We'll probably want to announce multiple services. We should at least default to announcing `_libp2p` (and only announce `_ipfs` if configured to do so).

discovery/mdns.md
```diff
@@ -0,0 +1,94 @@
...
+    
+The **additional records** of the response contain the peer's discovery details
+
+    <peer-id>.<service-name> SRV ... <port> <host-name>
+    <peer-id>.<service-name> TXT ...
+    <host-name>              A <ipv4 address>
+    <host-name>              AAAA <ipv6 address>
+   
+
+Multiple A and AAAA records are expected. The `TXT` is not needed for IPFS peer discovery, but is required by DNS-SD.
```
Stebalien: So, we'll need it for multiaddrs. We're able to get away with this for now but not once we enable QUIC by default. Ideally, we'd *only* use TXT records. However, we have to supply an A or AAAA record and also have to supply a port. Personally, I'm all for using our TCP port and real IP addresses when announcing but mostly ignoring them when accepting (unless we get *no* TXT records in which case we can *try* the IP/PORT from the other records).

Note: We'd probably want to use `dnsaddr=/ip4/.../tcp/...` as the format for TXT records.

richardschneider: TXT record `dnsaddr=/full/address` is a problem.  Strings must be less than **128 bytes**.
 
`dnsaddr=/ip6/2406:e001:1467:1:ecec:9ca6:XXXX:XXXX/tcp/4009/ipfs/QmTp1fwYnJxc5rUEVM2xowcDiyR61J7yvmBRUQzT4g8MHs` almost reaches this limit.  If the peer-id hashing algorithm changes to sha2-512 then the byte limit is exceeded.

I strongly suggest that TXT dnsaddr is not used.

discovery/mdns.md
richardschneider: I'm happy with whatever the community decides for the name.  My preference is `_p2p`.

discovery/mdns.md
richardschneider: Also, we are just trying to discover peers on the local link. So to my thinking an IP address, port and peer ID  is all that is needed.

whyrusleeping: Can we send multiple records?

Stebalien: Well, we also have protocols like QUIC and may add other random protocols in the future. We also *do* need to put the peer ID somewhere.

It looks like the standard way to do this is to break it into multiple records. You can see how DKIM does it here: https://serverfault.com/questions/255580/how-do-i-enter-a-strong-long-dkim-key-into-dns

We could also use a base64 encoded binary multiaddr but I'd rather not.

discovery/mdns.md
Stebalien: SGTM.

richardschneider:
@Stebalien 
> We also do need to put the peer ID somewhere.

The peer id is the first label of the SRV record's name.

richardschneider:
I'm not sure that the spec on TXT record allows multiple strings with the same property name.  I know for certain that most "mdns browsers" only show the first/last value,

I'll dig into the specs and report back.

discovery/mdns.md
```diff
@@ -0,0 +1,102 @@
...
+
+The goal is to allow peers to discover each other when on the same local network with zero configuration. 
+MDNS uses a multicast system of DNS records; this allows all peers on the local network to see all query responses.
+
+Conceptually, it is very simple.  When a peer starts (or detects a network change), it sends a query for all peers. 
+As responses come in, the peer adds the other peers information into is local database of peers.
+
+## Definitions
+
+`service-name` is the DNS-SD service name for all IPFS peers. It is defined as `_ipfs._udp.local`.
```
gpestana: suggestion: DNS-SD -> DNS service discovery (DNS-SD)

richardschneider:
@Stebalien I hope all you concerns have been answered, `dnsaddr=...` is now part of the spec.

If you're happy, I start on the worked examples.

richardschneider:
Should we specifically exclude loopback addresses (127.0.0.1 and ::1) from being discovered?

Stebalien:
> Should we specifically exclude loopback addresses (127.0.0.1 and ::1) from being discovered?

No, they're actually quite useful for finding multiple nodes on the same machine. However, we *do* want to avoid advertising addresses on the wrong interfaces. We should also filter incoming advertisements but that gets a bit tricker. At a minimum, we could filter incoming localhost advertisements to localhost.

discovery/mdns.md
```diff
@@ -14,12 +14,11 @@ As responses come in, the peer adds the other peers information into is local da
 
 `service-name` is the DNS-SD service name for all peers. It is defined as `_p2p._udp.local`.
 
-`host-name` is the name of the peer.  It is derived from the peer's ID and `p2p.local`, for example 
-`Qmid.p2p.local`.
+`host-name` is the fully qualified name of the peer.  It is derived from the peer's name and `p2p.local`.
 
-`peer-id` is the ID of the peer.  It normally is the base-58 enconding of the hash of the peer's public key.
+`peer-name` is the case-insenstive ID of the peer and less than 64 characters.  It is normally is the base-32 encoding of peer's ID.
```
lidel: I guess until https://github.com/ipfs/go-ipfs/issues/5287 is solved this would this require a manual step of converting back to Base58, yes?

richardschneider: > I guess until ipfs/go-ipfs#5287 is solved this would require apps to do a manual step of converting back to Base58, yes?

`peer-name` is just a unique name, we should not assume it's the peer id. The `dnsaddr` ends with the peer ID.

> What if PeerID is longer than 64 characters? Should we split after 63 characters like noted in ipfs/in-web-browsers#89 ?

I think **Workaround A: Split at 63rd character** makes sense.

discovery/mdns.md
```diff
@@ -0,0 +1,93 @@
...
+Conceptually, it is very simple.  When a peer starts (or detects a network change), it sends a query for all peers. 
+As responses come in, the peer adds the other peers information into is local database of peers.
+
+## Definitions
+
+`service-name` is the DNS-SD service name for all peers. It is defined as `_p2p._udp.local`.
+
+`host-name` is the fully qualified name of the peer.  It is derived from the peer's name and `p2p.local`.
+
+`peer-name` is the case-insenstive ID of the peer and less than 64 characters.  It is normally is the base-32 encoding of peer's ID.
```
Kubuxu: This limits PeerID to 320bits which might become a problem if ever we need to switch to 512bit hash function (just mentioning for you to consider).

lidel: This is similar problem we have with using CIDs in subdomains.

One of proposed workarounds in https://github.com/ipfs/in-web-browsers/issues/89#issue-341357014 is to "Split at 63rd character"

richardschneider: Good idea. I'll make the changes.

Kubuxu: It might be useful to create a separate doc for DNS encoding of CIDs. As it is already present in two places.

discovery/mdns.md
```diff
@@ -0,0 +1,93 @@
...
+extremely usefull for network administrators to discover what is running on the 
+network.
+
+### Meta Query
+
+This allows discovery of all services.  The question is `_services._dns-sd._udp.local PTR`.
+
+A peer responds with the answer
+
+    _services._dns-sd._udp.local PTR <service-name>
```
vyzo: missing underscore in the end of line.

discovery/mdns.md
vyzo: Is there a technical reason to limit the peer id size?

lidel: @vyzo  IIUC we need to comply with length limits from  [RFC 1034](https://tools.ietf.org/html/rfc1034#page-7):

> Each node has a label, which is zero to 63 octets in length.  [#](https://tools.ietf.org/html/rfc1034#page-7)

> To simplify implementations, the total number of octets that represent a
domain name (i.e., the sum of all label octets and label lengths) is
limited to 255. [#](https://tools.ietf.org/html/rfc1034#page-8)

CIDv1(sha-256) in Base32 fits, but hashes with longer outputs can't be used as a single label and need to be split.

tomaka:
Ideally a link to theses specs should be added in `4-architecture.md`.

discovery/mdns.md
richardschneider: @vasco-santos The peer-name is used as a DNS label.  A DNS label must be case insensitive, less than 63 characters and US-ASCII.

richardschneider:
@Stebalien @diasdavid  What is needed to get some approval and then merge it?

diasdavid:
@richardschneider it would be great to give it a proper review (I can do it probably later next week). One strong case to merge would be if the implementations were both done (probably they are already) and that interoperability has been tested.

tomaka:
> @richardschneider it would be great to give it a proper review (I can do it probably later next week). One strong case to merge would be if the implementations were both done (probably they are already) and that interoperability has been tested.

Right now the latest `ipfs/go-ipfs:master` Docker image doesn't seem to be conformant at all.
Sniffing traffic with Wireshark, it seems to generate two queries in a row, with the service name `_ipfs-discovery._udp.local`, and never seems to generate any answer.

richardschneider:
@tomaka Yes, go-ipfs mdns has been broken for a long time.  Background discussion at [Future of mDNS discovery](https://github.com/libp2p/libp2p/issues/28).

The "novel" idea here, is that we first write the spec and then implement it!

jacobheun:
The state of this spec looks reasonable to me for us to get this merged in and marked as Draft. rust-libp2p looks to already have implemented this. It would be ideal to get our MDNS implementations in go and js to an interoperable and better state.

richardschneider:
How should a private network interact with MDNS?  I think the `service-name` should change from `_p2p._udp.local` to **_p2p-*x*.udp.local** , where *x* is the fingerprint of the private network in base16.

discovery/mdns.md
raulk: Addressed.

raulk:
In the interest of advancing this, I'm listing TODOs based on above comments:

- [ ] support loopback addresses to enable local process discovery.
- [ ] context-sensitive address advertisement based on network interface.
- [ ] private network discovery.

@richardschneider – if you agree on points 1 & 2, do you have enough bandwidth to enhance the spec?

Re: private networks, SGTM unless we want to hide private network membership somehow. If that was the case, you wouldn't be using mDNS at all.

Should we enforce that nodes belonging to a pnet only respond to queries for that pnet?

richardschneider:
@raulk 

- The spec does not prevent loopback, it was just a question.
- I think we should postpone "context-sensitive address advertisement based on network interface". My feeling is that a node can filter on TXT dnsaddr already (like swarm filters)
- By using a private network  specific `service-name` we get for free that nodes belonging to a pnet only respond to queries for that pnet.

It's midnight my time, I'll update the doc for a private network tomorrow.

raulk:
@richardschneider that's great, thanks!

richardschneider:
@raulk  IMHO this is ready to go.

Stebalien:
(i.e., let's just iterate in new PRs so we have something we can work with _now_)

Issue #83, opened 2018-08-02
The current Gossipsub spec does not correspond to the current implementation of `GossipSubRouter` in `go-floodsub`


mhchia:
We have two versions of pubsub router now, [`FloodSubRouter`](https://github.com/libp2p/go-floodsub/blob/master/floodsub.go#L30) and [`GossipSubRouter`](https://github.com/libp2p/go-floodsub/blob/master/gossipsub.go#L57) respectively. It is clear that `FloodSubRouter` just floods messages. However, the current `GossipSubRouter` does not perform the same way as the [Gossipsub spec](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/README.md) described. It is more like performing normal gossiping on each topic. Its algorithm is quite the same as the protocol in [gerbil-simsub](https://github.com/vyzo/gerbil-simsub#the-gossipsub-protocol). For me, it is quite confusing that the implementation is a lot different from the spec.

IMO the name "gossipsub" is suitable for the current implementation of `GossipSubRouter` in the repo `go-floodsub`, i.e. the gerbil-simsub's protocol. And the one described in the [Gossipsub spec](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/README.md) should be the spec for "Proximity Aware Epidemic PubSub". So it will be clearer if we have specs for "gossipsub" and "Proximity Aware Epidemic PubSub" respectively.

Is there a plan already that the current gossipsub implementation will be used, or it will be changed to "Proximity Aware Epidemic PubSub" soon? It is actually discussed in https://github.com/libp2p/libp2p/issues/33#issuecomment-399171178 and I really appreciate the answer from @whyrusleeping. Just want to confirm again, since "Proximity Aware Epidemic PubSub" seems to be more experimental, for its broadcast tree, random walks for joining, and passive view exchange. In contrast, the current "gossipsub" in `go-floodsub` seems more safe.

Thank you for reading!

whyrusleeping:
@mhchia (also responded in the other issue)

Sorry for the delay, was at the decentralized web summit all last week.

The gossipsub spec draft is still a draft (review welcome!) and should probably renamed something other than gossipsub (because its much more than a simple gossip).

[...]

Issue #81, opened 2018-07-25
GossipSub: GetNodes message handling


jannikluhn:
In the [Joining the Overlay](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub#joining-the-overlay) section of the GossipSub specification a `GETNODES` message type is mentioned, but not how receivers should handle it. My guess is that they should respond with a `NODES` message containing a subset of their active and/or passive list. Is that correct? If so, what's the point of the `ttl` field in the `NODES` message?

whyrusleeping:
Hey @jannikluhn, sorry for the delay. Been a busy couple of weeks. I'll get this and your other question answered soon!

vyzo:
Ok, I think there is some confusion here -- the gossipsub spec that got checked in doesn't correspond to the protocol implementation. 
Rather, it is based on an earlier proposal for epidemic proximity aware pubsub, which is planned for the next iteration of the protocol.

raulk:
This misrepresentation got fixed with #84.

PR #84, opened 2018-08-08
gossipsub is not episub

vyzo:
The README in gossipsub is not a spec for the extant implementation (and actual design of gossipsub).
It is instead based on an earlier proposal for episub, a follow up protocol to be implemented on top of gossipsub.

This has created a lot of confusion, see:
- #81 
- #83 
- https://github.com/libp2p/go-floodsub/issues/92

This pr immediately rectifies the situation as it moves the readme into episub.md and creates a stub for the forthcoming spec of the actual gossipsub protocol.


mgoelzer:
LGTM

PR #87, opened 2018-08-19
gossipsub spec

vyzo:
The long awaited gossipsub spec has arrived.

### DIFF
--- a/pubsub/gossipsub/README.md
+++ b/pubsub/gossipsub/README.md
@@ -12,27 +12,30 @@ specialized routers, which may add protocol messages and gossip in
 order to provide behaviour optimized for specific application
 profiles.
 
-<!-- toc -->
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Contents**
 
-- [Implementation status](#implementation-status)
-- [In the beginning was floodsub](#in-the-beginning-was-floodsub)
-  * [Ambient Peer Discovery](#ambient-peer-discovery)
-  * [Flood routing](#flood-routing)
-  * [Retrospective](#retrospective)
-- [Controlling the flood](#controlling-the-flood)
-  * [randomsub: A random message router](#randomsub-a-random-message-router)
-  * [meshsub: An overlay mesh router](#meshsub-an-overlay-mesh-router)
-  * [gossipsub: The gossiping mesh router](#gossipsub-the-gossiping-mesh-router)
-- [The gossipsub protocol](#the-gossipsub-protocol)
-  * [Control messages](#control-messages)
-  * [Router state](#router-state)
-  * [Topic membership](#topic-membership)
-  * [Message processing](#message-processing)
-  * [Heartbeat](#heartbeat)
-  * [Control message piggybacking](#control-message-piggybacking)
-  * [Protobuf](#protobuf)
+- [gossipsub: An extensible baseline pubsub protocol](#gossipsub-an-extensible-baseline-pubsub-protocol)
+  - [Implementation status](#implementation-status)
+  - [In the beginning was floodsub](#in-the-beginning-was-floodsub)
+    - [Ambient Peer Discovery](#ambient-peer-discovery)
+    - [Flood routing](#flood-routing)
+    - [Retrospective](#retrospective)
+  - [Controlling the flood](#controlling-the-flood)
+    - [randomsub: A random message router](#randomsub-a-random-message-router)
+    - [meshsub: An overlay mesh router](#meshsub-an-overlay-mesh-router)
+    - [gossipsub: The gossiping mesh router](#gossipsub-the-gossiping-mesh-router)
+  - [The gossipsub protocol](#the-gossipsub-protocol)
+    - [Control messages](#control-messages)
+    - [Router state](#router-state)
+    - [Topic membership](#topic-membership)
+    - [Message processing](#message-processing)
+    - [Heartbeat](#heartbeat)
+    - [Control message piggybacking](#control-message-piggybacking)
+    - [Protobuf](#protobuf)
 
-<!-- tocstop -->
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 ## Implementation status
 
@@ -46,8 +49,8 @@ profiles.
 
 The initial pubsub experiment in libp2p was `floodsub`.
 It implements pubsub in the most basic manner, with two defining aspects:
-- ambient peer discovery.
-- most basic routing; flooding.
+- ambient peer discovery; and
+- most basic routing: flooding.
 
 ### Ambient Peer Discovery
 
@@ -55,8 +58,8 @@ With ambient peer discovery, the function is pushed outside the scope
 of the protocol. Instead, it relies on ambient connection events to
 perform peer discovery via protocol identification. Whenever a new
 peer is connected, the protocol checks to see if the peer implements
-floodsub, and if so it sends a hello packet announcing the topics it
-is currently subscribing.
+floodsub, and if so it sends a hello packet that announces the topics 
+that it is currently subscribing to.
 
 This allows the peer to maintain soft overlays for all topics of
 interest. The overlay is maintained by exchanging subscription
@@ -94,12 +97,15 @@ highly desirable properties:
 
 The problem however is that messages don't just follow the minimum
 latency paths; they follow all edges, thus creating a flood. The
-outbound degree of the network is unbounded. This creates a problem
-for individual densely connected nodes, as they may have a large
-number of connected peers and cannot afford the bandwidth to forward
-all these pubsub messages.  Similary, the amplification factor is only
-bounded by the sum of degrees of all nodes in the overlay, which
-creates a scaling problem for densely connected overlays at large.
+outbound degree of the network is unbounded, whereas we want it to be
+bounded in order to reduce bandwidth requirements and increase 
+decentralization and scalability. In other words, this unbounded
+outbound degree creates a problem for individual densely connected 
+nodes, as they may have a large number of connected peers and cannot
+afford the bandwidth to forward all these pubsub messages.  Similary,
+the amplification factor is only bounded by the sum of degrees of all
+nodes in the overlay, which creates a scaling problem for densely
+connected overlays at large.
 
 
 ## Controlling the flood
@@ -111,10 +117,10 @@ globally controls the amplification factor.
 ### randomsub: A random message router
 
 Let's first consider the simplest bounded floodsub variant, which we
-call `randomsub`. In this construction, the router is still stateless
+call `randomsub`. In this construction, the router is still stateless,
 apart from a list of known peers in the topic. But instead of
-forwarding messages to all peers, it forwards to a random subset up to
-`D` peers, where `D` is the desired degree of the network.
+forwarding messages to all peers, it forwards to a random subset of up 
+to `D` peers, where `D` is the desired degree of the network.
 
 The problem with this construction is that the message propagation
 patterns are non-deterministic. This results in extreme message route
@@ -158,7 +164,7 @@ admissible mesh degree bounds.
 
 ### gossipsub: The gossiping mesh router
 
-The meshsub router offsers a baseline construction with good amplification
+The meshsub router offers a baseline construction with good amplification
 control properties, which we augment with _gossip_ about message flow.
 The gossip is emitted to random subsets of peers not in the mesh, similar
 to randomsub, and it allows us to propagate _metadata_ about message flow
@@ -171,7 +177,7 @@ The router can use this metadata to improve the mesh, for instance an
 [episub](episub.md) router built on top of gossipsub can create
 epidemic broadcast trees.  Beyond that, the metadata can restart
 message transmission at different points in the overlay to rectify
-downstream message loss. Or it can simply jump hops oppurtunistically
+downstream message loss. Or it can simply jump hops opportunistically
 and accelerate message transmission for peers who are at some distance
 in the mesh.
 
@@ -191,33 +197,33 @@ towards them.
 ### Control messages
 
 The protocol defines four control messages:
-- `GRAFT`: graft a mesh link; notifies the peer that it has been added to the local mesh view.
-- `PRUNE`: prune a mesh link; notifies the peer that it has been removed from the local mesh view.
-- `IHAVE`: gossip; notifies the peer that the following messages were recently seen and are available on request.
-- `IWANT`: requests the transmission of messages announced in an `IHAVE` message.
+- `GRAFT`: graft a mesh link; this notifies the peer that it has been added to the local mesh view.
+- `PRUNE`: prune a mesh link; this notifies the peer that it has been removed from the local mesh view.
+- `IHAVE`: gossip; this notifies the peer that the following messages were recently seen and are available on request.
+- `IWANT`: request transmission of messages announced in an `IHAVE` message.
 
 ### Router state
 
 The router maintains the following state:
 - `peers`: a set of all known peers; `peers.gossipsub` denotes the gossipsub peers
    while `peers.floodsub` denotes the floodsub peers.
-- `mesh`: the overlay meshes as a map of topic to lists of peers.
+- `mesh`: the overlay meshes as a map of topics to lists of peers.
 - `fanout`: the mesh peers to which we are publishing to without topic membership,
-   as a map of topic to list of peers.
-- `seen`: this is the timed message id cache, which tracks seen messages.
+   as a map of topics to lists of peers.
+- `seen`: this is the timed message ID cache, which tracks seen messages.
 - `mcache`: a message cache that contains the messages for the last few
    heartbeat ticks.
 
-The message cache is a data structure that stores windows of message ids
+The message cache is a data structure that stores windows of message IDs
 and the corresponding messages. It supports the following operations:
 - `mcache.put(m)`: adds a message to the current window and the cache.
-- `mcache.get(id)`: retrieves a message from the cache by its id, if it is still present.
-- `mcache.window()`: retrieves the message id for messages in the current history window.
+- `mcache.get(id)`: retrieves a message from the cache by its ID, if it is still present.
+- `mcache.window()`: retrieves the message ID for messages in the current history window.
 - `mcache.shift()`: shifts the current window, discarding messages older than the
    history length of the cache.
 
-The timed message id cache is the flow control mechanism. It tracks
-the message ids of seen message for the last couple of minutes. It is
+The timed message ID cache is the flow control mechanism. It tracks
+the message IDs of seen messages for the last couple of minutes. It is
 separate from `mcache` for implementation reasons in Go (the `seen`
 cache is inherited from the pubsub framework), but they could be the
 same data structure.
@@ -237,8 +243,8 @@ router, as part of the pubsub api:
 Note that the router can publish messages without topic membership. In order
 to maintain stable routes in that case, it maintains a list of peers for each
 topic it has published in the `fanout` map. If the router does not publish any
-messages for some time, then the `fanout` peers for the topic are forgotten, so
-this is soft state.
+messages of a topic for some time, then the `fanout` peers for that topic are 
+forgotten, so this is soft state.
 
 Also note that as part of the pubsub api, the peer emits `SUBSCRIBE`
 and `UNSUBSCRIBE` control messages to all its peers whenever it joins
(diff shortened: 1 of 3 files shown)


whyrusleeping:
This is great! Thank you @vyzo!

whyrusleeping:
Hey @mhchia, Here is the gossipsub spec we've been talking about :) Please take a look at let us know if anything seems weird.

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,373 @@
...
+  subscribed to the topic. If it is not subscribed, it responds with a `PRUNE(topic)`
+  control message.
+- On `PRUNE(topic)` it removes the peer from `mesh[topic]`.
+- On `IHAVE(ids)` it checks the `seen` set and requests uknown messages with an `IWANT`
+   message.
+- On `IWANT(ids)` it forwards all request messages that are present in `mcache` to the
+   requesting peer.
+
+When the router publishes a message that originates from the router itself (at the
+applicaiton layer), then it proceeds similar to the payload reaction:
```
jannikluhn: applicaiton -> application

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,373 @@
...
+```
+The parameters of the algorithm are `D` which is the target degree,
+and two relaxed degree parameters `D_low` and `D_high` which represent
+admissible mesh degree bounds.
+
+### gossipsub: The gossiping mesh router
+
+The meshsub router offsers a baseline construction with good amplification
+control properties, which we augment with _gossip_ about message flow.
+The gossip is emited to random subsets of peers not in the mesh, similar
```
jannikluhn: emited -> emitted

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,373 @@
...
+
+Let's first consider the simplest bounded floodsub variant, which we
+call `randomsub`. In this construction, the router is still stateless
+apart from a list of known peers in the topic. But instead of
+forwarding messages to all peers, it forwards to a random subset up to
+`D` peers, where `D` is the desired degree of the network.
+
+The problem with this construction is that the message propagation
+patterns are non-deterministic. This results to extreme message route
+instability which is an undesirable property for many applications.
```
jannikluhn: Is "message route instability" just the fact that there are no stable path along which messages are routed? If so, could you clarify what the problem with that is?

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,373 @@
...
+### meshsub: An overlay mesh router
+
+Nonetheless, the idea of limiting the flow of messages to a random
+subset of peers is solid. But instead of randomly selecting peers on a
+per message basis, we can form an overlay mesh where each peer
+forwards to a subset of its peers on a stable basis. We construct a
+router in this fashion, dubbed `meshsub`.
+
+Each peer maintains its own view of the mesh for each topic, which is
+a list of bidirectional links to other peers.  That is, in steady
```
jannikluhn: What's the advantage of requiring links to be bidirectional? Just simplicity?

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,373 @@
...
+- `PRUNE`: prune a mesh link; notifies the peer that it has been removed from the local mesh view.
+- `IHAVE`: gossip; notifies the peer that the following messages were recently seen and are available on request.
+- `IWANT`: requests the out of band transmission of messages announced in an `IHAVE` message.
+
+### Router state
+
+The router maintains the following state:
+- `peers`: a set of all known peers; `peers.gossipsub` denotes the gossipsub peers
+   while `peers.floodsub` denotes the floodsub peers.
+- `mesh`: the overlay meshes as a map of topic to lists of peers.
```
jannikluhn: Is it correct that all set elements in `mesh` are also in `peers.gossipsub`, but not necessarily the other way around?

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,373 @@
...
+- On `IHAVE(ids)` it checks the `seen` set and requests uknown messages with an `IWANT`
+   message.
+- On `IWANT(ids)` it forwards all request messages that are present in `mcache` to the
+   requesting peer.
+
+When the router publishes a message that originates from the router itself (at the
+applicaiton layer), then it proceeds similar to the payload reaction:
+- it forwards the message to every peer in `peers.floodsub[topic]`.
+- if it is subscribed to the topic, then it must have a set of peers in `mesh[topic]`,
+  to which the message is forwarded.
```
jannikluhn: "it must have a set of peers in `mesh[topic]`": This seems to be not true in the (slightly pathological, but possible) case when a peer has received `PRUNE`s from all its mesh peers, but hasn't sent any `GRAFT`s yet. This could be fixed by immediately `GRAFT`ing when a `PRUNE` would lead to `|mesh[topic]| < D_low`.

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,373 @@
...
+```
+
+Gossip is emitted by selecting peers for each topic that are not already part
+of the mesh:
+```
+for each topic in mesh+fanout:
+  let mids be mcache.window[topic]
+  if mids is not empty:
+    select D peers from peers.gossipsub[topic]
+    for each peer not in mesh[topic]
```
jannikluhn: This loops over the just selected peers, correct?

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,373 @@
...
+router, as part of the pubsub api:
+- On `JOIN(topic)` the router joins the topic. In order to do so, it
+  selects `D` peers from `peers.gossipsub[topic]`, adds them to `mesh[topic]`
+  and notifies them with a `GRAFT(topic)` control message. If it already has
+  `fanout` peers in the topic, then it selects those peers as the
+  initial mesh peers.
+- On `LEAVE(topic)` the router leaves the topic. It notifies the peers in
+  `mesh[topic]` with a `PRUNE(topic)` message and forgets `mesh[topic]`.
+
+Note that the router can publish messages without topic membership. In order
```
jannikluhn: Is there a relevant application for publishing messages one is not subscribed to? To me it seems like this makes the protocol more complicated for very little gain.

To simplify it while keeping the option of publishing without membership, one could just use "randomsub" which would be equivalent as long as the time between those messages is greater than the fanout TTL.

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,373 @@
...
+maintains a topic view of its direct peers only. Whenever a peer
+disconnects, it is removed from the overlay.
+
+Ambient peer discovery can be driven by arbitrary external means, which
+allows orthogonal development and no external dependencies for the protocol
+implementation.
+
+These are a couple of options we are exploring as canonical approaches
+for the discovery driver:
+- DHT lookups using provider records.
```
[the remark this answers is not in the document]
vyzo: It means we utilize provider records for the topic.  I will reword it a bit.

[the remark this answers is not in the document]
vyzo: So basically each peer that subscribes to the topic, announces a provider record named after the topic.
Peers interested in discovery perform a DHT lookup with `findProviders` to retrieve the other peers.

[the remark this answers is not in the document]
vyzo: Updated the wording.

pubsub/gossipsub/README.md
vyzo: The problem is that message start taking random paths that are different in for each message, which results in reordeding and weird timing.
It's still a correctly functioning pubsub router, it's just undesirable to pathologically reorder stuff.

pubsub/gossipsub/README.md
vyzo: Yes, simplicity.  In fact the very first version I wrote had unidirectional links, but it turned out that it was just complexity that didn't add anything to the protocol.

pubsub/gossipsub/README.md
vyzo: yes, correct.

pubsub/gossipsub/README.md
vyzo: The set could be currently be empty indeed, in which case we will not forward messages until the next heartbeat adds some links.

pubsub/gossipsub/README.md
vyzo: yes -- prose pseudocode :)

pubsub/gossipsub/README.md
vyzo: Well, we still want to keep stable routes for non-subscribing publishers, hence the fanout set.

Re: applications: any streaming application will have this property, the source doesn't need to subscribe at all.

pubsub/gossipsub/README.md
vyzo: I will add some words to this extend.

pubsub/gossipsub/README.md
vyzo: Nonetheless you are right that we could immediately graft some links if we try to publish on an empty set (because of prunes).

pubsub/gossipsub/README.md
vyzo: Another application is when you have a cluster utilizing pubsub and allow clients to publish to the cluster by using pubsub. The clients are not interested in the pubsub chatter in that case.

vyzo:
cc @jamesray1 as well.

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,375 @@
...
+## Implementation status
+
 - Go: [libp2p/go-floodsub#67](https://github.com/libp2p/go-floodsub/pull/67) (experimental)
 - JS: not yet started
 - Rust: not yet started
-- Gerbil: [vyzo/gerbil-simsub](https://github.com/vyzo/gerbil-simsub) (very first implementation)
+- Gerbil: [vyzo/gerbil-simsub](https://github.com/vyzo/gerbil-simsub) (simulator)
+
+
+## In the beginning was floodsub
```
jamesray1: Lol

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,375 @@
...
+- it is straightforward to implement.
+- it minimizes latency; messages are delivered across minimum latency
+  paths, modulo overly connectivity.
+- it is highly robust; there is very little maintenance logic or state.
+
+The problem however is that messages don't just follow the minimum
+latency paths; they follow all edges, thus creating a flood. The
+outbound degree of the network is unbounded, restricted solely from
+connectivity. This creates a problem for individual densely connected
+nodes, as they may have large number of connected peers and cannot
```
jamesray1: a large number

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,375 @@
...
+### randomsub: A random message router
+
+Let's first consider the simplest bounded floodsub variant, which we
+call `randomsub`. In this construction, the router is still stateless
+apart from a list of known peers in the topic. But instead of
+forwarding messages to all peers, it forwards to a random subset up to
+`D` peers, where `D` is the desired degree of the network.
+
+The problem with this construction is that the message propagation
+patterns are non-deterministic. This results to extreme message route
```
jamesray1: results in

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,375 @@
...
+highly desirable properties:
+- it is straightforward to implement.
+- it minimizes latency; messages are delivered across minimum latency
+  paths, modulo overly connectivity.
+- it is highly robust; there is very little maintenance logic or state.
+
+The problem however is that messages don't just follow the minimum
+latency paths; they follow all edges, thus creating a flood. The
+outbound degree of the network is unbounded, restricted solely from
+connectivity. This creates a problem for individual densely connected
```
jamesray1: The phrase, "restricted solely from connectivity", is a bit confusing. Being unbounded, it is unrestricted, whereas we want it to bounded and restricted to what a node needs to connect to.

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,375 @@
...
+with floodsub, as it accepts floodsub peers and behaves like floodsub
+towards them.
+
+### Control messages
+
+The protocol defines four control messages:
+- `GRAFT`: graft a mesh link; notifies the peer that it has been added to the local mesh view.
+- `PRUNE`: prune a mesh link; notifies the peer that it has been removed from the local mesh view.
+- `IHAVE`: gossip; notifies the peer that the following messages were recently seen and are available on request.
+- `IWANT`: requests the out of band transmission of messages announced in an `IHAVE` message.
```
jamesray1: Just to clarify about this out of band transmission, I think at the moment no in-band transmission is specified, right? It leaves it to be externally done via [peer routing](https://github.com/libp2p/specs/blob/master/4-architecture.md#41-peer-routing), e.g. Kademlia or mDNS.

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,375 @@
...
+   add the peers to fanout[topic]
+```
+
+Gossip is emitted by selecting peers for each topic that are not already part
+of the mesh:
+```
+for each topic in mesh+fanout:
+  let mids be mcache.window[topic]
+  if mids is not empty:
+    select D peers from peers.gossipsub[topic]
```
jannikluhn: Is there a reason why this `D` is the same as the target mesh degree `D`? I think it would make sense to be able to configure them separately because they fulfill different purposes: One might want to decrease the "eager D" (to use episub terminology) to save bandwidth, but increase the "lazy D" to still be safe (so that nodes still get all messages with high probability even if all of their few mesh peers drop out).

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,375 @@
...
+messages are not further forwarded. It also never forwards a message
+back to the source or the peer that forwarded the message.
+
+### Retrospective
+
+Evaluating floodsub as a viable pubsub protocol reveals the following
+highly desirable properties:
+- it is straightforward to implement.
+- it minimizes latency; messages are delivered across minimum latency
+  paths, modulo overly connectivity.
```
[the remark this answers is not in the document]
vyzo: Nah, i'll just fix the typos. Thanks for spotting those!

pubsub/gossipsub/README.md
vyzo: Well, there is no bound from the point of view of the node, and it is restricted by the connectivity -- but I could use a better word.

pubsub/gossipsub/README.md
vyzo: Perhaps out of band was the wrong word.
The `IWANT` asks for the message, which gets transmitted directly as a normal message within the peer stream.
I'll reword to not call it out of band.

pubsub/gossipsub/README.md
vyzo: There is no specific reason for using D, I just wanted to reduce the number of parameters.
But you are right, this can be different from D, and there is merit in having a separate parameter for this.

pubsub/gossipsub/README.md
vyzo: Just dropped out of band here, it's very much in-band :)

pubsub/gossipsub/README.md
vyzo: I'll add some words to this effect, as this could be an important parameter.

pubsub/gossipsub/README.md
vyzo: Removed the offending phrase.

pubsub/gossipsub/README.md
vyzo: Add a paragraph about this.

pubsub/gossipsub/README.md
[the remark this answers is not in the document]
mhchia: Big thanks for the clear explanation!

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,386 @@
...
+   as a map of topics to lists of peers.
+- `seen`: this is the timed message ID cache, which tracks seen messages.
+- `mcache`: a message cache that contains the messages for the last few
+   heartbeat ticks.
+
+The message cache is a data structure that stores windows of message IDs
+and the corresponding messages. It supports the following operations:
+- `mcache.put(m)`: adds a message to the current window and the cache.
+- `mcache.get(id)`: retrieves a message from the cache by its ID, if it is still present.
+- `mcache.window()`: retrieves the message ID for messages in the current history window.
```
raulk: nit: message **IDs** for better clarity.

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,386 @@
...
+The protocol defines four control messages:
+- `GRAFT`: graft a mesh link; this notifies the peer that it has been added to the local mesh view.
+- `PRUNE`: prune a mesh link; this notifies the peer that it has been removed from the local mesh view.
+- `IHAVE`: gossip; this notifies the peer that the following messages were recently seen and are available on request.
+- `IWANT`: request transmission of messages announced in an `IHAVE` message.
+
+### Router state
+
+The router maintains the following state:
+- `peers`: a set of all known peers; `peers.gossipsub` denotes the gossipsub peers
```
raulk: IIUC, these are distinct circuits that do not interact with one another, i.e. the gossipsub operations will not interact with the floodsub ones, and viceversa. In fact, these could even be fully distinct sets with no overlapping peers, right?

If that's the case, it would be great to highlight it. Because the processes below do refer to both of them, so it leads to one thinking that they are somehow interrelated. In fact, at first sight I thought the `fanout` and `floodsub` fulfilled similar roles. Been banging my head a little bit 🙃 

On a related note, if two peers expose both gossipsub and floodsub, they'd participate in both circuits?

pubsub/gossipsub/README.md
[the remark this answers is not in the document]
vyzo: ok

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,386 @@
...
+The message cache is a data structure that stores windows of message IDs
+and the corresponding messages. It supports the following operations:
+- `mcache.put(m)`: adds a message to the current window and the cache.
+- `mcache.get(id)`: retrieves a message from the cache by its ID, if it is still present.
+- `mcache.window()`: retrieves the message ID for messages in the current history window.
+- `mcache.shift()`: shifts the current window, discarding messages older than the
+   history length of the cache.
+
+The timed message ID cache is the flow control mechanism. It tracks
+the message IDs of seen messages for the last couple of minutes. It is
```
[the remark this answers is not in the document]
vyzo: It's two minutes, but non-normative. A router could use a different time, this is chosen to approximate the propagation delay in the overlay with some healthy margin.

pubsub/gossipsub/README.md
```diff
@@ -1,7 +1,386 @@
...
+
+The message cache is a data structure that stores windows of message IDs
+and the corresponding messages. It supports the following operations:
+- `mcache.put(m)`: adds a message to the current window and the cache.
+- `mcache.get(id)`: retrieves a message from the cache by its ID, if it is still present.
+- `mcache.window()`: retrieves the message ID for messages in the current history window.
+- `mcache.shift()`: shifts the current window, discarding messages older than the
+   history length of the cache.
+
+The timed message ID cache is the flow control mechanism. It tracks
```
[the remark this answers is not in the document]
vyzo: sure.

pubsub/gossipsub/README.md
vyzo: There are no distinct circuits, the same messages propagate across the overlay regardless of whether it's a gossipsub or floodsub peer.
The difference is that there is no control pruning for floodsub peers, we just forward the messages to them.

pubsub/gossipsub/README.md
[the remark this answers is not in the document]
vyzo: I'll change it to say explicitly "two minutes"

vyzo:
Last Call for Comments -- the merge train will depart tomorrow.

vyzo:
> ... wondering if you've thought about building in reputation and scoring algorithms. 

Sure, that's something we could do in extended gossipsub routers -- but it's not in the simple baseline protocol we want then :)

> For example, I could see malicious actors joining the network to flood routers with large IWANT messages, or other types of attacks.

Yeah, it could help with quite a few things.

pubsub/gossipsub/README.md
raulk: Yeah, sorry. What I meant is that the gossipsub processes don't interact with the floodsub processes directly. The entrypoint to the "circuits" is the same: the incoming message, but they get treated differently by each protocol with no cross-usage of resources.

jamesray1:
Yep, one feature extension at a time!

pubsub/gossipsub/README.md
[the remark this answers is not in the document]
vyzo: Good point, will add.

Issue #32, opened 2017-08-15
Service Discovery instead of Protocol Negotiation


Stebalien:
Currently, there's a *lot* of overhead when each connection is established opening up multiple streams to figure out what services the other peer supports. This also tends to be a bit racy and *really* messy.

Instead, we could introduce a push-based service discovery system. On connect, a peer would announce what protocols it supports up-front (the list shouldn't be that large) and whenever a new service gets added/removed, the peer would tell its peers about this change. We could then avoid an entire round trip when opening new streams because we already know the protocols our peers support. This would also remove the need for a lot of our peer probing code (we could just have the host keep track of which connections support which protocols).

---

This would also make it possible to implement a pseudo message-oriented system on-top of our stream-oriented communication primitives. Why? Because most consumers of libp2p in go-ipfs are already (mostly) message oriented. That is, we have, for almost *every* service:

1. Something to keep track of peers and metadata about them (and keep track of the status of the connection).
2. Go routines to accept messages from peers and pipe it into a central event loop.
3. Go routines to send messages back out to peers. Note: they *are* sent back to peers in-order, this is the only non-message oriented part of this and why implementing this on-top of streams actually makes sense.

We shouldn't have to implement this over-and-over for every service; this is where a lot of our leaks and race conditions come from...

If we had this central service registry, we could use it to transparently manage all streams, opening them as necessary and possibly even managing associated metadata.

dryajov:
@Stebalien I like the general idea and I believe this is to some extent along the same lines as https://github.com/ipfs/notes/issues/237. We need a way to tell what a peer can do and who it knows about (connected to?) and I believe @vyzo and I have both independently proposed something very similar to this while working on the circuit implementations.

whyrusleeping:
We do currently announce what protocols we support up front (via identify). The issue i think is that other protocols dont bother waiting for identify to finish. Great suggestion overall, i agree.

RangerMauve:
How does one use the identify API from `js-ipfs`?

dryajov:
@RangerMauve afaik identify is a lower level libp2p protocol and we don't expose it with the http API.

RangerMauve:
@dryajov js-ipfs in the browser can run a full IPFS node which uses `js-libp2p`. Where exactly in libp2p would this call be? Would it be somewhere inside `swarm`?

dryajov:
@RangerMauve https://github.com/libp2p/js-libp2p-swarm/blob/master/src/connection.js#L35...L52

[...]

PR #95, opened 2018-10-09
[WIP] [RFC] Multistream-2.0

Stebalien:
Here's a draft of multistream-2.0 (+ a retrospective that you can skip).

Note: *NONE* of this is set in stone (or even sand), this is for discussion.

**PLEASE JUMP TO: https://github.com/libp2p/specs/pull/95#issuecomment-454373749**

### DIFF
--- a/multistream-2.0/spec.md
+++ b/multistream-2.0/spec.md
@@ -27,8 +27,8 @@ anything.
    `multistream/advertise` To inform the initiator of *it's* mapping between
    protocols and contextual IDs.
 
-Second, this document proposes 2 auxiliary protocols that can be used with the 4
-multistream protocols to actually negotiate protocols. These are *primarily*
+Second, this document proposes an auxiliary protocol that can be used with the 4
+multistream protocols to actually negotiate protocols. This is *primarily*
 useful (a) in packet-based protocols (without sessions) and (b) when initially
 negotiating a transport session (before protocols have been advertised and the
 stream multiplexer has been configured).
@@ -38,10 +38,6 @@ stream multiplexer has been configured).
    negotiate a protocol, use it, and then return to multistream. It also allows
    us to speculatively choose a single protocol and then drop back down to
    multistream if that doesn't work.
-2. `speculative-stream`: A speculative stream "multiplexer" where the initiator
-   can speculatively initiate multiple streams and the receiver must select at
-   most one and discard the others. On a bidirectional stream, the receiver will
-   inform the initiator of the selected sub-stream, collapsing the state.
    
 All peers *must* implement `multistream/use` and *should* implement
 `serial-stream`. This combination will allow us to apply a series of quick
@@ -54,9 +50,11 @@ Notes:
    should be a *protocol*. Given the `serial-stream` protocol, this shouldn't be
    an issue as we can run as many sub-protocols over the same stream as we want.
 2. To reduce RTTs, all protocols are unidirectional.
-3. These protocols were *also* designed to eventually support packet protocols
-   (the other reason to be unidirectional and a strong motivator for the
-   `speculative-stream` and `serial-stream` protocols).
+3. These protocols were *also* designed to eventually support packet protocols.
+4. We considered a `speculative-stream` protocol where the initiator
+   speculatively starts multiple streams and the receiver acts on at most one.
+   This would have allowed for 0-RTT worst-case protocol negotiation but was
+   deemed too complicated for inclusion in the core spec.
 
 ### Multistream Advertise
 
@@ -124,52 +122,6 @@ This protocol has *also* been designed to be hardware friendly:
    IDs are chosen by the *receiver* means that the receiver can reuse the same
    IDs for all connected peers (reusing the same hardware routing table).
    
-### Speculative Stream
-
-
-The `speculative-stream` protocol allows an initiator to start multiple streams in
-parallel while telling the receiver to only *act* on one of them. This:
-
-1. Allows us to "negotiate" each stream using the other multistream protocols.
-   That is, each message/sub-stream recursively uses multistream.
-2. Pack data into the initial packet to shave off a RTT in many cases.
-3. Support packet transports out of the box where round-trips may not be
-   possible.
-
-Each message in this protocol consists of:
-
-```
-<stream number (varint)>
-<data length (varint)>
-```
-
-The where the receiver can transition to a single one of these streams by
-sending:
-
-```
-<stream number>
-0
-```
-
-And the initiator responds the same way to finish off the transition.
-
-This aborts all the other streams, allowing the chosen stream to completely take
-over the channel.
-
-Note: A *simple* implementation of this protocol would simply send a sequence of
-protocols as `<stream number 1><length><multistream/use>...<stream number
-2><length><multistream/use>...` and then wait for the other side to select the
-appropriate protocol.
-
-* [ ] Q: The current framing system is dead simple but inefficient in some
-      cases. Specifically, one can't just (a) read a *single* header and then
-      (b) jump to the desired sub-stream. Alternatives include:
-    * Have a single header that maps stream numbers to offsets and lengths. This
-      way, one could jump to the correct section immediately.
-    * Have a single list of "sections", no stream numbers. Stream numbers would be
-      inferred by index. This is slightly smaller but not very flexible.
-* [ ] Q: Just do something simpler?
-
 ### Serial Stream
 
 The `serial-stream` protocol is the simplest possible stream multiplexer.
@@ -196,8 +148,7 @@ makes implementing this protocol trivial, even in hardware.
 
 Why: This allows us to:
 
-1. Try protocols and fall back on others (we can also use `speculative-stream`
-   for this).
+1. Try protocols and fall back on others.
 2. More importantly, it allows us to speak a bunch of protocols before setting
    up a stream multiplexer. Specifically, we can use this for
    `multistream/advertise` to send an advertisement as early as possible.
@@ -277,72 +228,48 @@ Now that we're using multistream 2.0, the initiator will send, in a single
 packet:
 
 ```
-<multistream/use (multicodec)><speculative-stream (multicodec)>        // use multistream/use to select speculative-stream
-  <0 (stream number, varint)><len (varint)>                            // in alt stream 0
-    <multistream/use (multicodec)><secio (multicodec)>                 // select SECIO
-      <initial secio packet...>                                        // initiate SECIO
-  <1 (stream number, varint)><len (varint)>                            // in alt stream 1
-    <multistream/use (multicodec)><tls (multicodec)>                   // select TLS
-      <initial tls packet...>                                          // initiate TLS
+<multistream/use (multicodec)><serial-stream (multicodec)>                 // use serial-stream to make the stream recoverable
+  <len>                                                                    // serial-stream message framing
+    <multistream/use (multicodec)><multistream/advertise (multicodec)>     // select advertise protocol
+      supported security protocols...                                      // 
+  -1                                                                       // return to multistream (EOF)
+
+<multistream/use (multicodec)><serial-stream (multicodec)>                 // open a new serial-stream
+  <len>
+    <multistream/use (multicodec)><tls (multicodec)>                       // select TLS
+      <initial tls packet...>                                              // initiate TLS
 ```
-
-The code to do this will likely look roughly like:
-
-```go
-streams := multistream.XOR(stream, ProtocolTLS, ProtocolSecIO)
-var wg sync.WaitGroup
-wg.Add(2)
-var (
-  secioConn, tlsConn net.Conn
-  secioErr,  tlsErr  error
-)
-go func() {
-  defer wg.Done()
-  secioConn, tlsErr = tls.Upgrade(streams[0])
-  ...
-}()
-
-go func() {
-  defer wg.Done()
-  tlsConn, tlsErr = secio.Upgrade(streams[1])
-  ...
-}()
-
-wg.Wait()
-
-switch {
-case tlsErr == nil:
-  return tlsConn
-case secioConn == nil:
-  return secioConn
-default:
-  return (some error)
-}
-```
-
 
 The receiver will respond with:
 
 ```
-<multistream/use (multicodec)><speculative-stream (multicodec)>        // use multistream/use to select speculative-stream
-  <1 (stream number, varint)>0                                         // choose stream 1
+<multistream/use (multicodec)><serial-stream (multicodec)>                 // respond to serial stream
+  <len>
+    <multistream/use (multicodec)><multistream/advertise (multicodec)>     // select advertise protocol
+      security protocols...
+  -1                                                                       // return to multistream (EOF)
 
-<multistream/use (multicodec)><tls (multicodec)>                       // respond to the "use tls" protocol
-<tls response...>                                                      // speak tls
+<multistream/use (multicodec)><serial-stream (multicodec)>                 // respond to second serial stream
+  0                                                                        // transition to a normal stream.
+<multistream/use (multicodec)><tls (multicodec)>                           // select TLS
+  <response tls packet...>                                                 // complete TLS handshake
 ```
 
-The speculative stream handler will likely just try each stream in-order,
-selecting the first stream that ends up negotiating a known protocol. More
-advanced implementations may allow for speculative stream *handlers* to select
-from within multiple known protocols. However, this is unlikely to be necessary
-for a while.
+This:
+
+1. Responds to the advertisement, also advertising available security protocols.
+2. Accepts the TLS stream.
+3. Finishes the TLS handshake.
+
+If the receiver had *not* supported TLS, it would have reset the serial-stream.
+In that case, the initiator would have used the protocols advertised by the
+receiver to select an appropriate security protocol.
 
 Finally, the initiator will finish the TLS negotiation, send a advertise packet,
-*optimistically* negotiate yamux (it could also use speculative-stream to
-negotiate both at the same time but let's not), and sends the DHT request.
+*optimistically* negotiate yamux, and sends the DHT request.
 
 ```
-  <1 (stream number)>0                                                     // choose stream 1
+  0                                                                        // transition to a normal stream.
 
 <tls client auth...>                                                       // finish TLS
 
@@ -386,3 +313,7 @@ And the receiver will send:
 Note: Ideally, we'd be able to avoid the optimistic yamux negotiation. However,
 to do that, some protocol information will have to be embedded in the TLS
 negotiation and exposed through a connection-level `Stat` method.
+
+Alternatively, we could choose to include this information in the advertisement
+sent *before* the security transport. However, that has some security
+implications.
--- a/multistream-2.0/spec.md
+++ b/multistream-2.0/spec.md
@@ -20,7 +20,8 @@ anything.
 1. `multistream/advertise`: Inform the remote end about which protocols we
    speak. This should partially replace the current identify protocol.
 2. `multistream/multicodec`: Selects the stream's protocol using a multicodec.
-3. `multistream/string`: Selects the stream's protocol using a string protocol name.
+3. `multistream/string`: Selects the stream's protocol using a string protocol
+   name.
 4. `multistream/dynamic`: Selects the stream's protocol using a protocol ID
    defined by the *receiver*, valid for the duration of the "session"
    (underlying connection). To use this, the *receiver* must have used the
--- a/multistream-2.0/spec.md
+++ b/multistream-2.0/spec.md
@@ -49,7 +49,9 @@ Notes:
 1. The "ls" feature of multistream has been removed. While useful, this really
    should be a *protocol*. Given the `serial-stream` protocol, this shouldn't be
    an issue as we can run as many sub-protocols over the same stream as we want.
-2. To reduce RTTs, all protocols are unidirectional.
+2. All multistream-2 protocols are unidirectional. On a bidirectional stream,
+   these protocols are run once in each direction with the receiver mirroring
+   the initiator.
 3. These protocols were *also* designed to eventually support packet protocols.
 4. We considered a `speculative-stream` protocol where the initiator
    speculatively starts multiple streams and the receiver acts on at most one.
--- a/multistream-2.0/spec.md
+++ b/multistream-2.0/spec.md
@@ -139,7 +139,7 @@ The protocol is:
 
 Where the header is:
 
-* -2 - Send a reset and return to multistream. All queued data (remote and
+* -2 - Abnormal End: Send a reset and return to multistream. All queued data (remote and
   local) should be discarded.
 * -1 - Close: Send an EOF and return to multistream.
 *  0 - Rest: Ends the reuse protocol, transitioning to a direct stream.
--- a/multistream-2.0/spec.md
+++ b/multistream-2.0/spec.md
@@ -19,13 +19,13 @@ anything.
 
 1. `multistream/advertise`: Inform the remote end about which protocols we
    speak. This should partially replace the current identify protocol.
-2. `multistream/use`: Selects the stream's protocol using a multicodec.
-3. `multistream/dynamic`: Selects the stream's protocol using a string protocol name.
-4. `multistream/contextual`: Selects the stream's protocol using a protocol ID
+2. `multistream/multicodec`: Selects the stream's protocol using a multicodec.
+3. `multistream/string`: Selects the stream's protocol using a string protocol name.
+4. `multistream/dynamic`: Selects the stream's protocol using a protocol ID
    defined by the *receiver*, valid for the duration of the "session"
    (underlying connection). To use this, the *receiver* must have used the
    `multistream/advertise` To inform the initiator of *it's* mapping between
-   protocols and contextual IDs.
+   protocols and dynamic IDs.
 
 Second, this document proposes an auxiliary protocol that can be used with the 4
 multistream protocols to actually negotiate protocols. This is *primarily*
@@ -39,7 +39,7 @@ stream multiplexer has been configured).
    us to speculatively choose a single protocol and then drop back down to
    multistream if that doesn't work.
    
-All peers *must* implement `multistream/use` and *should* implement
+All peers *must* implement `multistream/multicodec` and *should* implement
 `serial-stream`. This combination will allow us to apply a series of quick
 connection upgrades (e.g., to multistream 3.0) with no round trips and no funny
 business (learn from past mistakes).
@@ -59,34 +59,34 @@ Notes:
 ### Multistream Advertise
 
 Unspeced (for now). Really, we just need to send a mapping of protocol
-names/codecs to contextual IDs (and may be some service discovery information).
+names/codecs to dynamic IDs (and may be some service discovery information).
 This is the subset of identify needed for protocol negotiation.
 
 ### Multistream Use
 
-The `multistream/use` protocol is simply two varint multicodecs: the
+The `multistream/multicodec` protocol is simply two varint multicodecs: the
 multistream-use multicodec followed by the multicodec for the protocol to be
 used. This protocol supports unidirectional streams. If the stream is
 bidirectional, the receiver must acknowledge a successful protocol negotiation
 by responding with the same multistream-use protocol sequence.
 
 Every stream starts with multistream-use. Every other protocol defined here will
-be assigned a multicodec and selected with `multistream/use.`
+be assigned a multicodec and selected with `multistream/multicodec.`
 
 This protocol should *also* be trivial to optimize in hardware simply by prefix
 matching (i.e., matching on the first N (usually 16-32) bits of the
 stream/message).
 
-### Multistream Dynamic
+### Multistream String
 
-The `multistream/dynamic` protocol is like the `multistream/use` protocol
+The `multistream/string` protocol is like the `multistream/multicodec` protocol
 *except* that it uses a string to identify the protocol. To do so, the initiator
 simply sends a varint length followed by the name of the protocol.
 
-Including the `multistream/use` portion, the initiator would send:
+Including the `multistream/multicodec` portion, the initiator would send:
 
 ```
-<multistream/use><multistream/dynamic><length(varint)><name(string)>
+<multistream/multicodec><multistream/string><length(varint)><name(string)>
 ```
 
 Note: This used to use a fixed-width 16 bit number for a length. However, a
@@ -94,9 +94,9 @@ varint *really* isn't going to cost us much, if anything, in terms of
 performance as most protocol names will be <= 128 bytes long. On the other hand,
 using different number formats everywhere *will* cost us in terms of complexity.
 
-### Multistream Contextual
+### Multistream Dynamic
 
-The `multistream/contextual` protocol is used to select a protocol using a
+The `multistream/dynamic` protocol is used to select a protocol using a
 *receiver specified*, session-ephemeral protocol ID. These IDs are analogues of
 ephemeral ports.
 
@@ -106,19 +106,19 @@ In this protocol, the stream initiator sends a varint ID specified by the
 Format:
 
 ```
-<multistream/use><multistream/contextual><id(varint)>
+<multistream/multicodec><multistream/dynamic><id(varint)>
 ```
 
 The ID 0 is reserved for saying "same protocol" on a bidirectional stream. The
-receiver of a bidirectional stream can't reuse the same contextual ID that the
-initiator used as this contextual ID is relative *to* the receiver. Really, this
+receiver of a bidirectional stream can't reuse the same dynamic ID that the
+initiator used as this dynamic ID is relative *to* the receiver. Really, this
 last rule *primarily* exists to side-step the TCP simultaneous connect issue.
 
 This protocol has *also* been designed to be hardware friendly:
 
 1. Hardware can compare the first 16 bits of the message against
-   `<multistream/use><multistream/contextual>`.
-2. It can then route the message based on the contextual ID. The fact that these
+   `<multistream/multicodec><multistream/dynamic>`.
+2. It can then route the message based on the dynamic ID. The fact that these
    IDs are chosen by the *receiver* means that the receiver can reuse the same
    IDs for all connected peers (reusing the same hardware routing table).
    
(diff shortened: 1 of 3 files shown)


Stebalien:
Note: `multistream/choose` is actually an *xor* operator and `serial-multiplex` is actually an *and* operator. We may want to just call them `multistream/xor` and `multistream/and` but I figured that may be even *more* confusing.

Stebalien:
> Would be nice to get some example negotiation scenarios to make it a bit easier to reason about

I agree and @vyzo gave the same feedback. Working on it.

magik6k:
So the first pass looks good, I need to dig into libp2p code to get a bit more context to convince myself, that there are no other ways to make this less complicated

(Also, can exploit the fact that quic is still experimental and just drop multistream v1 compat on quic connections?)

Stebalien:
> (Also, can exploit the fact that quic is still experimental and just drop multistream v1 compat on quic connections?)

Yes and I plan on doing exactly that :).

> (Also, can exploit the fact that quic is still experimental and just drop multistream v1 compat on quic connections?)

> So the first pass looks good, I need to dig into libp2p code to get a bit more context to convince myself, that there are no other ways to make this less complicated

There probably are (although I'd like to keep it modular; as it stands, implementations can drop sub-protocols).

We can probably either remove or simply not implement the speculative-stream protocol. It's a nice to have, especially for future packet-switched networks, but really not necessary for a first-pass.

We need multistream/dynamic and serial-stream for parity with the current multistream and we need multistream/contextual for cheeper streams.

We can probably drop the explicit multistream/use protocol and just say "in multistream, all protocols begin with a multicodec specifying the protocol". That's *effectively* multistream/use, it just drops the explicit multistream/use multicodec. That is, `multistream/dynamic` would begin with `<multistream/dynamic (multicodec)>...` instead of `<multistream/use (multicodec)><multistream/dynamic (multicodec)>...`.

raulk:
Gonna review this soon. Meanwhile, just jotting down a couple of things I'd like to see in our next iteration of multistream (aside from the usual suspects like less chattiness).

## 1. Upfront negotiation

Nodes should be able to share a list of protocols they support during session establishment. This is useful for applications that knowingly support a small list of protocol-version tuples (e.g. Ethereum). A protocol like the below would allow for async 1-RTT negotiation, where the otherwise amortised cost per-stream is paid once upfront.

### Example flow

Upon connection, both nodes exchange a list of protocols they support in lexicographical order. This list can be plaintext (inefficient), gzipped (better), a bloom filter (non-deterministic), or something else.

Alongside each entry, they signal an version agreement strategy (byte value), e.g. `STRICT`, `SEMVER`, `ANY_VERSION`, etc.

Both peers intersect the lists by:

1. Deleting the protocols they do not support.
2. Resolving version ambiguity/conflict by applying the version agreement strategy for the protocol (resolution behaviour to be defined, e.g. for `SEMVER` we fallback to the oldest).

The output is the identical on both sides: an ordered list of agreed protocols, 0-based. Both parties now build a multiplexing table `int => protocol`. 

Henceforth, when opening a stream, instead of sending the full name of the protocol, they send the `int`.

**Dealing with confusion (fallback).** If a node arrives at an undefined result (e.g. different libp2p versions with different logic for the same strategy -- although this is bad), they request the multiplexing table from the peer, and accept it as valid. If both parties request the listing, the connection is terminated.

## 2. Protocol indexing

In cases where upfront negotiation is unfeasible (e.g. too many protos), peers must keep track of the order of proto selection during the lifetime of a session. When opening a stream for a previously selected protocol, they must send the index instead of the full protocol name.

tomaka:
I don't know whether that's actually a problem, but one consequence of remembering the list of protocols that a node supports is that nodes can no longer decide to drop support for protocols at runtime.

raulk:
@tomaka I believe that case is minor enough we can return an error interactively when the peer attempts to select the now-unsupported protocol. 

Also: the good thing about upfront negotiation is that we're already storing protos in the peerstore. If we store the multiplexing table, when reestablishing a connection with that peer later, we can exchange merkle hashes of the table, and bypass negotiation altogether if they match.

Stebalien:
> I don't know whether that's actually a problem, but one consequence of remembering the list of protocols that a node supports is that nodes can no longer decide to drop support for protocols at runtime.

@tomaka that can be fixed by sending negative protocol announcements (or just responding with "I don't speak that" later).

Basically, you just can't reuse protocol IDs.

---

> In cases where upfront negotiation is unfeasible (e.g. too many protos), peers must keep track of the order of proto selection during the lifetime of a session. When opening a stream for a previously selected protocol, they must send the index instead of the full protocol name.

Completely agree (and we currently do this, luckily).

> 1. Upfront negotiation

So, technically, we already do this (identify). However, we don't currently *wait* for this negotiation to complete. Unfortunately, due to the lack of a `serial-stream`-like protocol, we have to wait to setup a stream multiplexer before we can run identify. With this proposal, we can piggy-back a protocol announcement along with the end of the crypto handshake.

> If we store the int-indexed multiplexing table, when reestablishing a connection with that peer later, we can exchange merkle hashes of the table, and bypass negotiation altogether if they match, i.e.

Really, we can probably just send the entire table. It'll probably be a packet, maybe two. State between reconnects feels like a bug waiting to happen. I mean, we could do the whole hash-dance (mmm, IPLD) however, a few packets is generally cheeper than a round-trip.

> The output is the identical on both sides: an ordered list of agreed protocols, 0-based. Both parties now build a multiplexing table int => protocol.

The protocols may not be symmetric. Really, we should just think of them as "endpoints" rather than protocols.

With that in mind, this proposal doesn't try to merge these lists. Instead, each side just sends *their own* mapping and expects the other side to use it when establishing inbound connections. Really, these mappings are equivalent to dynamic port mappings.

> Dealing with confusion (fallback). If a node arrives at an undefined result (e.g. different libp2p versions with different logic for the same strategy -- although this is bad), they request the multiplexing table from the peer, and accept it as valid. If both parties request the listing, the connection is terminated.

What case is this trying to cover?

> Alongside each entry, they signal a version agreement strategy (byte value), e.g. STRICT, SEMVER, ANY_VERSION

So, I'm not convinced about protocol versioning. Semver was designed for a centralized world where every versioned thing is maintained by a single party (linearized changes). However, most web protocols don't work that way; they evolve over time as different parties add different extensions/features. This is why browsers now use and recommend feature detection. For unavoidable breaking changes, I'd just change the protocol itself (e.g., `dht1`, `dht2`, etc.).

A flexible alternative is to allow an optional "user-data" object to be associated with a protocol advertisement. Now, the user-data *could* just be a version, but I'd recommend against that.

Really, I'd like to go further and split protocol/service advertisements (not really in-scope for *this* discussion but I guess all these systems tend to overlap). That is:

* Each side advertises *protocols* to (a) generate the contextual ID mappings and (b) obviate the need for protocol negotiation.
* Each side advertises *services* where each service may be accompanied by an optional service description "object" (opaque data). (I'd love to use IPLD for this but that may be pushing it).

For example, I may speak the relay protocol but may not be willing to relay connections for others. In this case, I'd advertise the relay protocol but not the service (or I could advertise the service but say that I only accept terminal connections in the service description).

vyzo:
> Nodes should be able to share a list of protocols they support during session establishment. 

We need a mechanism for update as well; it's not just dropping protocols, but also adding -- for instance the daemon adds protocol handlers dynamically. Furthermore, these protocols can be added after the connection has been established, so we need a push/update protocol to update the contextual tables.

multistream-2.0/spec.md
```diff
@@ -0,0 +1,403 @@
...
+This document *does not*, in fact, propose a protocol *negotiation* protocol.
+Instead, it proposes a set of stream/protocol management protocols that can be
+composed to flexibly negotiate protocols.
+
+First, this document proposes 4 protocol "negotiation" protocols. "Negotiation"
+is in quotes because none of these protocols actually involve negotiating
+anything.
+
+1. `multistream/advertise`: Inform the remote end about which protocols we speak
+   and. This should partially replace the current identify protocol.
```
vyzo: incomplete sentence. and? 

Is this the equivalent of `identify/push`?

multistream-2.0/spec.md
```diff
@@ -0,0 +1,403 @@
...
+stream multiplexer has been configured).
+
+1. `serial-stream`: A simple stream "multiplexer" that can multiplex multiple
+   streams *in serial* over the same connection. That is, it allows us to
+   negotiate a protocol, use it, and then return to multistream. It also allows
+   us to speculatively choose a single protocol and then drop back down to
+   multistream if that doesn't work.
+2. `speculative-stream`: A speculative stream "multiplexer" where the initiator
+   can speculatively initiate multiple streams and the receiver must select at
+   most one and discard the others.
```
vyzo: presumably there is  a mechanism to inform the peer of the actual selection.

multistream-2.0/spec.md
```diff
@@ -0,0 +1,403 @@
...
+2. To reduce RTTs, all protocols are unidirectional.
+3. These protocols were *also* designed to eventually support packet protocols
+   (the other reason to be unidirectional and a strong motivator for the
+   `speculative-stream` and `serial-stream` protocols).
+
+### Multistream Advertise
+
+Unspeced (for now). Really, we just need to send a mapping of protocol
+names/codecs to contextual IDs (and may be some service discovery information).
+Basically, identify.
```
vyzo: identify does quite a bit more: provides addresses (and updates), observed addrs, keys, etc... so we are not replacing it entirely, this is just the protocol mapping.

vyzo:
I am wondering a bit about the dynamic/contextual identifiers. It seems that these only come in play within user streams, which are themselves within the multiplexer. This means that the multiplexer will need to be informed of these protocol assignments itself.

vyzo:
It's not clear to me how the selection works with user protocols.
The user wants to open a stream using either the old protocol dht identifier or the new one, and this happens within the multiplexer.
Can we have an example on how this would work?
It's not clear how we save either bytes or RTTs in user protocol negotiation.

multistream-2.0/spec.md
Stebalien: Yes. Fixed.

Stebalien: > Is this the equivalent of identify/push?

Yes except:

1. It's designed to be significantly more efficient.
2. It only covers protocol advertisement. I'd like to try to avoid monolithic protocols in the future as small protocols like this are easier to mix/match/upgrade.

multistream-2.0/spec.md
Stebalien: Yes (described in the protocol description)

Stebalien:
> I am wondering a bit about the dynamic/contextual identifiers. It seems that these only come in play within user streams, which are themselves within the multiplexer. This means that the multiplexer will need to be informed of these protocol assignments itself.

Yes. That's one of the reasons this protocol set defines `multistream/advertise` instead of deferring to identify.

Basically, every new connection will initiate a "session" and all streams multiplexed over that connection will, by default, belong to the same session. They're pretty similar to unix process groups. All streams belonging to the same session will use the same contextual IDs.

I'm proposing binding these to *sessions* rather than *peers* as I'd rather not tie *state* to peers.

---

> It's not clear to me how the selection works with user protocols.

Usually, each side will have communicated which protocols it speaks using the advertise protocol. This way, you get a 0-RTT negotiation (because both sides already know which protocols the other speaks).

If they *haven't* communicated this information yet, for some reason, they can use the `speculative-stream` protocol to speak *both* protocols and let the other side decide which stream it wants to keep.

In practice, I don't expect to run into this case.

Stebalien:
This also ties into https://github.com/libp2p/specs/issues/32.

raulk:
>> Upfront negotiation
> So, technically, we already do this (identify). However, we don't currently wait for this negotiation to complete. Unfortunately, due to the lack of a serial-stream-like protocol, we have to wait to setup a stream multiplexer before we can run identify. With this proposal, we can piggy-back a protocol announcement along with the end of the crypto handshake.

I'd phrase this differently. Currently, we don't _negotiate_, we only _enumerate_. Both parties discover what their peer supports, but they agree on nothing. By introducing a deterministic intersection scheme, both parties can agree on the subset of protocols supported for that session, which allows them to select stream protocols with a number.

> State between reconnects feels like a bug waiting to happen.

Hmmm, I'm not so sure. QUIC supports session re-establishment, and if we intend to leverage it (which I assume we do), we'll need a resilient wire protocol that can detect when the state has changed on either end (e.g. protocols have been added or removed).

We can't always rely on parties dropping their cached QUIC sessions when their state changes. So by adding a merkle/hash-based verification, a QUIC reconnection could bypass the security handshake, confirm that the previous multiplexing table still applies, and go straight into live traffic.

> I mean, we could do the whole hash-dance (mmm, IPLD) however, a few packets is generally cheeper than a round-trip.

Could be, but if we're performing the same _big_ upfront protocol agreement dance upon reconnection over and over again, we're not leveraging the benefits of resumption at layer 4 (QUIC) in layer 7.

> The protocols may not be symmetric. Really, we should just think of them as "endpoints" rather than protocols.

You're right, and a recent case is AutoNAT, where peers can request dial backs from others, without offering them themselves. Bottom line: both peers understand AutoNAT, but only one of them offers a service.

The point about _protocols vs. endpoints_ is a good one. In Ethereum devp2p this distinction does not exist. A protocol is simply a set of messages both parties understand.

If we ported our asymmetric case to the devp2p model (for the sake of exploration), both peers would advertise AutoNAT, and they'd open a stream to exchange their capabilities (e.g. dialback enabled) as part of a AutoNAT `HELLO` message. I feel this style is more coherent with peer-to-peer philosophy, whereas in our model there's a vestige of client-server interaction.

This ties into what you said below in your comment regarding service advertisements. I'll think about this further 👍 

> What case is this trying to cover?

Future-proofing. I'm thinking that the logic behind agreement strategies can evolve, so libp2p v7 may specify an agreement strategy that libp2p v6 doesn't understand, and we need a fallback path for that case. The approach I floated may introduce an attack vector though. We may be better off versioning `multistream/advertise` itself, and creating a new version when the agreement strategies change.

> So, I'm not convinced about protocol versioning. Semver was designed for a centralized world where every versioned thing is maintained by a single party (linearized changes).

I agree. I would simplify to an int. So if a node supports both relay 01 and 02, it'd advertise: `/p2p/relay/{01,02}`. If a more modern node supports `/p2p/relay/{02,03}`, they would both agree on `02`. The good thing is that we remove the assumptions/obligations semver carries and in most cases, we even shave bytes in the serialised format. And a single protocol handler function can handle 01, 02, and 03.

> A flexible alternative is to allow an optional "user-data" object to be associated with a protocol advertisement.

IMO, this approach is suitable for node records in a DHT or a discovery gadget. It can get too heavy for in-band protocol negotiation, no?

In devp2p, protocol-specific capabilities are exchanged as part of a `HELLO` message for that protocol. So you only incur in that cost if you're actually interested in that protocol at all. The tradeoff is an extra RTT. With circuit relay, the dance would be as follows:

1. Node A and B exchange protocol advertisements containing `/p2p/relay`.
2. Node A and B exchange relay-specific `HELLO` messages containing their capabilities.

Stebalien:
So, this really looks like it comes down to endpoints versus protocols versus services:

* Endpoint: A "socket" or "port" you connect to. In our case, a named endpoint.
* Protocol: The language you speak over a stream. Multiple endpoints can speak the same protocol.
* Service: A service can listen on one or more endpoints to fulfill requests.

In libp2p, we really don't negotiate protocols. Instead, we just pick the (named) endpoint that we know speaks the right protocol.

In TCP, endpoints are ports. In HTTP, endpoints are paths.

>  Each party discovers what the peer supports, but they agree on nothing AFAIK.

Correct.

I guess this really is the protocol versus endpoint issue. We don't actually negotiate protocols, we advertise endpoints endpoints.

Up till now, there really hasn't been anything to *agree* on. Each side just picks the endpoint it wants to talk to and does its thing.

> By introducing a deterministic intersection scheme, both parties can agree on the subset of protocols supported for that session, which allows them to select stream protocols with a number.

So, we don't actually need to do that to use numbers. *This* spec just lets both sides pick the numbers they want the other side to use when talking to them. That's why I said they're effectively ephemeral ports where each side sends *their* ephemeral port mapping to the other.

This actually makes stream routing nice and simple because all incoming streams use *our* chosen IDs.

> > State between reconnects feels like a bug waiting to happen.

> Hmmm, I'm not so sure. QUIC supports session resumption, and if we intend to leverage it (which I assume we do), we'll need a resilient wire protocol that can detect when the state has changed on either end (e.g. protocols have been added or removed).

So, we'll still see QUIC session resumption as a new connection. Really, to apply the hash-check, we *need* to see this as a new connection (we need to know if we need to re-negotiate).

However, you're right. We will be storing state through reconnects anyways (e.g., QUIC sessions) and can probably allow peers (at their option) to pick up where they left off in terms of the currently supported protocols.

> I feel this style is more coherent with peer-to-peer philosophy, whereas in our model there's a vestige of client-server interaction.

I'd say that our system is a strict super set. That is, you *can* have both sides listen on the protocol/endpoint, even if they aren't willing to perform any *services*, but you don't have to.

> I may have missed it. Where do we memorise protocol indices? From what I see, neither BasicHost nor Multistream do this.

Sorry, I meant the first part, not the second. That is, if we successfully negotiate protocol A, we'll just *use* it in the future instead of negotiating. We don't assign numbers anywhere.

> In fact, I just discovered that each stream selects /multistream/1.0.0, hence requiring 2-RTT to open. Up until now, I had thought we negotiated multistream only once: when we opened the connection.

Yeah... Note, we *usually* (after we've shared our protocols), use a lazy negotiation that doesn't actually wait for anything. Unfortunately, this is actually unsound (https://github.com/multiformats/go-multistream/issues/20).

But it's even *worse* than that. We don't even *negotiate* multistream 1.0.0, if the other side asks for multistream 2.0.0, we abort instead of replying with "na". This is why *this* spec introduces the upgrade dance that negotiates multistream 1.0 before negotiating multistream 2.0.

> Future-proofing. I'm thinking that the logic behind agreement strategies can evolve, so libp2p v7 may specify an agreement strategy that libp2p v6 doesn't understand, and we need a fallback path for that case. The approach I floated may introduce an attack vector though. 

Got it. I guess I just still don't know why we actually *need* to agree (usually).

> I agree. I would simplify to an int. So if a node supports both relay 01 and 02, it'd advertise: /p2p/relay/{01,02}. If a more modern node supports /p2p/relay/{02,03}, they would both agree on 02. The good thing is that we remove the assumptions/obligations semver carries and in most cases, we even shave bytes in the serialised format. And a single protocol handler function can handle 01, 02, and 03.

So my question is, at *this* layer, can we just treat those versions as opaque strings and the endpoints as distinct (and let the peers figure it out on a protocol by protocol basis)?

From my perspective, we're trying to solve the issue where we have some number of mutually exclusive endpoints/protocols. However, mutual exclusion may have nothing to do with monotonically increasing version numbers. For example, yamux versus mplex.

> IMO, this approach is suitable for node records in a DHT or a discovery gadget. It can get too heavy for in-band protocol negotiation, no?

So, one of my motivations here was distinguishing between "service" and "endpoint". That is, a single service may want to expose multiple endpoints but it's still a single logical "service". Really, we could probably deduce services based on endpoints/protocols and then just say hello as devp2p does (or even have both sides expose `/servicename/hello` endpoints).

Regardless, we don't have to tackle this here.

marten-seemann:
The problem of assigning code points to protocol descriptor strings reminds me of HPACK (or QPACK, for QUIC). Maybe I’m just hanging out too much with the HTTP folks in the QUIC Working Group, but the problems actually looks similar: In HTTP, you have HTTP header values that are sent with every request and every response, and HPACK compresses them by assigning code points and only sending these code points. HPACK has the additional nice property that it takes care of a DoS, where one peer floods the other peer with new mappings by evicting old values from that table.

I’m wondering if it make sense to build a static table of protocol identifiers here. There are certain protocols that every peer would support (ping, identify, etc.) and protocols that most peers would support. It seems wasteful to send all those identifiers at the beginning of every connection. However, we’d probably want to version the static table (and kind of run it like an IANA registry). Building such a table would require us to disentangle the two functions of `multistream/advertise`: 1. the announcement of the mapping of protocol name => code point and 2. declaring which of these protocols are actually supported.

The problem of session resumption is a hard one. It seems like potentially every session will arrive at a unique table, depending on the order that peers register new protocols. If we adopt @raulk's suggestion to identify these tables by a hash (not sure how a Merkle-tree based solution would look like), one could imagine each peer keeping a mapping of hash => table around for some time. However, what happens when such a mapping is evicted? We'd then have to fall back to renegotiating, which will cost us one RTT. Assuming that the mapping itself fits into a few packets, we might actually be better of NOT storing any table, and just renegotiate every time (trading bandwidth for latency).

multistream-2.0/spec.md
```diff
@@ -0,0 +1,403 @@
...
+* [ ] Q: Technically, the first multicodec is redundant. However, it acts as a
+      magic byte that allows us to figure out what's going on. Should we keep
+      it? We could just start all streams with a single multicodec representing
+      the protocol
+* [ ] Q: Should we somehow distinguish between initiator and receiver? Should we
+      distinguish between bidirectional and unidirectional? We could even bit
+      pack these options into a single byte and use this instead of the leading
+      multicodec...
+
+### Multistream Dynamic
```
marten-seemann: Do we really have a use case for this? Or wouldn't it be easier for peers to first announce a new code point, and then use it with `multistream/use`?

multistream-2.0/spec.md
```diff
@@ -0,0 +1,403 @@
...
+   an issue as we can run as many sub-protocols over the same stream as we want.
+2. To reduce RTTs, all protocols are unidirectional.
+3. These protocols were *also* designed to eventually support packet protocols
+   (the other reason to be unidirectional and a strong motivator for the
+   `speculative-stream` and `serial-stream` protocols).
+
+### Multistream Advertise
+
+Unspeced (for now). Really, we just need to send a mapping of protocol
+names/codecs to contextual IDs (and may be some service discovery information).
```
marten-seemann: We also need the option to add a mapping later on, right?
I'm ok with not specifying the wire format (for now), but we should define **exactly** what we expect to be getting out of `multistream/advertise`.

tomaka:
> Assuming that there are no other use cases I forget in my list above, we'd add a whole lot of complexity just to smoothen the transition from secio to TLS.

In practice you also need WebRTC/WebSockets for inside browsers.
Also, that would assume we're never going to switch away from QUIC. From what I understand, one of the major selling points of libp2p is its modularity.

marten-seemann:
> > Assuming that there are no other use cases I forget in my list above, we'd add a whole lot of complexity just to smoothen the transition from secio to TLS.
>
> In practice you also need WebRTC/WebSockets for inside browsers.
Also, that would assume we're never going to switch away from QUIC. From what I understand, one of the major selling points of libp2p is its modularity.

I was talking about the transition to TLS/TCP, not about QUIC.

Can you elaborate on WebRTC/WebSockets? I'm not familiar with how these are used in web browsers. How would `serial-stream` and `speculative-stream` be applied there?

multistream-2.0/spec.md
```diff
@@ -0,0 +1,403 @@
...
+In this protocol, the stream initiator sends a varint ID specified by the
+*receiver* to the receiver. This is a *unidirectional* protocol.
+
+Format:
+
+```
+<multistream/use><multistream/contextual><id(varint)>
+```
+
+The ID 0 is reserved for saying "same protocol" on a bidirectional stream. The
```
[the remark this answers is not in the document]
Stebalien: We should probably just say "ID 0 means YES".

The issues are:

1. These protocols all have unidirectional variants.
2. Who initiated the stream isn't, unfortunately, always unambiguous.
3. These contextual IDs are *relative* and, even if you know my contextual ID for a protocol, I may not know yours.

Given these three issues:

1. I can't use *your* ID because I may not know it.
2. I can't use *mine* because your multistream muxer is expecting *your* IDs.

Therefore, I'm using 0.

However, there's probably a better way to say this.

multistream-2.0/spec.md
```diff
@@ -0,0 +1,403 @@
...
+using different number formats everywhere *will* cost us in terms of complexity.
+
+### Multistream Contextual
+
+The `multistream/contextual` protocol is used to select a protocol using a
+*receiver specified*, session-ephemeral protocol ID. These IDs are analogues of
+ephemeral ports.
+
+In this protocol, the stream initiator sends a varint ID specified by the
+*receiver* to the receiver. This is a *unidirectional* protocol.
```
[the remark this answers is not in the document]
Stebalien: That is, *all* of these protocols can be run on unidirectional streams. I'll just get rid of those statements. They used to be important because some of these protocols worked over unidirectional steams and others didn't but that's not the case anymore.

multistream-2.0/spec.md
Stebalien: Privacy, flexibility, feature parity with multistream 1.0. For example, I might listen on a non-enumerable set of protocols.

Concretely: We can use multistream to connect to peers within peers by connecting to the multistream endpoint `/p2p/InnerPeerId`.

> Or wouldn't it be easier for peers to first announce a new code point, and then use it with multistream/use

Note: multistream/use is for multicodecs. For custom protocols, we'd use multistream/contextual.

multistream-2.0/spec.md
Stebalien: I agree. We need a way to add/remove mappings (although we shouldn't reuse them within a session).

multistream-2.0/spec.md
```diff
@@ -0,0 +1,403 @@
...
+   `multistream/advertise` To inform the initiator of *it's* mapping between
+   protocols and contextual IDs.
+
+Second, this document proposes 2 auxiliary protocols that can be used with the 4
+multistream protocols to actually negotiate protocols. These are *primarily*
+useful (a) in packet-based protocols (without sessions) and (b) when initially
+negotiating a transport session (before protocols have been advertised and the
+stream multiplexer has been configured).
+
+1. `serial-stream`: A simple stream "multiplexer" that can multiplex multiple
```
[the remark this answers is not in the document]
Stebalien: The idea is to define a simple multiplexer that every single implementation will support, forever. This is designed to be the *dumbest*, *simplest*, *worst* multiplexer in the world so we can have *some* multiplexer while we negotiate a *real* multiplexer.

multistream-2.0/spec.md
```diff
@@ -0,0 +1,403 @@
...
+Each message in this protocol consists of:
+
+```
+<stream number (varint)>
+<data length (varint)>
+```
+
+The initiator can transition to a single one of these streams by sending:
+
+```
```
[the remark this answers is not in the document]
Stebalien: Ah, damn, that was wrong. It should have said *receiver*. That is, all streams remain live until the receiver selects one.

multistream-2.0/spec.md
```diff
@@ -0,0 +1,403 @@
...
+The protocol is:
+
+```
+<header (signed 16 bit int)>
+<body>
+```
+
+Where the header is:
+
+* -2 - Send a reset and return to multistream. All queued data (remote and
```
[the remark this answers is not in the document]
Stebalien: > edit: or is this to say -2 is a reset?

Yes.

Stebalien:
> I’m wondering if it make sense to build a static table of protocol identifiers here.

We have these via multicodecs and `multistream/use`.

> I'm wondering if we can find a different term than "stream" for serial-stream and speculative-stream. It's kind of confusing to have multiple layers of streams. Maybe we should reserve "stream" for the streams that actually carry application data (i.e. the yamux/mplex/spdy/QUIC streams).

I agree but I can't think of any better names.

I originally called these `serial-multiplex` and `multistream/choose` respectively. I kind of like `serial-multiplex` but wanted the two protocol names to match.

> Assuming that there are no other use cases I forget in my list above, we'd add a whole lot of complexity just to smoothen the transition from secio to TLS.

This, unfortunately, means we'd have a *hard* dependency on:

1. Security transports with built-in protocol negotiation (and some way to advertise supported protocols).
2. Putting the security protocol in the Miltiades (or only supporting one).

However, these really aren't assumptions we can make in practice. We *want* to be able to support arbitrary security transports and will need to run multiple security transports on the same port/endpoint.

For example, relay. When I connect to a peer over relay, we start out with a plain-text channel and need to negotiate the security transport.

So, we need *something*. However, we can probably simplify/replace the speculative protocol (I just kind of like it from a design standpoint).

Stebalien:
Thinking through this, the speculative stream protocol isn't actually going to work as described because there's no clear cutoff by which the receiver has to pick a stream.

(also, it should have said that the receiver picks the stream, not the initiator)

I believe we need to either:

1. Have some way to say "I have nothing more to say until we pick a final stream".
2. Restrict the protocol to one "packet" of speculative data. Currently, we technically allow data to flow in both directions indefinitely before the receiver needs to "choose" a stream. However, we could say that the initiator gets to send one batch of speculative data and the receiver needs to pick after receiving this batch, no round trips.

The second solution simplifies the protocol significantly but it *does* remove some a nice use-case. For example, the current protocol (with fix 1) would allow us to negotiate multiple protocols with multiple round-trips on the two speculative streams before finally picking one. The second solution force us to pick up-front.

Stebalien:
Ok, I've dropped speculative stream for now. I agree it's *really* complicated and doesn't pull it's weight. It may cost us a round-trip occasionally (when upgrading the network) but that's not the end of the world.

We can always introduce it later (using serial-stream to speculatively use speculative-stream).

Stebalien:
Questions from discussion concerning serial-stream:

1. Should we just *always* use TLS? TLS allows sending additional information (e.g., an advertise packet).
2. Should we use out-of-band information to avoid uncertainty entirely? That is, should we just assume that we know what security/multiplex protocol the peer speaks.

Questions with respect to advertise:

1. Should we remember our peer's protocols instead of re-advertising each time? Can we just do this as a separate protocol?

multistream-2.0/spec.md
```diff
@@ -0,0 +1,319 @@
...
+  local) should be discarded.
+* -1 - Close: Send an EOF and return to multistream.
+*  0 - Rest: Ends the reuse protocol, transitioning to a direct stream.
+* >0 - Data: The header indicates the length of the data.
```
jhiesey: `>0` is interpreted as formatting, not the literal `>` character

multistream-2.0/spec.md
```diff
@@ -0,0 +1,319 @@
...
+Instead, it proposes a set of stream/protocol management protocols that can be
+composed to flexibly negotiate protocols.
+
+First, this document proposes 4 protocol "negotiation" protocols. "Negotiation"
+is in quotes because none of these protocols actually involve negotiating
+anything.
+
+1. `multistream/advertise`: Inform the remote end about which protocols we
+   speak. This should partially replace the current identify protocol.
+2. `multistream/use`: Selects the stream's protocol using a multicodec.
```
jhiesey: I'm not a fan of these names. How about `multistream/use-muilticodec`, `multistream/use-dynamic`, `multistream/use-contextual`?

multistream-2.0/spec.md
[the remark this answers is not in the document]
jhiesey: So the distinction is that -2 is an abnormal end and -1 is a normal end?

multistream-2.0/spec.md
```diff
@@ -0,0 +1,319 @@
...
+`serial-stream`. This combination will allow us to apply a series of quick
+connection upgrades (e.g., to multistream 3.0) with no round trips and no funny
+business (learn from past mistakes).
+
+Notes:
+
+1. The "ls" feature of multistream has been removed. While useful, this really
+   should be a *protocol*. Given the `serial-stream` protocol, this shouldn't be
+   an issue as we can run as many sub-protocols over the same stream as we want.
+2. To reduce RTTs, all protocols are unidirectional.
```
jhiesey: What does this mean in practice?

multistream-2.0/spec.md
Stebalien: Renamed use -> multicodec, dynamic -> string, contextual -> dynamic.

multistream-2.0/spec.md
Stebalien: Updated.

multistream-2.0/spec.md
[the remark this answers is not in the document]
Stebalien: Yes. Updated.

jhiesey:
LGTM. @whyrusleeping your thoughts?

multistream-2.0/spec.md
```diff
@@ -0,0 +1,390 @@
...
+<len>/multistream/2.0.0\n
+```
+
+in which case it'll switch to multistream 2.0.
+
+Importantly: When we switch to multistream 2.0, we'll tag the connection (and
+any sub connections) with the multistream version. This way, we never have to do
+this again.
+
+## TCP Simultaneous Open
```
Stebalien: @vyzo, @jhiesey 

This is my proposal for handling simultaneous open. TL;DR: The connection has two unidirectional streams until they're joined.

@vyzo IIRC, this is slightly different from the protocol we discussed as we *always* use it, even if we have no reason to believe we performed a simultaneous connect.

multistream-2.0/spec.md
```diff
@@ -187,6 +187,75 @@ Importantly: When we switch to multistream 2.0, we'll tag the connection (and
...
+1. The receiver sends: `<multistream/multicodec><duplex-stream>1<the same 32 bytes of randomness>`
+
+If we end up in a situation where both peers want to be the initiator of a
+single pair of unidirectional streams, the peer that picks the *lower* random ID
+should back off and act as the receiver.
+
+### Usage
+
+We treat each new TCP connection as a pair of unidirectional streams and use
+this protocol to bind them together.
```
vyzo: I think we need full detail on the wire for this, all the way up to secio negotiation (both cases).

multistream-2.0/spec.md
vyzo: LGTM but we need more wire detail on the examples.

So let's have two examples with full protocol detail, one for the common case and one for simultaneous open.

Stebalien: SGTM.

multistream-2.0/spec.md
```diff
@@ -0,0 +1,412 @@
...
+### Serial Stream
+
+The `serial-stream` protocol is the simplest possible stream multiplexer.
+Unlike other stream multiplexers, `serial-stream` can only multiplex streams
+in *serial*. That is, it has to close the current stream to open a new one.
+
+The protocol is:
+
+```
+<header (signed 16 bit int)>
```
Kubuxu: Endianess of this integer should be defined or endianness of all integers in the document should be defined.

At this time I'm assuming network byte order/big-endian.

Stebalien: Yes. Network order.

multistream-2.0/spec.md
```diff
@@ -0,0 +1,412 @@
...
+
+1. The side that wants to be the "initiator" of a duplex stream sends a
+   "initiate stream ID" message (where ID is randomly generated 256 bit number).
+2. The receiver sends back "receive stream ID" on a different unidirectional stream.
+3. The two streams are now joined.
+
+More specifically,
+
+1. The initiator generates a 32 byte random ID (`ID`).
+2. The initiator negotiates the `duplex-stream` protocol and then sends `0<ID>` (`0` is a single 0 byte).
```
Kubuxu: `0x00<id>`  `0x01<id>`
Hex numbers make it obvious that it is a single byte.

raulk:
Multistream is foundational in libp2p and I'm keen to iterate on it. Here are further considerations after letting the proposal sink in for a few weeks:

1. We commonly refer to multistream as a multiplexer. This leads to confusion, as it's really a _protocol selector_: nothing in multistream 1.0/2.0 enables parallel streams or conversations over the same conduit, _per se_ (without the involvement of a real multiplexer like yamux).
    * I wouldn't call `serial-stream` a multiplexer, rather an _interactive selection_ protocol.
    * Not wanting to start a naming war, but to me `multiselect 2.0` is a more precise term to capture what's happening here.

2. It's unclear to me how implementations are expected to decide which protocols to use, when, what the API will look like, and if these mechanisms will be exposed to the user at all. I think we need a normative choreography in the spec to make this actionable.

3. Often it's necessary to open a stream for a protocol that's already been selected before during a the lifetime of a session (e.g. imagine a stream pool for kad-dht). We should make this very efficient, without incurring in the cost of an upfront advertisement (`multistream-{dynamic,advertise}` pair).
    * Proposed solutions: implementations track unique protocol agreements throughout a session, and assign them a 2-byte index starting from 0. Via a `multistream-index` protocol, any party can open a new stream for a pre-used protocol by sending its 2-byte index.

4. Peers MAY store tables transmitted through `multistream-advertise`. Upon a future reconnection, they can bypass the cost of sending the table again by serialising the last known state for the counterparty (we must define a canonical format, protobuf map?), digesting it and sending a multihash. The party can ACK, or NACK by sending the updated protocol table if the older one no longer stands.

5. Centralising protocol definitions under multicodec looks unscalable. Consequently I expect `multistream-codec` to be underused in the wild (except for maybe the protocols the libp2p community maintains). Even if we went this path, it's unclear how protocol versioning would work (we certainly wouldn't assign a multicodec for each version).

6. We should cover how multi~stream~select 2.0 deals with out-of-band service advertisements. Right now we assume pure uncertainty (not knowing which protocols the other party knows). We should define an interface such that discovery drivers can feed known services to multistream. 
    * This is just around the corner, with Ethereum 2.0 developing discovery v5 on libp2p (which includes support for [Ethereum Node Records](https://eips.ethereum.org/EIPS/eip-778) that carry service advertisements), Bluetooth experimentation by @tomaka (i.e. Bluetooth SDP), and mDNS interest.

bigs:
@Stebalien @raulk Added a proposal in `packet-oriented.md` for some extensions, modifications for the packet oriented use case. If this isn't the appropriate venue, I'm happy to move!

PR #96, opened 2018-10-10
Add specs for mplex

tomaka:
The first commit copy-pastes the specs from https://github.com/libp2p/mplex
This repo should be deleted after this PR is merged.

The second commit adds my changes.

Most notably, I tried to clarify some ambiguities remaining after https://github.com/libp2p/mplex/pull/3

cc @Stebalien

### DIFF
--- a/mplex/README.md
+++ b/mplex/README.md
@@ -2,7 +2,7 @@
 
 > This repo contains the spec of mplex, the friendly Stream Multiplexer (that works in 3 languages!)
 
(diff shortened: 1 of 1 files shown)


Stebalien:
I actually added this here (https://github.com/libp2p/specs/tree/master/mplex) during the spec writing session. I couldn't do it though a PR as I had to merge unrelated git histories. Can you base your second commit off of that?

PR #97, opened 2018-10-10
identify spec

vyzo:


identify/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+This is the public key of the peer, marshalled in binary form as specicfied
+in [peer-ids](../peer-ids).
+
+
+### listenAddrs
+
+These are the addresses on which the peer is listening as multi-addresses.
+
+### observedAddr
```
tomaka: There are two possibilities here: this can be either the stack of protocols the listener used in order to reach back the dialer, or the stack of protocols we think the dialer used to reach us.
For the sake of forward compatibility, I think it'd be a nice idea to define this more precisely, and I'd be in favour of the former.

vyzo: Added a `source` disambiguator, per discussion in meatspace.

tomaka: Per discussion I'm not satisfied with the wording, because I'd like to disambiguate what is returned when `p2p-circuit` is used, but I can't find any appropriate wording.

vyzo: We can add an example for circuit addresses specifically, as they seem to be the contentious issue.

vyzo: Added some examples, hopefully it's clearer now.

identify/README.md
```diff
@@ -0,0 +1,56 @@
...
+the existence of a NAT and its public address.
+
+For example, in the case of a TCP/IP transport the observed addresses will be of the form
+`/ip4/x.x.x.x/tcp/xx`. In the case of a circuit relay connection, the observed address will
+be of the form `/p2p/QmRelay/p2p-circuit`. In the case of onion transport, there is no
+observable source address.
+
+### protocols
+
+This is a list of protocols supported by the peer.
```
raulk: not ordered in any specific manner.

identify/README.md
```diff
@@ -0,0 +1,56 @@
+# Identify v1.0.0
+
+The identify protocol is used to identify a peer and its capabilities.
```
raulk: can we unpack the definition of "identify a peer"? the recursive definition is confusing for a spec

identify/README.md
```diff
@@ -0,0 +1,56 @@
...
+  repeated bytes listenAddrs = 2;
+  optional bytes observedAddr = 4;
+  repeated string protocols = 3;
+}
+```
+
+### protocolVersion
+
+The protocol version identifies the family of protocols used by the peer.
+The current protocol version is `ipfs/0.1.0`; if the protocol does not match
```
raulk: with the current logic, the micro/patch version can be different. only major and minor need to be equal.

identify/README.md
```diff
@@ -0,0 +1,56 @@
...
+the protocol used by the initiating peer, then the connection is considered
+unusable and the peer must close the connection.
+
+### agentVersion
+
+This is a free-form string, identifying the implementation of the peer.
```
raulk: can we encourage a particular format, _even if non-normative_? the default is `go-libp2p/<semver>`, and IPFS changes this to `"go-ipfs/" + version.CurrentVersionNumber + "/" + version.CurrentCommit`.

https://github.com/ipfs/go-ipfs/blob/master/core/core.go#L99

identify/README.md
vyzo: right, but do we need to say that?

identify/README.md
vyzo: uhm ok; me loves recursion.

identify/README.md
vyzo: nah, let's let people call their agents whatever they want!

vyzo: although an example might be useful.

identify/README.md
vyzo: i really can't think of a non-verbose way of expressing this.

identify/README.md
raulk: example sounds good

identify/README.md
vyzo: it's a list on the wire, there is no set datatype.

identify/README.md
vyzo: query, not interrogate.

identify/README.md
vyzo: yeah, and it's not really any better than the current wording.

identify/README.md
dryajov: might be worth clarifying that this are multistream protocols (or protocol strings)?

identify/README.md
```diff
@@ -0,0 +1,58 @@
...
+### publicKey
+
+This is the public key of the peer, marshalled in binary form as specicfied
+in [peer-ids](../peer-ids).
+
+
+### listenAddrs
+
+These are the addresses on which the peer is listening as multi-addresses.
+
```
richardschneider: Does the multiaddress also include `/ipfs/Qm...`?  If so, is `/p2p` also supported?

Stebalien: It doesn't include it. Well, it shouldn't.

daviddias:
Please make sure to take https://github.com/libp2p/js-libp2p-switch/issues/78 into account

yusefnapora:
@vyzo this looks pretty solid to me - should we merge?

I don't have the context behind @daviddias comment about the js / go interop, so maybe we need to address the changes made there?

vyzo:
We also have `identify/push` now, which is the active update variant of the protocol and should be in the spec.

vyzo:
@yusefnapora feel free to edit!

yusefnapora:
@vyzo I added a bit about identify/push - lmk if it looks alright. If so, I think we should press the button 😃

yusefnapora:
hehe, fair enough :) I'll let you do the honor of pressing the big green button then

vyzo:
Any objects to merging this?

identify/README.md
```diff
@@ -0,0 +1,79 @@
...
+When a peer's basic information changes, for example, because they've obtained a new
+public listen address, they can use `identify/push` to inform others about the new
+information.
+
+The push variant works by opening a stream to each remote peer you want to update, using
+`/ipfs/id/push/1.0.0` as the protocol id string. When the remote peer accepts the stream,
+the local peer will send an `Identify` message and close the stream.
+
+## The Identify Message
```
Stebalien: We should probably note that any missing fields should be ignored (so we can do a partial identify push).

vyzo: @yusefnapora want to add it?

vyzo:
Alright, let's merge!

PR #100, opened 2018-10-10
docs: add peer id spec

mgoelzer:
This updates the peer ID spec to explain what keypairs are supported and how peer IDs are encoded  for each key type.  Thanks to @Stebalien for figuring this out with me.


mgoelzer:
In the spirit of learning and spreading knowledge, I have chosen @jhiesey and @raulk as reviewers of this PR.

peer-ids/peer-ids.md
```diff
@@ -3,7 +3,7 @@
 ## Keys
 
 
-Everything is based on a very simple protobuf in [libp2p/go-libp2p-crypto/pb/crypto.proto#L5](https://github.com/libp2p/go-libp2p-crypto/blob/master/pb/crypto.proto#L5):
+Our key pairs are transmitted on the wire using a simple protobuf defined in [libp2p/go-libp2p-crypto/pb/crypto.proto#L5](https://github.com/libp2p/go-libp2p-crypto/blob/master/pb/crypto.proto#L5):
```
vyzo: the private key is _not_ transmitted on the wire; this is for saving the keys to disk.

peer-ids/peer-ids.md
```diff
@@ -0,0 +1,95 @@
+# Spec:  Peer Ids and Keys
+
+## Keys
+
+
+Our key pairs are transmitted on the wire using a simple protobuf defined in [libp2p/go-libp2p-crypto/pb/crypto.proto#L5](https://github.com/libp2p/go-libp2p-crypto/blob/master/pb/crypto.proto#L5):
+
+```
```
vyzo: perhaps add a `protobuf` after the backticks so that it is syntax highlighted.

mgoelzer: Good point - thanks!  Fixed

Stebalien:
We should probably say: Implementations SHOULD support RSA and Ed25519. Implementations MAY support Secp256k1 and ECDSA but nodes using those keys may not be able to connect to all nodes.

marten-seemann:
+1 for using normative language

whyrusleeping:
Any update here? I'd love to have some specs around here to link to ;)

peer-ids/peer-ids.md
```diff
@@ -0,0 +1,117 @@
...
+## Peer Ids
+
+Here is the process by which we generate peer id's based on the public/private keypairs described above:
+
+  1. Encode the public key into the protobuf.
+  2. Serialize the protobuf containing the public key into bytes using the [canonical protobuf encoding](https://developers.google.com/protocol-buffers/docs/encoding).
+  3.  If the length of the serialized bytes <= 42, then we compute the "identity" multihash of the serialized bytes.  In other words, no hashing is performed, but the [multihash format is still followed](https://github.com/multiformats/multihash) (byte plus varint plus serialized bytes).  The idea here is that if the serialized byte array is short enough, we can fit it in a multihash proto without having to condense it using a hash function.
+  4. If the length is >42, then we hash it using it using the SHA256 multihash.
+
+For more information, refer to this block in [libp2p/go-libp2p-peer/peer.go](https://github.com/libp2p/go-libp2p-peer/blob/master/peer.go):
```
marten-seemann: Same here. I think the text already describes the logic pretty well, so we don't need to cite this comment.

peer-ids/peer-ids.md
```diff
@@ -0,0 +1,117 @@
...
+
+Four key types are supported:
+ - RSA
+ - Ed25519
+ - Secp256k1
+ - ECDSA
+
+Implementations SHOULD support RSA and Ed25519. Implementations MAY support Secp256k1 and ECDSA, but nodes using those keys may not be able to connect to all other nodes.
+
+Keys are passed around in code as byte arrays.  Keys are encoded within these arrays differently depending on the type of key.
```
marten-seemann: That seems like an implementation decision. Remove this sentence?

peer-ids/peer-ids.md
```diff
@@ -0,0 +1,117 @@
...
+
+### RSA
+
+We encode the public key using the DER-encoded PKIX format.
+
+We encode the private key as a PKCS1 key using ASN.1 DER.
+
+To sign a message, we first hash it with SHA-256 and then sign it using the RSASSA-PKCS1-V1.5-SIGN from RSA PKCS#1 v1.5.
+
+See [libp2p/go-libp2p-crypto/rsa.go](https://github.com/libp2p/go-libp2p-crypto/blob/master/rsa.go) for details
```
marten-seemann: Same here.

peer-ids/peer-ids.md
```diff
@@ -0,0 +1,117 @@
...
+We do not do any special additional encoding for Ed25519 public keys.
+
+The encoding for Ed25519 private keys is a little unusual. There are two formats that we encourage implementors to support:
+
+ - Preferred method is a simple concatenation:  `[private key bytes][public key bytes]` (64 bytes)
+ - Older versions of the libp2p code used the following format:  `[private key][public key][public key]` (96 bytes).  If you encounter this type of encoding, the proper way to process it is to compare the two public key strings (32 bytes each) and verify they are identical.  If they are, then proceed as you would with the preferred method.  If they do not match, reject or error out because the byte array is invalid.
+
+Ed25519 signatures follow the normal Ed25519 standard.
+
+See [libp2p/go-libp2p-crypto/ed25519.go](https://github.com/libp2p/go-libp2p-crypto/blob/master/ed25519.go) for details
```
marten-seemann: And here.

peer-ids/peer-ids.md
```diff
@@ -0,0 +1,117 @@
...
+
+### ECDSA
+
+We encode the public key using ASN.1 DER.
+
+We encode the private key using DER-encoded PKIX.
+
+To sign a message, we hash the message with SHA 256, and then sign it with the ECDSA standard algorithm, then we encode it using DER-encoded ASN.1.
+
+See [libp2p/go-libp2p-crypto/ecdsa.go](https://github.com/libp2p/go-libp2p-crypto/blob/master/ecdsa.go) for details.
```
marten-seemann: And here.

peer-ids/peer-ids.md
```diff
@@ -0,0 +1,117 @@
...
+
+See [libp2p/go-libp2p-crypto/ed25519.go](https://github.com/libp2p/go-libp2p-crypto/blob/master/ed25519.go) for details
+
+### Secp256k1
+
+We use the standard Bitcoin EC encoding for Secp256k1 public and private keys.
+
+To sign a message, we hash the message with SHA 256, then sign it using the standard Bitcoin EC signature algorithm (BIP0062), and then use standard Bitcoin encoding.
+
+See [libp2p/go-libp2p-crypto/secp256k1.go](https://github.com/libp2p/go-libp2p-crypto/blob/master/secp256k1.go) for details.
```
marten-seemann: And here.

peer-ids/peer-ids.md
```diff
@@ -0,0 +1,117 @@
...
+ - Preferred method is a simple concatenation:  `[private key bytes][public key bytes]` (64 bytes)
+ - Older versions of the libp2p code used the following format:  `[private key][public key][public key]` (96 bytes).  If you encounter this type of encoding, the proper way to process it is to compare the two public key strings (32 bytes each) and verify they are identical.  If they are, then proceed as you would with the preferred method.  If they do not match, reject or error out because the byte array is invalid.
+
+Ed25519 signatures follow the normal Ed25519 standard.
+
+See [libp2p/go-libp2p-crypto/ed25519.go](https://github.com/libp2p/go-libp2p-crypto/blob/master/ed25519.go) for details
+
+### Secp256k1
+
+We use the standard Bitcoin EC encoding for Secp256k1 public and private keys.
```
marten-seemann: A link to the document where this standard encoding is defined would be helpful.

peer-ids/peer-ids.md
```diff
@@ -0,0 +1,117 @@
...
+
+See [libp2p/go-libp2p-crypto/rsa.go](https://github.com/libp2p/go-libp2p-crypto/blob/master/rsa.go) for details
+
+### Ed25519
+
+Ed25519 specifies the exact format for keys and signatures, so we do not do much additional encoding, except as noted below.
+
+We do not do any special additional encoding for Ed25519 public keys.
+
+The encoding for Ed25519 private keys is a little unusual. There are two formats that we encourage implementors to support:
```
marten-seemann: This seems like an implementation decision, so we probably don't need to specify it.

peer-ids/peer-ids.md
```diff
@@ -0,0 +1,117 @@
...
+	required bytes Data = 2;
+}
+
+message PrivateKey {
+	required KeyType Type = 1;
+	required bytes Data = 2;
+}
+```
+
+As should be apparent from the above code block, this proto simply encodes for transmission a public/private key pair along with an enum specifying the type of keypair.
```
marten-seemann: Is there any situation where we want to transmit the `PrivateKey`? That seems... dangerous. If not, we don't need to specify the `PrivateKey` here at all.

raulk: Yeah, storage of private key is implementation specific, so no need to cover them in this doc I think.

raulk:
@yusefnapora if you wanna do some spec herding, this is a quick win I think. Pretty good consensus.

peer-ids/peer-ids.md
```diff
@@ -0,0 +1,117 @@
...
+
+## How Keys are Encoded and Messages Signed
+
+Four key types are supported:
+ - RSA
+ - Ed25519
+ - Secp256k1
+ - ECDSA
+
+Implementations SHOULD support RSA and Ed25519. Implementations MAY support Secp256k1 and ECDSA, but nodes using those keys may not be able to connect to all other nodes.
```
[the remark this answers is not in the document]
marten-seemann: I think SHOULD is correct here, because might be reasons not to support them, e.g. security or performance considerations.
If an implementer really wants to produce a useless implementation, he'll find plenty of ways to do so...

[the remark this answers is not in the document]
raulk: That lack of definition is not good for a spec. If you leave this open ended, one implementor could choose to support RSA only, another Ed25519 only, and their implementations won't interoperate. Choosing _your_ private key type based on arbitrary criteria is a different topic than interoperability.

yusefnapora:
I did a quick edit to remove references to private keys and serialization on disk. I also removed links to go code and added some links to e.g. the RSA signing spec, etc.

@mgoelzer do you mind if I push changes to this branch? I put up a new one here with the edits: https://github.com/libp2p/specs/blob/edit/peer-ids/peer-ids/peer-ids.md

but it might be easier to discuss if I drop the commits here.

raulk:
@yusefnapora gonna jump in and say yes. In the interest of moving forward, push to the branch in this PR. Thanks!

peer-ids/peer-ids.md
```diff
@@ -0,0 +1,117 @@
...
+The second is for generating peer ids; this is discussed in the section below.
+
+## Peer Ids
+
+Here is the process by which we generate peer id's based on the public/private keypairs described above:
+
+  1. Encode the public key into the protobuf.
+  2. Serialize the protobuf containing the public key into bytes using the [canonical protobuf encoding](https://developers.google.com/protocol-buffers/docs/encoding).
+  3.  If the length of the serialized bytes <= 42, then we compute the "identity" multihash of the serialized bytes.  In other words, no hashing is performed, but the [multihash format is still followed](https://github.com/multiformats/multihash) (byte plus varint plus serialized bytes).  The idea here is that if the serialized byte array is short enough, we can fit it in a multihash proto without having to condense it using a hash function.
+  4. If the length is >42, then we hash it using it using the SHA256 multihash.
```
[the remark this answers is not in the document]
yusefnapora: I added a bit about base58btc, but didn't mention multibase, since we hadn't defined it yet in the doc. Should I bring it up? I think if people are likely to expect Peer Ids to use multibase we should clarify.

peer-ids/peer-ids.md
Stebalien: Not entirely. We do want users to be able to port keys from one implementation to another.

peer-ids/peer-ids.md
Stebalien: Unfortunately, users do need to be able to take their private keys with them (especially because we use these for things like IPNS).

yusefnapora: It's true that removing the private key format from this doc leaves a gap. We still need to specify somewhere how we handle them.

We could bring back the private key references and add a call-out at the top of the doc that they're not related to peer-id calculation and are shown for reference.

Stebalien: Really, we should probably rename this doc to the "libp2p key spec" and make peer ID calculation a part of that.

yusefnapora: 👍 for that

peer-ids/peer-ids.md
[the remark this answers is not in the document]
yusefnapora: Maybe a libp2p language implementation MUST support RSA and Ed25519, but we call out that an individual libp2p application or network can make its own call on what to support. So if you have security or performance concerns for your deployment, you can disable RSA and Ed25519 and use whatever you need.

If a language implementation doesn't happen to support both of the baseline keys, that's fine, but we would consider it a missing feature. Whereas we'd consider an implementation without Secp256k1 "feature complete", since that's not a MUST.

Either way, we should make it clear that not supporting them will limit your interoperability. Which might be fine for some use cases, as @marten-seemann says.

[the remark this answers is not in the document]
raulk: I think overcomplicating this. These are specs for **implementors**; what the application using an implementation chooses or not to do is irrelevant here. 

My gripe is that, as an implementor, I look at this statement and it's totally useless because I don't know what to do (MUST), and where I'm given optionality (SHOULD/COULD), I don't know what the implications are.

peer-ids/peer-ids.md
```diff
@@ -0,0 +1,80 @@
...
+ - SECIO handshake
+
+The second is for generating peer ids; this is discussed in the section below.
+
+## Peer Ids
+
+Here is the process by which we generate peer ids based on the public keys described above:
+
+  1. Encode the public key into the protobuf.
+  2. Serialize the protobuf containing the public key into bytes using the [canonical protobuf encoding](https://developers.google.com/protocol-buffers/docs/encoding).
```
arnetheduck: `1=0,2=bytes` and `2=bytes,1=0` are two equally valid encodings for an RSA key. Indeed, so is `1=2,2=bytes,1=0` given that the last value overrides the first, per the protobuf spec. Protobuf is not deterministic in its encoding - this would be a problem in general.

yusefnapora: Hmm, that's true. From the linked doc:

> When a message is serialized, there is no guaranteed order for how its known or unknown fields should be written. 

Are we doing anything to enforce deterministic field ordering? If not, maybe we should be...

yusefnapora: Some notes about the lack of deterministic serialization for protobuf: https://gist.github.com/kchristidis/39c8b310fd9da43d515c4394c3cd9510

This is a thorny issue... a libp2p implementor could do everything according to the spec and still end up with peer ids that other peers wouldn't validate, if their protobuf implementation happened to order the keys differently.

arnetheduck: It helps to think of a protobuf message as a hash map, essentially - semantically it's a very similar situation - the last thing you put in remains, and the order is unspecified.

In fact, that's cheap and easy way to implement it - I'd expect dynamic languages with strong dictionary implementations to do it this way. 

Locking down field order solves the issue, but I'm not sure I'd call it protobuf any more at that point - perhaps protobuf-compatible.

yusefnapora: Yeah, I'm not crazy about requiring semantics on top of protobuf - at that point I'd rather just specify a different serialization that is deterministic. Then you have a backwards compatibility problem though...

@marten-seemann do you have any concerns about the non-deterministic encoding for this public key structure? Asking because it gets embedded in the certificate extension for the TLS 1.3 spec you're working on.

yusefnapora:
For those watching this thread, I pushed a couple commits:

- private keys are back! as discussed with @Stebalien the serialized private key format is important for implementors, since keys should be portable across implementations.
  - I added a note that private keys shouldn't be transmitted, and the storage format is a reference for implementors
- I changed the wording around keys being "passed around in byte arrays in code" to make it clear that you can do whatever you want in-memory, but you MUST use the described encodings when using the protobufs for transmission and storage.
- I made RSA and Ed25519 `MUST`-haves, per @raulk's notes.

While I'm here, let me drop a link to ~my soundcloud~ the [concept doc PR](https://github.com/libp2p/docs/pull/30) I'm working on for the docs site - it's not blocked on this, but it would be nice to link to the merged spec instead of a PR 😄

peer-ids/peer-ids.md
raulk: @Stebalien @marten-seemann would appreciate your views here. I think @arnetheduck has surfaced a legitimate concern. Likely it hasn’t been an issue so far because of the simplicity of the protobuf, which makes it highly unlikely that serialisers will play tricks. But indeed, a lack of determinism is dangerous.

Stebalien: We generally require that fields be ordered by tag. I believe all(?) implementations do this and it should usually "just work". However, this _spec_ may need to specify this.

A more general solution is to tell users to never re-serialize. That is, when you get a public key from the network, you can read it but you shouldn't _re_-serialize it when computing the peer ID. Instead, you should keep the serialized bytes around and use those. That would also solve https://github.com/libp2p/go-libp2p-crypto/issues/51.

raulk: Alternatively, this protobuf is so simple we can just remove the dependency on protobuf and inline the encoding/decoding procedure by retrofitting the protobuf representation, enforcing the order and determinism.

raulk: We could then state: “you MAY use a protobuf encoder to serialise the key, if and only if XYZ. Else you MUST use an alternative means to enforce XYZ in order to guarantee interoperability and universality.”

Stebalien: Either way but I _would_ like to be able to add additional fields in the future (which is problematic with respect to re serializing).

peer-ids/peer-ids.md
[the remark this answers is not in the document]
tomaka: > what the application using an implementation chooses or not to do is irrelevant here.

The concrete issue is: we receive a handshake containing an RSA key. Is it mandatory to accept it? Both Rust and Substrate would be very glad if it wasn't.

peer-ids/peer-ids.md
arnetheduck: > When a message is serialized, there is no guaranteed order for how its known or unknown fields should be written.

The way I interpret this line, specially since it's explicitly stated in their spec, is that regardless of what implementations happen to doing today, there's no guarantee that they will do so tomorrow. The absolutely easiest way to implement a protobuf library in JavaScript for example is to use an object (I've done that for some custom projects) - with only 2 fields you have a pretty good 50% chance that the order will be correct. You then get to roll the dice every time a new version of chrome is out ;)

Another correct encoding of an RSA key is simply `2=bytes` - in protobuf 3 there's no distinction between a default value (0) and an explicitly set value, so implementations are effectively not required to spell it out. In protobuf2, the situation is similar, though you can detect a missing field in the generated API. They were able to do this change in the high-level language between v2 and v3 without changing the wire spec, largely because of nuances like this.

The only safe thing to say is that you can use a protobuf decoder - these will always be compatible.

I think stating that you may use a protobuf encoder would be shooting future users in the foot, specially if the aim is to create a standard that is language and implementation agnostic. This would be a concern, since we're considering a new libp2p impl in a new language (Nim).

> inline the encoding/decoding procedure

For a message like this, the key is a varint that includes a few bits for field type + field number and then a varint for the value. the second one is again typed varint, and bytes means a varint for length+the bytes themselves (afair). Only the encoding needs to be locked down (it would be a bit unusual, as far as specs go, though).

> never re-serialize

One of the reasons I've seen people use protobuf is to have the ability to re-serialize - it was a design goal of the language, and manifests itself as the "UnknownFields" map in the C++ API, for example. Prohibiting this takes the spec another step away from this being protobuf. Granted, it's likely not the first thing someone will want to do with a peer id specifically..

peer-ids/peer-ids.md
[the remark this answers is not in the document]
raulk: IMO Your implementation MUST support it. Your implementation MAY allow the user to enable/disable specific key types via configuration.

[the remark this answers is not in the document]
tomaka: But this spec is supposed to define what happens on the wire, because that's all that matters for compatibility.
To SHOULD is the right way to put it, then.

peer-ids/peer-ids.md
Stebalien: > The way I interpret this line, specially since it's explicitly stated in their spec, is that regardless of what implementations happen to doing today, there's no guarantee that they will do so tomorrow.

Specifically, it says:

> Serialization order is an implementation detail and the details of any particular implementation may change in the future.

(i.e., implementation defined)

Annoyingly, they actually changed this. It [used to say](https://web.archive.org/web/20140215063318/https://developers.google.com/protocol-buffers/docs/encoding#order):

> While you can use field numbers in any order in a .proto, when a message is serialized its known fields should be written sequentially by field number, as in the provided C++, Java, and Python serialization code.

Which is why we relied on this. All fully-featured and correct implementations ensured their field orders matched.

> in protobuf 3 there's no distinction between a default value (0) and an explicitly set value

We currently require proto2 (these fields are actually marked as _required_ which doesn't exist in proto3) but yeah, implementers may ignore this.

> One of the reasons I've seen people use protobuf is to have the ability to re-serialize - it was a design goal of the language, and manifests itself as the "UnknownFields" map in the C++ API, for example. Prohibiting this takes the spec another step away from this being protobuf. Granted, it's likely not the first thing someone will want to do with a peer id specifically..

IMO, that's a design goal when _editing_. Any edits will change the peer ID anyways.

---

My worry with saying "this is not protobuf but you can use a protobuf decoder" is that people _will_ use protobuf encoders _anyways_ because they'll usually "just work". Then, when the underlying protobuf implementation changes, bad things will happen.

peer-ids/peer-ids.md
```diff
@@ -0,0 +1,139 @@
...
+
+The second is for generating peer ids; this is discussed in the section below.
+
+## Peer Ids
+
+Here is the process by which we generate peer id's based on the public/private keypairs described above:
+
+  1. Encode the public key into the protobuf.
+  2. Serialize the protobuf containing the public key into bytes using the [canonical protobuf encoding](https://developers.google.com/protocol-buffers/docs/encoding).
+  3.  If the length of the serialized bytes <= 42, then we compute the "identity" multihash of the serialized bytes.  In other words, no hashing is performed, but the [multihash format is still followed](https://github.com/multiformats/multihash) (byte plus varint plus serialized bytes).  The idea here is that if the serialized byte array is short enough, we can fit it in a multihash verbatim without having to condense it using a hash function.
```
arnetheduck: This step looks like it could be simplified so that the hashing is always performed - this would simplify peer id management by making them uniform and at the same time disconnect the fingerprint from the actual public key, minimizing the unnecessary spread of key material.

Stebalien: Take a look at https://github.com/libp2p/specs/issues/138.

Note: Unfortunately, this isn't something we can "just change" because we have users.

peer-ids/peer-ids.md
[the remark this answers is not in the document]
raulk: @tomaka and I cycled ideas back and forth via chat, as it was easier. We concluded that:
1. Specs cover behaviour insofar interoperability is concerned.
2. Implementations MUST support a RSA and Ed25519, but just like transports are pluggable and one can configure ciphersuites in TLS, implementations MAY allow the user to configure which key types their host will support. So while RSA (source of concern) MUST be supported, the user can choose to disable it.
3. libp2p aims to put the choice in user’s hands, in the same way that e.g. Go ships with RSA but it’s up to the user to use it.

Thanks for everyone who participated.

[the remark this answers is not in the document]
marten-seemann: For an endpoint it’s indistinguishable if its peer supports RSA but disabled RSA support or if it uses an implementation that doesn’t support RSA in the first place. Therefore, I see little value in making it a requirement to support a certain encryption scheme.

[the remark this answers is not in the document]
raulk: @marten-seemann this is about creating clarity for _implementors_ of the spec. Splintering a cloudy SHOULD into clear and detailed MUSTs and MAYs goes a long way here.

peer-ids/peer-ids.md
raulk: @Stebalien 

> My worry with saying "this is not protobuf but you can use a protobuf decoder" is that people will use protobuf encoders anyways because they'll usually "just work". Then, when the underlying protobuf implementation changes, bad things will happen.

We can include test vectors for that. It's fair for implementors to use protobuf encoders, and I expect that to be the dominant implementation approach. However, if between version X and Y a specific encoder changed in a way that it broke our requirement, that's when the implementor would have to craft a manual encoding. It's a fair trade-off. We can provide guidance in this spec (in the form of "implementor's notes").

> https://web.archive.org/web/20140215063318/https://developers.google.com/protocol-buffers/docs/encoding#order
> Which is why we relied on this. All fully-featured and correct implementations ensured their field orders matched.

Woah, thanks for tracing that back 😓 

> Either way but I would like to be able to add additional fields in the future (which is problematic with respect to re serializing).

IIUC, you'd like nodes to tolerate unrecognised fields and use the struct's byte string verbatim when calculating the peer ID?

Isn't this dangerous? This creates a surface for polymorphic peer IDs, where node `A` with pubkey `P` could produce infinite peer ID (or craft specific ones) for itself – and hence break protocols that rely on peer IDs – merely by setting random fields in the protobuf.

peer-ids/peer-ids.md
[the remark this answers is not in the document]
marten-seemann: It's not enforceable by the peer anyway, so whatever language we put into the documents, implementors will do what implementors see necessary to do. Requiring something that's not necessary for the function of the protocol seems like a wrong use of a normative MUST to me.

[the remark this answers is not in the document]
raulk: If you read the original statement from the viewpoint of an implementer, you will have no clue what to do, as everything is optional. You'd end up guessing or digging into the code of other implementations, which defeats the purpose of writing a spec altogether.

Our specs are useless if they don't provide a baseline framework for interoperability. In this case, this is achieved by making all implementations support RSA and Ed25519.

Under this arrangement, if peer A (implementation X) shares an RSA key with peer B (implementation Y), and the latter rejects it, peer A can be certain that peer B has disabled support for RSA keys.

peer-ids/peer-ids.md
Stebalien: > Isn't this dangerous? This creates a surface for polymorphic peer IDs, where node A with pubkey P could produce infinite peer ID (or craft specific ones) for itself – and hence break protocols that rely on peer IDs – merely by setting random fields in the protobuf.

This should be just fine.

1. You can probably already do this. We don't make any effort to ensure that the `data` part of the key is completely deterministic.
2. Any change to the key metadata will, effectively, create a new identity. IMO, that's _exactly_ what we want.

raulk: > We don't make any effort to ensure that the data part of the key is completely deterministic.

Could you elaborate?

> Any change to the key metadata will, effectively, create a new identity. IMO, that's exactly what we want.

That feels wrong. Identity should be derived from cryptographic material (a pubkey), not from random metadata.

yusefnapora:
Sorry for sleeping on this for a while everyone :)

I added a few commits to address some feedback. I think the most significant is the note about deterministic protobuf encoding. It basically says determinism is "desirable" and you should try to make it happen, but doesn't call it out as a MUST. Without requiring a bunch of changes to the protobuf spec, that might be the best we can do, but if anyone has a better way to put this, I'm all ears.

peer-ids/peer-ids.md
yusefnapora: Not sure exactly what @Stebalien is referring to about the data field, but the spec describes one case where the same public key could lead to different data fields (and thus peer ids). 

Since there are two supported methods for encoding Ed25519 keys, you could end up with different peer ids for the same key depending on which you choose.

yusefnapora:
@Stebalien @arnetheduck @raulk - could you guys help me figure out the resolution to the deterministic encoding problem?

If we definitely want to require a consistent / canonical encoding for peer ids, then I think I should write up a precise spec that requires a certain field ordering, etc. And we can have some tests that ensure your encoder handles edge cases well.

But @Stebalien mentions that, because we're also not guaranteeing a canonical encoding for the key `Data` field, it's not really relevant at this layer. And that it's fine to have multiple valid peer ids for the same key, which seems contentious. 

I can make up some arguments in favor of this view; for one, we can extend the `PublicKey` message in the future without having to decide now how to encode unknown fields in our "not-quite-protobuf" format. It could also be up to an application to decide whether to reject multiple peer ids that all derive from the same public key. It's also much simpler, of course, since the easiest problem to solve is the one you don't have.

@Stebalien could you elaborate a bit on your view? I think we should figure out if this is a blocker or not to merging the spec.

raulk:
Re: deterministic key serde. I suggest we specify the format as proto3 + the extra requirements to reach a deterministic result (ordering, no unrecognised fields, no duplicates last wins, etc.) We should add an implementers note in the form of a SHOULD recommendation to use an OOTB protobuf encoder, and provide test vectors. We should also provide a boiled down serde spec in BNF or similar form. It’ll be super simple, the schema is so simple and constrained we can express the serialised form separately without alluding to proto3 at all.

Re: using unrecognised fields for peer ID calculation, I’d like to hear what use cases you had in mind @stebalien. In my view, the peer ID should be derived from the pubkey modulo metadata, if any. I don’t think user-defined metadata should yield a different identity. Seems like opening a trivial attack vector for sybils.

Stebalien:
> In my view, the peer ID should be derived from the pubkey modulo metadata, if any. I don’t think user-defined metadata should yield a different identity.

Note: I'm really not expecting much if any _user_ defined metadata. However, we may want to add new fields in the future and it's hard to do this if we don't include them in the hash.

1. We (IPFS/IPLD) would like to to be able to convert peer IDs to CIDs (and, ideally, fetch keys over bitswap).
2. Currently, peer IDs always map to _exactly_ one public key. If we both allow other metadata _and_ don't hash it, we'll need to handle merging metadata.
3. The un-hashed metadata won't be authenticated. When we connect to a peer, they send us their key. We then compute the peer ID of that key and check if it matches the expected one. If we ignore metadata, we could end up with inconsistent metadata views.

> I don’t think user-defined metadata should yield a different identity. Seems like opening a trivial attack vector for sybils.

There are two cases:

* The key's _owner_ generates multiple identities. This is a non-issue, they can already generate more keys.
* An attacker generates multiple identities for a victim: We'll detect this on connect because we'll compute the victim's peer ID from the key _they_ give us. Ideally, keys would be self signed but our use-case for these keys is simple enough that this probably isn't that big of an issue (we should keep using this system for identifying libp2p peers but should come up with something better for identity).

Stebalien:
> But @Stebalien mentions that, because we're also not guaranteeing a canonical encoding for the key Data field, it's not really relevant at this layer. And that it's fine to have multiple valid peer ids for the same key, which seems contentious.

> Same Key

IMO, "same key" should mean "same bytes". That is, if I change _anything_ about the bytes of the serialized key, I get a new key.

I'm concerned that not all key formats will have a "canonical" encoding, some libraries may strip certain metadata while others preserve it, etc. This will lead to hard-to-track-down bugs. This has already been a real pain for us in IPLD and our solution there is to avoid re-serializing unless we change something.

My unconcern with having multiple valid peer IDs for the same underlying cryptographic key is that there's likely nothing we can do to stop this. I haven't audited our key formats/algorithms but it's likely possible to make _some_ small changes to a public key and have it continue to work with the private key.

yusefnapora:
@raulk, what do you think of the header in https://github.com/libp2p/specs/pull/100/commits/6c4a58793732a06145edc960611df06e4c6865ec - I think the markdown table is a decent way to present the main status information.

I figure we can hammer out details here and I'll write up a little spec for the header format once we like it.

yusefnapora:
Also, @raulk, @vyzo & @Stebalien I nominated you guys as the Interest Group for this one 😄

Others are welcome to join in if they like

Zolmeister:
I would like to reiterate the value in migrating to embedded-key base32 encoded canonical representation of PeerIDs
Ref. https://github.com/libp2p/specs/issues/139

yusefnapora:
How do we feel about merging this? 

https://github.com/libp2p/specs/pull/100/commits/870b71ac046b2774be8a90644d398d45c2c35ed8 kicks the deterministic encoding issue down the road by saying a future version of the spec may require more strict encoding than the protobuf spec, but until then, don't extend the `PublicKey` message and use a protobuf encoder that writes fields in order.

@mgoelzer @Stebalien @raulk @vyzo

folex:
Sorry for commenting on an old PR, but I have not found an answer anywhere else.

What's the motivation for number `42` here?
> keys that serialize to at most 42 bytes must be hashed using the "identity" multihash codec.

When one would expect keys to be longer than 42 bytes?

Thank you!

Winterhuman:
I believe it's the size of ed25519 public keys inside of protobuf encoding, since ed25519 keys are supposed to be inlined into PeerIDs

folex:
uh, so it's 32 + 10 bytes. 

30% for serializaton is such a big number, I didn't even think about that.

thank you, that makes it clearer🙏

mxinden:
And I vaguely remember that 42 would be small enough to then fit into a single DNS segment. Though I might be misremembering that.

PR #106, opened 2018-10-24
SECIO spec

bigs:
Introduces a spec for SECIO. Various crypto algorithms living in `*-libp2p-crypto` have been omitted for the time being.


secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Protocol
+
+### Proposal Generation
+
+SECIO channel negotiation begins with a proposal phase. Both sides generate a
+proposal as follows:
+
+*Note: the public key should be serialized per the `Bytes` method from
+[go-libp2p-crypto](https://godoc.org/github.com/libp2p/go-libp2p-crypto#Key).*
```
Stebalien: Let's point at https://github.com/libp2p/specs/pull/100 (once it's merged)

secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+```
+Propose{
+  Rand: <16 secure random bytes>,
+  Pubkey: <public key bytes>,
+  Exchanges: <comma separated string of supported key exchanges>,
+  Ciphers: <comma separated string of supported ciphers>,
+  Hashes: <comma separated string of supported hashes>,
```
Stebalien: Let's put the actual protobuf here (and say that it's a protobuf).

secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+}
+```
+
+Both peers serialize this message and send it over the wire. Upon deserializing
+their peer's message, they verify that the pubkey matches that described by the
+peer's peer ID (NOTE: this is sometimes only possible for the peer *initiating*
```
Stebalien: Given that this is just the *spec*, I'd just say:

> At this point, each peer should check that the public key matches the expected peer ID, if known.

Or something like that. We don't need to explicitly tell the programmer how to implement the protocol step-by-step (i.e., "calculate the peer ID and store it for later").

secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+Given our preference, we now sort through each of the `Exchanges`, `Ciphers`,
+and `Hashes` provided by both peers, selecting the first item from our preferred
+peer's set that is also shared by the other peer.
+
+### Key Exchange
+
+Now the peers prepare a key exchange. Both peers generate an ephemeral key based
+on the agreed upon exchange (currently support is only available for elliptic
+curve algorithms). Ephemeral keys are generated via
+[go-libp2p-crypto](https://godoc.org/github.com/libp2p/go-libp2p-crypto#GenerateEKeyPair).
```
Stebalien: We should actually spec this.

secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+Then, generate the `Exchange`:
+
+```
+Exchange{
+  Epubkey: <ephemeral pubkey>,
+  Signature: <sign corpus with local private key>,
+}
+```
```
Stebalien: Again, let's just use the protobuf definition.

secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+expect them to have generated with their public key. Peers should close the
+connection if the signature does not validate.
+
+### Key Stretching
+
+Peers now generate their shared secret based on the function generated by the
+ephemeral key generation, passing it the remote peer's ephemeral key. With the
+shared secret generated, both peers stretch the key using the algorithm
+described by
+[go-libp2p-crypto](https://godoc.org/github.com/libp2p/go-libp2p-crypto#KeyStretcher).
```
Stebalien: Again, we need to spec this.

secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+[`StretchedKeys`](https://godoc.org/github.com/libp2p/go-libp2p-crypto#StretchedKeys)
+objects.
+
+### Initiate Secure Channel
+
+With the cipher and HMAC signer created, the secure channel is ready to be
+opened. Each packet is of the form:
+
+```
+[uint32 length of packet | encrypted body | hmac signature of encrypted body]
```
Stebalien: * We encode the lengths as big-endian (network-order).
* Need to specify what "encrypt" means.
* Need to specify how we mac.

secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+```
+k1, k2 := KeyStretcher(sharedSecret)
+```
+
+With `k1` and `k2` computed, swap the two values if the remote peer is the
+preferred peer. After swapping if necessary, `k1` becomes the local peer's key
+and `k2` the remote peer's key.
+
+Each peer now generates a MAC key and cipher for the remote and local keys
+generated in the previous step using the `MacKey` and `CipherKey` from the
```
tomaka: Need to specify the order within the result of the stretching.

secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+SECIO allows participating peers to support a subset of the following
+algorithms.
+
+### Exhchanges
+
+- P-256
+- P-384
+- P-521
+
+### Ciphers
```
raulk: Ciphers used for key stretching and for message encryption once SECIO channel is established.

secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+## Table of Contents
+
+- [Algorithm Support](#algorithm-support)
+
+## Algorithm Support
+
+SECIO allows participating peers to support a subset of the following
+algorithms.
+
+### Exhchanges
```
raulk: Elliptic curves used for ephemeral key generation.

secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+- P-384
+- P-521
+
+### Ciphers
+
+- AES-256
+- AES-128
+- Blowfish
+
+### Hashes
```
raulk: Hashes used for key stretching, and for HMACs once SECIO channel is established.

secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+- SHA-256
+- SHA-512
+
+## Data Structures
+
+The SECIO wire protocol features two message types defined in the
+[protobuf description language](https://github.com/libp2p/go-libp2p-secio/blob/master/pb/spipe.proto).
+These two messages, `Propose` and `Exchange` are the only serialized types
```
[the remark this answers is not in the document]
raulk: Nevermind, I see this feedback is recurrent below.

secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+```
+k1, k2 := KeyStretcher(sharedSecret)
+```
+
+With `k1` and `k2` computed, swap the two values if the remote peer is the
+preferred peer. After swapping if necessary, `k1` becomes the local peer's key
+and `k2` the remote peer's key.
+
+Each peer now generates a MAC key and cipher for the remote and local keys
```
raulk: These are not generated here. The MAC key is already generated during key stretching above. The cipher was selected above when comparing Propose messages. 

What the Go implementation does at this point is prepare the constructs for HMAC and encryption/decryption. I think that's what needs to be stated.

secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+With the cipher and HMAC signer created, the secure channel is ready to be
+opened. Each packet is of the form:
+
+```
+[uint32 length of packet | encrypted body | hmac signature of encrypted body]
+```
+
+The first packet transmitted by each peer must be the remote peer's nonce. Peers
+validate that the remote peer sent them their nonce, closing if unsuccessful.
```
raulk: The target of validation is actually the (d)encryption and HMAC'ing logic. We use a known value (nonce) to send a predictable message so we can validate that both parties have set up the encrypted channels correctly.

secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+- [Lars Gierth](https://github.com/lgierth)
+
+## Implementations
+
+- [js-libp2p-secio](https://github.com/libp2p/js-libp2p-secio)
+- [go-secio](https://github.com/libp2p/go-libp2p-secio)
+
+## Table of Contents
+
+- [Algorithm Support](#algorithm-support)
```
raulk: ToC is missing entries.

secio/README.md
richardschneider: Does the padding style for encrypt need to be specified?

Stebalien: Really, we should specify *everything*. Ideally, we'd point to an RFC. However, we should optimize for merging something that's correct rather than waiting for something that's perfect.

secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+### Ciphers
+
+- AES-256
+- AES-128
+- Blowfish
+
+### Hashes
+
+- SHA-256
+- SHA-512
```
richardschneider: Current go implementation does not use dashes in the hash name; see [al.go](https://github.com/libp2p/go-libp2p-secio/blob/e4910f5adf3debfcd76f64efea6d1e3dd96756d4/al.go#L33).

bigs:
yep @whyrusleeping just lower priority than testbed work at the moment. lotta good feedback re: shaping this into a more formal spec

secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+The peers serialize these and write them over the wire. Upon receiving the
+remote peer's `Exchange`, validate the signature by computing the `corpus` you
+expect them to have generated with their public key. Peers should close the
+connection if the signature does not validate.
+
+### Key Stretching
+
+Peers now generate their shared secret based on the function generated by the
```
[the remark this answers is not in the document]
richardschneider: I think its official name is ECSVDP-DH

secio/README.md
richardschneider: I'm assuming that the length includes the encrypted body and the hmac signature.

secio/README.md
GriffinMB: Why is a deprecated cipher (Blowfish) supported? From the docs:

>  Blowfish is a legacy cipher and its short block size makes it vulnerable to birthday bound attacks (see https://sweet32.info). It should only be used where compatibility with legacy systems, not security, is the goal.

> Deprecated: any new system should use AES (from crypto/aes, if necessary in an AEAD mode like crypto/cipher.NewGCM) or XChaCha20-Poly1305 (from golang.org/x/crypto/chacha20poly1305).

Stebalien: :+1:

secio/README.md
```diff
@@ -0,0 +1,160 @@
...
+connection. If the peer doesn't know the remote peer's ID, they can compute and
+store the remote peer ID for later use.
+
+### Determining Roles and Algorithms
+
+Next, the peers use a deterministic formula to compute their roles in the coming
+exchanges. Each peer computes:
+
+```
+oh1 := sha256(concat(remotePeerPubKeyBytes, myNonce))
```
yusefnapora: This is the first use of the word "nonce". Does it refer to the `rand` field of the `Propose` message?

Stebalien: Yes.

yusefnapora:
Hey all, I took a stab at addressing the review feedback here. I'd love if people with more experience could give it a review.

One thing I didn't address was using more standard nomenclature (e.g. KDF, ECDH), as I'm not super confident I'd use the terms correctly. I might read up a bit more and give that a go for the next round.

secio/README.md
GriffinMB: It would also be worth noting what mode the AES ciphers should use. From the Go implementation, it looks like it's CTR!

Also, more generally, is this an implementation of any well-reviewed specification? If not, is there a reason why? I would imagine there are a number of specs for low-overhead, encrypted channels which may prevent future pitfalls.

Stebalien: This is definitely not well-reviewed and will be will be deprecated in favor of TLS quite soon. See: https://github.com/libp2p/go-libp2p-tls/.

(go-ipfs now has experimental support which we'll likely upgrade to default support after a release or two).

GriffinMB: Gotcha, makes sense! And that answers my follow-up questions as well :)

yusefnapora:
I'm going to go ahead and merge this, since I think it's in a decent state. If I added anyone to the interest group (@Stebalien, @richardschneider, @tomaka, @raulk) that doesn't want to be there, let me know.

PR #108, opened 2018-11-07
Add minimal Kademlia DHT spec

raulk:
~~This was a complicated birth.~~

~~TODO:~~

- [ ] ~~Specify that we haven't implemented PINGs.~~
- [ ] ~~Deviations in bucket behaviour from baseline Kademlia. Since we don't use PINGs, we don't test the least recently seen peer. We evict it blindly, thus causing a high degree of [bucket thrashing](https://github.com/libp2p/go-libp2p-kad-dht/issues/194) and not observing the heuristic "the longer a peer has been around, the more likely it is to stay around".~~
- [ ] ~~Revisit RPC messages section. Copy protobufs and explain each field meticulously.~~
- [ ] ~~Resolve all feedback below.~~

--- 

_Update by @mxinden on 2021-05-14._

This pull request adds a **minimal** specification for the libp2p Kademlia DHT protocol. It is a minimal version and thus while striving for correctness it does not strive for completeness. The version proposed in this pull request sets a base layer specification to be expanded in the future.

Areas worth expanding in future pull requests:

- Client mode (non publicly reachable nodes not advertising the kademlia protocol ID) as well as its implications to `FIND_NODE` (see https://github.com/libp2p/specs/pull/108/files#r657950178).
- Detailed record validation
- Default value adjustments and reasoning
- Detailed request handling
- Expanded k-bucket routing table management
- ...


kad-dht/README.md
```diff
@@ -0,0 +1,288 @@
...
+}
+```
+
+## Value lookups
+
+When looking up an entry in the DHT, the implementor should collect at least `Q`
+(quorum) responses from distinct nodes to check for consistency before returning
+an answer.
+
+Should the responses be different, the `Validator.Select()` function is used to
```
jhiesey: What about `Validator.VerifyRecord`? It's called on both put and get in the js version.

kad-dht/README.md
```diff
@@ -0,0 +1,288 @@
...
+* the set of peers we've already queried (`Pq`) and the set of next query candidates sorted by distance from `K` in ascending order (`Pn`).
+* the set of peers with outdated values (`Po`). 
+
+**Initialization**: seed `Pn` with the `α` peers from our routing table we know are closest to `K`, based on the XOR distance function.
+
+**Then we loop:**
+
+*WIP (raulk): lookup timeout.*
+
+1. If we have collected `Q` or more answers, we cancel outstanding requests, return `best`, and we notify the peers holding an outdated value (`Po`) of the best value we discovered, by sending `PUT_VALUE(K, best)` messages. _Return._
```
jhiesey: There's also a termination condition where we run out of peers to query without getting `Q` answers

kad-dht/README.md
jhiesey: Ah nevermind, you cover this below. Still might not hurt to mention validation here.

kad-dht/README.md
```diff
@@ -0,0 +1,288 @@
...
+The record is rejected if the validation fails.
+
+## Provider records
+
+Nodes must keep track of which nodes advertise that they provide a given key
+(CID). These provider advertisements should expire, by default, after 24 hours.
+These records are managed through the `ADD_PROVIDER` and `GET_PROVIDERS`
+messages.
+
+_WIP (jhiesey): explain what actually happens when `Provide()` is called._
```
jhiesey: Would you like me to add this section?

kad-dht/README.md
```diff
@@ -0,0 +1,288 @@
+# libp2p Kademlia DHT specification
+
+The Kademlia Distributed Hash Table (DHT) subsystem in libp2p is a DHT
+implementation largely based on the Kademlia [0] whitepaper, augmented with
+notions from S/Kademlia [1], Coral [2] and mainlineDHT \[3\].
```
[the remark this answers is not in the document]
raulk: I had to escape it because, for some reason, Github elided it when unescaped. Didn't dig into why, though. Settled for the easy path.

kad-dht/README.md
```diff
@@ -0,0 +1,288 @@
...
+## Kademlia routing table
+
+The data structure backing this system is a k-bucket routing table, closely
+following the design outlined in the Kademlia paper [0]. The default value for
+`k` is 20, and the maximum bucket count matches the size of the SHA256 function,
+i.e. 256 buckets.
+
+The routing table is unfolded lazily, starting with a single bucket a position 0
+(representing the most distant peers), and splitting it subsequently as closer
+peers are found, and the capacity of the nearmost bucket is exceeded.
```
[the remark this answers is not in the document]
raulk: Good call, will double check

kad-dht/README.md
raulk: 👍

kad-dht/README.md
raulk: Good call!

kad-dht/README.md
raulk: Yes, please ;-) Everywhere that's marked `jhiesey` are areas I left for you ;-)

kad-dht/README.md
```diff
@@ -0,0 +1,288 @@
...
+```
+
+## Value lookups
+
+When looking up an entry in the DHT, the implementor should collect at least `Q`
+(quorum) responses from distinct nodes to check for consistency before returning
+an answer.
+
+Should the responses be different, the `Validator.Select()` function is used to
+resolve the conflict and select the _best_ result.
```
vasco-santos: We should notice here that the `Validator.Select()` is associated with a specific namespace [libp2p/go-libp2p-record/validator.go#L56](https://github.com/libp2p/go-libp2p-record/blob/master/validator.go#L56). 

Moreover, I have been thinking about this for a while in JS land. In case we do not have a `Validator.Select()` for the namespace of a key being used, shouldn't we fallback to a default `Select` function? In JS land, when using arbitrary keys with unknown namespaces it failed to get the record. We ended up changing to selecting the first record, but I don't know if that is the best approach.

I believe the same happens with `Validate()`.

kad-dht/README.md
```diff
@@ -0,0 +1,288 @@
...
+or more multiaddrs, that provider info is added to the peerbook and the peer is
+added as a provider for the CID.
+* `PING() -> ()` Tests connectivity to destination node. Currently never sent.
+
+# Appendix A: differences in implementations
+
+The `addProvider` handler behaves differently across implementations:
+  * in js-libp2p-kad-dht, the sender is added as a provider unconditionally.
+  * in go-libp2p-kad-dht, it is added once per instance of that peer in the
+    `providerPeers` array.
```
tomaka: That doesn't really say what the difference between implementations is.
Also the idea of the specs is to remove these differences.

kad-dht/README.md
```diff
@@ -0,0 +1,288 @@
...
+* `PUT_VALUE(key bytes, value Record) -> ()` Validates `value` and, if it is
+valid, stores it in the datastore.
+* `GET_PROVIDERS(key bytes) -> (providerPeers PeerInfo[], closerPeers
+PeerInfo[])` Verifies `key` is a valid CID. Returns `providerPeers` if in the
+providers cache, and an array of closer peers.
+* `ADD_PROVIDER(key, providerPeers PeerInfo[]) -> ()` Verifies `key` is a valid
+CID. For each provider `PeerInfo` that matches the sender's id and contains one
+or more multiaddrs, that provider info is added to the peerbook and the peer is
+added as a provider for the CID.
+* `PING() -> ()` Tests connectivity to destination node. Currently never sent.
```
tomaka: Shouldn't that be removed entirely?

kad-dht/README.md
```diff
@@ -0,0 +1,288 @@
...
+
+## Record keys
+
+Records in the DHT are keyed by CID [4]. There are intentions to move to
+multihash [5] keys in the near future, as certain CID components like the
+multicodec are redundant.
+
+## Interfaces
+
+The libp2p Kad DHT implementation satisfies the routing interfaces:
```
tomaka: Is that really relevant to a spec? That looks pretty specific to Go to me, or to a specific set of programming languages that are capable of fulfilling it. In particular the Rust code has no interest in following these interfaces.

tomaka:
It should also mention how the substreams that use the protocol behave.
Is it expected to open one substream per request? Or should implementations make an effort to only use one at a time and send requests one by one? Should endpoints close their substream after a successful response, or can they continue sending more requests?
(note: I actually have these questions right now for Rust, as I have no clue)

kad-dht/README.md
```diff
@@ -0,0 +1,288 @@
...
+_WIP (jheisey): consider just dumping a nicely formatted and simplified
+protobuf._
+
+See [protobuf
+definition](https://github.com/libp2p/go-libp2p-kad-dht/blob/master/pb/dht.proto)
+
+On any error, the entire stream is reset. This is probably not the behavior we
+want.
+
+* `FIND_NODE(key bytes) -> (nodes PeerInfo[])` Finds the `ncp` (default:6) nodes
```
tomaka: Is `key` a multihash in binary format?

kad-dht/README.md
```diff
@@ -0,0 +1,288 @@
...
+closest to `key` from the routing table and returns an array of `PeerInfo`s. If
+a node with id equal to `key` is found, returns only the `PeerInfo` for that
+node.
+* `GET_VALUE(key bytes) -> (record Record, closerPeers PeerInfo[])` If `key` is
+a public key (begins with `/pk/`) and the key is known, returns a `Record`
```
tomaka: Begins with `/pk/`, so it's a string?

kad-dht/README.md
```diff
@@ -0,0 +1,288 @@
...
+containing that key. Otherwise, returns the `Record` for the given key (if in
+the datastore) and an array of `PeerInfo`s for closer peers.
+* `PUT_VALUE(key bytes, value Record) -> ()` Validates `value` and, if it is
+valid, stores it in the datastore.
+* `GET_PROVIDERS(key bytes) -> (providerPeers PeerInfo[], closerPeers
+PeerInfo[])` Verifies `key` is a valid CID. Returns `providerPeers` if in the
```
tomaka: Well, here is says it's a CID.

kad-dht/README.md
```diff
@@ -0,0 +1,288 @@
...
+
+* `FIND_NODE(key bytes) -> (nodes PeerInfo[])` Finds the `ncp` (default:6) nodes
+closest to `key` from the routing table and returns an array of `PeerInfo`s. If
+a node with id equal to `key` is found, returns only the `PeerInfo` for that
+node.
+* `GET_VALUE(key bytes) -> (record Record, closerPeers PeerInfo[])` If `key` is
+a public key (begins with `/pk/`) and the key is known, returns a `Record`
+containing that key. Otherwise, returns the `Record` for the given key (if in
+the datastore) and an array of `PeerInfo`s for closer peers.
+* `PUT_VALUE(key bytes, value Record) -> ()` Validates `value` and, if it is
```
tomaka: `Record` is not defined.

tomaka (dismissed):

tomaka:
Kademlia is by far the protocol for which I've suffered the most when implementing into Rust, because of all the differences between the well-documented Kademlia algorithm and the way it is implemented in libp2p. I think we should focus more on the latter and not copy-paste what people can already find on Google.

tomaka:
Also, writing the Kademlia specs is a tremendous task. I don't expect a complete spec to be less than 2k lines of code, and expecting a single person to write them is very optimistic.

raulk:
@tomaka 

> Kademlia is by far the protocol for which I've suffered the most when implementing into Rust, because of all the differences between the well-documented Kademlia algorithm and the way it is implemented in libp2p. I think we should focus more on the latter and not copy-paste what people can already find on Google.

That was the spirit of this spec: to focus on differential areas vs. regurgitating how Kademlia works (as conveyed in the spec intro). Hence it covers provider records, public key records, conflict resolution, peer correction, etc. which is specific to _libp2p_ Kad DHT.

> I don't expect a complete spec to be less than ~2k lines

Could you enumerate what other aspects are worth covering? Aspects that are unique in libp2p. We don't want to clone the Kademlia and friends literature.

---

Regarding all the data model comments, there's a request in the doc for @jhiesey to replace these descriptions with the protobuf.

kad-dht/README.md
raulk: I had second thoughts when dumping the interface here, as it should be non-normative as you say. However, it helps bind things together, as it specifies the public calls supported by this component along with its inputs and outputs, i.e. the expected public surface of this component. Also, each exposed behaviour is specified at some point in the doc.

IMHO, we do need to capture an abstract interface outline, and I used Go nomenclature and copied our existing one. I'm open to changing this.

tomaka:
> Could you enumerate what other aspects are worth covering? Aspects that are unique in libp2p. We don't want to clone the Kademlia and friends literature.

Well, the 2k would include actually covering Kademlia, which I think we should do anyway, just not urgently.

I think there should be more explanation as to what actually happens on the network, rather than just dumping a protobuf definition file.
For example, which fields need to be field for which kind of RPC query? The format of the fields? Should nodes relay multiaddresses in their raw format, or produce error/ignore the ones it can't decode? Is there a limit to the number of multiaddresses that a peerinfo should contain?
That's just from the top of my head.
The reason why I didn't implement record store in rust-libp2p at the time is that the `Record` definition was way too vague to be helpful. I'd expect more help from a specs.

raulk:
@tomaka I agree that the RPC section needs redoing. The idea is to copy the protobufs, as these are normative for interoperability, and explain how each field is used and serialised (especially for the `bytes` types, which can be _anything_). @jhiesey, are you planning to tackle this?

I do recognise you've implemented this from scratch, and therefore your feedback is valuable. However, in all honesty, I don't see the value in reinventing the wheel and re-specifying the Kademlia baseline in this spec. I rather make it pre-required reading (like I've done), and build on top of it.

In a nutshell, I think of this spec as a _diff_ between baseline Kademlia and our implementation, which:
* deviates from Kademlia in some aspects, e.g. the way we manage buckets, the lack of PINGs, etc. (this needs specifying!)
* cherry-picks ideas from Coral, mainlineDHT, etc.

Maybe you can compile a list of all the areas you tripped over, and we can make sure to cover them?

Also, Kademlia is abstract, in the sense that it doesn't specify wire messages, RPCs, timeouts, etc. So our spec should define those aspects very clearly.

tomaka:
cc the second part of this comment: https://github.com/libp2p/specs/issues/111#issuecomment-442211341

kad-dht/README.md
raulk: @jhiesey

kad-dht/README.md
raulk: @jhiesey

kad-dht/README.md
raulk: @jhiesey

kad-dht/README.md
raulk: @jhiesey

kad-dht/README.md
raulk: @jhiesey

raulk:
@jhiesey tagged you in the comments that need your attention; let's try to push this through! ;-)

jhiesey:
@raulk sorry for not getting to this before now! Will work on this today.

kad-dht/README.md
jhiesey: I'm inclined to leave this in the spec for now, as handlers are implemented in both JS and Go, and it may turn out to be useful soon as an end-to-end connection test. If we decide we don't want it then let's remove it from the code too.

kad-dht/README.md
jhiesey: We have separate repos defining at least two of these interfaces abstractly; see https://github.com/libp2p/interface-peer-routing https://github.com/libp2p/interface-content-routing

Admittedly this isn't required for interoperability, but I think we should at least suggest relevant interfaces for the DHT's public API.

kad-dht/README.md
```diff
@@ -0,0 +1,288 @@
...
+
+The routing table is unfolded lazily, starting with a single bucket a position 0
+(representing the most distant peers), and splitting it subsequently as closer
+peers are found, and the capacity of the nearmost bucket is exceeded.
+
+## Alpha concurrency factor (α)
+
+The concurrency of node and value lookups are limited by parameter `α`, with a
+default value of 3. This implies that each lookup process can perform no more
+than 3 inflight requests, at any given time.
```
anacrolix: I'm not sure why the spec has this. In real-world implementations, a concurrency factor much, much higher is required to be reasonably fast.

kad-dht/README.md
```diff
@@ -0,0 +1,288 @@
...
+
+## Record keys
+
+Records in the DHT are keyed by CID [4]. There are intentions to move to
```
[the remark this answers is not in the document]
anacrolix: How does this work if the hash used changes? It seems that this allows you to choose a new hash to generate a new location on the DHT, that still references the same actual object.

kad-dht/README.md
anacrolix: Can validation be deferred to consumers of the DHT? It's not really a requirement to participate in it?

kad-dht/README.md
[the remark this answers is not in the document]
jhiesey: Right, all locations will end up changing if we change the hash. So we'll have to have a new (incompatible) DHT protocol at that point.

kad-dht/README.md
jhiesey: It's not strictly necessary, no. But it would be nice if nodes can throw out clearly bogus records instead of storing them. So this should be suggested but optional for now.

kad-dht/README.md
jhiesey: Right, this is a bug.

kad-dht/README.md
jhiesey: You're right, this probably doesn't belong in the spec. However, with the implementation of https://github.com/libp2p/go-libp2p-kad-dht/issues/146 landing, the number of outgoing requests is multiplied by a factor of the number of paths (currently default of 10). That will change the behavior in practice.

We'll have to do some testing to determine what this should be set to.

@raulk what do you think?

kad-dht/README.md
jhiesey: OK I think I've addressed them all :)

jhiesey:
I've addressed the issues I see. Left some more comments too.

kad-dht/README.md
jhiesey: That is, it would be nice if the node that receives a `PUT_VALUE` can do some sanity checking.

jhiesey:
What's the status on this @raulk? Anything I can do to help?

kad-dht/README.md
anacrolix: Hm, per the [comment on the refactor proposal](https://github.com/libp2p/research-dht/pull/8#issuecomment-453289582), the DHT node implementation could call a registered handler on receiving a `PUT_VALUE`, which does whatever it wishes with the data.

kad-dht/README.md
```diff
@@ -236,33 +257,119 @@ This process is repeated as many times per run as configuration parameter
...
+// for a key value pair
+message Record {
+	// The key that references this record
+	bytes key = 1;
+
+	// The actual value this record is storing
+	bytes value = 2;
+
+	// Note: These fields were removed from the Record message
+	// hash of the authors public key
```
anacrolix: At the network/RPC level, none of these last 3 fields exist in the Record, should they be removed from a spec?

kad-dht/README.md
anacrolix: I think shoehorning the DHT node implementation into these external interfaces is causing it to take on unnecessary complexity. Using the DHT for routing (to fit the interface) should be trivial to provide with a type that wraps the DHT node. This will free up some very bizarre methods and requirements that currently exist directly on the node implementation. Perhaps the spec should say that "The DHT node implementation may implement or provide the required features to implement the following interfaces:" or something to that effect.

kad-dht/README.md
jhiesey: Yes, that's the idea. We could move this validation outside the DHT itself.

kad-dht/README.md
jhiesey: Yes they should be removed

kad-dht/README.md
jhiesey: I'm inclined to agree, yes

jhiesey:
After reading @anacrolix 's feedback on this and [my refactor proposal](https://github.com/libp2p/research-dht/pull/8), I think we should simplify this DHT node spec substantially and move a bunch of the hairier stuff into separate discovery modules with their own specs.

Mikerah:
I noticed that some of the links the the bibliography were behind Springer's paywall. It would be awesome to provide the links to these papers from the author's website for example. I think this would increase the accessibility of the spec.

kad-dht/README.md
```diff
@@ -0,0 +1,395 @@
...
+	FindProvidersAsync(context.Context, cid.Cid, int) <-chan pstore.PeerInfo
+}
+
+// PeerRouting is a way to find information about certain peers.
+//
+// This can be implemented by a simple lookup table, a tracking server,
+// or even a DHT (like herein).
+type PeerRouting interface {
+	// FindPeer searches for a peer with given ID, returns a pstore.PeerInfo
+	// with relevant addresses.
```
alexh: Why is addresses plural?

anacrolix: Because peers can have multiple addresses: different ports, protocols, locations etc.

kad-dht/README.md
```diff
@@ -0,0 +1,395 @@
...
+	repeated Peer providerPeers = 9;
+}
+```
+
+Any time a relevant `Peer` record is encountered, the associated multiaddrs
+are stored in the node's peerbook.
+
+These are the requirements for each `MessageType`:
+* `FIND_NODE`: `key` must be set in the request. `closerPeers` is set in the
+response; for an exact match exactly one `Peer` is returned; otherwise `ncp`
```
jacobheun: `ncp` should be `k` (20) per kademlia and not 6, correct?

cmr:
S/Kademlia requires either "crypto puzzles" (proofs of work) or a centralized authority issuing certificates to prevent sybil and eclipse attacks from generating lots of node IDs. With node ID's being so easy to generate, how is this DHT protected against abuse?

kad-dht/README.md
```diff
@@ -0,0 +1,395 @@
...
+## Node lookups
+
+_WIP (raulk)._
+
+## Bootstrap process
+The bootstrap process is responsible for keeping the routing table filled and
+healthy throughout time. It runs once on startup, then periodically with a
+configurable frequency (default: 5 minutes).
+
+On every run, we generate a random node ID and we look it up via the process
```
Warchant: >we generate a random node ID

Specify how node ID is generated.
Is it the same as peer ID?

kad-dht/README.md
```diff
@@ -0,0 +1,395 @@
...
+* [Raúl Kripalani](https://github.com/raulk)
+* [John Hiesey](https://github.com/jhiesey)
+
+## Distance function (dXOR)
+
+The libp2p Kad DHT uses the **XOR distance metric** as defined in the original
+Kademlia paper [0]. Peer IDs are normalised through the SHA256 hash function.
+
+For recap, `dXOR(sha256(id1), sha256(id2))` is the number of common leftmost
+bits between SHA256 of each peer IDs. The `dXOR` between us and a peer X
```
bertrandfalguiere: Hi. I'm trying to write a bit of doc about the DHT here: https://github.com/ipfs/docs/pull/240 
I still need to wrap my head around some things.

But shouldn't dXOR be "256 minus the number of common leftmost bits"?
The spec implies dXOR(id1,id1) = 256.  I would expect dXOR(id1,id1) = 0 an input should have a distance of 0 with itself.
The spec implies dXOR(0000001111...,00000000000.......) = 6.  I would expect dXOR(0000001111...,00000000000.......) = 250, as a big distance would imply a big difference.
Am I missing something here?

Stebalien: This spec is currently incorrect. The actual distance is just the one in the kademlia paper: `id1 xor id2`. This section is confusing that with the how a peer's bucket is calculated. _That_ uses the number of shared bits:

```
Bucket 0: no shared bits.
Bucket 1: 1 shared bit.
Bucket 2: 2 shared bits.
...
Last bucket: everything else
```

Stebalien: (fixed)

kad-dht/README.md
```diff
@@ -0,0 +1,386 @@
...
+           with the peer who sent it in `Pb`.
+		2. Otherwise, we resolve the conflict by calling `Validator.Select(best,
+           new)`:
+			1. If the new value wins, store it in `best`, and mark all formerly
+               “best" peers (`Pb`) as _outdated peers_ (`Po`). The current peer
+               becomes the new best peer (`Pb`).
+			2. If the new value loses, we add the current peer to `Po`.
+	2. If successful without a value, the response will contain the closest
+       nodes the peer knows to the key `K`. Add them to the candidate list `Pn`,
+       except for those that have already been queried.
```
tmakarios, line 156:
```suggestion
	2. If successful with or without a value, the response will contain the closest
       nodes the peer knows to the key `K`. Add them to the candidate list `Pn`,
       except for those that have already been queried.
```

As the specification is currently worded, if (for example) `Q > α` and all of the first `α` queries return a value, then `Pn` will be empty, but `Q` answers won't yet have been collected.

kad-dht/README.md
```diff
@@ -0,0 +1,413 @@
...
+(default: 10 seconds), which upon firing, aborts the run.
+
+## RPC messages
+
+Remote procedure calls are performed by:
+
+1. Opening a new stream.
+2. Sending the RPC request message.
+3. Listening for the RPC response message.
+4. Closing the stream.
```
mxinden, line 278: From what I can tell, given my limited go-libp2p knowledge, [go-libp2p-kad-dht opens a new stream for each RPC request](https://github.com/libp2p/go-libp2p-kad-dht/blob/59321226c166b2da2ea96078e439a3e0fc34e63c/message_manager.go#L195). @aschmahmann could you confirm this?

I can confirm that this is the case for rust-libp2p.

See https://github.com/libp2p/specs/pull/108#issuecomment-438821336 for more details.

mxinden: Friendly ping @aschmahmann. Can you confirm my assumption above?

Stebalien: We actually leave the stream open for a bit (just in case we need to make more requests? Not sure how much it helps at our network scale). So the answer is:

1. Open a stream.
2. Make any number of requests on the stream (pipelined).
3. Close the stream when done.

mxinden: Thanks @Stebalien. Specified in https://github.com/libp2p/specs/pull/108/commits/6ae2551281df0a7c5f6dc7ec704f47cdffb4a84e.

For what it is worth rust-libp2p never re-uses streams, but allows for more than one request on an incoming stream.

kad-dht/README.md
mxinden: Thanks @tmakarios for pointing this out.

https://github.com/libp2p/specs/pull/108/commits/0126af14220d2935b6d0a173e65e3d73273f122a adjusts the specification accordingly.

For the record, both go-libp2p and rust-libp2p sent `closerPeers` detached from whether they sent `record`.

https://github.com/libp2p/go-libp2p-kad-dht/blob/eac1b5e1ad7a90ffee7a7d4b0650e04a621fbb84/handlers.go#L55-L91

https://github.com/libp2p/rust-libp2p/blob/c5bcada2c2a1e34b8e39b885f10ed7c2d00e8373/protocols/kad/src/behaviour.rs#L1779-L1804

kad-dht/README.md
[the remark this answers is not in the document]
mxinden: Fixed with https://github.com/libp2p/specs/pull/108/commits/c21ca3a0f3110946e86362e4d67782cb356b6882.

kad-dht/README.md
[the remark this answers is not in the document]
mxinden: Keys are sent as-is on the wire. Kademlia XOR distance is calculated via the key's sha256 hash.

Marking as resolved. Please re-open in case I am misunderstanding the question.

kad-dht/README.md
mxinden: Thanks @jhiesey. Documented in https://github.com/libp2p/specs/pull/108/commits/6c1c2248dc02979367615d8ca55b952a0dd7bdcf.

kad-dht/README.md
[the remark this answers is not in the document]
mxinden: I would like to keep this unspecified for now, leaving concrete k-bucket strategies up to the implementations, only adding a minimal specification via this pull request. https://github.com/libp2p/specs/pull/108/commits/defe08df0e6a684342412b135740ef54e6a33dae removes the paragraph. I am happy to discuss more specific k-bucket strategies in follow up pull requests. 

In case you object, please speak up :)

kad-dht/README.md
mxinden: This is removed with https://github.com/libp2p/specs/pull/108/commits/f240a22be4d93f873d154533e99d393102f3377c, due to reasons mentioned above. Please follow up in case you would still like to see this documented.

kad-dht/README.md
mxinden: https://github.com/libp2p/specs/pull/108/commits/e1442fa3f109944c8eda8368c781ef5935572806 deprecates the `PING` message type with the following description:

> * `PING`: Deprecated message type replaced by the dedicated [ping
  protocol][ping]. Implementations may still handle incoming `PING` requests for
  backwards compatibility. Implementations must not actively send `PING`
  requests.

Please comment on this pull request in case you would like to not deprecate `PING` messages.

Also see https://github.com/libp2p/go-libp2p-kad-dht/issues/31

> If we decide we don't want it then let's remove it from the code too.

For backwards compatibility I suggest to keep the handling logic. As far as I can tell go-libp2p and rust-libp2p never actively send `PING` messages.

kad-dht/README.md
mxinden: https://github.com/libp2p/specs/pull/108/commits/f05ee26ed4636a23f00ed86eeddf963ebb4de689 removes the DHT interfaces, capturing its structure and naming in the structure of the description itself. The `Validator` interface is kept as is, though the requirement of strictly implementing the interface is loosened. 

Let me know in case you have any objections.

kad-dht/README.md
mxinden: https://github.com/libp2p/specs/pull/108/commits/6909f8a97685c3f60679dbe7c9e81a5ab411bcf7 rewords the validation description, allowing implementations to use any mechanism for validation, including validation through consumers of the DHT. I am resolving this conversation for now. Let me know if you would prefer further adjustments of the section.

kad-dht/README.md
mxinden: For now, I would like to keep the concurrency factor default value at `3`. After all, implementations may diverge from the default values suggested in this specification. I think it is well worth it to extend this specification with a more in-depth discussion on query concurrency, though I would prefer this discussion to happen in a follow-up pull request.

Please let me know in case you object.

kad-dht/README.md
mxinden: I am keeping `author` and `signature` in the protobuf definition, commented out, to make sure protobuf field numbers are not reused for a different purpose in the future.

kad-dht/README.md
mxinden: :+1: done in https://github.com/libp2p/specs/pull/108/commits/f05ee26ed4636a23f00ed86eeddf963ebb4de689.

mxinden:
> Protocol Id is not specified. Implementations use `/ipfs/kad/1.0.0`. IMO should drop `ipfs` prefix, thus lets use `/kad/1.0.0`

@Warchant I am reluctant to document a specific default Protocl ID in this specification. I don't think there is much use in a global libp2p DHT. Instead in my opinion DHTs should be application specific (e.g. IPFS / Polkadot / ...). Please let me know in case you think otherwise.

kad-dht/README.md
mxinden: Simplified by using `peer ID` with https://github.com/libp2p/specs/pull/108/commits/a9ec52376b0cd923f8f1c6a8e6919c10d763805f.

kad-dht/README.md
```diff
@@ -0,0 +1,386 @@
...
+	Select(key string, values [][]byte) (int, error)
+}
+```
+
+`Validate()` is a pure function that reports the validity of a record. It may
+validate a cryptographic signature, or else. It is called on two occasions:
+
+1. To validate incoming values in response to `GET_VALUE` calls.
+2. To validate outgoing values before storing them in the network via
+   `PUT_VALUE` calls.
```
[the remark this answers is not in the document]
mxinden: Done in https://github.com/libp2p/specs/pull/108/commits/6909f8a97685c3f60679dbe7c9e81a5ab411bcf7.

kad-dht/README.md
```diff
@@ -0,0 +1,386 @@
...
+
+Similarly, `Select()` is a pure function that returns the best record out of 2
+or more candidates. It may use a sequence number, a timestamp, or other
+heuristic to make the decision.
```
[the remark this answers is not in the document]
mxinden: Rephrased in https://github.com/libp2p/specs/pull/108/commits/6909f8a97685c3f60679dbe7c9e81a5ab411bcf7. Happy to add more details on validation requirements, though I would prefer to merge a minimal version of the specification via this pull request first. Please comment on the pull request in case you object.

mxinden:
> _Update by @mxinden on 2021-05-14._
> 
> This pull request adds a **minimal** specification for the libp2p Kademlia DHT protocol. It is an early version and thus while striving for correctness it does not strive for completeness. The version proposed in this pull request sets a base layer specification to be expanded in the future.
> 
> Areas worth expanding in future pull requests:
> 
> - Detailed record validation
> - Default value adjustments and reasoning
> - Detailed request handling
> - Expanded k-bucket routing table management
> - ...

I would argue that a minimal correct but incomplete specification is better than no specification (status quo). With that in mind I am proposing to merge this pull request. I would very much appreciate peoples reviews.

kad-dht/README.md
```diff
@@ -0,0 +1,427 @@
...
+
+The record is rejected if the validation fails.
+
+#### Putting values
+
+To _put_ a value the DHT finds `k` or less closest peers to the key of the value
+using the `FIND_NODE` RPC (see [peer routing section](#peer-routing)), and then
+sends a `PUT_VALUE` RPC message with the record value to each of the peers.
+
+#### Getting values
```
Stebalien, line 140: 1. Most of this logic should go into a "query" section where we discuss how we find the closest peers.
2. This is now completely wrong. We (a) find the closest peers then (b) stop. We don't care about the number of values, that's not relevant.

mxinden: > where we discuss how we find the closest peers.

This is described in `### Peer routing`.

> This is now completely wrong. We (a) find the closest peers then (b) stop. We don't care about the number of values, that's not relevant.

I am not sure I follow. Could you expand? Just to avoid confusion, the procedure to find the closest peers is described in `### Peer routing`. Are you implying that go-libp2p-kad-dht dropped the notion of quorums for value lookups? rust-libp2p still offers specifying a quorum when looking up a value which is relevant outside of the ipfs space e.g. in Polkadot.

kad-dht/README.md
```diff
@@ -0,0 +1,395 @@
...
+	// Used to return a value
+	// PUT_VALUE, GET_VALUE
+	Record record = 3;
+
+	// Used to return peers closer to a key in a query
+	// GET_VALUE, GET_PROVIDERS, FIND_NODE
+	repeated Peer closerPeers = 8;
+
+	// Used to return Providers
+	// GET_VALUE, ADD_PROVIDER, GET_PROVIDERS
```
mxinden:
```suggestion
	// ADD_PROVIDER, GET_PROVIDERS
```

This seems to be a copy-paste error from the above `closerPeers` field. Am I missing something?

kad-dht/README.md
```diff
@@ -0,0 +1,427 @@
...
+key, the DHT finds the closest peers to the key using the `FIND_NODE` RPC (see
+[peer routing section](#peer-routing)), and then sends an `ADD_PROVIDER` RPC with
+its own `PeerInfo` to each of these peers.
+
+Each peer that receives the `ADD_PROVIDER` RPC should validate that the
+received `PeerInfo` matches the sender's `peerID`, and if it does, that peer
+must store the `PeerInfo` in its datastore.
+
+_Getting_ the providers for a given key is done in the same way as _getting_ a
+value for a given key (see [getting values section](#getting-values)) expect
```
mxinden, line 252:
```suggestion
value for a given key (see [getting values section](#getting-values)) except
```

kad-dht/README.md
```diff
@@ -0,0 +1,427 @@
...
+
+In all cases, the distance between two keys is `XOR(sha256(key1),
+sha256(key2))`.
+
+## DHT operations
+
+The libp2p Kademlia DHT offers the following types of routing operations:
+
+- **Peer routing** - _Finding_ the closest nodes to a given key (`FIND_NODE`).
+- **Value routing** - _Putting_ a value to the nodes closest to the value's key
```
petar, line 71: This is not a routing operation (as the section title suggests). Also the description is not accurate.
PUT_VALUE puts a value on the node it is called on. (Not on the "closest node to the key").

kad-dht/README.md
```diff
@@ -0,0 +1,427 @@
...
+  (`PUT_VALUE`) and _getting_ a value by its key from the nodes closest to that
+  key (`GET_VALUE`).
+- **Content routing** - _Adding_ oneself to the list of providers for a given
```
petar, line 74: Same comment as above.

kad-dht/README.md
```diff
@@ -0,0 +1,427 @@
...
+  key at the nodes closest to that key (`ADD_PROVIDER`) and _getting_ providers
+  for a given key from the nodes closest to that key (`GET_PROVIDERS`).
+
+In addition the libp2p Kademlia DHT offers the auxiliary _bootstrap_ operation.
+
+### Peer routing
+
+The below is one possible algorithm to find nodes closest to a given key on
+the DHT. Implementations may diverge from this base algorithm as long as they
```
petar, line 83: "Implementations may diverge from this base algorithm": so what's the point of describing this algorithm if our implementation is different and other implementations don't have to follow this?

kad-dht/README.md
```diff
@@ -0,0 +1,427 @@
...
+continue to adhere to the wire format.
+
+Let's assume we’re looking for nodes closest to key `Key`. We then enter an
+iterative network search.
+
+We keep track of the set of peers we've already queried (`Pq`) and the set of
+next query candidates sorted by distance from `Key` in ascending order (`Pn`).
+At initialization `Pn` is seeded with the `α` peers from our routing table we
```
petar, line 91: Our implementation, is seeded with K peers.

kad-dht/README.md
```diff
@@ -0,0 +1,427 @@
...
+know are closest to `Key`, based on the XOR distance function (see [distance
+definition](#distance)).
+
+Then we loop:
+
+1. > The lookup terminates when the initiator has queried and gotten responses
+   from the k (see [#replication-parameter-k]) closest nodes it has seen.
```
petar, line 98: In our implementation, a lookup actually terminates when a given percent (e.g. 80%) of the beta (another parameter) peers closest to the target have been contacted.

kad-dht/README.md
```diff
@@ -0,0 +1,427 @@
...
+
+   (See Kademlia paper [0].)
+
+   The lookup might terminate early in case the local node queried all known
+   nodes, with the number of nodes being smaller than `k`.
+2. Pick as many peers from the candidate peers (`Pn`) as the `α` concurrency
```
petar, line 104: When do you do the picking? Our implementation does this after a prior request completes and the state is updated.

mxinden: I am not sure I follow. Steps 1 - 4 are executed in a loop, thus picking happens at the very start as well as each time a response is received (3) and the termination criterion (1) is still false. As far as I can tell, this is inline with what you describe above. Does that make sense @petar?

kad-dht/README.md
```diff
@@ -0,0 +1,427 @@
...
+## Definitions
+
+### Replication parameter (`k`)
+
+The amount of replication is governed by the replication parameter `k`. The
+default value for `k` is 20.
+
+### Kademlia routing table
+
+The data structure backing this system is a k-bucket routing table, closely
```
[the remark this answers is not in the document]
mxinden: > So, what is this a spec of?

Good point. This should be a specification of the protocol. Not a description of a specific implementation.

> E.g. it is an implementation detail that we use kbucket data structure.

Thanks. 5a0cb410dd67874d49aa2b2816474e4d8ad11b9b refines this section, treating the data structure as an implementation detail. The section now contains your generic description using key prefix lengths.

> Other places in our stack use XOR-tries instead (which is generally recommended).

I was not aware of XOR-tries. I will take a deeper look. With 5a0cb410dd67874d49aa2b2816474e4d8ad11b9b the specification suggests XOR-tries as one possible data structure.

kad-dht/README.md
mxinden: Changed via 60f4c8d31dda8c4652ef8f6d9b36185f06c72306. Let me know what you think @petar.

kad-dht/README.md
mxinden: Changed via 60f4c8d31dda8c4652ef8f6d9b36185f06c72306. Let me know what you think @petar.

mxinden (commented):
Thanks @petar for the review! Very much appreciated.

I have addressed all your comments. Would you mind taking another look?

mxinden:
@petar friendly ping. Would you mind taking another look?

@aschmahmann could you give this pull request a review as well?

kad-dht/README.md
petar: Sounds good. I would just elaborate slightly, something like: "may diverge as long as they adhere to the wire format and make progress towards the target key"

kad-dht/README.md
petar: That sounds fine. What you've written will work. It'll just wait quite a bit before it is ready to terminate, but it is correct.

kad-dht/README.md
petar: You are right... I overlooked. This looks fine.

petar:
@mxinden Looks much better. Thank you!

kad-dht/README.md
mxinden: Added in https://github.com/libp2p/specs/pull/108/commits/b07409173c9efbde4b31deb87b3c30b17c0c7379.

kad-dht/README.md
```diff
@@ -0,0 +1,453 @@
...
+behaviour similar to Kademlia-based libraries.
+
+Code snippets use a Go-like syntax.
+
+## Definitions
+
+### Replication parameter (`k`)
+
+The amount of replication is governed by the replication parameter `k`. The
+default value for `k` is 20.
```
aschmahmann, line 46: When we use default throughout this document what do we mean? Is it the IPFS DHT (i.e. only applies to network parameters), the defaults in go-libp2p-kad-dht, or something else? If it's go-libp2p-kad-dht then  how problematic is it if some of the algorithms/query approaches are different then the ones mentioned here?

aschmahmann (changes requested):
Thanks, this is much improved. Left a few comments/questions.

kad-dht/README.md
```diff
@@ -0,0 +1,453 @@
...
+healthy throughout time. It runs once on startup, then periodically with a
+configurable frequency (default: 5 minutes).
+
+On every run, we generate a random peer ID and we look it up via the process
+defined in [peer routing](#peer-routing). Peers encountered throughout the
+search are inserted in the routing table, as per usual business.
+
+This process is repeated as many times per run as configuration parameter
+`QueryCount` (default: 1). Every repetition is subject to a `QueryTimeout`
+(default: 10 seconds), which upon firing, aborts the run.
```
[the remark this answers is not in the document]
mxinden: :+1: c755a419228709a8b17809a3c4f983ac3d26659b includes self lookup.

> This isn't what go-libp2p does.

Neither does rust-libp2p. It first does a lookup for the local key and then tries to [probabilistically generate a key for each bucket](https://github.com/libp2p/rust-libp2p/blob/e8fed53598696a45a26866408534cfa186b23d4a/protocols/kad/src/behaviour.rs#L1063-L1075).

> I can give the brief approach taken in go-libp2p-kad-dht if you're interested.

Yes.

kad-dht/README.md
```diff
@@ -0,0 +1,453 @@
...
+
+    // The actual value this record is storing
+    bytes value = 2;
+
+    // Note: These fields were removed from the Record message
+    //
+    // Hash of the authors public key
+    // optional string author = 3;
+    // A PKI signature for the key+value+author
+    // optional bytes signature = 4;
```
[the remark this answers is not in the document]
mxinden: I kept them in to give some background on what these numbers were originally used for. In my eyes, with the note at the top `// Note: These fields were removed from the Record message` this shouldn't confuse anyone.

Not a strong opinion. Would you prefer this to be changed?

kad-dht/README.md
```diff
@@ -0,0 +1,453 @@
...
+    // GET_VALUE, ADD_PROVIDER, GET_PROVIDERS
+    repeated Peer providerPeers = 9;
+}
+```
+
+These are the requirements for each `MessageType`:
+
+* `FIND_NODE`: In the request `key` must be set to the binary `PeerId` of the
+  node to be found. `closerPeers` is set in the response; for an exact match
+  exactly one `Peer` is returned; otherwise `k` closest `Peer`s are returned.
```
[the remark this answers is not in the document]
mxinden: As in the ID of the peer matches the target `key` in the original `FIND_NODE` request. Does that make sense @aschmahmann?

mxinden (commented):
Thanks @aschmahmann for the thorough review, especially with the additional background.

I replied to all of your comments. Would you mind taking another look?

kad-dht/README.md
```diff
@@ -0,0 +1,453 @@
...
+default value of 3. This implies that each lookup process can perform no more
+than 3 inflight requests, at any given time.
+
+## DHT operations
+
+The libp2p Kademlia DHT offers the following types of operations:
+
+- **Peer routing**
+
+  - Finding the closest nodes to a given key via `FIND_NODE`.
```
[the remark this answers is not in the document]
aschmahmann: > On FIND_NODE a DHT server returns a single DHT client if and only if the DHT client's peer ID matches the target.

Not quite, if we happen to know about a peer whose peerID is an exact match for the FIND_NODE query key then we add them to our list of servers to return. https://github.com/libp2p/go-libp2p-kad-dht/blob/6fff2a3b391f73da7a1c7558e27725ebe730e7ba/handlers.go#L256

> In my eyes documenting this distinction without documenting the DHT client / server mode extension (see your comment on rust-libp2p) would be inconsistent.

I guess that's fair. If all nodes were servers this distinction wouldn't really matter. A follow up PR is fine. Note: client mode isn't just for unreachable peers, it's generally recommended for "low quality" peers. This could mean low power, low bandwidth, low network uptime, etc. Basically peers that if they were servers would do more harm then good probably shouldn't be servers but may still want to make queries.

kad-dht/README.md
aschmahmann: Maybe recommended is better.

What I was trying to distinguish was between network-wide recommendations (e.g. k=20) and implementation-specific recommendations (e.g. alpha = 3).

My thinking was that network-wide parameters are much easier to give recommendations for. Client stuff is more a "here's one way to do it". Maybe this distinction is too in the weeds though. Your call

kad-dht/README.md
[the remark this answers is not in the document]
aschmahmann: I'm not really sure why we don't always return the `k` closest peers but by https://github.com/libp2p/go-libp2p-kad-dht/blob/6fff2a3b391f73da7a1c7558e27725ebe730e7ba/handlers.go#L256

the only time we return a single peer is when Peer A asks Peer B "do you know peer B" and they say "yes, it's me". If Peer A asks Peer B "do you know peer C" they will say "yes here's there address along with `k` peers closer to them"

kad-dht/README.md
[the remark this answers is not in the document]
aschmahmann: Your call. Was thinking it might make the long spec a little shorter was all.

kad-dht/README.md
```diff
@@ -0,0 +1,453 @@
...
+    `GET_VALUE`.
+
+- **Content provider advertisement and discovery**
+
+  - Adding oneself to the list of providers for a given key at the nodes closest
+    to that key by finding the closest nodes via `FIND_NODE` and then adding
+    oneself via `ADD_PROVIDER`.
+
+  - Getting providers for a given key from the nodes closest to that key via
+    `GET_PROVIDERS`.
```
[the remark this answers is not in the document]
aschmahmann: > Is this good enough?

Probably?

> Am I missing something @aschmahmann?

Nope, I guess it's just an obvious sort of thing. It could have been designed that `GET_VALUE` only returned the value/error instead of also sending back closest peers but this way is more efficient.

kad-dht/README.md
```diff
@@ -0,0 +1,453 @@
...
+
+To _put_ a value the DHT finds `k` or less closest peers to the key of the value
+using the `FIND_NODE` RPC (see [peer routing section](#peer-routing)), and then
+sends a `PUT_VALUE` RPC message with the record value to each of the peers.
+
+#### Value retrieval
+
+When _gettting_ a value in the DHT, the implementor should collect at least `Q`
+(quorum) responses from distinct nodes to check for consistency before returning
+an answer.
```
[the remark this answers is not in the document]
aschmahmann: I liked your wording "Implementations may diverge from this base algorithm as long as they adhere
to the wire format and make progress towards the target key." in that it indicated what we care about vs what we don't.

If you think it's obvious that all of this quorum stuff is optional client behavior and people can do whatever they want then feel free to leave it as is. If you think it's not really obvious, then it might be worth adding some wording around how when getting values you can abort whenever you feel confident you have the latest value, for example using a quorum.

kad-dht/README.md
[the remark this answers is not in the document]
mxinden: Thanks for looking this up. dab454963595e670ae4d254d125cab4b1f3fa5ab simplifies the specification, always returning the `k` closest peers. I don't think we should document / require the peculiarity of go-libp2p-kad-dht returning a single peer when oneself was looked up. Sounds good?

kad-dht/README.md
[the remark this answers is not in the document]
mxinden: 324f9155e57af15bca9c373ae823bf8f7710f810 extends the section, stressing the fact that quorum is one mechanism among many to determine when a query is finished. Let me know what you think.

kad-dht/README.md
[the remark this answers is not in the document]
aschmahmann: I think that's right, or if it's important we can catch it in the follow up PR. We should investigate historical context here and see if a change is merited in go-libp2p-kad-dht.

kad-dht/README.md
```diff
@@ -0,0 +1,453 @@
...
+
+	// Select selects the best record from the set of records (e.g., the
+	// newest).
+	//
+	// Decisions made by select should be stable.
+	Select(key string, values [][]byte) (int, error)
+}
+```
+
+`Validate()` is a pure function that reports the validity of a record. It may
```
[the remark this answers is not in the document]
aschmahmann: Ya, it's probably fine. Once we've got something we can start to clean it up, including removing language specific details that seem extraneous in a spec.

kad-dht/README.md
```diff
@@ -0,0 +1,439 @@
...
+On every run, we generate a random peer ID and we look it up via the process
+defined in [peer routing](#peer-routing). Peers encountered throughout the
+search are inserted in the routing table, as per usual business.
+
+This is repeated as many times per run as configuration parameter `QueryCount`
+(default: 1). In addition, to improve awareness of nodes close to oneself,
+implementations should include a lookup for their own peer ID.
+
+Every repetition is subject to a `QueryTimeout` (default: 10 seconds), which
+upon firing, aborts the run.
```
aschmahmann, line 287: Should include our usual, here's one simple way of doing this. Especially since neither go nor rust libp2p's implementations actually do this.

aschmahmann (approved):
Added one more comment, and I'm sure there are nits to be had in places but this is way better than nothing and given we are planning a follow up PR to iron out more details of the spec (e.g. support for clients) I'm happy here. Thanks for all the hard work 😄

My 2c are that as the specs evolve, including making an IPFS DHT spec, we can start to separate out the segments on, basic wire formats/rules ("the spec"), how to build compatible networks ("how to use the spec"), and how client implementations might work ("how to work with spec compliant networks").

kad-dht/README.md
mxinden: :+1: added via https://github.com/libp2p/specs/pull/108/commits/3e6f8f5e9a98b7078fb9ae7020179aa7d2c86947.

mxinden:
@tomaka @alanshaw you are both still tracked as "requesting changes". I have addressed your comments and suggestions above. Could you either remove your request for changes or give this pull request another review?

BigLep:
@mxinden : given this has had lots of eyeballs and iterations, I would suggest we give a cutoff date for comments (e.g., EOD 2021-06-29).  Otherwise we merge.  Further improvements can always be made after if necessary.

mxinden:
:tada: After more than two years, this is ready to be merged. Thanks to everyone involved! :tada:

PR #119, opened 2018-12-28
Add info about limiting the size of the data and topicIDs fields.

jamesray1:


Stebalien:
Let's consider just limiting the size of the *entire* message. That should be simpler to validate (given that we can't easily validate data/topic limits until we finish reading the message).

pubsub/README.md
```diff
@@ -90,7 +92,10 @@ the peer won't be able to send identical messages (e.g. keepalives) within the
 timecache interval, as they will get rejected as duplicates."
 
 The `topicIDs` field specifies a set of topics that this message is being
-published to.
+published to. To prevent the size of this field from growing indefinitely,
+it is suggested that that the application using an implementation of this
+library spec should have some kind of storage economics (see e.g.
+[here](https://ethresear.ch/t/draft-position-paper-on-resource-pricing/2838).
```
vyzo: that's probably too much information, is it even relevant to recommend storage economics?

jamesray1: Maybe not so much for non-blockchain applications.

pubsub/README.md
```diff
@@ -92,14 +90,23 @@ the peer won't be able to send identical messages (e.g. keepalives) within the
...
-[here](https://ethresear.ch/t/draft-position-paper-on-resource-pricing/2838).
+published to.
 
 Note that messages are currently *not* signed. This will come in the near
 future.
 
+The size of the `Message` should be limited, say to 1 MiB, but could also
+be configurable, for more information see 
+[issue 118](https://github.com/libp2p/specs/issues/118), while messages should be
+rejected if they are over this size (this isn't done as of Dec 2018).
```
vyzo: this is incorrect, messages over 1MB _are_ rejected, by virtue of a read error and closing the stream.

vyzo: at least in go-libp2p-pubsub.

PR #123, opened 2019-01-18
Add protocols for IPv6 address resolution

dlrobertson:
Add [NDISC] and [DHCPv6] to the section discussing existing
protocols.

[NDISC]: https://tools.ietf.org/html/rfc4861
[DHCPv6]: https://tools.ietf.org/html/rfc3315


dlrobertson:
Just a small nit. The doc includes ARP && DHCP but not the IPv6 address resolution protocol nor DHCPv6.

PR #124, opened 2019-01-21
Specify specifc types for implementation interoperability

AgeManning:
I would like to propose adding specific data types for the pubsub protocol such that new implementations are interoperable with current implementations (specifically the go implementation). 

This example specifies the protobuf bytes `sequence_number` of pubsub to be a 64-bit big-endian uint as is implemented in the go-implementation. 

If this is desired, I can continue this PR with all other information that is sent across the wire using the go-implementation as the standard.

### DIFF
--- a/pubsub/README.md
+++ b/pubsub/README.md
@@ -80,8 +80,8 @@ The `seqno` field is a 64-bit big-endian uint that is a linearly increasing
 number that is unique among messages originating from each given peer. No two
 messages on a pubsub topic from the same peer should have the same `seqno`
 value, however messages from different peers may have the same sequence number,
-so this number alone cannot be used to address messages. (Notably the
-'timecache' in use by the floodsub implementation contains a `message_id`,
+so this number alone cannot be used to address messages. Notably the
+'timecache' in use by the go implementation contains a `message_id`,
 which is constructed from the concatenation of the `seqno` and the `from`
 fields. This `message_id` is then unique among messages. It was also proposed
 in [#116](https://github.com/libp2p/specs/issues/116) to use a `message_hash`,
--- a/pubsub/README.md
+++ b/pubsub/README.md
@@ -76,15 +76,15 @@ done to allow content to be routed through a swarm of pubsubbing peers.
 The `data` field is an opaque blob of data, it can contain any data that the
 publisher wants it to.
 
-The `seqno` field is a linearly increasing number that is unique among messages
-originating from each given peer. No two messages on a pubsub topic from the
-same peer should have the same `seqno` value, however messages from different
-peers may have the same sequence number, so this number alone cannot be used to
-address messages. (Notably the 'timecache' in use by the floodsub
-implementation contains a `message_id`, which is constructed from the
-concatenation of the `seqno` and the `from` fields. This `message_id` is then
-unique among messages. It was also proposed in
-[#116](https://github.com/libp2p/specs/issues/116) to use a `message_hash`,
+The `seqno` field is a 64-bit big-endian uint that is a linearly increasing
+number that is unique among messages originating from each given peer. No two
+messages on a pubsub topic from the same peer should have the same `seqno`
+value, however messages from different peers may have the same sequence number,
+so this number alone cannot be used to address messages. (Notably the
+'timecache' in use by the floodsub implementation contains a `message_id`,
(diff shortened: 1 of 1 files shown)


tomaka:
We should remove "linearly increasing", as it makes the protocol attackable.

AgeManning:
> We should remove "linearly increasing", as it makes the protocol attackable.
@tomaka - I notice you set the sequence to 20 random bytes in rust-libp2p. I didn't understand the attack logic. A message is unique via (peerId, seq_no). If someone knows future sequence numbers, do they not also have to duplicate the `peerId` to DOS future messages on the network?

vyzo:
we use message signing now (which needs to be spec'ed), which nullifies any possible attack vectors on the sequence number.

pubsub/README.md
```diff
@@ -76,15 +76,15 @@ done to allow content to be routed through a swarm of pubsubbing peers.
...
-address messages. (Notably the 'timecache' in use by the floodsub
-implementation contains a `message_id`, which is constructed from the
-concatenation of the `seqno` and the `from` fields. This `message_id` is then
-unique among messages. It was also proposed in
-[#116](https://github.com/libp2p/specs/issues/116) to use a `message_hash`,
+The `seqno` field is a 64-bit big-endian uint that is a linearly increasing
+number that is unique among messages originating from each given peer. No two
+messages on a pubsub topic from the same peer should have the same `seqno`
+value, however messages from different peers may have the same sequence number,
+so this number alone cannot be used to address messages. (Notably the
```
vyzo: can we get rid of the parentheses here?

pubsub/README.md
```diff
@@ -76,15 +76,15 @@ done to allow content to be routed through a swarm of pubsubbing peers.
...
+'timecache' in use by the floodsub implementation contains a `message_id`,
```
vyzo: this is not just floodsub, it is common in all routing algorithms in `go-libp2p-pubsub` so I would use different wording here (say "the go implementation").

pubsub/README.md
AgeManning: This commit only adds `is a 64-bit big-endian uint` to the first sentence. All the other changes are just from re-formatting the paragraph to the standard line width. 
I'm happy to make further modifications to the old text if you like.

pubsub/README.md
AgeManning: See previous comment.

AgeManning:
@vyzo - For this PR, I was mainly asking if we wanted to include more specifications in the spec. For example - the `topicId` is a protobuf String. I think the go implementation uses plain strings for `topicId`'s and the rust implementation uses hashes for the `topicId`. 
Can we standardize these? If we use hashes, which hash?

If we are using message signing now, then I guess the `from` field should also be specified as the public key for the signing algorithm. 

My intention here is not to be pedantic, but to make it easier and clear for the implementations to all agree on the data being sent to each other, such that each implementation can function on the same network.

vyzo:
The `topicID`s are plain strings in all implementations as far as i know, this was discussed a while ago.
As for message signing, yes, this needs to be spec'ed. I will update the spec to reflect the implementation of message signing in go.

AgeManning:
@vyzo - Thanks for the clarification. I'll leave you to it. If I find any further specifications that I think may need to be added, I'll let you know. 

@tomaka - Have I misunderstood the rust implementation? It seems to send `TopicHash`'s for messages and topic strings for subscriptions. For other clients, should these both be topic strings?

https://github.com/libp2p/rust-libp2p/blob/615dd3332fd4d14d9f41ffa69ee4beb841aece98/protocols/floodsub/src/protocol.rs#L100

https://github.com/libp2p/rust-libp2p/blob/615dd3332fd4d14d9f41ffa69ee4beb841aece98/protocols/floodsub/src/protocol.rs#L113

tomaka:
The story behind `TopicHash` is that I implemented it as a hash because the documentation said it was a hash. Then someone noticed that in Go it was actually not a hash, but that using a hash would be a good idea. Some people then wanted to have a new version of floodsub where it is a hash, while some others were preferring backwards compatibility. That was six months ago. No decision has been taken yet.

jamesray1:
I've also been using a `TopicHash` in [this gossipsub impl](https://github.com/libp2p/rust-libp2p/pull/767) as well as a `MsgHash` instead of a `MsgId`.

Issue #125, opened 2019-01-21
Clarification for the ambient peer discovery section of the gossipsub spec


Mikerah:
At the beginning of the ambient peer discovery section, it starts with 

> With ambient peer discovery, the function is pushed outside the scope of the protocol. 

Is this meant to say that floodsub doesn't depend on a particular method of peer discovery, e.g. a DHT, but depends on peers themselves to announce themselves to other peers. Then, the protocol will check if that peer is running floodsub or gossipsub and responds accordingly? Not really sure what function is pushed outside the scope of the protocol is suppose to mean.

raulk:
It's a rather cryptic way of saying that finding pubsub peers is a _passive_ action, that relies on the discovery mechanism that the libp2p node is using. This can be peer routing via DHT, rendezvous, local discovery, etc.

go-libp2p-pubsub [registers a notifee](https://github.com/libp2p/go-libp2p-pubsub/blob/master/notify.go#L18) with the libp2p stack and reacts to new connections being made [by opening a stream](https://github.com/libp2p/go-libp2p-pubsub/blob/master/comm.go#L59) for the currently supported pubsub routing protocol IDs (gossipsub, floodsub).

If the other party ACKs the protocol negotiation, we know we have found a pubsub peer and we add them to the overlay.

raulk:
@Mikerah does that help?

Mikerah:
@raulk, yes it does. Thanks

PR #127, opened 2019-01-22
Added a clarification to the Ambient Peer Discovery section

Mikerah:
I asked about the Ambient Peer Discovery section in gossipsub in issue #125. @raulk responded with a more details explanation. I added the clarifications to the gossipsub specification.

### DIFF
--- a/pubsub/gossipsub/README.md
+++ b/pubsub/gossipsub/README.md
@@ -54,12 +54,13 @@ It implements pubsub in the most basic manner, with two defining aspects:
 
 ### Ambient Peer Discovery
 
-In ambient peer discovery, discovering other pubsub peers is a passive action. 
-In other words, pubsub peers discover other peers by relying on 
-the discovery mechanism that peer is using, e.g. DHT, rendezvous,etc. Whenever a new
-peer is connected, the protocol checks to see if the peer implements
-a pubsub variant such as floodsub, and if so, it sends a hello packet that announces the topics
-that it is currently subscribing to. As such, the functionality of peer discovery is pushed outside of the pubsub protocol.
+With ambient peer discovery, the function is pushed outside the scope of the
+protocol. Instead, the mechanism for discovering peers is provided for by the
+environment. In practice, this can be embodied by DHT walks, rendezvous
+points, etc. This protocol relies on the ambient connection events produced by
+such mechanisms. Whenever a new peer is connected, the protocol checks to see
+if the peer implements floodsub and/or gossipsub, and if so, it sends it a
+hello packet that announces the topics that it is currently subscribing to.
 
 This allows the peer to maintain soft overlays for all topics of
 interest. The overlay is maintained by exchanging subscription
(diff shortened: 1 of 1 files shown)


pubsub/gossipsub/README.md
```diff
@@ -54,11 +54,11 @@ It implements pubsub in the most basic manner, with two defining aspects:
 
 ### Ambient Peer Discovery
 
-With ambient peer discovery, the function is pushed outside the scope
-of the protocol. Instead, it relies on ambient connection events to
-perform peer discovery via protocol identification. Whenever a new
+In ambient peer discovery, discovering other pubsub peers is a passive action.
```
vyzo: please don't change the original sentence; you can add a clarifying sentence on top.

Mikerah: I think the original sentence is lacking. It doesn't specify what function is pushed outside the scope of the protocol. If the function you are referring to is peer discovery then you should state it. Otherwise, the sentence doesn't make much sense.

vyzo: It makes perfect sense to me, i am not going to change it... but I am fine with adding a clarification.

Mikerah: Of course it makes perfect to you, you wrote the specification. However, when writing specifications, you aren't writing them for yourself but for others. If you add a clarification for what "the function" pertains to, then you can keep the sentence with the added clarification.

vyzo: the function is the topic of the section; ambient peer discovery.

raulk: BTW – we’re missing an intro here. That’s my fault! @vyzo, @mikerah is implementing gossipsub-js from scratch, and she would like to rely on the spec as much as possible, but is having to fall back to the implementation very frequently for clarifications. So I think her fresh view will bring maturity and help us make this important spec more accessible to a wider audience.

Mikerah: Any updates on this PR?

Mikerah: I think it's quite circular. Upon a first reading of it, it's quite confusing. I only started to get it after multiple reads and opening an issue in this repository. I'm pretty sure there are others that might feel the same way.

vyzo: If you feel that way, you are very welcome to add a clarifying sentence :)

Mikerah: I updated your sentence to be more clear.

vyzo: you are still changing the original sentence, which I very much like.  Please don't.

raulk: I think the new sentence is imprecise in the sense that "discovery of pubsub peers" is too coarse grained. This can be decomposed in "discovering a peer" and "identifying it's a compatible pubsub peer". I do see how the original sentence is confusing. May I suggest this change?

```suggestion
With ambient peer discovery, the mechanism for discovering peers is provided for by the environment. In practice, this can be embodied by DHT walks, rendezvous points, etc. This protocol relies on the ambient connection events produced by such mechanisms. Whenever a new peer is connected, the protocol checks to see if the peer implements floodsub and/or gossipsub, and if so, it sends it a hello packet that announces the topics that it is currently subscribing to. 
```

vyzo: And can we please leave my sentence unmolested? I am issuing a veto here :)

raulk: Molestation reverted 😛 I added a few clarifying sentences in https://github.com/libp2p/specs/pull/127/commits/ecf4623b13bba68e108185b7c78ae787094ad3d0. LGTY?

raulk:
Thanks @Mikerah and @vyzo!

PR #147, opened 2019-02-20
Add a see also section with Whisper v2.0

jamesray1:


jamesray1:
I would consider it relevant, since the Web3 Foundation rejected giving a grant for gossipsub development, with the reason that they are taking a different direction with messaging protocols, and referring me upon my querying to https://github.com/w3f/messaging.

vyzo:
I don't think this relevant to the pubsub spec at all, really.

jamesray1:
It's an alternative decentralised Privacy-Preserving Communication protocol for web3 that is intending to integrate with libp2p. I think saying it is not relevant to pubsub and closing this after such a short time seems hasty.

vyzo:
I know what whisper is, but this is the libp2p pubsub specification. Whisper has nothing to do with this spec.

Issue #111, opened 2018-11-14
Specify peer ID hash function in public key


Stebalien:
In the beginning, peer IDs were base58 encoded sha256 multihashes. However, being able to embed a key into a peer ID was rather tempting so we introduced a IDFromEd25519PublicKey function. This function created a *different* peer ID that embedded the public key.

Unfortunately,

1. Users had to remember to call this.
2. If a user called the normal `IDFromPublicKey` function, they'd get a different peer ID for the same key. This obviously isn't good as our security transports derive peer IDs from public keys instead of *validating* peer IDs (this makes it impossible to mis-validate).

Therefore, I submitted https://github.com/libp2p/go-libp2p-peer/pull/30 which:

1. Removed IDFromEd25519PublicKey.
2. Changed inline peer IDs to just inline the protobuf (instead of introducing an entirely new format).
3. Auto-inlined peer ids <= 42 bytes.

Unfortunately, this last bit appears to be a problem... At the time, I was under the impression that nobody had actually gotten around to *using* ed25519 keys as inlining didn't really work (libp2p would still use the sha256 peer IDs). However, it turns out that OpenBazaar was using them anyways.

Second, in retrospect, this wasn't even the optimal solution. Really, we also need to support arbitrary hash functions but this only adds support for the identity hash function.

---

So, my proposal here is to:

1. Revert that change. I'm pretty sure OpenBazaar is the only ed25519 user in the wild so that should be safe.
2. Introduce a new field in the public/private key protobuf that explicitly specifies the hash function (defaulting to "sha256" if missing).

Open question: Should we use this as an opportunity to switch to multibase-enabled peer IDs now? One significant issue with the current embedding method is that it makes it *slightly* harder to switch to multibase peer IDs (cc @lidel). I've previously proposed switching to CID-based peer IDs but that may be more work than it's worth (a motivation would be transfering keys over bitswap, easier to support large quantum-safe keys, format flexibility, etc.).

lidel:
Current representation of PeerIDs is base58, which is case-sensitive and thus not Origin-safe.  **In the context of IPFS  CIDs switching to base32** (https://github.com/ipfs/ipfs/issues/337) the time feels right to solve Origin problem for IPNS websites as well. 

So the ask from Web Browsers WG is:

- IPNS paths (peer/keyIDs) should have Origin-safe base32 identifiers –  https://github.com/ipfs/go-ipfs/issues/5287
[...]

PR #151, opened 2019-02-28
libp2p TLS 1.3 Handshake

marten-seemann:
In the last couple of weeks I spent a lot of time designing a TLS 1.3 based handshake for libp2p. This document has come a long way, and I managed to remove a lot of the complexity, while at the same time preserving the future extensibility which I originally built into this version of the protocol.

This is a proposal how we can run the libp2p handshake using TLS 1.3 in the future. Since we're currently only using secio in production, we are completely free to design things the way we want, without having to worry about backwards compatibility. That means that **now** is the right time to criticize every single aspect of this protocol.

I already implemented a proof of concept in https://github.com/libp2p/go-libp2p-tls/pull/20, feel free to play around with that code as well.

### DIFF
--- a/tls/design considerations.md
+++ b/tls/design considerations.md
@@ -19,7 +19,7 @@ The handshake protocol described here:
 
 ### TLS 1.3 - What about older versions?
 
-The handshake protocol requires TLS 1.3 support. This means that the handshake between to peers that have never communicated before will typically complete in just a single roundtrip. With older TLS versions, a handshake typically takes two roundtrips. By not specifying support for older TLS versions, we increase perfomance and simplify the protocol.
+The handshake protocol requires TLS 1.3 support. This means that the handshake between two peers that have never communicated before will typically complete in just a single roundtrip. With older TLS versions, a handshake typically takes two roundtrips. By not specifying support for older TLS versions, we increase performance and simplify the protocol.
 
 
 ### Why we're not using the host key for the certificate
@@ -44,4 +44,4 @@ An earlier version of this document included a version negotiation mechanism. Wh
 To keep things simple, the current proposal does not include a version negotiation mechanism. A future version of this protocol might:
 
 1. Change the format in which the keys are transmitted. A x509 extension has an ID (the Objected Identifier, OID), so we can use a new OID if we want to change the way we encode information. x509 certificates allow use to include multiple extensions, so we can even send the old and the new version during a transition period. In the handshake protocol defined here, peers are required to skip over extensions that they don't understand.
-2. For more involved changes, a new version might (ab)use the SNI field field in the ClientHello to announce support for new versions. To allow for this to work, the current version requires clients to send anything in the SNI field and server to completely ignore this field, no matter what its contents are.
+2. For more involved changes, a new version might (ab)use the SNI field in the ClientHello to announce support for new versions. To allow for this to work, the current version requires clients to send anything in the SNI field and server to completely ignore this field, no matter what its contents are.
--- a/tls/design considerations.md
+++ b/tls/design considerations.md
@@ -1,0 +1,47 @@
+# Design considerations for the libp2p TLS Handshake
+
+## Requirements 
+
+There are two main requirements that prevent us from using the straightforward way to run a TLS handshake (which would be to simply use the host key to create a self-signed certificate).
+
+1. We want to use different key types: RSA, ECDSA, and Ed25519, Secp256k1 (and maybe more in the future?).
+2. We want to be able to send the key type along with the key (see https://github.com/libp2p/specs/issues/111).
+
+The first point is problematic in practice, because Go currently only supports RSA and ECDSA certificates. Support for Ed25519 was planned for Go 1.12, but was deferred recently, and the Go team is now evaluating interest in this in order to prioritze their work, so this might or might not happen in Go 1.13. I'm not aware of any plans for Secp256k1 at the moment.
+The second requirement implies that we might want add some additional (free-form) information to the handshake, and we need to find a field to stuff that into.
+
+The handshake protocol described here:
+* supports arbitrary keys, independent from what the signature algorithms implemented by the TLS library used
+* defines a way how future versions of this protocol might be negotiated without requiring any out-of-band information and additional roundtrips
+
+
+## Design Choices
+
+### TLS 1.3 - What about older versions?
+
+The handshake protocol requires TLS 1.3 support. This means that the handshake between to peers that have never communicated before will typically complete in just a single roundtrip. With older TLS versions, a handshake typically takes two roundtrips. By not specifying support for older TLS versions, we increase perfomance and simplify the protocol.
+
+
+### Why we're not using the host key for the certificate
+
+The current proposal uses a self-signed certificate to carry the host's public key in the libp2p Public Key Extension. The key used to generate the self-signed certificate has no relationship with the host key. This key can be generated for every single connection, or can be generated at boot time.
+
+One optimisation that was considered when designing the protocol was to use the libp2p host key to generate the certificate in the case of RSA and ECDSA keys (which we can assume to be supported signature schemes by all peers). That would have allowed us to strip the host key and the signature from the key extension, in order to
+
+1. reduce the size of the certificate and
+2. reduce the number of signature verifications the peer has to perform from 2 to 1.
+
+The protocol does not include this optimisation, because
+
+1. assuming that the peer uses an ECDSA key for generating the self-signed certificate, this only saves about ~150 bytes if the host key is an ECDSA key as well, and it even slightly increases the size of the certificate in case of a RSA host key. Furthermore, for ECDSA keys, the size of all handshake messages combined is less than 900 bytes, so having a slightly larger certificate won't require us to send more (TCP / QUIC) packets.
+2. For a client, the number of signature verifications shouldn't pose a problem, since it controls the rate of its dials. Only for servers this might be a problem, since a malicious client could force a server to waste resources on signature verification. However, this is not a particularly interesting DoS vector, since the client's certificate is sent in its second flight (after receiving the ServerHello and the server's certificate), so it requires the attacker to actually perform most of the TLS handshake, including encrypting the certificate chain with a key that's tied to that handshake.
+
+
+### Versioning - How we could roll out a new version of this protocol in the future
+
+An earlier version of this document included a version negotiation mechanism. While it is a desireable property to be able to change things in the future, it also adds a lot of complexity.
+
+To keep things simple, the current proposal does not include a version negotiation mechanism. A future version of this protocol might:
+
+1. Change the format in which the keys are transmitted. A x509 extension has an ID (the Objected Identifier, OID), so we can use a new OID if we want to change the way we encode information. x509 certificates allow use to include multiple extensions, so we can even send the old and the new version during a transition period. In the handshake protocol defined here, peers are required to skip over extensions that they don't understand.
+2. For more involved changes, a new version might (ab)use the SNI field field in the ClientHello to announce support for new versions. To allow for this to work, the current version requires clients to send anything in the SNI field and server to completely ignore this field, no matter what its contents are.
--- a/tls/tls.md
+++ b/tls/tls.md
@@ -27,14 +27,14 @@ Note for clients: Since clients complete the TLS handshake immediately after sen
 ### libp2p Public Key Extension
 
 In order to prove ownership of its host key, an endpoint sends two values:
-- the public key corresponding to its host key
-- a signature performed using the host key
+- the public host key
+- a signature performed using the private host key
 
-The public key corresponding to the host key allows the peer to calculate the peer ID of the peer it is connecting to. Clients MUST verify that the peer ID derived from the certificate matches the peer ID they intended to connect to, and MUST abort the connection if it there is a mismatch.
+The public host key allows the peer to calculate the peer ID of the peer it is connecting to. Clients MUST verify that the peer ID derived from the certificate matches the peer ID they intended to connect to, and MUST abort the connection if it there is a mismatch.
 
-The peer signs the public key that it used to generate the certificate carrying the libp2p Public Key Extension using the its host key. This signature provides cryptographic proof that the peer was in possession of the private key at the time the certificate was signed. Peers MUST verify the signature, and abort the connection attempt if signature verification fails.
+The peer signs the public key that it used to generate the certificate carrying the libp2p Public Key Extension using the its private host key. This signature provides cryptographic proof that the peer was in possession of the private host key at the time the certificate was signed. Peers MUST verify the signature, and abort the connection attempt if signature verification fails.
 
-The public key and the signature are ANS.1-encoded into the SignedKey data structure, which is carried in the libp2p Public Key Extension. The libp2p Public Key Extension is a x509 extension with the Object Identier 1.3.6.1.4.1.XXX.YYY.
+The public host key and the signature are ANS.1-encoded into the SignedKey data structure, which is carried in the libp2p Public Key Extension. The libp2p Public Key Extension is a x509 extension with the Object Identier 1.3.6.1.4.1.XXX.YYY.
 
 TODO: Nothing will break if we just use an arbitrary value for XXX. However, if we want to do things correctly, [OID](https://en.wikipedia.org/wiki/Object_identifier) PENs should be registered with [IANA](https://pen.iana.org/pen/PenApplication.page). 
 
@@ -45,7 +45,7 @@ SignedKey ::= SEQUENCE {
 }
 ```
 
-The publicKey field of SignedKey contains the public key of the endpoint, encoded using the following protobuf.
+The publicKey field of SignedKey contains the public host key of the endpoint, encoded using the following protobuf.
 
 ```protobuf
 enum KeyType {
--- a/tls/tls.md
+++ b/tls/tls.md
@@ -30,9 +30,9 @@ In order to prove ownership of its host key, an endpoint sends two values:
 - the public key corresponding to its host key
 - a signature performed using the host key
 
-The public key allows the peer to calculate the peer ID of the peer it is connecting to. Clients MUST verify that the peer ID derived from the certificate matches the peer ID they intended to connect to, and MUST abort the connection if it there is a mismatch.
+The public key corresponding to the host key allows the peer to calculate the peer ID of the peer it is connecting to. Clients MUST verify that the peer ID derived from the certificate matches the peer ID they intended to connect to, and MUST abort the connection if it there is a mismatch.
 
-The peer signs its public key using the its host key. This signature provides cryptographic proof that the peer was in possession of the private key at the time the certificate was signed. Peers MUST verify the signature, and abort the connection attempt if signature verification fails.
+The peer signs the public key that it used to generate the certificate carrying the libp2p Public Key Extension using the its host key. This signature provides cryptographic proof that the peer was in possession of the private key at the time the certificate was signed. Peers MUST verify the signature, and abort the connection attempt if signature verification fails.
 
 The public key and the signature are ANS.1-encoded into the SignedKey data structure, which is carried in the libp2p Public Key Extension. The libp2p Public Key Extension is a x509 extension with the Object Identier 1.3.6.1.4.1.XXX.YYY.
 
--- a/tls/tls.md
+++ b/tls/tls.md
@@ -34,9 +34,7 @@ The public host key allows the peer to calculate the peer ID of the peer it is c
 
 The peer signs the public key that it used to generate the certificate carrying the libp2p Public Key Extension using the its private host key. This signature provides cryptographic proof that the peer was in possession of the private host key at the time the certificate was signed. Peers MUST verify the signature, and abort the connection attempt if signature verification fails.
 
-The public host key and the signature are ANS.1-encoded into the SignedKey data structure, which is carried in the libp2p Public Key Extension. The libp2p Public Key Extension is a x509 extension with the Object Identier 1.3.6.1.4.1.XXX.YYY.
-
-TODO: Nothing will break if we just use an arbitrary value for XXX. However, if we want to do things correctly, [OID](https://en.wikipedia.org/wiki/Object_identifier) PENs should be registered with [IANA](https://pen.iana.org/pen/PenApplication.page). 
+The public host key and the signature are ANS.1-encoded into the SignedKey data structure, which is carried in the libp2p Public Key Extension. The libp2p Public Key Extension is a x509 extension with the Object Identier 1.3.6.1.4.1.53594.1.1.
 
 ```asn1
 SignedKey ::= SEQUENCE {
--- a/tls/tls.md
+++ b/tls/tls.md
@@ -2,14 +2,14 @@
 
 ## Introduction
 
-This document describes how [TLS 1.3](https://tools.ietf.org/html/rfc8446) is used to secure libp2p connections. Endpoints authenticates to their peers by encoding their public key into a x509 certificate extension. The protocol described here allows peers to use arbitrary key types, not constrained to those for which signing of a x509 certificates is specified.
+This document describes how [TLS 1.3](https://tools.ietf.org/html/rfc8446) is used to secure libp2p connections. Endpoints authenticate to their peers by encoding their public key into a x509 certificate extension. The protocol described here allows peers to use arbitrary key types, not constrained to those for which signing of a x509 certificates is specified.
 
 
 ## Handshake Protocol
 
 The libp2p handshake uses TLS 1.3 (and higher). Endpoints MUST NOT negotiate lower TLS versions.
 
-During the handshake, peers authenticate each other’s identity as described in [Peer Authentication](#peer-authentication). Endpoints MUST verify the peer's identy. Specifically, this means that servers MUST require clients authentication during the TLS handshake, and MUST abort a connection attempt if the client fails to provide the requested authentication information.
+During the handshake, peers authenticate each other’s identity as described in [Peer Authentication](#peer-authentication). Endpoints MUST verify the peer's identity. Specifically, this means that servers MUST require client authentication during the TLS handshake, and MUST abort a connection attempt if the client fails to provide the requested authentication information.
 
 
 ## Peer Authentication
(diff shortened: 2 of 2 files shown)


anacrolix:
Could you include why we want TLS 1.3-like handshakes? Is it because of 0-RTT, or just keeping up with the times?

tls/tls.md
```diff
@@ -0,0 +1,69 @@
...
+The signature provides cryptographic proof that the peer was in possession of the private key at the time the certificate was signed. Peers MUST verify the signature, and abort the connection attempt if signature verification fails.
+
+The public key and the signature are ANS.1-encoded into the SignedKey data structure, which is carried in the libp2p Public Key Extension. The libp2p Public Key Extension is a x509 extension with the Object Identier 1.3.6.1.4.1.XXX.YYY.
+
+TODO: Nothing will break if we just use an arbitrary value for XXX. However, if we want to do things correctly, [OID](https://en.wikipedia.org/wiki/Object_identifier) PENs should be registered with [IANA](https://pen.iana.org/pen/PenApplication.page). 
+
+```asn1
+SignedKey ::= SEQUENCE {
+  publicKey BIT STRING,
+  signature BIT STRING
```
Kubuxu: What is the content of the signature? As in, what is being signed?

tls/tls.md
```diff
@@ -0,0 +1,69 @@
...
+
+
+## Peer Authentication
+
+In order to be able use arbitrary key types, peers don’t use their host key to sign the x509 certificate they send during the handshake. Instead, the host key is encoded into the [libp2p Public Key Extension](#libp2p-public-key-extension), which is carried in a self-signed certificate. The key used to generate and sign this certificate SHOULD NOT be related to the host's key. Endpoints MAY generate a new key and certificate for every connection attempt, or they MAY reuse the same key and certificate for multiple connections. Endpoints MUST choose a key that will allow the peer to verify the certificate (i.e. choose a signature algorithm that the peer supports), and SHOULD use a key type which allows for efficient signature computation and which reduces the combined size of the certificate and the signature.
```
magik6k: Do we do this only because Go doesn't sepport some protocols in it's TLS implementation?

marten-seemann:
@anacrolix it’s in the design considerations. TLS 1.3 just takes a single round trip, older versions take two. And there’s little reason to roll out a new Handshake protocol on top of an old TLS version anyway.

tls/tls.md
marten-seemann: That's one reason, the other is https://github.com/libp2p/specs/issues/111.

tls/tls.md
marten-seemann: It's the public key. I added the missing sentence.

tls/tls.md
```diff
@@ -0,0 +1,69 @@
...
+
+### libp2p Public Key Extension
+
+In order to prove ownership of its host key, an endpoint sends two values:
+- the public key corresponding to its host key
+- a signature performed using the host key
+
+The public key allows the peer to calculate the peer ID of the peer it is connecting to. Clients MUST verify that the peer ID derived from the certificate matches the peer ID they intended to connect to, and MUST abort the connection if it there is a mismatch.
+
+The peer signs its public key using the its host key. This signature provides cryptographic proof that the peer was in possession of the private key at the time the certificate was signed. Peers MUST verify the signature, and abort the connection attempt if signature verification fails.
```
Kubuxu: Q: Which public key is signed?
Q: I feel like it should somehow be bound to TLS session we are establishing. Otherwise, this proof could be "stolen" and reused by someone else.

marten-seemann: It's the public key used to generate the certificate. I'll update the PR to clarify that.
Why do you want to bind it to the session that we're establishing? What we're doing here is basically creating a certificate chain, with the host key at the root. Just that we're not using x509 certificates, due to the limitations described earlier.

yusefnapora: I initially thought that the `publicKey` field of `SignedKey` was referring to the host key. I think it's this sentence which is a bit ambiguous: 

> The peer signs its public key using the its host key.

 Since you were just talking about the public key for the peer ID in the prior paragraph, I assumed that's what "public key" referred to here.

It's clearer down below, where it says "The publicKey field of SignedKey contains the public key of the endpoint", although it wouldn't hurt to spell out that the endpoint key is the one used to generate the certificate.

Since it is the endpoint key being signed, if you wanted to have the proof be bound to the session as @Kubuxu suggests, you could just generate a new key and self-signed certificate for each session.  Is that correct @marten-seemann?

tls/design considerations.md
```diff
@@ -0,0 +1,47 @@
...
+The handshake protocol described here:
+* supports arbitrary keys, independent from what the signature algorithms implemented by the TLS library used
+* defines a way how future versions of this protocol might be negotiated without requiring any out-of-band information and additional roundtrips
+
+
+## Design Choices
+
+### TLS 1.3 - What about older versions?
+
+The handshake protocol requires TLS 1.3 support. This means that the handshake between to peers that have never communicated before will typically complete in just a single roundtrip. With older TLS versions, a handshake typically takes two roundtrips. By not specifying support for older TLS versions, we increase perfomance and simplify the protocol.
```
tomaka: Most importantly, it would make us vulnerable to downgrade attacks.

tls/design considerations.md
```diff
@@ -0,0 +1,47 @@
...
+2. For a client, the number of signature verifications shouldn't pose a problem, since it controls the rate of its dials. Only for servers this might be a problem, since a malicious client could force a server to waste resources on signature verification. However, this is not a particularly interesting DoS vector, since the client's certificate is sent in its second flight (after receiving the ServerHello and the server's certificate), so it requires the attacker to actually perform most of the TLS handshake, including encrypting the certificate chain with a key that's tied to that handshake.
+
+
+### Versioning - How we could roll out a new version of this protocol in the future
+
+An earlier version of this document included a version negotiation mechanism. While it is a desireable property to be able to change things in the future, it also adds a lot of complexity.
```
tomaka: What about the protocol string in `multistream-select`? That's a version negotiation mechanism.

tls/tls.md
```diff
@@ -0,0 +1,69 @@
...
+TODO: Nothing will break if we just use an arbitrary value for XXX. However, if we want to do things correctly, [OID](https://en.wikipedia.org/wiki/Object_identifier) PENs should be registered with [IANA](https://pen.iana.org/pen/PenApplication.page). 
+
+```asn1
+SignedKey ::= SEQUENCE {
+  publicKey BIT STRING,
+  signature BIT STRING 
+}
+```
+
+The publicKey field of SignedKey contains the public key of the endpoint, encoded using the following protobuf.
```
tomaka: That seems redundant with https://github.com/libp2p/specs/pull/100

tls/tls.md
Kubuxu: I'm still not clear about which keys, public keys and certificates are being sent over and in which fields.

I don't see cleanly that the chain confirming ownership of key used in TLS handshake is bound to a PeerID.
I have few ideas regarding possible attacks, and possible reuse in case TLS key is leaked but first I need to understand this part fully.

Kubuxu: I'm reading it again so bear with me.

tls/tls.md
```diff
@@ -0,0 +1,69 @@
...
+The certificate MUST contain the [libp2p Public Key Extension](#libp2p-public-key-extension). If this extension is missing, endpoints MUST abort the connection attempt. The certificate MAY contain other extensions, implementations MUST ignore extensions with unknown OIDs.
+
+Note for clients: Since clients complete the TLS handshake immediately after sending the certificate (and the TLS ClientFinished message), the handshake will appear as having succeeded before the server had the chance to verify the certificate. In this state, the client can already send application data. If certificate verification fails on the server side, the server will close the connection without processing any data that the client sent.
+
+
+
+### libp2p Public Key Extension
+
+In order to prove ownership of its host key, an endpoint sends two values:
+- the public key corresponding to its host key
```
Kubuxu: So my main issue is this sentence, I don't understand the difference between `public key` and `host key`.

burdges:
As Noise is more flexible than TLS 1.3, we might support specific Noise handshakes for specific nodes roles, but TLS 1.3 supports cipher suite agility and conceivably helps obfuscate traffic, but only if integrated with the web-like transports.

As a naive first approximation, we might consider (a) replacing secio with Noise in a way exploits Noise's flexibility, (b) integrate TLS 1.3 with the web-like transports in a way that really permits blending in with web traffic, and (c) correctly permit both Noise and TLS 1.3 to manage authentication using key provided by the application.  In fact, there is an enormous nasty can of worms involved in doing traffic obfuscation like I suggest in doing (b), for which the Tor project recruits an army of young, idealistic, and talented developers to write pluggable transports, and for which U.S. State Dept pays them like $2 M per year.  We should not compete with them here, so maybe the correct solution for obfuscation is actually TLS 1.3 or Noise running over Tor pluggable transports.  <shrug>

We're left with TLS 1.3 being the only secure protocol with cipher suite agility, and Noise providing more handshake flexibility.  Increasingly, we view cipher suite agility as harmful, based in part on how hard doing it correctly proved for TLS, so modern protocols should dump cipher suite agility and simply version instead.  IPFS appears rather committed to cipher suite agility though.  <shrug>

tls/design considerations.md
```diff
@@ -0,0 +1,47 @@
...
+## Design Choices
+
+### TLS 1.3 - What about older versions?
+
+The handshake protocol requires TLS 1.3 support. This means that the handshake between to peers that have never communicated before will typically complete in just a single roundtrip. With older TLS versions, a handshake typically takes two roundtrips. By not specifying support for older TLS versions, we increase perfomance and simplify the protocol.
+
+
+### Why we're not using the host key for the certificate
+
+The current proposal uses a self-signed certificate to carry the host's public key in the libp2p Public Key Extension. The key used to generate the self-signed certificate has no relationship with the host key. This key can be generated for every single connection, or can be generated at boot time.
```
raulk: IMO another reason for not using the host key directly is to achieve Perfect Forward Secrecy.

tls/tls.md
raulk: I think generating a new key per session should be a MUST, otherwise Perfect Forward Secrecy might be compromised.

tls/tls.md
marten-seemann: Sorry for the confusion here. You sign the public key that was used to generate the certificate with your host private key.

The reason is that the key used to generate the certificate is used by TLS during the handshake to sign the CertificateVerify message. By signing the public key of the certificate we get a chain of verified signatures.

Using the host's public key would be a bad idea. An attacker could just copy the contents of the libp2p Public Key Extension and use that to impersonate this peer. We really need an operation that proves that the peer actually meant to use the key used by TLS.

tls/design considerations.md
marten-seemann: That works for TCP, but not for QUIC. In QUIC, the first packet sent by the client already contains the ClientHello.

tls/design considerations.md
marten-seemann: TLS 1.3 uses ephemeral Diffie-Hellman for the key exchange mechanism, so it's always PFS, no matter what kind of certificate you use.

tls/tls.md
marten-seemann: Let me expand on my [earlier comment](https://github.com/libp2p/specs/pull/151#discussion_r261432354). In TLS 1.3, the client send a key_share extension in the ClientHello. This key share is the client's ephemeral DH key (i.e. it is a fresh value for every connection). The server sends its part of the ephemeral DH key share in the key_share extension in the ServerHello.
Since TLS 1.3 decouples key exchange mechanism from the signature algorithms, TLS 1.3 handshakes are always forward secure, no matter what kind of certificate you use.

By the way, you might have heard of [eTLS 1.3](https://www.etsi.org/deliver/etsi_ts/103500_103599/10352303/01.01.01_60/ts_10352303v010101p.pdf). It's supposed to be used by businesses who want middleboxes to be able to decode their traffic (and who don't care about their customer's privacy). It removes PFS from the protocol by having the server send a static key share in the ServerHello. The sad thing is, there's no immediate way to detect a server that's misbehaving in that way, other than running heuristics over multiple connections.

marten-seemann:
@burdges
> As Noise is more flexible than TLS 1.3, we might support specific Noise handshakes for specific nodes roles, but TLS 1.3 supports cipher suite agility and conceivably helps obfuscate traffic, but only if integrated with the web-like transports.

I'm a bit skeptical about obfuscation. As you note, obfuscation done right is really hard, and we even fail at the most basic aspects. The first one is that you'd have to run your node on port 443, a request to a non-standard HTTPS port will immediately stick out. Second, there's the ALPN extension which is sent unencrypted in the ClientHello, and I'd like to avoid putting a fake value into that field. And then there are hard parts like traffic patterns, which are very characteristic for HTTPS traffic, which will probably be impossible to replicate in libp2p.

> As a naive first approximation, we might consider (a) replacing secio with Noise in a way exploits Noise's flexibility, (b) integrate TLS 1.3 with the web-like transports in a way that really permits blending in with web traffic, and (c) correctly permit both Noise and TLS 1.3 to manage authentication using key provided by the application.

The advantage of TLS 1.3 at this point is that we can use the same handshake logic on TCP and on QUIC. Noise looks really interesting and we might be able to use it on top of TCP, but it's not obvious how to integrate it into the QUIC handshake.

marten-seemann:
@Kubuxu I just pushed an update that (hopefully) makes it clear which key I'm referring to. Can you please have another look?

Kubuxu:
It is much clearer now.

One issue I see (and it is a small issue that depends on our threat model) is if ever the private key for the outside certificate was leaked, it gives the attacker an ability to impersonate owner of the host key forever.

What do you think about this?

Kubuxu:
I've created this diagram to make it easier to understand: 
libp2p-tls

<details>
<summary>Source</summary>

```
@startuml
top to bottom direction
object "Cert Key" as certKey


package "x509 Certificate" {
object "Certificate" as cert {
  libp2p Public Key Extensions
}
certKey -[#blue]-> cert : signs

object "libp2p Public Key Extension" as pke {
    hostPubKey : PublicKey
    signature
}

object signature
}

cert --* pke
pke --* signature

object PeerID {
  hostKey
}
PeerID -[#blue]> signature : signs
pke .> PeerID : hostKey proves 
signature .> certKey :  prooves
certKey -[hidden]-> signature
certKey -[hidden]-> pke
certKey -[hidden]-> cert


@enduml
```
</details>

marten-seemann:
@Kubuxu: Thanks for the diagram!

> One issue I see (and it is a small issue that depends on our threat model) is if ever the private key for the outside certificate was leaked, it gives the attacker an ability to impersonate owner of the host key forever.

That's true, but I'm wondering if this is a realistic threat. After all, losing your host key would also allow an attacker to impersonate you forever (obviously). If at all, the host key will be more vulnerable to be compromised, since it's typically written to disk (at least for IPFS) and loaded into memory, where as the key for the outer certificate only needs to exist in memory.
That being said, an easy mitigation for this would be to attach an expiration date to the signature, so an attacker could only impersonate the host until that signature expires. Not sure if the additional complexity is justified. What do you think?

tls/tls.md
```diff
@@ -0,0 +1,69 @@
...
+
+The public host key allows the peer to calculate the peer ID of the peer it is connecting to. Clients MUST verify that the peer ID derived from the certificate matches the peer ID they intended to connect to, and MUST abort the connection if it there is a mismatch.
+
+The peer signs the public key that it used to generate the certificate carrying the libp2p Public Key Extension using the its private host key. This signature provides cryptographic proof that the peer was in possession of the private host key at the time the certificate was signed. Peers MUST verify the signature, and abort the connection attempt if signature verification fails.
+
+The public host key and the signature are ANS.1-encoded into the SignedKey data structure, which is carried in the libp2p Public Key Extension. The libp2p Public Key Extension is a x509 extension with the Object Identier 1.3.6.1.4.1.XXX.YYY.
+
+TODO: Nothing will break if we just use an arbitrary value for XXX. However, if we want to do things correctly, [OID](https://en.wikipedia.org/wiki/Object_identifier) PENs should be registered with [IANA](https://pen.iana.org/pen/PenApplication.page).
```
raulk: I requested a PEN number and will report back as soon as we have it (ETA: 1 week).

tls/tls.md
Stebalien: > The key used to generate and sign this certificate SHOULD NOT be related to the host's key.

This precludes the optimization we talked about where a peer MAY derive the certificate key from the host key and the verifier MAY skip verifying the signature in the extension if the key matches the certificate key.

I guess _SHOULD_ allows implementations to do what they want, but we may want to say that implementations MUST NOT reject certificates that use the host key.

tls/tls.md
```diff
@@ -0,0 +1,69 @@
...
+	ECDSA = 3;
+}
+
+message PublicKey {
+	required KeyType Type = 1;
+	required bytes Data = 2;
+}
+```
+
+TODO: PublicKey.Data looks underspecified. Define precisely how to marshal the key.
```
Stebalien: Some discussion here: https://github.com/libp2p/specs/pull/100/files.

tls/tls.md
Stebalien: I think we still need to specify how it's encoded. Also, WRT my comment (https://github.com/libp2p/go-libp2p-tls/pull/20#discussion_r263695084) I'd kind of like to sign `libp2p-tls-certificate:$PublicKey`.

tls/tls.md
marten-seemann: Yes, SHOULD means that you're not supposed to do it, unless you think you have good reasons for it. I've described some measurements I did in the Design Considerations, section Why we're not using the host key for the certificate, which led me to conclude that this optimization is not worth it. Do you agree?

tls/tls.md
marten-seemann: We have a bunch of specs that only live in PRs now, but don't seem to be blocked by anything major. We should get them merged some time, so we can properly refer to them.

tls/tls.md
Stebalien: It probably isn't but there's no reason to preclude it. Basically, I just don't want some over-zealous dev to reject these keys.

marten-seemann: Sure, that's why it's a SHOULD. SHOULD implies that you can't rely on the peer following the recommendation, so I don't think any additional text is needed.

tls/design considerations.md
```diff
@@ -0,0 +1,47 @@
...
+2. For a client, the number of signature verifications shouldn't pose a problem, since it controls the rate of its dials. Only for servers this might be a problem, since a malicious client could force a server to waste resources on signature verification. However, this is not a particularly interesting DoS vector, since the client's certificate is sent in its second flight (after receiving the ServerHello and the server's certificate), so it requires the attacker to actually perform most of the TLS handshake, including encrypting the certificate chain with a key that's tied to that handshake.
+
+
+### Versioning - How we could roll out a new version of this protocol in the future
+
+An earlier version of this document included a version negotiation mechanism. While it is a desireable property to be able to change things in the future, it also adds a lot of complexity.
+
+To keep things simple, the current proposal does not include a version negotiation mechanism. A future version of this protocol might:
```
vyzo: that's a bit of a hole here, let's at least make it compatible with negotiation in the future.

tls/tls.md
raulk: They came back very quickly. Our PEN assignment is 53594. As of today, line number 213928 in https://www.iana.org/assignments/enterprise-numbers/enterprise-numbers.

tls/tls.md
marten-seemann: I added the prefix.

tls/design considerations.md
Stebalien: Ideally, that's what (2) below provides. We can use the SNI field as it's sent in the first packet.

tls/design considerations.md
jacobheun:
```suggestion
The handshake protocol requires TLS 1.3 support. This means that the handshake between two peers that have never communicated before will typically complete in just a single roundtrip. With older TLS versions, a handshake typically takes two roundtrips. By not specifying support for older TLS versions, we increase performance and simplify the protocol.
```

tomaka:
For what it's worth, the `rustls` library (Rust-TLS), which is used by `quinn`, the only viable QUIC implementation in Rust at the moment, seems to always reject self-signed certificates on purpose (https://github.com/ctz/rustls/issues/143).

If I were to implement this on a short time frame, I'd generate a CA certificate, embed it in the source code as a trusted source, and sign the ephemeral certificates from this CA.

tls/tls.md
```diff
@@ -0,0 +1,67 @@
...
+
+Endpoints MUST NOT send a certificate chain that contains more than one certificate. The certificate MUST have NotBefore and NotAfter fields set such that the certificate is valid at the time it is received by the peer. When receiving the certificate chain, an endpoint MUST check these conditions and abort the connection attempt if the presented certificate is not yet valid or if it is expired.
+
+The certificate MUST contain the [libp2p Public Key Extension](#libp2p-public-key-extension). If this extension is missing, endpoints MUST abort the connection attempt. The certificate MAY contain other extensions, implementations MUST ignore extensions with unknown OIDs.
+
+Note for clients: Since clients complete the TLS handshake immediately after sending the certificate (and the TLS ClientFinished message), the handshake will appear as having succeeded before the server had the chance to verify the certificate. In this state, the client can already send application data. If certificate verification fails on the server side, the server will close the connection without processing any data that the client sent.
+
+
+
+### libp2p Public Key Extension
```
tomaka: What's the exact string name of the extension to use when generating the certificate?

marten-seemann: I'm not sure I understand the question. You can find the OID further down in the document.

marten-seemann:
> For what it's worth, the rustls library (Rust-TLS), which is used by quinn, the only viable QUIC implementation in Rust at the moment, seems to always reject self-signed certificates on purpose (ctz/rustls#143).

This sounds like something that should be fixed in rustls. Accepting self-signed certificate should be a basic feature of any TLS library. I really don't want to increase the length of the certificate chain to work around that shortcoming (thereby increasing the number of signature verifications to be performed during each handshake as well as significantly increase the number of bytes transmitted). Any estimate how long would it take to get this fixed in rustls?

tomaka:
> Accepting self-signed certificate should be a basic feature of any TLS library.

That's debatable. I'm pretty sure the authors of rustls would disagree.

Stebalien:
They probably don't want to blindly trust self-signed certificates but I doubt they'd object to a custom certificate validator if cleanly implemented.

marten-seemann:
@tomaka I'd like to merge this PR very soon. What do you think is the best way forward here?

tls/design considerations.md
```diff
@@ -0,0 +1,47 @@
+# Design considerations for the libp2p TLS Handshake
+
+## Requirements 
+
+There are two main requirements that prevent us from using the straightforward way to run a TLS handshake (which would be to simply use the host key to create a self-signed certificate).
+
+1. We want to use different key types: RSA, ECDSA, and Ed25519, Secp256k1 (and maybe more in the future?).
+2. We want to be able to send the key type along with the key (see https://github.com/libp2p/specs/issues/111).
+
+The first point is problematic in practice, because Go currently only supports RSA and ECDSA certificates. Support for Ed25519 was planned for Go 1.12, but was deferred recently, and the Go team is now evaluating interest in this in order to prioritze their work, so this might or might not happen in Go 1.13. I'm not aware of any plans for Secp256k1 at the moment.
```
burdges: You should fork the required library and use Ed25519 using agl's libraries.  ECDSA works but why bother.  RSA should not even be considered here.

Stebalien: Really, we might wan to rephrase this in terms of ecosystem support. That is, TLS is a rather complicated protocol and we don't want to force libp2p implementations to add support for new key types.

marten-seemann: > You should fork the required library and use Ed25519 using agl's libraries.

I considered that, but concluded that's it's not a good solution. It would mean pulling in a forked x509 package, and in order to use that one, we'd also have to fork tls just to change the import path.

tomaka: As a general remark, in my opinion whether or not a library exists in the ecosystem should only be a very minor consideration when writing specs. You want your specs to do *the right thing*, not the thing that can be implemented the most easily.

tomaka: Which is why I didn't answer your question "Any estimate how long would it take to get this fixed in rustls?" (https://github.com/libp2p/specs/pull/151#issuecomment-475864990)
To me how long something takes to implement is not relevant when discussing specs.

burdges: Filo has a branch at https://github.com/golang/go/issues/25355   I'm surprised this takes google so long given agl's role in Ed25519 being accepted by standards bodies.

tls/design considerations.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+### Versioning - How we could roll out a new version of this protocol in the future
+
+An earlier version of this document included a version negotiation mechanism. While it is a desireable property to be able to change things in the future, it also adds a lot of complexity.
+
+To keep things simple, the current proposal does not include a version negotiation mechanism. A future version of this protocol might:
+
+1. Change the format in which the keys are transmitted. A x509 extension has an ID (the Objected Identifier, OID), so we can use a new OID if we want to change the way we encode information. x509 certificates allow use to include multiple extensions, so we can even send the old and the new version during a transition period. In the handshake protocol defined here, peers are required to skip over extensions that they don't understand.
+2. For more involved changes, a new version might (ab)use the SNI field in the ClientHello to announce support for new versions. To allow for this to work, the current version requires clients to send anything in the SNI field and server to completely ignore this field, no matter what its contents are.
```
Stebalien: What was our motivation for using SNI instead of a custom extension again?

marten-seemann: A custom extension would definitely be the cleaner way to implement this. However, we‘d have to fork the Go standard library implementation and add support for that ourselves. Not sure how this would work out in JS- and Rust-land.

marten-seemann:
@tomaka: Is there any way to define a custom certificate verifier in rustls. Or if that’s not possible, could you temporarily add the certificate to the root CAs until it is verified?
Would using a different TLS library be an option? I saw that there are multiple TLS 1.3 implementations in Rust.

tls/design considerations.md
Stebalien: Given that we don't even _need_ this yet, we can probably cross that bridge when we get there. Honestly, I'd kind of like to be able to use the SNI field for the expected server peer ID (for multiplexing multiple peers through some kind of reverse proxy).

tomaka:
> Is there any way to define a custom certificate verifier in rustls. Or if that’s not possible, could you temporarily add the certificate to the root CAs until it is verified?

It is possible to have a custom verifier, but once a connection is established it is not possible to know which certificate it corresponded to (cc https://github.com/libp2p/rust-libp2p/issues/211#issuecomment-475879633).
However that can probably be fixed by tweaking rustls.

Again, the specs shouldn't be based on whether they are easy to implement.

marten-seemann:
@tomaka Are you ok with merging the specs as they are now, or do you think we need to change anything?

tomaka:
I'm fine with merging that!

My "regret" is that the libp2p handshake will still be distinguishable from an HTTP handshake, but there's no way to fix that as far as I know.

Kubuxu:
@tomaka I'm not sure if it will be distinguishable, TLS1.3 establishes encrypted (not authenticated) channel after ClientHello and ServerHello. Certificate, CertificateRequest are sent in a separate message: EncryptedExtensions.

I think this should make libp2p handshake and https handshake roughly the same.

marten-seemann:
@tomaka: The cleartext messages of the handshake will be identical to an HTTP handshake if we use the same ALPN (the spec doesn't say anything about which ALPN to use, so we're free to specify that later). However, as long as people don't run their nodes on port 443, and connect from random port numbers, libp2p traffic can be easily distinguished from genuine HTTP traffic just by looking at the IP header.

PR #161, opened 2019-04-01
WIP connmgr v2 specification

raulk:
Heavy WIP. Open to high-level feedback.

- [X] introduction
- [X] high-level overview
- [ ] quota allocation
- [ ] protocol enrollment
- [ ] stream scoring
- [ ] locking
- [ ] bursting
- [ ] regulation procedure
- [ ] environment sensing
- [ ] observability

### DIFF
--- a/connmgr/README.md
+++ b/connmgr/README.md
@@ -1,0 +1,148 @@
+# proposal: connection manager v2
+
+Revision: r0; 2019-04-01 author: @raulk
+
+_NOTE: code listings are Go-style pseudo-code_
+
+**Contents**
+
+{{TOC}}
+
+## motivation
+
+Resource management is an essential, cross-cutting concern for all libp2p
+components. Physical resources subject to management include file descriptor
+usage, bandwidth, memory, cpu, etc. Management can take place in a reactive or
+proactive fashion. While hard limit enforcement can be delegated to the
+environment, judicious resource usage/pressure remains a responsiblity of
+libp2p itself.
+
+Generally speaking, we consider the responsibilities of resource management to
+include at least:
+
+* global resource pool modelling.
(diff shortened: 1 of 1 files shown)


vasco-santos:
Awesome!

I and @jacobheun will need to work in improving the connection manager in `js-libp2p` for Q2, as a consequence of enabling the DHT and dealing with way more peers now. This way, we should be able to get involved in this discussion.

JustMaier:
It seems like there could be some interesting overlap between the connection manager and the [proposed v2 peerstore](https://github.com/libp2p/specs/issues/164). Ideally, peer metrics and connection manager determined quotas could be retained in the peerstore, and if the peerstore becomes a source of address quality/confidence, the information would be very valuable to the connection manager as it allocates resources.

Either way, I think that evolving peerstore around the same time as connection manager could make a lot of sense.

raulk:
@JustMaier – thanks for your interest in this spec. I think you're dead on with your intuitions about the alignment between peerstore v2 and connection manager v2. Have a look at the [RFC for the go-libp2p eventbus](https://github.com/libp2p/go-libp2p/issues/653) too – I think you'll find it interesting. All these proposals share a common theme of adopting a reactive, event-driven system comprised of autonomous, discrete services with loose coupling amongst each other.

jacobheun:
I've started to put down some of my notes on Connection Management over at https://github.com/libp2p/notes/issues/13#issue-463852516 to avoid polluting this PR.

meyer9:
Would the protocol enrollment section explain how to accept peers based on the protocols they support? For example, say I want to have 10 peers supporting the IPFS protocol and 10 peers supporting the FooBar protocol.

PR #162, opened 2019-04-08
Update README.md

decanus:


mxinden:
@decanus could you add more context to your patch proposal? I am having difficulties understand the motivation of the change.

mxinden:
Closing here. @decanus feel free to comment in case you would like to pick this back up.

PR #168, opened 2019-05-20
Connections & upgrading

yusefnapora:
There's still a ways to go on this, but so far it covers the basic protocol negotiation & connection upgrade process. 

One thing I'm not sure is correct is the roles of dialer and listener in the negotiation process. I've written it out with both sides sending the initial multistream header simultaneously, followed by the dialing peer proposing and the listener either accepting or rejecting. Does it work the other way as well, with the listening peer proposing protocols?


JustMaier:
Hey @yusefnapora, I just wanted to say thanks for your efforts on this. As someone trying to grapple what exactly is happening in the connection lifecycle, this is super helpful and this documentation is much needed.

yusefnapora:
Thanks for the kind words @JustMaier! I pushed a few more paragraphs just now, so hopefully that's helpful as well. Let me know if anything is hard to parse or understand :)

connections/README.md
```diff
@@ -173,20 +193,128 @@ upgrade process is complete, and both peers are able to use the resulting libp2p
...
+Maintaining a large number of persistent connections can cause issues with some
+network environments and can lead to resource exhaustion and erratic behavior.
+
+It's highly recommended that libp2p implementations maintain an upper bound on
+the number of open connections. Doing so while still maintaining robust
+performance and connectivity will likely require implementing some kind of
+priority mechanism for selecting which connections are the most "expendable"
+when you're near the limit.
+
+Resource allocation, measurement and enforcement policies are all an active area
```
JustMaier: I'm interested in this discussion, where can I find this?

yusefnapora: There's a proposal for go-libp2p here: https://github.com/libp2p/go-libp2p/issues/635 - it's not clear yet whether a cross-platform spec is needed, although we may want one if we start exposing metrics about resource consumption, so everything uses the same units.

yusefnapora:
I think this is ready to review - I ought to have flipped the draft switch earlier, but got a bit blindsided by moving to NC.

I requested reviews from a few people, but anyone interested is welcome to review, of course. Let me know if you want to be in the Interest Group - otherwise I'll start volunteering some people this week 😄

connections/README.md
```diff
@@ -0,0 +1,320 @@
...
+It's recommended that libp2p implementations provide a persistent metadata
+storage interface that contains at minimum the peer id and last known valid
+addresses for each peer. This allows you to more easily "catch
+back up" and rejoin a dense network between invocations of your libp2p
+application without having to rely on a few bootstrap nodes and random DHT walks
+to build up a routing table.
+
+Even during a single invocation of an application, you're likely to benefit from
+an in-memory metadata storage facility, which will allow you to cache addresses
+for connection resumption. Desining a storage interface which can be backed by
```
JustMaier: Small typo here `Desining`

connections/README.md
```diff
@@ -0,0 +1,325 @@
...
+Even during a single invocation of an application, you're likely to benefit from
+an in-memory metadata storage facility, which will allow you to cache addresses
+for connection resumption. Designing a storage interface which can be backed by
+memory or persistent storage will let you swap in whichever is appropriate for
+your use case and stage of development.
+
+For examples, see [go-libp2p-peerstore][go-libp2p-peerstore] and
+[js-peer-book][js-peer-book].
+
+#### Connection Limits
```
vasco-santos: Should we link this to the connection manager spec? I know it is WIP, but most of this information will also be there

connections/README.md
```diff
@@ -0,0 +1,341 @@
...
+
+Let's walk through the diagram above. The peer initiating the connection is
+called the **Initiator**, and the peer accepting the connection is the
+**Responder**.
+
+The Initiator first opens a channel to the Responder. This channel could either be a
+new connection or a new stream multiplexed over an existing connection.
+
+Next, both peers will send the multistream protocol id to establish that they
+want to use multistream-select. Note that both sides may send the initial
```
bigs: -"Note that b"
+B

connections/README.md
```diff
@@ -0,0 +1,341 @@
...
+
+Streams are routed to application-defined handler functions based on their
+protocol id string. Incoming stream requests will propose a protocol id to use
+for the stream using `multistream-select`, and the peer accepting the stream
+request will determine if there are any registered handlers capable of handling
+the protocol. If no handlers are found, the peer will respond to the proposal
+with `"na"`.
+
+When registering protocol handlers, it's possible to use a custom predicate or
+"match function", which will receive incoming protocol ids and return `true` if
```
bigs: return a boolean communicating whether or not the protocol is supporte

connections/README.md
```diff
@@ -0,0 +1,341 @@
...
+#### Connection Limits
+
+Maintaining a large number of persistent connections can cause issues with some
+network environments and can lead to resource exhaustion and erratic behavior.
+
+It's highly recommended that libp2p implementations maintain an upper bound on
+the number of open connections. Doing so while still maintaining robust
+performance and connectivity will likely require implementing some kind of
+priority mechanism for selecting which connections are the most "expendable"
+when you're near the limit.
```
bigs: maybe note that protocols based on connectionless protocols (aka QUIC via UDP) can be suspended and resumed with much lower overhead

connections/README.md
```diff
@@ -0,0 +1,341 @@
...
+
+| Event        | Description                               |
+|--------------|-------------------------------------------|
+| Connected    | A new connection has been opened          |
+| Disconnected | A connection has closed                   |
+| OpenedStream | A new stream has opened over a connection |
+| ClosedStream | A stream has closed                       |
+| Listen       | We've started listening on a new address  |
+| ListenClose  | We've stopped listening on an address     |
+
```
bigs: in the future, we're eager to explore a services style model to allow subsystems to communicate. maybe doesn't need to be in this draft directly, but something to think about!

connections/README.md
```diff
@@ -328,6 +350,45 @@ baseline would be:
...
+perhaps eventually replace) the current [notification system][go-net-notifee].
+
+One of the near-term goals of the event bus refactor is to more easily respond
+to changes in the protocols supported by a remote peer. Those changes are
+communicated over the wire using the [identify/push protocol][identify-push].
+Using an event bus allows other, unrelated components of libp2p (for
+example, a DHT module) to respond to changes without tightly coupling components
+together with direct dependencies.
+
+While the event bus refactoring is specific to go-libp2p, a future spec may
```
tomaka: Fwiw, that's something that rust-libp2p would definitely never implement.

connections/README.md
```diff
@@ -328,6 +350,45 @@ baseline would be:
...
+standardize event types used to communicate information across key libp2p
+subsystems, and may possibly require libp2p implementations to provide an
```
tomaka: Fwiw, that's something that rust-libp2p would definitely never implement.

yusefnapora: Good to know 😄

I honestly don't know if there's value in specifying the event system, since it feels like an implementation detail. The reason I added this note is that there's some ongoing work now (in go-libp2p) to dynamically change the DHT behavior in response to events from another protocol (identify). It seems likely that other kinds of "loosely coupled" interactions are likely to emerge in go-libp2p once the event bus is in place. So really I just want something to refer to when writing about those kinds of interactions, e.g. "the DHT may do X in response to Y events from component Z". 

I think actually spec'ing the mechanism for event delivery is probably too much, and maybe even referring to "events" at all is too implementation-specific.

yusefnapora:
Thanks for the feedback and kind words everyone!

PR #169, opened 2019-05-21
libp2p specification framework – lifecycle: maturity level and status

raulk:
A minimal viable meta-specification of the maturity process for libp2p specifications.

Next step is to specify the header for libp2p specification documents. It will carry at least: stage, author, current revision.

---

Context: some implementors are understandably frustrated because our spec process is too heavyweight. I suspect we are suffering from the psychological effect of not modelling spec maturity levels, so everything is either a binary “perfect or nothing”. Introducing spec maturity levels will dislodge stuff by allowing us merge docs that aren’t perfect, but are on their path to becoming so, and that’s fine.


raulk:
CC @warchant @kamilsa @creationix.

00-framework-01-maturity.md
```diff
@@ -0,0 +1,118 @@
...
+  * motivation: why this specification is relevant, and how it advances the
+    status quo.
+  * scope and rationale: what areas of the technical system the specification
+    impacts.
+  * goals: what we expect to achieve (positively and negatively) as a result
+    of implementing the specification.
+  * expected feature set: a summary/enumeration of features the spec provides.
+  * tentative technical directions: how are we planning to materialise the
+    specification in terms of system design.
+
```
vasco-santos: Should we cover a checklist of expected points needed in order to get in stage 2? 

Examples: finish proof of work, implementation, ...

I think it would be valuable to have a clear notion of progress

00-framework-01-maturity.md
```diff
@@ -0,0 +1,118 @@
...
+The specification document itself is technically complete. It defines wire
+level formats for interoperability, error codes, algorithms, data structures,
+heuristics, behaviours, etc., in a way that it is sufficient to enable
+contributors to develop an interoperable implementation.
+
+There is at least ONE implementation conforming to the specification. That
+implementation serves as the Reference Implementation.
+
+The promotion from a `Working Draft` to a `Candidate Recommendation` is done
```
vasco-santos: What is the criterium for achieving this? Does everyone need to agree?

00-framework-01-maturity.md
raulk: The way I see it, the author will need to enhance the spec so that it covers the requirements of a Candidate Recommendation:

```
The specification document itself is technically complete. It defines wire
level formats for interoperability, error codes, algorithms, data structures,
heuristics, behaviours, etc., in a way that it is sufficient to enable
contributors to develop an interoperable implementation.
```

And also there would need to be ONE conforming implementation.

At that stage, the author can submit a PR to promote the spec from Working Draft to Candidate Recommendation.

00-framework-01-maturity.md
raulk: Gonna answer more generically in the pull requests, as other questions too relate to consensus and the review process.

yusefnapora:
Clarity on how to progress between levels is needed, that's very true. Especially if we're committing to X days for review and so on.

raulk:
@jacobheun @vasco-santos How do you feel about the following:

> Upon submission of an `Initial Working Draft`, a minimum of five (5) libp2p contributors are required to express interest and commitment to steward and advise the author(s) throughout the specification process. The resulting group will constitute the _Interest Group_, formed by consensus, barring blocking, binding community feedback. We encourage the _Interest Group_ to be heterogeneous yet relevant, and hold representation for libp2p implementation teams across various languages. Changes in the membership of an _Interest Group_ are possible at any time. While we don't enumerate the reasons comprehensively, they include waning dedication/commitment of members, changes in technical relevance, or violations of the [community code of conduct](https://github.com/ipfs/community/blob/master/code-of-conduct.md).

> The _Interest Group_ will be responsible for awarding the review approvals necessary to transition the specification across stages.

00-framework-01-maturity.md
vasco-santos: My idea would be to allow the author to do an iterative work with several PRs, and the complete checklist would increase over time. 

With this, people checking the specs docs will have a better understanding of the current state of the spec without digesting all the content.

vasco-santos:
@raulk that approach seems good to allow a smooth iteration between levels 👍

00-framework-01-maturity.md
raulk: I like the ability to track progress in that manner. I don't want to define a global checklist, though, because we'll very quickly find an exceptional case where the checklist is unfit. Perhaps a tradeoff is to have the author introduce a spec-relevant checklist during the working draft phase?

00-framework-01-maturity.md
raulk: https://github.com/libp2p/specs/pull/169#issuecomment-494459635

whyrusleeping:
Calling them levels makes me think you always want to move on to the next level, but that's not the case here. Maybe we should use a word that doesn't imply succession like that

raulk:
@whyrusleeping the spec refers to the progression as _stages_. We used the term _levels_ in discussion, but that's inaccurate.

00-framework-01-maturity.md
```diff
@@ -0,0 +1,118 @@
...
+    status quo.
+  * scope and rationale: what areas of the technical system the specification
+    impacts.
+  * goals: what we expect to achieve (positively and negatively) as a result
+    of implementing the specification.
+  * expected feature set: a summary/enumeration of features the spec provides.
+  * tentative technical directions: how are we planning to materialise the
+    specification in terms of system design.
+
+The `Initial Working Draft` shall be reviewed by the community in no more than 3
```
vyzo: that's too strict, let's say a week or two.

00-framework-01-maturity.md
```diff
@@ -0,0 +1,118 @@
...
+level formats for interoperability, error codes, algorithms, data structures,
+heuristics, behaviours, etc., in a way that it is sufficient to enable
+contributors to develop an interoperable implementation.
+
+There is at least ONE implementation conforming to the specification. That
+implementation serves as the Reference Implementation.
+
+The promotion from a `Working Draft` to a `Candidate Recommendation` is done
+via a *Pull Request* that is reviewed by the libp2p community in no more than
+8 working days.
```
vyzo: again, deadlines are too strict. Make that two weeks at least.

jacobheun:
> Upon submission of an Initial Working Draft, a minimum of five (5) libp2p contributors are required to express interest and commitment to shepherd and advise the author(s) throughout the specification process.

Would the Interest Group members be specified in the header of the spec? Also, 5 seems excessive, especially for the Initial Working Draft. I could see this being more reasonable for the two recommendation stages. We have historically dropped the ball on moving things forward. I think having designated individuals will help, but I also think the barrier to entry should be lower for Working Drafts. Maybe start with 3 and recommendations require 5? 

Do all Interest Group members need to approve the PR before merge? I think majority rules might be good here. PRs that have open request for changes, might be a reasonable blocker, as long as they don't get stale (at which point they should be dismissed imo).

00-framework-01-maturity.md
```diff
@@ -0,0 +1,118 @@
...
+specification reaching the `Candidate Recommendation` stage.
+
+The transition from a `Candidate Recommendation` or a `Recommendation` stage
+to the `Deprecated` stage is performed via a *Pull Request* that is reviewed
+by the libp2p community in no more than 5 working days.
+
+## Stage 5: Abandoned Draft
+
+In order to motivate accountability, efficiency and order, a specification
+that stays in the `Working Draft` stage for over 4 months of its initial
```
vyzo: 4 months? That's just arbitrary and too soon for the velocity we have with those things.
Make it 6 months minimum, and not an automatic transition.

00-framework-01-maturity.md
raulk: Covered here: f65572f77dbee36638b1bfd63906b24cb8d61227. Let me know what you think.

00-framework-01-maturity.md
raulk: That's where the extensions come in. I really don't want specs to linger for 6 months with zero activity. I'd rather notice this earlier and establish checkpoints for extension. A spec is given a cumulative runway of 12 months; I think it's reasonable.

00-framework-01-maturity.md
raulk: Ok, I'll bump this up to 10 working days (2 weeks).

00-framework-01-maturity.md
raulk: Ok, I'll bump this up to 5 days (1 working week). Two weeks is too much to ingest an idea for incubation.

vyzo:
I would like to second @jacobheun in the 5 persons involved in the spec, that's too much.

00-framework-01-maturity.md
vasco-santos: Yes, I agree with your proposal!

00-framework-01-maturity.md
vyzo: I am concerned with us having the bandwidth to review the spec, and a proper review might take time.

raulk:
@vyzo @jacobheun I made IG headcount incremental as the working draft matures: https://github.com/libp2p/specs/pull/169/commits/ad14bff31980ea35c9709e42ae8b57099c9739cd

00-framework-01-maturity.md
```diff
@@ -0,0 +1,155 @@
...
+acceptance, and level of technical detail. It is somewhat inspired in the W3C
+Process [0].
+
+  1. `Working Draft`
+  2. `Candidate Recommendation`
+  3. `Recommendation`
+  4. `Deprecated`
+  5. `Abandoned Draft`
+
+## Stage 1: Working draft
```
vyzo: I agree with @whyrusleeping, let's not call it stages. Otherwise this would imply progression from 1->5.

00-framework-01-maturity.md
```diff
@@ -0,0 +1,155 @@
...
+and evolving their specification by self-approving their own *Pull Requests*.
+
+To facilitate open progress tracking and observability, as the `Working Draft`
+evolves, the author(s) SHOULD assemble a checklist of items that are pending
+specification, explicitly stating which items are compulsory for promoting the
+spec to a `Candidate Recommendation`.
+
+As a `Working Draft` evolves and shows promise to exit this stage towards a
+`Candidate Recommendation`, the _Interest Group_ shall be expanded by two (2)
+additional members, comprising a total of five (5).
```
vyzo: we barely have 5 people in all of libp2p team...

raulk:
Ok, I’ll call it “maturity status” then, or simply remove the numbering. Or I’ll segregate this into two fields: maturity (draft, candidate, recommendation) and status (active, deprecated, abandoned/terminated).

yusefnapora:
I like the separate maturity and status fields - that seems a bit clearer.

00-framework-01-maturity.md
```diff
@@ -0,0 +1,155 @@
...
+
+There are at least TWO implementations conforming to the specification, with
+demonstrated cross-interoperability. This is the supreme stage in the
+lifecycle of a specification.
+
+The promotion from a `Candidate Recommendation` to a `Recommendation` is done
+via a *Pull Request* that is reviewed by the _Interest Group_, allowing a
+minimum of 10 working days to elapse to collect feedback from the libp2p
+community at large.
+
```
bigs: may be worth having a note that some things, like multistream, will be a requirement. so perhaps the language of this stage could reflect it being either recommended or required.

00-framework-01-maturity.md
raulk: The libp2p community of implementers and stakeholders is way beyond the group we employ at Protocol Labs. IGs should be diverse so as to integrate different viewpoints from the community. When posting a draft, it’s fair to nudge the people that might want to participate.

00-framework-01-maturity.md
raulk: “Recommendation” is used in the same sense as W3C. It doesn’t mean it’s optional; it means it is endorsed. To model your specific concern, I think it’s appropriate to write a spec that describes the minimal baseline interoperable libp2p stack, which would refer to MS as a MUST.

00-framework-01-maturity.md
lanzafame: I would go with what the w3c process uses, `Levels`.

momack2:
Seems cool! My main question is where these PRs are going to go in your current folder structure (are they organized by stage or content?)

raulk:
@momack2 I'm trying to modularise the spec framework. Specs repository structure and document naming are elements worth speccing separately. So far we have converged to placing top-level specs in the root and subsystem specs in directories. People seem to be comfortable with this organisation, and we might want to formalise it in a spec.

raulk:
I see positive feedback for splintering "maturity stage" into separate "maturity level" and "status" fields. Here's the approach I'm taking:

* Maturity level: Working Draft, Candidate Recommendation, Recommendation.
* Status: Active, Deprecated, Terminated.

00-framework-01-maturity.md
raulk: See https://github.com/libp2p/specs/pull/169#issuecomment-494741011.

gpestana:
This is very nice stuff! Are abandoned candidate recommendations moved to `Deprecated` stage?

raulk:
I believe I've now addressed your comments about the impreciseness of the term "stages", and the sequential numbering.

We now have a "maturity level" (progressive) and a "status" field. See this commit: https://github.com/libp2p/specs/pull/169/commits/9b630b90a44b294354dd6a5ee1aa7df79454764f.

raulk:
@gpestana 

> This is very nice stuff! Are abandoned candidate recommendations moved to Deprecated stage?

I've now split this up into two fields: maturity level and status. Could you check if the new revision is clearer?

yusefnapora:
This is much improved since my initial hasty approval 😄

@raulk how would you like to apply this framework to the existing PRs? 

I'd like to get #106 and #100 merged soon. Perhaps it's best to pull them both in as Candidate Recommendations. Then we can have a sort of "practice round" of promoting them to Recommendations so we can get a feel for that process, and hopefully improve them some more while doing so.

raulk:
@yusefnapora 

> This is much improved since my initial hasty approval 😄

We have been talking about this for a long time, so I understand your excitement in merging it in ;-)

> @raulk how would you like to apply this framework to the existing PRs?

Off the top of my head, go over all outstanding PRs and:

1. Nominate 3 members for the initial Interest Group, based on the most active participants so far (you should be a part of all of them). Get them to ACK. Ping them out of band (IRC, Slack) to expedite.
2. Categorise as 1A (Working Draft, Active).
3. Add a checklist for the items that were controversial, or are still being discussed.
4. Merge.

> I'd like to get #106 and #100 merged soon.

I think #106 (SecIO) can be a 3A (Recommendation, Active). For #100, we can merge a spec that reflects the status quo as a 3A, with a notice that a refinement is on the way as a 1A (open a new issue to address the serialisation discussion?). Or we can merge it as a Working Draft to begin with.

gpestana:
> > This is very nice stuff! Are abandoned candidate recommendations moved to Deprecated stage?
> 
> I've now split this up into two fields: maturity level and status. Could you check if the new revision is clearer?

Yes, the status and the table showing its possible transitions are helpful. looks good to me! 👍

gpestana:
> > This is very nice stuff! Are abandoned candidate recommendations moved to Deprecated stage?
> 
> I've now split this up into two fields: maturity level and status. Could you check if the new revision is clearer?

Yes, the status and the table showing its possible transitions are helpful. looks good to me! 👍

raulk:
Thank you all for your feedback! @yusefnapora – would you like to take the lead on a small `00-framework` meta-specification for the document header? IMO, it should contain at least:

* title
* author(s)
* revision
* lifecycle stage (in the header form introduced in https://github.com/libp2p/specs/pull/169/commits/b229eb10c7c7a926f81bf5cea2f7394a3295a30d)
* interest group members

marten-seemann:
Why is an important spec like this merged after just two days of discussion?
I'm currently on a conference and haven't had the time to review the document yet. In a company where people are frequently traveling for work, 48 hours seems like a way too short time for deciding on the process how we evolve specs.

raulk:
@marten-seemann the consensus has been overwhelmingly positive so far. We really need to be nimbler; a lot of people are frustrated with our general heavyweightedness (me included). Also, we've had prior discussions in this direction, so nothing here should be controversial. Feel free to review the PR after the fact, or open a new PR to modify this work.

---

If anybody else feels they weren't able to get their input in in time, please say so. We can revert this PR and subject it to another round of reviews if necessary!

yusefnapora:
I added a commit to #100 with a first-pass at a document header. If anyone has opinions about that sort of thing, please check it out and let me know. Once we have something we like we can add a little spec for the header format.

raulk:
@yusefnapora – that LGTM!

lanzafame:
@raulk I would like to reiterate @marten-seemann point but also propose the solution of explicitly stating the timeframe in which a contributor has to make a comment/suggestion. Even if consensus is reached before the end of the timeframe, the PR shouldn't be merged until the timeframe is complete, in case someone hasn't had the chance to comment, i.e. like @marten-seemann on this PR. The timebox will cut the long tail off getting things through, whilst also allowing contributors to feel that they had the opportunity to contribute. Personally, I would have like to see this PR open for at least 4-5 days so that there is ample time for it to makes its way around the globe. This PR only had one and half round-trips between EU and Asia, and due to other priorities, I wasn't able to even see if my comments from the first day were responded before the PR was merged. (sidenote: I have had look now and its all good) 

I agree that things speeding up is a good thing but there needs to be a predictable timing to how things go so that people can schedule the time in their weeks to review and do a good job of things instead of rushing because they don't know when something will be merged, because consensus was formed on the other side of the planet, whilst they were sleeping.

Issue #139, opened 2019-01-31
Proposal to migrate PeerIDs to match Tor v3 addresses


Zolmeister:
Moved to: https://github.com/libp2p/specs/pull/172

I propose adapting Tor's [.onion v3 spec](https://gitweb.torproject.org/torspec.git/tree/rend-spec-v3.txt#n2029) for all PeerIDs moving forward:

```
peer_id = base32(PUBKEY | CHECKSUM | VERSION)
CHECKSUM = SHA3("peerid checksum" | PUBKEY | VERSION)[:2]

where:
  - PUBKEY is the 32 bytes ed25519 public key
  - VERSION is a 1-byte version field (default value '\x03')
  - "peerid checksum" is a constant string
  - CHECKSUM is truncated to two bytes

Examples (invalid checksum):
  pg6mmjiyjmcrsslvykfwnntlaru7p5svn6y2ymmju6nubxndf4pscryd
  sp3k262uwy4r2k3ycr5awluarykdpag6a7y33jxop4cs2lu5uz5sseqd
  xa4r2iadxm55fbnqgwwi5mymqdcofiu3w6rpbtqn7b2dyn7mgwj64jyd
```

Main points:
  - base32 for DNS
    - always, ux consistency. no multibase
  - version byte at end
    - using a common prefix (`Qm`) makes it much harder for people to distinguish addresses
    - better vanity addresses
  - has checksum
  - fixed encryption key type
    - public key is always inline
    - changing encryption algorithm requires version byte increment, forces consistency within PeerID versions
  - allows development of 'hidden services' (encrypted DHT records) a la Tor hidden services

Prior discussion:
[Tor mailing list thread](https://lists.torproject.org/pipermail/tor-dev/2017-January/011816.html)

Solves design constrains in https://github.com/libp2p/specs/issues/138
This solves https://github.com/libp2p/specs/issues/111 without adding the hash function to the protobuf

Zolmeister:
Migration path:
  1. Add `Ed25519_inline` key type to protobuf
      - All new keys will have this type
```
enum KeyType {
  RSA = 0;
  Ed25519 = 1;
  Secp256k1 = 2;
  Ed25519_inline = 3;
}
```
  
  2. `Ed25519_inline` will always use the `peer_id` calculation above
    - 'implicit' hash function - solves https://github.com/libp2p/specs/issues/111
    - no conflict with previous implementation https://github.com/libp2p/specs/issues/138

Stebalien:
[...]

Issue #138, opened 2019-01-29
Peer ID Calculation History And Resolution


Stebalien:
Forks off #111 to focus on the backwards compatibility issue instead of the CID/PID design space.

## Summary

Peer ID calculation has changed a couple of times over the past year.

1. Initially, peer IDs were always sha2-256 multihashes of the public key.
2. From go-ipfs 0.4.16-0.4.18, we used the identity multihash to "inline" small keys (specifically, ed25519 keys) into the peer ID under the assumption that nobody was using them.
3. We reverted (2) in 0.4.19 dev as OpenBazaar was using ed25519 keys before 0.4.16.
4. **We are now planning on re-introducing (2) as:
    a. Textile is also using ed25519 keys and is relying on 0.4.18 behavior.
    b. OpenBazaar is using a forked go-ipfs network. We are adding a package-level flag so they can restore pre-0.4.16 behavior in their forked build.**
5. Finally, we need to agree on a migration path such that all IPFS nodes use the same method to calculate peer IDs.

PR: https://github.com/libp2p/go-libp2p-peer/pull/42

## Goals

This all grew out of several requirements and wants:

**Requirements**

* We need to provide a way to for users to select a hash function other than sha256 when computing their peer ID.
* Given a public key, peer ID calculation must be deterministic.
* There must be a case-insensitive way to express peer IDs appearing in IPNS paths (for browsers). Currently, peer IDs are always base58 encoded which is not case-insensitive.

**Wants**

* We'd like a way to "inline" public keys into peer IDs (if the public key is small enough to fit comfortably). This will allow encrypting a message to a peer without needing to look up their key first.
* We'd like to be able to fetch keys with bitswap.

## Definitions

1. Inlining peer ID (working title...): A peer ID from which the associated
   public key can be extracted.
2. sha256 peer ID: A peer ID created using the sha256 hash function.

Note: "inlining" peer IDs look like `1...`, sha256 peer IDs look like `Qm...`.

## Events

Below is an exhaustive history of this issue:

 1. In the beginning, peer IDs were sha256 multihashes of peer IDs.
 2. At some point, we added support for ed25519 keys in go-libp2p. The plan was
    to embed them in peer IDs
    (https://github.com/libp2p/go-libp2p-crypto/pull/5). However, we punted on
    that. We never added support to go-ipfs
    (https://github.com/ipfs/go-ipfs/issues/3625).
 3. OpenBazaar started using ed25519 keys anyways.
[...]

PR #172, opened 2019-05-28
Create 1A-tor-address-encoding-and-encrypted-peer-info.md

Zolmeister:
1A – Working Draft / Active

Tor Address Encoding and Encrypted `PeerInfo`

### Context:

IPNS addresses are encoded as:

```
peer_id = base58(Multihash_SHA256(protobuf(RSA_PUBLIC_KEY)
# e.g. QmNUhKfcGJyQJnZu3AKn8NoiomDwDCRBicgqPt1YRqJBCz
```

And added to the DHT like so (~ I think):

```
PUT(peer_id, PeerInfo, signature)
GET(peer_id) -> PeerInfo, signature
```

The proposal is to encode IPNS addresses as [Tor rend-spec-v3 onion addresses](https://gitweb.torproject.org/torspec.git/tree/rend-spec-v3.txt#n2029):

```
peer_id = base32(PUBKEY | CHECKSUM | VERSION)
CHECKSUM = SHA3("peerid checksum" | PUBKEY | VERSION)[:2]

where:
  - PUBKEY is the 32 bytes ed25519 public key
  - VERSION is a 1-byte version field (default value '\x03')
  - "peerid checksum" is a constant string
  - CHECKSUM is truncated to two bytes

# e.g. pg6mmjiyjmcrsslvykfwnntlaru7p5svn6y2ymmju6nubxndf4pscryd
```

And added to the DHT like so:

```
PUT(BlindKey, Encrypt(PeerInfo, DeriveKey(peer_id)), signature)
GET(BlindKey) -> EncryptedPeerInfo, signature
```

Where `BlindKey` is derived from the ed25519 keypair using [Tors derivation scheme](https://gitweb.torproject.org/torspec.git/tree/rend-spec-v3.txt#n2161), except without key rotation

Note that `BlindKey` derivation _only_ works for ed25519 keypairs

`Encrypt` and `DeriveKey` follow Tors [hidden service descriptor encryption key derivation](https://gitweb.torproject.org/torspec.git/tree/rend-spec-v3.txt#n1424) (again without shared random)

### Motivation

Adversaries can:
  - discover all services by walking the DHT
  - monitor changes to `PeerInfo` of all services
  - monitor `PeerInfo` requests of other members in the network

### Scope and Rationale

IPNS address calculation, IPNS DHT read/write and validation

### Goals

Significantly improve the privacy of IPNS services

### Expected Feature Set: a summary/enumeration of features the spec provides.

Encrypted `PeerInfo` records and base32 IPNS addressing with inlined signing keys

### Tentative Technical Directions

Add a ney `KeyType` to the public key protobuf

### Address Tradeoffs

  - base32 for DNS
    - always, ux consistency. no multibase
  - version byte at end
    - using a common prefix (`Qm`) makes it much harder for people to distinguish addresses
    - better vanity addresses
  - has checksum
  - fixed signing key type
    - public key is always inline
    - changing signing algorithm requires version byte increment, forces consistency within PeerID versions

### Additional Notes

  - Original writeup: https://github.com/libp2p/specs/issues/139
  - All of this is heavily based on [Tor rend-spec-v3](https://gitweb.torproject.org/torspec.git/tree/rend-spec-v3.txt) which contains much of the implementation details
  - `PUT(Hash(PeerID), Encrypt(PeerInfo, PeerID))` (proposed elsewhere)
    - cannot be signed, because signature verification would require revealing `PeerID` to parties before inserting into the DHT
  - Unlike Tor, `BlindKeys` are not rotated daily using a shared random value
    - The consquences of this have not been explored thorougly
      - One point though is that the position within the DHT does not change over time (so certain network nodes could be targeted for DOS)
  - Tor additionally supports a second encryption layer for records for access control (a 'password' to access the service after querying the DHT)
  - Tor additionally pads their descriptors with `NUL` bytes to 10k, and 'stubs' extra fields to shield access controlled descriptors (which are longer)
  - I would have liked to have added the shared random `BlindKey` rotation, but could not find a suitable generation scheme
    - Tor uses a voting consensus system among their special `HSDir` servers
    - [Algorand](https://medium.com/algorand/algorand-releases-first-open-source-code-of-verifiable-random-function-93c2960abd61) does some neat tricks with VRF, but relies on crypto-assets

### DIFF
--- a/1A-tor-address-encoding-and-encrypted-peer-info.md
+++ b/1A-tor-address-encoding-and-encrypted-peer-info.md
@@ -1,0 +1,99 @@
+1A – Working Draft / Active
+
+Tor Address Encoding and Encrypted `PeerInfo`
+
+### Context:
+
+IPNS addresses are encoded as:
+
+```
+peer_id = base58(Multihash_SHA256(protobuf(RSA_PUBLIC_KEY)
+# e.g. QmNUhKfcGJyQJnZu3AKn8NoiomDwDCRBicgqPt1YRqJBCz
+```
+
+And added to the DHT like so (~ I think):
+
+```
+PUT(peer_id, PeerInfo, signature)
+GET(peer_id) -> PeerInfo, signature
+```
+
+The proposal is to encode IPNS addresses as [Tor rend-spec-v3 onion addresses](https://gitweb.torproject.org/torspec.git/tree/rend-spec-v3.txt#n2029):
+
+```
+peer_id = base32(PUBKEY | CHECKSUM | VERSION)
+CHECKSUM = SHA3("peerid checksum" | PUBKEY | VERSION)[:2]
+
+where:
+  - PUBKEY is the 32 bytes ed25519 public key
+  - VERSION is a 1-byte version field (default value '\x03')
+  - "peerid checksum" is a constant string
+  - CHECKSUM is truncated to two bytes
+
+# e.g. pg6mmjiyjmcrsslvykfwnntlaru7p5svn6y2ymmju6nubxndf4pscryd
+```
+
+And added to the DHT like so:
+
+```
+PUT(BlindKey, Encrypt(PeerInfo, DeriveKey(peer_id)), signature)
+GET(BlindKey) -> EncryptedPeerInfo, signature
+```
+
+Where `BlindKey` is derived from the ed25519 keypair using [Tors derivation scheme](https://gitweb.torproject.org/torspec.git/tree/rend-spec-v3.txt#n2161), except without key rotation
+
+Note that `BlindKey` derivation _only_ works for ed25519 keypairs
+
+`Encrypt` and `DeriveKey` follow Tors [hidden service descriptor encryption key derivation](https://gitweb.torproject.org/torspec.git/tree/rend-spec-v3.txt#n1424) (again without shared random)
+
+
+### Motivation
+
+Adversaries can:
+  - discover all services by walking the DHT
+  - monitor changes to `PeerInfo` of all services
+  - monitor `PeerInfo` requests of other members in the network
+
+### Scope and Rationale
+
+IPNS address calculation, IPNS DHT read/write and validation
+
+### Goals
+
+Significantly improve the privacy of IPNS services
+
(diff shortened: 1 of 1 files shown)


raulk:
@yusefnapora 	would you mind taking the lead in adjusting this doc to the new header format, and in identifying the appropriate members of the Interest Group?

yusefnapora:
Definitely. Thanks for making this @Zolmeister, we should definitely get an Interest Group formed up around this. @Stebalien @jhiesey @bigs, would any of you be interested?

To summarize a bit, it sounds like the end goal of this proposal is a more privacy-preserving IPNS, in which peers publish DHT records whose keys are a blind key (derived from their primary ed25519 identity) and whose values are encrypted peer info.

@Zolmeister I wasn't quite clear on why this requires a new `KeyType`. Are the derived keys a different type than ed25519?

Zolmeister:
@yusefnapora I'm not sure it does, but it seemed like a way to avoid a similar situation as https://github.com/libp2p/specs/issues/138, https://github.com/libp2p/specs/issues/111

yusefnapora:
I see, that makes sense. The inlining of keys is a bit of a problem at the moment.

I have a question for the group at large - are IPNS addresses required to also be peer ids? Or could this proposed encoding be considered a valid IPNS address, but not necessarily a peer id? That would let @Zolmeister define an encoding with as much metadata as necessary, and with the ed25519 key type restriction, which doesn't feel to me like it makes sense as part of the peer id definition.

Anyway, this sounds like an interesting direction to explore.

I'll be a bit delayed in my responses next week, as I'll be moving to a new house a few states away. But please `@` me if there's something I can help with.

Zolmeister:
Ping @vyzo @jamesray1 @daviddias @ivilata @richardschneider @lgierth @marten-seemann @dryajov 
I would appreciate some feedback here

vyzo:
How will this work in practice? It breaks compatibility with the existing system completely.

Zolmeister:
@vyzo Yes, these changes will not be backwards compatible with existing nodes.
Necessarily, because `PeerInfo` privacy is the feature being added.
Support for the `KeyType` will be added, and then at a later point can be made the default (once enough of the network has updated).

PR #171, opened 2019-05-27
libp2p specs framework: document header format

yusefnapora:
Following up with #169, this adds a spec for a small header to put at the top of our spec documents to convey the current lifecycle information, spec authors & interest group.

I've added these to the [peer id](https://github.com/libp2p/specs/pull/100) and [secio](https://github.com/libp2p/specs/pull/106) spec PRs, if anyone wants to see a real-world example.

In the spirit of the new lifecycle framework, I set the maturity level for this PR to `Working Draft`, so we can try sending this through the whole process.

Who wants to be in the Interest Group for this?

PR #173, opened 2019-05-29
relay/DCUtR: Add Direct Connection Upgrade through Relay protocol

vyzo:
~~still early draft, but it's an important subject that needs to gain some momentum.~~

> In this specification, we describe a synchronization protocol for direct
connectivity with hole punching that eschews signaling servers and utilizes
existing relay connections instead.

Status: Ready for review.


relay/DCUtR.md
```diff
@@ -0,0 +1,104 @@
...
+upgrade protocol as follows:
+1. `B` opens a stream to `A` using the `/libp2p/connect` protocol
+2. `B` sends to `A` a `Connect` message containing its observed (and possibly predicted)
+   addresses from identify and starts a timer to measure RTT of the relay connection.
+3. Upon receving the `Connect`, `A` responds back with a `Connect` message containing
+   its observed (and possibly predicted) addresses.
+4. Upon receiving the `Connect`, `B` sends a `Sync` message and starts a timer for
+   half the RTT measured from the time between sending the initial `Connect` and receiving
+   the response.
+5. Simultaneous Connect
```
raulk: Let's conduct reasonable research into how TCP hole punching works in the wild. I suspect we'll want to perform a few iterations of this process within a window of say, 10 seconds, before we declare a failure.

relay/DCUtR.md
```diff
@@ -0,0 +1,104 @@
...
+   - Upon receiving the `Sync`, `A` immediately starts a direct dial to B using the addresses
+     obtained from the `Connect` message.
+   - Upon expiry of the timer, `B` starts a direct dial to `A` using the addresses obtained
+     from the `Connect` message.
+6. If the connection is successful, then it is prioritized over the relay connection, which
```
raulk: We need to cover the stream migration procedure in this spec.

relay/DCUtR.md
```diff
@@ -0,0 +1,104 @@
+# Direct Connection Upgrade through Relay
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision    |
+|-----------------|---------------|--------|--------------------|
+| 1A              | Working Draft | Active | DRAFT, 2019-05-29  |
+
+Authors: vyzo
+
+Interest Group: raulk, stebalien, whyrusleeping
```
raulk: Let's follow the [suggested format](https://github.com/libp2p/specs/pull/171/files#diff-3f4c87e0067ecc0fb22bfffc359be3bdR9) here for this section. I know you had concerns about duplicating links, but not everybody reads this spec on github.com.

relay/DCUtR.md
```diff
@@ -0,0 +1,104 @@
...
+   - Upon receiving the `Sync`, `A` immediately starts a direct dial to B using the addresses
+     obtained from the `Connect` message.
+   - Upon expiry of the timer, `B` starts a direct dial to `A` using the addresses obtained
+     from the `Connect` message.
+6. If the connection is successful, then it is prioritized over the relay connection, which
+   can now be closed, possibly after a grace period.
+
+The purpose of the `Sync` message and `B`'s timer is to allow the two peers to synchronize
+so that they perform a simultaneous open that allows hole punching to succeed.
+
```
raulk: > 
```suggestion

### Inflight stream migration

TBD.

```

relay/DCUtR.md
```diff
@@ -0,0 +1,104 @@
...
+can keep using the relay connection as they were.
+
+## The Protocol
+
+Consider two peers, `A` and `B`. `A` wants to connect to `B`, which is
+behind a NAT and advertises relay addresses. `A` may itself be behind
+a NAT or be a public node.
+
+The protocol starts with the completion of a relay connection from `A`
+to `B`.  Upon observing the new connection, the inbound peer (here `B`)
```
raulk: There's a chance that `A` didn't know the current, full address set of `B`, e.g. `A` had stale addresses for `B` in its local address book that didn't include public addresses that `B` learnt of itself until later. So I think this protocol needs to be bi-directional for additional resiliency.

Also because `B` could be malicious and be actively pursuing something analogous to a "downgrade attack" at the transport layer, by picking a worse-performing transport deliberately, even if it knows of a better path to `A`.

relay/DCUtR.md
vyzo: This is still not finalized, and I'll be happy to let yusef be the steward of cosmetics :p

relay/DCUtR.md
vyzo: This should be covered by the upgrade protocol itself, that's the bidirectional part.

relay/DCUtR.md
vyzo: We don't have any yet...

relay/DCUtR.md
raulk: Why not do it yourself rather than pull someone else in?

relay/DCUtR.md
vyzo: I don't think that inflight stream migration is something we should cover in this protocol, it's a huge topic on its own.

relay/DCUtR.md
vyzo: I don't like boilerplate, especially in early drafts.
When we are ready to merge we can handle cosmetics.

relay/DCUtR.md
raulk: Yes, it needs to be specified, otherwise this whole thing is incomplete. We can incubate it in this spec and then spin it off.

vyzo: We don't necessarily _need_ any stream migration for the protocol to work.
We can simply open all new streams in the direct connection and simply garbage collect the relay connection when it no longer has any streams.
Or we can simply close it after a grace period and force new streams to be created.

raulk: > We can simply open all new streams in the direct connection 

This behaviour needs to be specified. Failing to specify the overall choreography makes this spec unactionable. "We now established a direct connection, now what?"

Maybe it's a naming thing. "Upgrade" implies the existing connection will evolve. If all we're intending to cover is the signalling and synchronisation, then this spec should be named accordingly.

relay/DCUtR.md
raulk: @vyzo: I don't understand why you adopt one part of the "boilerplate" and not another. This whole discussion is borderline childish.

The point of the specs process is to _merge early drafts_, so I was ready to approve this, but can't do so if the author fails to adhere to the spec format deliberately and without fair reason.

relay/DCUtR.md
raulk: I had misunderstood the spec on first read. This seems OK.

relay/DCUtR.md
raulk: We do need to specify what is the exit behaviour once the direct connection was established. Do we kill the old connection along with all streams? Do we emit the `Disconnected` and `ClosedStream` notifications? Do we emit the `Connected` notification and allow the protocols to initialise again?

vyzo: I am adding a paragraph to this regard.

relay/DCUtR.md
vyzo: The header spec hasn't even been finalized yet, there is nothing to adhere to yet.
Regardless I will add your boilerplate since you are so picky about it.

relay/DCUtR.md
```diff
@@ -0,0 +1,121 @@
...
+checks the addresses advertised by `A` via identify. If that set
+includes public addresses, then `A` _may_ be reachable by a direct
```
albrow: Isn't it possible that `A` may also be directly reachable at a private address if `A` and `B` are on the same local network?

vyzo: Yes it is possible, but that would have been dialed directly as the private addresses are still advertised with relay addresses.

raulk: I think @albrow has a point. @vyzo: while **that** _should_ be the case, if we want to be resilient and robust, this protocol should not make assumptions about how any other part of the system behaves. Usually those implicit assumptions make systems brittle.

Luckily our spec lifecycle process allows us to add this topic as an active discussion:

> To facilitate open progress tracking and observability, as the `Working Draft`
> evolves, the author(s) SHOULD assemble a checklist of items that are pending
> specification, explicitly stating which items are compulsory for promoting the
> spec to a `Candidate Recommendation`.

from: https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md

raulk:
@vyzo 	

@albrow and the 0x guys pointed us to the [Trickle ICE] spec, which seems like a relevant background read for us. I do expect our specs to be influenced by ICE – as a real-life, successful technology for coordinating hole punching between any two peers. 

We should aim to reference ICE WG material to back up our ideas and routines.

[Trickle ICE]: https://tools.ietf.org/html/draft-ietf-ice-trickle-21

relay/DCUtR.md
vyzo: Not making this assumption will make us dial private addresses in vain multiple times.
We already have a problem with that.

vyzo:
@raulk there is a reference to the ICE RFC already.

relay/DCUtR.md
vyzo: At best, we can consider dialing them in the bidirectional part of the protocol.

vyzo: Also, if A is public and B is private, we can't possibly be behind the same NAT.

vyzo: Furthermore, for the bidirectional part of the protocol we could check the public address of the other node. If that doesn't match our own, we can't possibly be behind the same NAT and dialing private addrs is pointless.

raulk:
@vyzo great, I missed that. What are the parallels between the algo we propose and the ICE procedure?

relay/DCUtR.md
yusefnapora: It would be nice to avoid dialing private addrs if we can avoid it though. Perhaps we could still exchange them, but in a separate field. Then they can be ignored unless your public address matches the other node and you infer that you're behind the same NAT. Or your implementation may be able to always ignore them, since they would have been dialed previously.

Anyway, I agree that we could punt on this for this round and discuss when we promote to candidate rec.

vyzo:
> What are the parallels between the algo we propose and the ICE procedure?

It's like ICE without a signalling server, and distributed STUN - we rely on public peers to tell us our observed addresses instead of using STUN servers.
Also note that ICE mainly caters to UDP, while we very much care for TCP.

yusefnapora:
I think the Trickle ICE spec @raulk linked to is an iteration on ICE that incrementally exchanges candidates instead of sending them all at once. Apparently this lets you start testing connectivity sooner

dryajov:
related previous discussion - https://github.com/libp2p/specs/pull/64

relay/DCUtR.md
```diff
@@ -0,0 +1,121 @@
...
+5. Simultaneous Connect
+   - Upon receiving the `Sync`, `A` immediately starts a direct dial to B using the addresses
+     obtained from the `Connect` message.
+   - Upon expiry of the timer, `B` starts a direct dial to `A` using the addresses obtained
+     from the `Connect` message.
+
+The purpose of the `Sync` message and `B`'s timer is to allow the two peers to synchronize
+so that they perform a simultaneous open that allows hole punching to succeed.
+
+If the direct connection is successful, then the peers should migrate
```
aarshkshah1992, line 85: @vyzo 

B can't close it' s connection with the Relay here, right ? What if there are other peers that would like to connect with B later to co-ordinate a hole punch ?

jacobheun: B should keep the connection to the relay for inbound dials. The migration mentioned here and its disconnect is for the virtual connection from B to A. Additionally, as the dialer, A should disconnect from the relay, unless it was also using it for inbound dials.

relay/DCUtR.md
```diff
@@ -0,0 +1,121 @@
...
+checks the addresses advertised by `A` via identify. If that set
+includes public addresses, then `A` _may_ be reachable by a direct
+connection, in which case `B` attempts a unilateral connection upgrade
+by initiating a direct connection to `A`.
+
+If the unilateral connection upgrade attempt fails or if `A` is itself a NATed peer that
+doesn't advertise public address, then `B` initiates the direct connection
+upgrade protocol as follows:
+1. `B` opens a stream to `A` using the `/libp2p/connect` protocol
+2. `B` sends to `A` a `Connect` message containing its observed (and possibly predicted)
```
aarshkshah1992, line 69: @vyzo @jacobheun  @Stebalien 

Does this mean that B sends it' s own observed addresses to A in the `Connect` message ? I think these will already be exchanged during the `Identify` handshake done upon creating the Relayed connection.

We can simply send empty `Ping` and `Pong` messages here to measure the RTT.

Stebalien: It's not necessary, but it may be nice to specify the preferred addresses anyways. It's also _really_ important that we end up picking the same transports.

relay/DCUtR.md
```diff
@@ -0,0 +1,121 @@
...
+
+The protocol starts with the completion of a relay connection from `A`
+to `B`.  Upon observing the new connection, the inbound peer (here `B`)
+checks the addresses advertised by `A` via identify. If that set
+includes public addresses, then `A` _may_ be reachable by a direct
+connection, in which case `B` attempts a unilateral connection upgrade
+by initiating a direct connection to `A`.
+
+If the unilateral connection upgrade attempt fails or if `A` is itself a NATed peer that
+doesn't advertise public address, then `B` initiates the direct connection
```
willscott, line 66: If `A` believes it is behind a NAT and will not be able to receive a direct connection from `B`, can `A` short circuit this process and directly initiate the upgrade protocol and/or send it's observed/predicted addresses such that the path to the `Sync` does not block until after the probes by `B` complete?

aarshkshah1992: @willscott 

In the code, we'll attempt a direct dial ONLY if we have a public non-Relay address for the peer. I think that's a good optimisation for now.

relay/DCUtR.md
marten-seemann: We don’t need to pick them, as long as we dial them in parallel, we can use our existing logic of racing connection attempts and using the connection that’s set up first.

relay/DCUtR.md
marten-seemann: I believe the feasibility of this has been shown by the experiments that Project Flare has performed earlier this year. Conclusion was: 3 retry attempts should work fine.

relay/DCUtR.md
```diff
@@ -0,0 +1,121 @@
...
+
+TBD
+
+## Implementation Considerations
+
+There are some difficulties regarding implementing the protocol, at least in `go-libp2p`:
+- the swarm currently has no mechanism for direct dials in the presence of existing connections,
+  as required by the upgrade protocol.
+- the swarm has no logic for prioritizing direct connections over relay connections
+- the current multistream select protocol is an interactive protocol that requires a single
```
marten-seemann, line 106: We now have a simultaneous connect protocol that works with multistream 1.0, so I believe this point has become moot since this spec was written.

relay/DCUtR.md
mxinden: @marten-seemann can you extend on how the Golang DCUTR implementation does the retries? Does it exchange `Connect` and `Sync` messages for each iteration?

The [Rust implementation](https://github.com/libp2p/rust-libp2p/pull/2076) only does a single iteration as of today.

relay/DCUtR.md
mxinden: Referencing stream migration protocol here https://github.com/libp2p/specs/issues/328.

mxinden:
@vyzo I will update this pull request with the latest changes in the Golang and Rust implementation. I hope you don't mind.

relay/DCUtR.md
mxinden: Addressed in https://github.com/libp2p/specs/pull/173/commits/dfc988c55fb49288e556c614ec334da75a2a3ac4.

vyzo:
yeah go for it

Stebalien (approved):

relay/DCUtR.md
mxinden: Resolving here with @marten-seemann 's answer. Feel free to unresolve in case you want to continue the discussion.

relay/DCUtR.md
mxinden: Agreed with @aarshkshah1992. Resolving here. Feel free to unresolve in case you want to continue the discussion @willscott.

relay/DCUtR.md
mxinden: https://github.com/libp2p/specs/pull/173/commits/fe64a21ddd9cae8b64f9b6160eb5787da5947bec documents the retry logic as discussed in https://github.com/libp2p/go-libp2p/pull/1057#issuecomment-897522382.

I am resolving here. Feel free to un-resolve or start a new thread on this topic.

relay/DCUtR.md
```diff
@@ -0,0 +1,158 @@
...
+to `B`.  Upon observing the new connection, the inbound peer (here `B`)
+checks the addresses advertised by `A` via identify. If that set
+includes public addresses, then `A` _may_ be reachable by a direct
+connection, in which case `B` attempts a unilateral connection upgrade
+by initiating a direct connection to `A`.
+
+If the unilateral connection upgrade attempt fails or if `A` is itself a NATed
+peer that doesn't advertise public address, then `B` initiates the direct
+connection upgrade protocol as follows:
+1. `B` opens a stream to `A` using the `/libp2p/dcutr` protocol.
```
mxinden, line 78: Note the protocol name `/libp2p/dcutr`.

relay/DCUtR.md
```diff
@@ -0,0 +1,158 @@
...
+   for half the RTT measured from the time between sending the initial `Connect`
+   and receiving the response.
+5. Simultaneous Connect
+   - Upon receiving the `Sync`, `A` immediately starts a direct dial to B using
+     the addresses obtained from the `Connect` message.
+   - Upon expiry of the timer, `B` starts a direct dial to `A` using the
+     addresses obtained from the `Connect` message.
+6. On failure go back to step (2), reusing the same stream opened in (1).
+   Inbound peers (here `B`) SHOULD retry twice (thus a total of 3 attempts)
+   before considering the upgrade as failed.
```
mxinden, line 94: Note the added retry logic. Also see FAQ further below for additional reasoning.

mxinden (commented):
I am proposing to merge this specification in its current state as a _Working Draft_.

Reviews welcome!

Below I am highlighting the two most notable recent changes.

relay/DCUtR.md
```diff
@@ -0,0 +1,158 @@
+# Direct Connection Upgrade through Relay
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision    |
+|-----------------|---------------|--------|--------------------|
+| 1A              | Working Draft | Active | r0, 2021-08-17     |
+
+Authors: [@vyzo]
+
+Interest Group: [@raulk], [@stebalien], [@whyrusleeping], [@mxinden]
```
marten-seemann, line 9:
```suggestion
Interest Group: [@raulk], [@stebalien], [@whyrusleeping], [@mxinden], [@marten-seemann]
```

relay/DCUtR.md
```diff
@@ -0,0 +1,158 @@
...
+the relay connection is closed.
+
+
+### RPC messages
+
+All RPC messages sent over a stream are prefixed with the message length in
+bytes, encoded as an unsigned variable length integer as defined by the
+[multiformats unsigned-varint spec][uvarint-spec].
+
+Implemntations SHOULD refuse encoded RPC messages (length prefix excluded)
```
marten-seemann, line 114:
```suggestion
Implementations SHOULD refuse encoded RPC messages (length prefix excluded)
```

relay/DCUtR.md
```diff
@@ -0,0 +1,158 @@
...
+
+message HolePunch {
+  enum Type {
+    CONNECT = 100;
+    SYNC = 300;
+  }
+
+  optional Type type=1;
+
+  repeated bytes ObsAddrs = 2;
```
marten-seemann, line 132: Should we reference the multiaddr spec to make clear that the `bytes` here is a serialized multiaddr?

relay/DCUtR.md
```diff
@@ -0,0 +1,158 @@
...
+
+- *Why exchange `CONNECT` and `SYNC` messages once more on each retry?*
+
+  Doing an additional CONNECT and SYNC for each retry prevents a flawed RTT
+  measurement on the first attempt to distort all following retry attempts.
+
+- *Why reuse the same stream for retries?*
+
+  Stream opening and stream protocol negotiation might distort the measured
+  round-trip-time. Reusing the stream from the first attempt allows cutting out
```
marten-seemann, line 146: I'm not sure I understand how. The RTT is determined as the time difference between the sent and received CONNECT messages. You can only send this message after the stream protocol has been negotiated. In case you pipeline this message, you'll only get a CONNECT in response if the protocol was accepted, so there shouldn't be any delay (at least no delay O(RTT)).
Are we worried about (probably sub-ms) processing delays introduced by multistream? If we are, should we also be worried about the time it takes to assemble the addresses to put in the CONNECT message, as well as the serialization delay of the protobuf?

relay/DCUtR.md
```diff
@@ -0,0 +1,158 @@
...
+   predicted) addresses from identify and starts a timer to measure RTT of the
+   relay connection.
+3. Upon receving the `Connect`, `A` responds back with a `Connect` message
+   containing its observed (and possibly predicted) addresses.
+4. Upon receiving the `Connect`, `B` sends a `Sync` message and starts a timer
+   for half the RTT measured from the time between sending the initial `Connect`
+   and receiving the response.
+5. Simultaneous Connect
+   - Upon receiving the `Sync`, `A` immediately starts a direct dial to B using
+     the addresses obtained from the `Connect` message.
```
marten-seemann, line 89: **a** direct dial to the **addresses** doesn't make sense. I think we should say that `A` should dial all (or at least all that make sense) addresses.

relay/DCUtR.md
marten-seemann: That change makes a lot of sense to me.

relay/DCUtR.md
```diff
@@ -0,0 +1,158 @@
...
+   - Upon expiry of the timer, `B` starts a direct dial to `A` using the
+     addresses obtained from the `Connect` message.
```
marten-seemann, line 91: We need to specify the roles of this simultaneous connect:
* for TCP we can avoid the additional round trip caused by the multistream sim open protocol (and with Protocol Select, there won't even be such a thing)
* for QUIC we need to assign which side sends the garbage UDP packets and which side actually dials, see https://github.com/libp2p/go-libp2p-quic-transport/pull/210

relay/DCUtR.md
```diff
@@ -0,0 +1,158 @@
...
+
+Authors: [@vyzo]
+
+Interest Group: [@raulk], [@stebalien], [@whyrusleeping], [@mxinden]
+
+[@vyzo]: https://github.com/vyzo
+[@raulk]: https://github.com/raulk
+[@stebalien]: https://github.com/stebalien
+[@whyrusleeping]: https://github.com/whyrusleeping
+[@mxinden]: https://github.com/mxinden
```
marten-seemann, line 15:
```suggestion
[@mxinden]: https://github.com/mxinden
[@marten-seemann]: https://github.com/marten-seemann
```

relay/DCUtR.md
marten-seemann: Would be fixed by #361.

relay/DCUtR.md
```diff
@@ -0,0 +1,158 @@
...
+connect peers behind NAT albeit with a high-latency, low-bandwidth
+connection.  Unfortunately, they are expensive to scale and maintain
+if they have to carry all the NATed node traffic in the network.
+
+It is often possible for two peers behind NAT to communicate directly
+by utilizing a technique called _hole punching_[1]. The technique
+relies on the two peers synchronizing and simultaneously opening
+connections to each other to their predicted external address. It
+works well for UDP, with an estimated 80% success rate, and reasonably
+well for TCP, with an estimated 60% success rate.
```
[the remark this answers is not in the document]
mxinden: I think they have been in the same ballpark, but I might as well be mistaken. Unfortunately I am unable to access the data from _project flare phase 1_. Either the data or my access seems to be removed.

@vyzo do you know more here?

relay/DCUtR.md
mxinden: The above makes no assumptions about the multiplexer and stream negotiation protocol in use.

E.g. say that a stream multiplexer needs a whole round-trip to open a new stream. If the negotiation protocol uses pipelining, but does not wait for the stream to be opened before accepting data, the measurement would be distorted.

I acknowledge that this example seems contrived, that it might be fine making the assumption and thus that there is no need for reusing the same stream.

> Are we worried about (probably sub-ms) processing delays introduced by multistream?

I don't think we need to care about same-machine processing delays. At least not at this point in time without proper measurements.

I am fine either way. I like the current version, not making any assumptions about the multiplexer and protocol negotiation protocol. I like the simplicity of opening a new stream on each attempt. With the additional context above, what do you prefer @marten-seemann?

marten-seemann: No really strong opinion, but I think I'd prefer to do use separate streams. It would simplify timeout handling, and allow peers to wait for a certain time between retries without keeping the stream open.

relay/DCUtR.md
mxinden: Merged, thus resolving.

relay/DCUtR.md
mxinden: Done via https://github.com/libp2p/specs/pull/173/commits/af0b9bbc13629b8a28cc0843e336af91fe7cd795.

relay/DCUtR.md
mxinden: Done via https://github.com/libp2p/specs/pull/173/commits/17f627532d9b4ddd54537d41ce38943e25d19bef.

relay/DCUtR.md
mxinden: :+1: changed to _opening a new stream on each retry_ with https://github.com/libp2p/specs/pull/173/commits/5943d3b34a3df937d56c8fa8308a2b5e01922795.

relay/DCUtR.md
```diff
@@ -0,0 +1,179 @@
...
+        protocols run on top of this TCP connection, `A` is assumed to be the
+        client and `B` the server.
+   - For QUIC:
+      - Upon receiving the `Sync`, `A` immediately attempts to directly connect
+        to `B`. `A` does so by dialing the addresses obtained from the `Connect`
+        message in parallel.
+      - Upon expiry of the timer, `B` starts to send UDP packets filled with
+        random bytes to the addresses obtained from the `Connect` message.
+        Packets should be sent in random intervals between 10 and 200 ms to each
+        address.
```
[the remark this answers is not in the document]
marten-seemann: I agree with the intention, but this might be a bit confusing, since `Connect` could also contain TCP addresses.

Maybe we should prefix the entire section (5.) with a "For every address obtained from the `Connect` message, ...", and then clean up the subpoints?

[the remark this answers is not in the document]
mxinden: Done via https://github.com/libp2p/specs/pull/173/commits/6f558f13aa4cbfe8458a3207764bd52e3f61a573.

vyzo:
We had much better experimental success rate, 90% for UDP/QUIC.
Maybe NATs have gotten better behaved in the 10-15 years since the
literature, or at least our sample had better behaved ones.

mxinden:
Thanks @vyzo.

https://github.com/libp2p/specs/pull/173/commits/cab60cced57e04a2b162b1b571187abc7d94bffa removes the concrete (outdated) success rate statements. My reasoning for removing them is the following: The results of _Project Flare Phase 1_ were convincing enough that it is worth finishing the _Project Flare_ project. We can only measure the real success rates once the protocols are widely deployed. Instead of outdated numbers in this spec from early on, I am in favor of removing them, maybe bringing them back once we have more data.

relay/DCUtR.md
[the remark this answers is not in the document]
mxinden: Discussion continued on https://github.com/libp2p/specs/pull/173#issuecomment-903693702.

relay/DCUtR.md
```diff
@@ -0,0 +1,163 @@
...
+        protocols run on top of this TCP connection, `A` is assumed to be the
+        client and `B` the server.
+   - For a QUIC address:
+      - Upon receiving the `Sync`, `A` immediately dials the address to `B`.
+      - Upon expiry of the timer, `B` starts to send UDP packets filled with
+        random bytes to `A`'s address. Packets should be sent repeatedly in
+        random intervals between 10 and 200 ms.
+      - This will result in a QUIC connection where `A` is the client and `B` is
+        the server.
+6. On successful establishment of a single connection does `A` cancel all
```
marten-seemann, line 104:
```suggestion
6. Once a single connection has been established, `A` SHOULD cancel all
```

relay/DCUtR.md
```diff
@@ -0,0 +1,163 @@
...
+   outstanding connection attempts. The peers should migrate to the established
+   connection by prioritizing over the existing relay connection. All new
+   streams should be opened in the direct connection, while the relay connection
+   should be closed after a grace period. Existing undefinite duration streams
```
marten-seemann, line 108: Not 100% sure, but I think this is the correct word:
```suggestion
   should be closed after a grace period. Existing indefinite duration streams
```

marten-seemann, line 108: I think this is more precise:
```suggestion
   should be closed after a grace period. Existing long-lived streams
```

marten-seemann (approved):

relay/DCUtR.md
mxinden: Resolved in https://github.com/libp2p/specs/pull/173/commits/8001cd9fe43719c559e192d4146f044ebcbb1aba.

mxinden:
Thanks to the many people involved here! :pray:

relay/DCUtR.md
```diff
@@ -0,0 +1,163 @@
...
+      - This will result in a TCP Simultaneous Connect. For the purpose of all
+        protocols run on top of this TCP connection, `A` is assumed to be the
+        client and `B` the server.
+   - For a QUIC address:
+      - Upon receiving the `Sync`, `A` immediately dials the address to `B`.
+      - Upon expiry of the timer, `B` starts to send UDP packets filled with
+        random bytes to `A`'s address. Packets should be sent repeatedly in
+        random intervals between 10 and 200 ms.
+      - This will result in a QUIC connection where `A` is the client and `B` is
+        the server.
```
wngr, line 103: From what I see, this whole mechanism would also fit nicely upgrading the relay connection to a direct WebRTC connection, if the peers would be allowed to exchange their SDP data here.
Would you be open in amending the spec?
(cc @mxinden)

mxinden: Yes, good point. We had this in mind, but as you said, it isn't mentioned anywhere. Given that the protocol uses protocol buffers, we could easily extend the messages to include additional data such as SDP payloads, or derive an SDP payload based on the information exchanged through the protocol.

Unfortunately there is no uniform way of speaking WebRTC across the many libp2p libraries (yet). In addition there is no specification yet (see https://github.com/libp2p/specs/issues/220 and https://github.com/libp2p/specs/pull/159). This is not to say that the project is not interested in adding WebRTC support in the future. Quite the opposite (see https://github.com/libp2p/specs/blob/master/connections/hole-punching.md and https://github.com/libp2p/specs/blob/master/ROADMAP.md#-unprecedented-global-connectivity).

With the above in mind, I am not sure whether it makes much sense to extend this paragraph with a section on WebRTC quite yet.

@wngr what do you think?

wngr: I think DCUtR would be a great way to add support for upgrading relayed connections to a direct WebRTC connection -- this just feels like the right abstraction, and the the alternative proposals so far appear inferior. Now I acknowledge that the big downside of this approach is that this requires a valid TLS certificate for the peer offering a WS endpoint, but I think that is a pill that can be swallowed.
In other words, _I think_ DCUtR is the right way to add support for upgrades to WebRTC (or allow exchanging arbitrary payloads here?), and I don't want to let the current opportunity window slide ;-).

(By the way, I hacked on an experimental webrtc transport for rust-libp2p which supports both browser apis (through wasm) and native; signalling is currently done via `p2p-webrtc-star`.)

mxinden: >  In other words, I think DCUtR is the right way to add support for upgrades to WebRTC (or allow exchanging arbitrary payloads here?), and I don't want to let the current opportunity window slide ;-).

:+1: 

> (By the way, I hacked on an experimental webrtc transport for rust-libp2p which supports both browser apis (through wasm) and native; signalling is currently done via p2p-webrtc-star.)

:rocket: that is great to hear. Mind opening a work-in-progress pull request on rust-libp2p @wngr?

wngr: My current WIP is at https://github.com/wngr/libp2p-webrtc; however I really want to replace the WS signalling server with a libp2p relay node; this is why I started adding my own custom (behaviour, transport) tuple on top of rust-libp2p, which very much is similar to dcutr on a higher level.
What's the state of your dcutr branch? Maybe it makes more sense to prototype it ontop of that?

mxinden: >  What's the state of your dcutr branch? Maybe it makes more sense to prototype it ontop of that?

You could leverage both https://github.com/libp2p/rust-libp2p/pull/2059 and https://github.com/libp2p/rust-libp2p/pull/2076. In case my understanding of WebRTC and SDP is correct, it solely needs to exchange a payload. If so (at least for now) you could just extend the Protobuf definition of the DCUTR protocol by a single field for that payload.

Happy to talk through this in person if that is preferred. Feel free to reach out via mail @wngr.

PR #175, opened 2019-06-11
Replace "main spec doc" with minimal index

### LINKED ISSUE #156: Improve outdated / misleading specs
This issue tracks our efforts to:

- identify outdated, misleading or incomplete areas of the existing specs
- visibly mark those areas as deprecated in the text
- replace and rewrite deprecated sections.

I'll be putting up a PR soon with some deprecation warnings and linking them back to this issue. The idea is for this issue to exist as long as deprecation warnings are still necessary, so we can have a place to discuss improvements and our action plan for rewriting / replacing the deprecated bits.

yusefnapora:
The main goal of this is to address #156 by moving the "main spec document" to an `archive` folder, replacing it with a minimal index into the specs that are currently living in subdirectories.

This is a fairly drastic move vs editing the main doc piecemeal to remove inaccuracies, but I think it's likely the cleanest way to provide visitors to the repo with a view into the currently active / supported specs. It also lets us keep the original document around as a reference, which I think is still useful, especially for getting your head around some of libp2p's design decisions and philosophy.

Over time we can add more contextual information to the README / index, such as splitting out sections for lifecycle stages and / or functionality (e.g. put the stream muxers together, etc).

closes #156


mgoelzer:
I'm a strong +1 on this.  Right now when people come to the specs repo, what they see is a set of numbered chapters and a table of contents that we all agree is not the content for people seeking specifications.  The content is not bad (it helps to motivate the problem), it's just not what we want to put front-and-center.  This PR fixes that problem.

yusefnapora:
I like the underscore, @jacobheun, that's a nice idea.

raulk:
This is a great improvement @yusefnapora. Thanks for championing this!

PR #176, opened 2019-06-20
add identify/delta to identify spec

yusefnapora:
This updates the spec to track the changes @raulk is adding in https://github.com/libp2p/go-libp2p/pull/659, which adds an `identify/delta` variant that only sends delta updates for protocol additions / removals (and maybe later, addresses). I'll keep this PR parked until the dust settles, but I figured I'd get a jump on it now.

@raulk am I correct that identify/push now always sends full updates?

### DIFF
--- a/identify/README.md
+++ b/identify/README.md
@@ -41,7 +41,8 @@ and spec status.
 
 ## Overview
 
-There are two variations of the identify protocol, `identify` and `identify/push`.
+There are three variations of the identify protocol, `identify`,
+`identify/push`, and `identify/delta`.
 
 ### `identify`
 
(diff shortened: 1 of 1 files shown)


identify/README.md
```diff
@@ -3,7 +3,8 @@
 The identify protocol is used to exchange basic information with other peers
 in the network, including addresses, public keys, and capabilities.
 
-There are two variations of the identify protocol, `identify` and `identify/push`.
+There are three variations of the identify protocol, `identify`,
+`identify/push`, and `identify/delta`.
```
raulk: we should probably specify the exact protocol IDs in a reference table here, even if we repeat them in prose below.

identify/README.md
```diff
@@ -28,9 +29,28 @@ The push variant works by opening a stream to each remote peer you want to updat
 the local peer will send an `Identify` message and close the stream.
 
 Upon recieving the pushed `Identify` message, the remote peer should update their local
```
raulk:
```suggestion
Upon receiving the pushed `Identify` message, the remote peer should update their local
```

jacobheun:
Does this really need to be a separate protocol? I'm not seeing the need here aside from the concern about older peers breaking, in which case I feel like this should just become `/ipfs/id/push/2.0.0`. Adding `delta` is ultimately going to result in `push` no longer being used.

raulk:
Librarian note: conversation continued in https://github.com/libp2p/go-libp2p/pull/659#issuecomment-504451553 and onwards.

identify/README.md
```diff
@@ -3,7 +3,16 @@
...
 in the network, including addresses, public keys, and capabilities.
 
-There are two variations of the identify protocol, `identify` and `identify/push`.
+There are three variations of the identify protocol, listed below:
+
+| Protocol         | Protocol Id           | Description                                    |
+|------------------|-----------------------|------------------------------------------------|
+| `identify`       | `/ipfs/id/1.0.0`      | Respond to queries for identifying info.       |
+| `identify/push`  | `/ipfs/id/push/1.0.0` | Send info proactively to inform about changes. |
+| `identify/delta` | `/p2p/id/delta/1.0.0` | Send partial updates to inform about changes.  |
```
Stebalien: Is this not just "identify push 1.1.0"?

(either way works for me, this just feels like identify push with an extra feature)

Stebalien:
This protocol makes absolutely no sense.

* Why are we reusing the protobuf from identify while _also_ saying that the new "delta" field is exclusive with the rest of the fields?
* Why doesn't this support updating addresses? Is the goal to use identify push for that and make this address only?

@raulk or @vyzo?

Stebalien:
> Why doesn't this support updating addresses? Is the goal to use identify push for that and make this address only?

Nvm. This part won't matter once we get signed peer addresses anyways.

Stebalien:
Proposed change: https://github.com/libp2p/specs/pull/253/files

aschmahmann:
@raulk @Stebalien what's the state of this PR?

mxinden:
Given that this effort has stalled and given that go-libp2p removed support for identify delta (https://github.com/libp2p/go-libp2p/pull/1975) I am closing here.

Please comment in case you disagree with the direction.

Issue #184, opened 2019-07-05
Specify the `insecure` security channel


raulk:
Several implementers (e.g. cpp-libp2p and py-libp2p) are using this security channel for interoperability tests against go-libp2p.

_While this is discouraged for production use, it should at least be functional. It is not._

The reason is that downstream components fail because we perform no handshake that covers identity exchange. Thus, the dialer knows the peer ID of the responder, but the latter never learns the peer ID of the dialer, causing panics and errors in places that _require_ the peer ID:

  * https://github.com/libp2p/go-libp2p-peerstore/issues/87
  * https://github.com/libp2p/go-libp2p/issues/452

Let's iterate on our implementation of `insecure`, and let's do that spec-first. @yusefnapora has stepped up to lead the charge.

In my view, the spec would be pretty brief.

1. We assume an exogenous protocol negotiation mechanism exists such that it selects `/plaintext/1.0.0.`.
2. Both peers exchange a protobuf message including the PeerID and their public key.
3. The dialer asserts that the peer ID matches its expectation and that it is indeed derived from the attached public key.
4. The responder can only assert the peer ID is coherent with the attached public key.

Stebalien:
We could also just send the keys only. But yeah, that sounds like the right approach.

yusefnapora:
It looks like the c++ implementation that Soramitsu are working on [sends just the keys](https://github.com/soramitsu/kagome/pull/166/files#diff-f097cad164242305e95eb79d1c442a18R18). 

@Warchant, in the PR above, does the responder derive a peer id for the initiator from their key?

yusefnapora:
So, personally, I would actually expect the plaintext protocol to not use keys at all, but still exchange peer ids. In which case you should be able to use arbitrary strings as peer ids, e.g. "insecure node 123" should be a valid plaintext peer id. 

The lack of keys is a pretty strong signal that this channel isn't secure 😄Whereas if a node were unknowingly misconfigured to allow plaintext connections, they could be presented with an arbitrary key and possibly trust it (e.g. by adding to the peerstore) without ever verifying. 

I don't know if that's actually problematic or not; just thought I'd mention. It's probably worth requiring keys just so that we can always assume that we have them, and avoid nil panics and so on.

Warchant:
@yusefnapora the [protocol is](https://github.com/soramitsu/kagome/blob/master/core/libp2p/security/plaintext/plaintext.hpp#L17):

1. Initiator immediately sends his public key to the other peer.
[...]

PR #186, opened 2019-07-08
Add draft spec for plaintext key exchange protocol

yusefnapora:
This adds a spec for exchanging keys and peer ids when setting up the plaintext "security" protocol, as discussed in #184.

This sends both peer ids and keys as @raulk outlined - we could potentially just send the keys as in @Warchant's c++ implementation and derive the ids.

Also, I bumped the version in the protocol id, since it seems like sending the key message over will break anyone using the existing `/plaintext/1.0.0` protocol id.

### DIFF
--- a/plaintext/README.md
+++ b/plaintext/README.md
@@ -9,9 +9,10 @@
 
 Authors: [@yusefnapora]
 
-Interest Group: TBD
+Interest Group: [@raulk], TBD
 
 [@yusefnapora]: https://github.com/yusefnapora
+[@raulk]: https://github.com/raulk
 
 See the [lifecycle document][lifecycle-spec] for context about maturity level
 and spec status.
@@ -42,7 +43,7 @@ process][conn-spec-conn-upgrade].
 ## Protocol Id and Version History
 
 The plaintext protocol described in this document has the protocol id of
-`/plaintext/1.1.0`. 
+`/plaintext/2.0.0`. 
 
 An earlier version, `/plaintext/1.0.0`, was implemented in several languages,
 but it did not include any exchange of public keys or peer ids. This led to
@@ -105,8 +106,8 @@ unsigned-varint spec][uvarint-spec].
 
 ### Exchange
 
-Once the plaintext protocol has been negotiated, both peers send an `Exchange`
-message containing their peer id and public key. 
+Once the plaintext protocol has been negotiated, both peers immediately send an
+`Exchange` message containing their peer id and public key.
 
 Upon receiving an `Exchange` message from the remote peer, each side will
 validate that the given peer id is consistent with the given public key by
@@ -114,7 +115,7 @@ deriving a peer id from the key and asserting that it's a match with the `id`
 field in the `Exchange` message.
 
 Dialing a peer in libp2p requires knowledge of the listening peer's peer id. As
-a result, the dialing peer also verifies that the peer id presented by the
+a result, the dialing peer ALSO verifies that the peer id presented by the
 listening peer matches the peer id that they attempted to dial. As the listening
 peer has no prior knowledge of the dialer's id, only one peer is able to perform
 this additional check.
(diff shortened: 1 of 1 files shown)


plaintext/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+This document describes the exchange of peer ids and keys that occurs when
+initiating a plaintext connection. This exchange happens after the plaintext
+protocol has been negotiated as part of the [connection upgrade
+process][conn-spec-conn-upgrade].
+
+## Protocol Id and Version History
+
+The plaintext protocol described in this document has the protocol id of
+`/plaintext/1.1.0`.
```
raulk: Since this is a breaking change, we should probably bump the version up to 2.0.0.

plaintext/README.md
```diff
@@ -0,0 +1,146 @@
...
+## Protocol Id and Version History
+
+The plaintext protocol described in this document has the protocol id of
+`/plaintext/1.1.0`. 
+
+An earlier version, `/plaintext/1.0.0`, was implemented in several languages,
+but it did not include any exchange of public keys or peer ids. This led to
+undefined behavior in parts of libp2p that assumed the presence of a peer id.
+
+As version `1.0.0` had no associated wire protocol, it was never specified.
```
raulk: 👍

plaintext/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Message Framing
+
+All messages sent over the wire are prefixed with the message length in bytes,
+encoded as an unsigned variable length integer as defined by the [multiformats
+unsigned-varint spec][uvarint-spec].
+
+### Exchange
+
+Once the plaintext protocol has been negotiated, both peers send an `Exchange`
```
raulk:
```suggestion
Once the plaintext protocol has been negotiated, both peers immediately send an `Exchange`
```

plaintext/README.md
```diff
@@ -0,0 +1,146 @@
...
+message containing their peer id and public key. 
+
+Upon receiving an `Exchange` message from the remote peer, each side will
+validate that the given peer id is consistent with the given public key by
+deriving a peer id from the key and asserting that it's a match with the `id`
+field in the `Exchange` message.
+
+Dialing a peer in libp2p requires knowledge of the listening peer's peer id. As
+a result, the dialing peer also verifies that the peer id presented by the
```
raulk:
```suggestion
a result, the dialing peer ALSO verifies that the peer id presented by the
```

raulk: I find stressing ALSO helps readability here.

plaintext/README.md
```diff
@@ -0,0 +1,146 @@
...
+listening peer matches the peer id that they attempted to dial. As the listening
+peer has no prior knowledge of the dialer's id, only one peer is able to perform
+this additional check.
+
+Once each side has received the `Exchange` message, they may store the public
+key and peer id for the remote peer in their local peer metadata storage (e.g.
+go-libp2p's [peerstore][go-libp2p-peerstore], or js-libp2p's
+[peer-book][js-peer-book]).
+
+Following delivery of `Exchange` messages, the plaintext protocol is complete.
```
raulk:
```suggestion
Following delivery and verification of `Exchange` messages, the plaintext protocol is complete.
```

plaintext/README.md
```diff
@@ -0,0 +1,146 @@
...
+No response is required, as it is assumed that peers will close the connection
```
raulk:
```suggestion
Should a verification or timeout error occur, the connection MUST be terminated abruptly.

The connection is now ready for insecure and unauthenticated data exchange. While we do exchange public keys upfront, replay attacks and forgery are trivial. We do not authenticate messages either. Therefore we reiterate the unsuitability for production usage of `/plaintext/2.0.0`.
```

plaintext/README.md
```diff
@@ -0,0 +1,146 @@
...
+> An insecure connection handshake for non-production environments.
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+|-----------------|---------------|--------|-----------------|
+| 1A              | Working Draft | Active | r0, 2019-05-27  |
+
+
+Authors: [@yusefnapora]
+
+Interest Group: TBD
```
raulk: Feel free to add me here.

Kubuxu:
I would be for adding a big disclaimer that this spec is only for development purposes and should never be shipped in actual application or library.

yusefnapora:
Awesome, thanks for all the reviews! I added everyone in the thread to the IG - please PR or ping me if you'd rather not be.  I also added another warning about not using in production - hopefully people get the message.

PR #196, opened 2019-07-29
multistream simultaneous open protocol

vyzo:
To unblock hole punching protocols.

### DIFF
--- a/connections/simopen.md
+++ b/connections/simopen.md
@@ -1,0 +1,72 @@
+# Simultaneous Open for multistream-select
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+|-----------------|---------------|--------|-----------------|
+| 1A              | Working Draft | Active | DRAFT           |
+
+Authors: [@vyzo]
+
+Interest Group: [@raulk], [@stebalien]
+
+[@vyzo]: https://github.com/vyzo
+[@raulk]: https://github.com/raulk
+[@stebalien]: https://github.com/stebalien
+
+See the [lifecycle document][lifecycle-spec] for context about maturity level
+and spec status.
+
+[lifecycle-spec]: https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md
+
+
+## Introduction
+
+In order to support direct connections through NATs with hole
+punching, we need to account for simultaneous open. In such cases,
+there is no single initiator and responder, but instead both peers act
+as initiators. This is breaks protocol negotiation in
+multistream-select, which assumes a single initator.
+
+This draft proposes a simple extension to the multistream protocol
+negotiation in order to select a single initator when both peers are
+acting as such.
+
+## The Protocol
+
+When a peer acting as the initiator enters protocol negotiation, it
+sends the string `iamclient` as first protocol selector. If the other
+peers is a responder or doesn't support the extension, then it
+responds with `na` and protocol negotiation continues as normal.
+
+If both peers believe they are the initiator, then they both send
+`iamclient`. If this is the case, they enter an initiator selection
+phase, where one of the peers is selected to act as the initiator. In
+order to do so, they both generate a random 256-bit integer and send
+it as response to the `iamclient` directive. The peer with the highest
+integer is selected to act as the initator and sends an `initiator`
+message. The peer with the lowest integer responds with `responder`
+message and both peers transition to protocol negotiation with a
+distinct initiator.
+
+The following schematic illustrates, for the case where A's integer is
+higher than B's integer:
+
+```
+A ---> B: iamclient
+B ---> A: iamclient
+A: generate random integer IA
+B: generate random integer IB
+A ---> B: {IA}
+B ---> A: {IB}
+A ---> B: initiator
+B ---> A: responder
+```
+
+In the unlikely case where both peers selected the same integer, they
+generate a fresh one and enter another round of the protocol.
+
+## Implementation Considerations
+
+The protocol is simple to implement and is backwards compatible with
+vanilla multistream-select.  In the common case of a single initiator,
+we can ensure that there there is no latency overhead by sending the
+`iamclient` message together with the multistream header.


tomaka:
For what it's worth, rust-libp2p only allows one simulatenous connection between two given nodes.

If a node detects that it has a second connection, the one with the lowest `PeerId` automatically has the dialing priority.

The other connection is automatically closed, but only once we make sure that the useful connection has been acknowledged by the remote (ie. the remote has sent at least one byte on it), otherwise there's a possible race condition where the other node closes the useful connection.

This was quite tricky to implement, but it avoids the round-trip.

vyzo:
We discussed a simple solution with @Stebalien , to basically pipeline `multistream,iamclient:random-id,protocol` in a single packet so as to avoid the negotiation altogether.
The only problem is the miniscule probability of selecting the same random integer, but we can throw our hands in the air in that case.
I will update the spec to reflect this.

vyzo:
The second, more robust solution, is to use a prefix for the integer when both sides declare `iamclient` and ignore protocols that are piggybacked (we discussed this also with @Stebalien).
Spec update forthcoming.

connections/simopen.md
```diff
@@ -0,0 +1,72 @@
+# Simultaneous Open for multistream-select
```
raulk: This spec only applies to connection bootstrapping, whereas multistream-select is used in other contexts too (stream-level protocol negotiation). So I'd title this differently, e.g.:

```
# Connection bootstrapping: handling TCP simultaneous open in multistream-select.
```

vyzo: Sure, we can change the title. Go ahead and pick something you like!

vyzo: Also, this works for UDP too!

raulk:
@vyzo Yup, I think that pipelining sequence could work. I'll continue reviewing tomorrow; thanks!

vyzo:
I pushed an update, which resolves the issues with pipelining.
The solution adopted was option 2, in which case the peers use a prefix for the random integer and can ignore the pipelined security protocol.

vyzo:
summoning @raulk @Stebalien for a second round of review.

connections/simopen.md
```diff
@@ -55,18 +59,30 @@ A ---> B: iamclient
...
-B ---> A: {IB}
+A ---> B: select:{IA}
+B ---> A: select:{IB}
 A ---> B: initiator
 B ---> A: responder
 ```
 
 In the unlikely case where both peers selected the same integer, they
-generate a fresh one and enter another round of the protocol.
+generate a fresh one and enter another round of the protocol. If
```
dirkmc: In the case where both peers select the same integer could they use their peer ids as a tie-breaker (eg lowest wins)?

jacobheun: multistream has no concept of peer ids which prevents us from doing this, otherwise we could avoid the random number altogether.

connections/simopen.md
```diff
@@ -0,0 +1,88 @@
...
+
+[lifecycle-spec]: https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md
+
+
+## Introduction
+
+In order to support direct connections through NATs with hole
+punching, we need to account for simultaneous open. In such cases,
+there is no single initiator and responder, but instead both peers act
+as initiators. This is breaks protocol negotiation in
```
jacobheun:
```suggestion
as initiators. This breaks protocol negotiation in
```

connections/simopen.md
tomaka: This look a bit like an implementation detail to me.
Rust-libp2p is using the lowest `PeerId` to determine who has priority in case of simultaneous connection.

raulk: We can make this an arbitrary byte string. The advantage of using a peer ID here is that, later (after handshaking), we _could_ validate if the peer was honest, and disconnect/blacklist it if dishonest. Rationale: if I'm dishonest, I can make myself win this challenge always by sending 0. Not that this will gain me any advantage that I can think of, but it's a nuance worth considering.

raulk: We could also mitigate this by making peers hash the received nonce (e.g. SHA256) prior to comparison.

raulk: @tomaka curious to learn you already have an implementation. Mind posting a pointer to the code?

vyzo: I don't think we should even have the concept of a peer id in multistream, this is at the bottom of the stack.
We could use hashing, but does it really gain anything?

tomaka: > @tomaka curious to learn you already have an implementation. Mind posting a pointer to the code?

Rust-libp2p doesn't implement UDP or hole punching, so we're not doing this in the scope of what this PR is about. Rust-libp2p only ever allows one connection between two given peers, and if we detect two active connections, the one with the lowest `PeerId` has the dialing priority and the other one is shut down.

Here's the code that handles an incoming connection for which we learned the `PeerId` of the remote: https://github.com/libp2p/rust-libp2p/blob/2fd941122a6c589501ad1c04b0f091f1337f5f49/core/src/nodes/network.rs#L1130-L1196

connections/simopen.md
```diff
@@ -0,0 +1,88 @@
...
+sends the string `iamclient` as first protocol selector. If the other
+peers is a responder or doesn't support the extension, then it
+responds with `na` and protocol negotiation continues as normal.
+
+If both peers believe they are the initiator, then they both send
+`iamclient`. If this is the case, they enter an initiator selection
+phase, where one of the peers is selected to act as the initiator. In
+order to do so, they both generate a random 256-bit integer and send
+it as response to the `iamclient` directive, prefixed with the
+`select:` string. The peer with the highest integer is selected to act
```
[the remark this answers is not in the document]
vyzo: good point, this needs to be encoded! Let's do base64.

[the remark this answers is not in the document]
vyzo: Updated to mandate base64.

connections/simopen.md
```diff
@@ -0,0 +1,89 @@
...
+sends the string `iamclient` as first protocol selector. If the other
+peers is a responder or doesn't support the extension, then it
+responds with `na` and protocol negotiation continues as normal.
+
+If both peers believe they are the initiator, then they both send
+`iamclient`. If this is the case, they enter an initiator selection
+phase, where one of the peers is selected to act as the initiator. In
+order to do so, they both generate a random 256-bit integer and send
+it as response to the `iamclient` directive, prefixed with the
+`select:` string. The integer is in big-endian format, encoded in base64.
```
aarshkshah1992, line 45: @vyzo Why do we need to encode to base64 here ?

marten-seemann (commented):
One question and one proposal:

* Do I understand correctly that we’ll end up with an additional round trip if the peer doesn’t support the `iamclient` protocol?
* Why don't we coalesce the initiator selection phase, i.e. send `iamclient/<random int>`? This would allow the peers to immediately assign roles when a simultaneous open occurs, saving one roundtrip.

connections/simopen.md
marten-seemann: I'd assume it's because multistream 1 is a text-based, human-readable protocol (with all the pros and cons that come with it).

That said, base64 seems like a suboptimal choice here. I propose switching to hexadecimal encoding. Encoding `math.MaxUint64` (18446744073709551615) is `0xFFFFFFFFFFFFFFFF` in hex and `MTg0NDY3NDQwNzM3MDk1NTE2MTU=` in base64.
Also, 256 bits seem excessive here. I think we can live with a collision probability of 2^-64, which would allow implementations to use a regular `uint64` instead of `math/big.Int`. We might even want to limit this to 52 (?) bits to make things easier in JavaScript (but I'm no expert on that).

tomaka: > Encoding math.MaxUint64 (18446744073709551615) is 0xFFFFFFFFFFFFFFFF in hex and MTg0NDY3NDQwNzM3MDk1NTE2MTU= in base64.

[I think it would be `//////////8=`](https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=ba311359861b6ab7459cc0345f57567b)

marten-seemann: Would it? I used https://www.base64encode.net/, and that gives the same result (except for the last character?) as `echo 18446744073709551615 | base64`. But maybe that's decoding the string "18446744073709551615" and not the number?

As `/` is a separator we use in multistream, so base64 encoding might get us into trouble, if we don't escape it (one of the perks of using a text-based protocol...).

aarshkshah1992:
@marten-seemann 

> One question and one proposal:
> 
> * Do I understand correctly that we’ll end up with an additional round trip if the peer doesn’t support the `iamclient` protocol?

- We will NOT end up with an extra roundtrip because the client sends a pipelined `multistream1.0|iamclient|noise` message to the remote peer and the remote peer will write a `multistream`,`na`, `noise` message back if it is NOT. simultaneous open.
If it is a simultaneous open and the remote peer does NOT support the `iamclient` protocol, things will fail just as they do today.


> * Why don't we coalesce the initiator selection phase, i.e. send `iamclient/<random int>`? This would allow the peers to immediately assign roles when a simultaneous open occurs, saving one roundtrip.

I think this is extremely hacky. multistream is supposed to be "yes"/"no" protocol. If we do this, we'd have to add special logic for the "iamclient" protocol where we say "hey multistream, read the integer after the / if you see the iamclient thing".
Any thoughts @Stebalien @jacobheun

aarshkshah1992:
@marten-seemann 

> One question and one proposal:
> 
> * Do I understand correctly that we’ll end up with an additional round trip if the peer doesn’t support the `iamclient` protocol?

- We will NOT end up with an extra roundtrip because the client sends a pipelined `multistream1.0|iamclient|noise` message to the remote peer and the remote peer will write a `multistream`,`na`, `noise` message back if it is NOT. simultaneous open.
If it is a simultaneous open and the remote peer does NOT support the `iamclient` protocol, things will fail just as they do today.


> * Why don't we coalesce the initiator selection phase, i.e. send `iamclient/<random int>`? This would allow the peers to immediately assign roles when a simultaneous open occurs, saving one roundtrip.

I think this is extremely hacky. multistream is supposed to be "yes"/"no" protocol. If we do this, we'd have to add special logic for the "iamclient" protocol where we say "hey multistream, read the integer after the / if you see the iamclient thing".
Any thoughts @Stebalien @jacobheun

connections/simopen.md
tomaka: Yes I guess you're encoding the string.
It doesn't make sense that encoding something in base64 would give something longer than encoding that same thing in base16

I think that `/` isn't a problem because of the `select:` prefix. Otherwise I'd suggest base58.

marten-seemann: Not opposed to base58, but not sure if that's easily accessible in all the languages we care about. In Go, we'd have to import a third-party library (it's not in the standard library).

Any thoughts on the length of this value?

vyzo:
@aarshkshah1992 @marten-seemann the go implementation has diverged from the spec; can you please update it to match what was implemented?

connections/simopen.md
```diff
@@ -0,0 +1,89 @@
...
+This draft proposes a simple extension to the multistream protocol
+negotiation in order to select a single initator when both peers are
+acting as such.
+
+## The Protocol
+
+When a peer acting as the initiator enters protocol negotiation, it
+sends the string `iamclient` as first protocol selector. If the other
+peers is a responder or doesn't support the extension, then it
+responds with `na` and protocol negotiation continues as normal.
```
mxinden, line 38: Note that the [rust-libp2p multistream-select implementation currently requires protocols to start with a `/`](https://github.com/libp2p/rust-libp2p/blob/40c4287bc9a95461889e10612519960a16a2874f/misc/multistream-select/src/protocol.rs#L73-L82) and thus the `iamclient` message is not backwards compatible with rust-libp2p as it would not parse as a valid protocol identifier.

The specification recommends protocols to start with a `/`, but does not require it. The rust-libp2p implementation is thus not in-line with the specification.

> Each protocol supported by a peer is identified using a unique string called a protocol id. While any string can be used, the conventional format is a path-like structure containing a short name and a version number, separated by / characters. For example: /mplex/1.0.0 identifies version 1.0.0 of the mplex stream multiplexing protocol. multistream-select itself has a protocol id of /multistream/1.0.0.

https://github.com/libp2p/specs/tree/master/connections#protocol-negotiation

I am posting this not to ask the specification to change, but only to raise awareness in case other implementations are as strict as rust-libp2p. (//CC @vasco-santos for the js-libp2p side).

mxinden:
> @aarshkshah1992 @marten-seemann the go implementation has diverged from the spec; can you please update it to match what was implemented?

@aarshkshah1992 @marten-seemann any updates here? Having an up-to-date spec would make it a lot easier to implement this extension in rust-libp2p.

connections/simopen.md
tomaka: > Note that the rust-libp2p multistream-select implementation currently requires protocols to start with a /

This was done because the code that parses multistream-select payloads is context-free.
In other words, if we want to parse the string `"na"`, the code that does so doesn't know whether it should interpret it as a "not available" response or as a protocol named `"na"`.

Stebalien: It's still reasonable to change this, it's not used in practice anyways.

Really, it would likely make more sense as `/libp2p/simultaneous-connect`.

mxinden: Though initially not intentioned, I am in favor of replacing `iamclient` with `/libp2p/simultaneous-connect`. This would both comply with the above mentioned convention to start (fake) protocol IDs with a `/` and be interoperable with the stateless parser in rust-libp2p.

I have opened https://github.com/libp2p/specs/pull/321 which would do the change in the specification. Please let me know what you think either on https://github.com/libp2p/specs/pull/321 or here.

connections/simopen.md
mxinden: This discussion seems to have stalled. The spec has settled on a random integer instead of a peer ID. Please un-resolve in case you would like to continue this discussion.

connections/simopen.md
```diff
@@ -0,0 +1,89 @@
+# Simultaneous Open for bootstrapping connections in multistream-select
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+|-----------------|---------------|--------|-----------------|
+| 1A              | Working Draft | Active | DRAFT           |
```
mxinden, line 5:
```suggestion
| 1A              | Working Draft | Active | r0, 2021-05-06 |
```

connections/simopen.md
```diff
@@ -0,0 +1,89 @@
...
+
+When a peer acting as the initiator enters protocol negotiation, it sends the
+string `/libp2p/simultaneous-connect` as first protocol selector. If the other
+peers is a responder or doesn't support the extension, then it responds with
+`na` and protocol negotiation continues as normal.
+
+If both peers believe they are the initiator, then they both send
+`/libp2p/simultaneous-connect`. If this is the case, they enter an initiator
+selection phase, where one of the peers is selected to act as the initiator. In
+order to do so, they both generate a random 256-bit integer and send it as
```
mxinden, line 43:
```suggestion
order to do so, they both generate a random 64-bit integer and send it as
```

The Golang implementation is using a 64-bit integer:

```Golang
	myNonce := binary.LittleEndian.Uint64(randBytes)
```

https://github.com/multiformats/go-multistream/blob/4e1d9a7e25b067b7527ca2af72ce9c06529ebfb7/client.go#L167

connections/simopen.md
```diff
@@ -0,0 +1,89 @@
...
+response to the `/libp2p/simultaneous-connect` directive, prefixed with the
+`select:` string. The integer is in big-endian format, encoded in base64. The
```
mxinden, line 45:
```suggestion
`select:` string. The integer is send in its base-10 string representation. The
```

The Golang implementation is sending the nonce in the base-10 string representation.

```Golang
		myselect := []byte(tieBreakerPrefix + strconv.FormatUint(myNonce, 10))
```

https://github.com/multiformats/go-multistream/blob/4e1d9a7e25b067b7527ca2af72ce9c06529ebfb7/client.go#L171

connections/simopen.md
```diff
@@ -0,0 +1,89 @@
...
+A ---> B: select:{IA}
+B ---> A: select:{IB}
+A ---> B: initiator
+B ---> A: responder
+```
+
+In the unlikely case where both peers selected the same integer, they
+generate a fresh one and enter another round of the protocol. If
+multiple rounds of the protocol result in the same integers, this is
+indicative of a bug and both peers should abort the connection.
```
mxinden, line 72:
```suggestion
In the unlikely case where both peers selected the same integer, connection establishment fails.
```

The Golang implementation does not retry. See https://github.com/multiformats/go-multistream/pull/42#discussion_r558755480.

mxinden (commented):
I have implemented simultaneous open for rust-libp2p (https://github.com/libp2p/rust-libp2p/pull/2066) with the help of this specification and the Golang implementation. I have treated the Golang implementation as a source of truth. According to basic smoke tests the two implementations are interoperable now. Below comments highlight the inconsistencies between the two.

With the upcoming go-libp2p release (https://github.com/libp2p/go-libp2p/issues/1090) this spec (+ the comments below) will be deployed in the wild.

That in mind, unless there are any objections, I will commit the chang-suggestions below and try to get this specification merged.

Let me know what you think.

connections/simopen.md
```diff
@@ -0,0 +1,89 @@
+# Simultaneous Open for bootstrapping connections in multistream-select
```
mxinden, line 1:
```suggestion
# Connection bootstrapping: Handling simultaneous open in multistream-select.
```

As per @raulk's comment in https://github.com/libp2p/specs/pull/196#discussion_r308391270.

mxinden (approved):
With the most recent commits the spec, the Golang implementation and the Rustlang implementation are now consistent (to the best of my knowledge).

Unless there are any objections I suggest to merge this pull request _as-is_ sometime next week.

Issue #195, opened 2019-07-26
Standardise Noise handshake for libp2p


raulk:
There have been isolated initiatives to implement a Noise handshake for libp2p, and it's time to standardise on a common approach for interoperability 🎉

## Prior art

* [rust-libp2p-noise](https://docs.rs/libp2p-noise/0.9.0/libp2p_noise/): experimental support for IX, IK, KK handshake patterns in rust-libp2p with protocol suite `Noise_{IK/IX/KK}_25519_ChaChaPoly_SHA256`.
* [proposal](https://github.com/libp2p/go-libp2p/issues/631) for a hack project to introduce the Noise IX handshake in go-libp2p.

## Some questions to answer

* Which handshake patterns we'll support? These are valid ones:
    * [IX](https://noiseexplorer.com/patterns/IX/): 1 RTT to be sound, exposes the initiator's public static key in clear text.
    * [IK](https://noiseexplorer.com/patterns/IK/): 1 RTT to be sound, encrypts initiator's public static key with the responder's static key, providing initiator privacy, but requires previous knowledge of the initiator's public key. Might be suitable to use against well-known bootstrap nodes.
    * [XX](https://noiseexplorer.com/patterns/XX/): 1.5 RTT to be sound, tradeoff is that no static key is exposed in plaintext without requiring prior knowledge of the responder's public key.

* libp2p key support. **Noise relies on ECDH operations and does not support RSA keys.**
    * We could maybe shim support for RSA keys with the NN handshake (only ephemeral keys, no static keys), and leveraging Noise message data to transmit a signature of the ephemeral key with our libp2p RSA key.
    * ~rust-libp2p does the above, but they use IX, IK, XX, and generates a Curve25519 key on the spot to fill in for the static key.~ (see https://github.com/libp2p/specs/issues/195#issuecomment-515586748 for correction)
    * The Noise framework does not mention P-256 and secp256k1 curves explicitly, but they're used in practice. See the [Bitcoin lightning network specs](https://github.com/lightningnetwork/lightning-rfc/blob/master/08-transport.md#noise-protocol-instantiation) and [implementation](https://github.com/lightningnetwork/lnd/blob/master/brontide/noise.go#L25).
    * I'm not sure how two nodes with non-matching key types would conduct a Noise handshake; we might have to fallback to NN in this case, with a linked identity.
[...]

PR #202, opened 2019-08-05
Draft of noise-libp2p spec

yusefnapora:
Hey all, this is what I've got so far to address #195 and specify a Noise security protocol for libp2p. It basically follows the outline I put in the [issue comments there](https://github.com/libp2p/specs/issues/195#issuecomment-517449710) and is very similar to the rust-libp2p implementation.

I'll fill in the Design Consideration sections soon, but didn't want to commit before we're done considering designs :)

Does this seem like a good direction? 

@raulk @tomaka @romanb - I added you all to the interest group. If anyone else wants in, let me know!

### DIFF
--- a/noise/README.md
+++ b/noise/README.md
@@ -306,7 +306,7 @@ Pipes](#optimistic-0-rtt-with-noise-pipes) and is described in that context belo
 
 ### Optimistic 0-RTT with Noise Pipes
 
-The Noise spec describes a [compound protocol][npf-compound-protocol] called
+The Noise spec describes a [compound protocol][npf-compound-protocols] called
 [Noise Pipes][npf-noise-pipes], which enables 0-RTT encryption in the optimistic
 case, while allowing peers to fallback to a different Noise protocol if their
 initial handshake attempt fails.
@@ -727,7 +727,7 @@ currently supported by the spec, while libp2p identity keys may be of other
 unsupported types like RSA.
 
 [peer-id-spec]: ../peer-ids/peer-ids.md
-[peer-id-spec-key-rules]: ../peer-ids/peer-ids.md#how-keys-are-encoded-and-messages-signed
+[peer-id-spec-signing-rules]: ../peer-ids/peer-ids.md#how-keys-are-encoded-and-messages-signed
 
 [conn-spec]: ../connections/README.md
 [multiselect-2-pr]: https://github.com/libp2p/specs/pull/95
--- a/noise/README.md
+++ b/noise/README.md
@@ -1,6 +1,6 @@
-# noise-libp2p - Channel Security Protocol
+# noise-libp2p - Secure Channel Handshake
 
-> A libp2p transport security protocol built with the Noise Protocol Framework.
+> A libp2p transport secure channel handshake built with the Noise Protocol Framework.
 
 | Lifecycle Stage | Maturity      | Status | Latest Revision |
 |-----------------|---------------|--------|-----------------|
@@ -8,17 +8,59 @@
 
 Authors: [@yusefnapora]
 
-Interest Group: [@raulk], [@tomaka], [@romanb]
+Interest Group: [@raulk], [@tomaka], [@romanb], [@shahankhatch], [@Mikerah]
 
 [@yusefnapora]: https://github.com/yusefnapora
 [@raulk]: https://github.com/raulk
 [@tomaka]: https://github.com/tomaka
 [@romanb]: https://github.com/romanb
+[@shahankhatch]: https://github.com/shahankhatch
+[@Mikerah]: https://github.com/Mikerah
+
 
 See the [lifecycle document][lifecycle-spec] for context about maturity level
 and spec status.
 
 [lifecycle-spec]: https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md
+
+
+## Table of Contents
+
+- [noise-libp2p - Secure Channel Handshake](#noise-libp2p---secure-channel-handshake)
+    - [Table of Contents](#table-of-contents)
+    - [Overview](#overview)
+    - [Negotiation](#negotiation)
+        - [Negotiating Whether to Use Noise](#negotiating-whether-to-use-noise)
+        - [Negotiating a Specific Noise Protocol](#negotiating-a-specific-noise-protocol)
+            - [Skipping Noise Protocol Negotiation to Prevent Deep Packet Inspection](#skipping-noise-protocol-negotiation-to-prevent-deep-packet-inspection)
+            - [Noise Socket Negotiation Data](#noise-socket-negotiation-data)
+            - [Noise Socket Negotiation Flow](#noise-socket-negotiation-flow)
+    - [The Noise Handshake](#the-noise-handshake)
+        - [Static Key Authentication](#static-key-authentication)
+        - [libp2p Data in Handshake Messages](#libp2p-data-in-handshake-messages)
+            - [The libp2p Signed Handshake Payload](#the-libp2p-signed-handshake-payload)
+        - [Supported Handshake Patterns](#supported-handshake-patterns)
+            - [XX](#xx)
+            - [IX](#ix)
+        - [Optimistic 0-RTT with Noise Pipes](#optimistic-0-rtt-with-noise-pipes)
+            - [IK](#ik)
+            - [XXfallback](#xxfallback)
+    - [Cryptographic Primitives](#cryptographic-primitives)
+        - [DH Functions](#dh-functions)
+        - [Cipher Functions](#cipher-functions)
+        - [Hash Functions](#hash-functions)
+    - [Valid Noise Protocol Names](#valid-noise-protocol-names)
+    - [Wire Format](#wire-format)
+        - [Noise Socket Handshake Messages](#noise-socket-handshake-messages)
+        - [Noise Socket Transport Messages](#noise-socket-transport-messages)
+        - [Noise Socket Encrypted Payloads](#noise-socket-encrypted-payloads)
+    - [libp2p Interfaces and API](#libp2p-interfaces-and-api)
+    - [Design Considerations](#design-considerations)
+        - [Why Noise Socket?](#why-noise-socket)
+        - [Separate Negotiation for Noise Protocols](#separate-negotiation-for-noise-protocols)
+        - [No Version Number in Protocol ID](#no-version-number-in-protocol-id)
+        - [Distinct Noise and Identity Keys](#distinct-noise-and-identity-keys)
+        - [Why Not Noise Signatures?](#why-not-noise-signatures)
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


shahankhatch:
If possible, could I please be added to the interest group? Thanks in advance!

Mikerah:
Count me in the interest group!

raulk:
I’d like to nominate @burdges and @zmanian for the interest group too. Read about the role of the interest group here: https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md

raulk:
Also @arnetheduck and @AgeManning, if they’re so inclined.

AgeManning:
If you'd like extra eyes on it, happy to help, however my contribution may be low as I'm traveling for the next stretch of time.

noise/README.md
```diff
@@ -0,0 +1,400 @@
...
+
+### Supported DH Functions
+
+noise-libp2p supports the 25519 DH functions [as defined in the Noise
+spec][npf-dh-25519].
+
+### Supported Cipher Functions
+
+noise-libp2p supports the ChaChaPoly cipher functions [as defined in the Noise
+spec][npf-cipher-chachapoly].
```
marten-seemann: AESGCM is a lot faster when running on hardware that has AES-support. For TLS, @kubuxu came up with a [nifty trick](https://github.com/libp2p/go-libp2p-tls/blob/46e9e9f9b00e3f896a17b8613bfb6efde5b177cc/crypto.go#L202-L222) to select AES if both peers have hardware support.

Is there any reason to exclude AESGCM here?

noise/README.md
```diff
@@ -0,0 +1,400 @@
...
+all cases, Alice initiates the first handshake attempt (either `XX` or `IK`).
+Bob MAY initiate an `XXfallback` handshake following a failed `IK` attempt. The
+`XXfallback` handshake is written in "Bob-initiated form," with the direction of
+the arrows reversed from the canonical (Alice-initiated) notation.
+
+```
+XX:
+  -> e
+  <- e, ee, s, es   [Alice's identity payload]
+  -> s, se          [Bob's identity payload]
```
marten-seemann: Are Alice and Bob the wrong way around here?

noise/README.md
```diff
@@ -0,0 +1,400 @@
...
+
+The`XX` pattern is used when the initiator (or in libp2p terms, the dialing peer)
+does not have knowledge of the static Noise key for the responder (the listening
+peer). As Noise static keys are distinct from libp2p identity keys, this is
+likely to be the case for the first Noise connection between two peers.
+
+Following a successful `XX` handshake, both peers may store the static Noise
+public key of the other. Future connections can attempt the `IK` handshake using
+the stored static key, which offers the benefits of zero-RTT encryption and
+fewer message round-trips.
```
marten-seemann: Am I right to assume that the security properties of zero-RTT data are equivalent to those of 0-RTT data in TLS1.3, i.e. can be replayed?

noise/README.md
```diff
@@ -0,0 +1,400 @@
...
+    different Noise protocols, provided the same hash algorithm is used in all of
+    them. (Reusing a Noise static key pair outside of Noise would require extremely
+    careful analysis to ensure the uses don't compromise each other, and security
+    proofs are preserved).
+
+All libp2p peers possess a cryptographic keypair which is used to [derive their
+peer id][peer-id-spec], which we will refer to as their "identity keypair." To
+avoid potential static key reuse, and to allow libp2p peers with any type of
+identity keypair to use Noise, noise-libp2p uses a separate static keypair for
+Noise that is distinct from the peer's identity keypair.
```
zmanian: Here is that way I interpret the relationship between the handshake static DH keypair and the libp2p signing keypair.

For each signing identity key, there will be at least be one or more handshake static keys. 

If the application is trying to initiate a connection to a specific identity key, they only determine if they have secure connection to that identity key after a handshake has taken place by receiving a certificate for the static public keys that were used in the connection.

noise/README.md
Kubuxu: @marten-seemann it was mostly about selecting ChaCha/Salsa on hardware that doesn't support it hardware AES. Hardware AES vs Salsa family are very similar in speed.

noise/README.md
```diff
@@ -0,0 +1,400 @@
...
+```
+
+The `pubkey` field contains the public libp2p identity key, encoded as described
+in the [peer id spec][peer-id-spec].
+
+The `signature` field contains a signature of the Noise static public key,
+produced by the private libp2p identity key. The Noise static key is encoded
+into a byte array according to the rules defined in [section 5 of RFC
+7748][rfc-7748-sec-5] and signed as described in the [peer id
+spec][peer-id-spec].
```
Kubuxu: We should be using domain separation, so the signed message in another domain (i.e. IPNS) for sure cannot be used in here. See https://github.com/libp2p/specs/blob/master/tls/tls.md#libp2p-public-key-extension

noise/README.md
marten-seemann: I remember that the difference was quite large. I did a quick Google search and found this article (unfortunately from 2014): https://www.zeitgeist.se/2014/08/23/optimize-aes-and-chacha20-usage-with-boringssl/. In that benchmark, AES beats ChaCha by a factor of 2, if the CPU has hardware support.

What's the reason for making ChaCha required and AES-GCM optional? Just for reference, that's the opposite of what [TLS1.3 is requiring](https://tools.ietf.org/html/rfc8446#section-9.1).

Warchant:
How about PSK? With PSK it would be possible to create private networks.

noise/README.md
shahankhatch: This is how I understand it as well. However, "signed identity payload" instead of "certificate" are the words I would lean towards. I believe that the "signed identity payload" may then be used for whichever key type and payload content is sent, which could be one or more certificates or other public key forms.

shahankhatch:
Regarding transport and encoding of post-handshake data, I haven't seen much/any reference to the chaining key that is obtained once the handshake completes, and which is used for the transport messages. Is this part of the protocol relevant?
https://noiseprotocol.org/noise.html#overview-of-handshake-state-machine

Kubuxu:
PSK could be a good idea (0 overhead private networks) and protection against the quantum break of forward secrecy.

noise/README.md
```diff
@@ -0,0 +1,400 @@
...
+non-issue.
+
+Using the [supported handshake patterns](#supported-handshake-patterns) and
+[supported DH function, cipher and
+hash](#supported-dh-functions-ciphers-and-hash-functions), the following are
+valid noise-libp2p protocol IDs:
+
+- `/noise/xx/25519/chachapoly/sha256/1.0.0`
+- `/noise/xxfallback/25519/chachapoly/sha256/1.0.0`
+- `/noise/ik/25519/chachapoly/sha256/1.0.0`
```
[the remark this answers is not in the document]
shahankhatch: Can someone please give some details of why the use of path-based protocol identifier is used? To me, the path here is not hierarchical, and also doesn't match the protocol identifier specified in the Noise protocol:
https://noiseprotocol.org/noise.html#protocol-names-and-modifiers
Why not use the Noise-specified protocol identifier after the "noise/" part of the path? The version can be used as the postfix.

raulk:
@shahankhatch good point. @yusefnapora – besides the handshake procedure, I believe this spec should also cover how the agreed key is used for encryption/decryption.

One aspect worth considering is [rekey](https://noiseprotocol.org/noise.html#rekey), but coordinating on this would require a meta-message of some kind on a dedicated stream (à la CRYPTO frame in QUIC). Might not be worth it for v1 of this secure channel.

noise/README.md
yusefnapora: thank you, yes :)

noise/README.md
yusefnapora: I just went with what's currently in rust-libp2p and figured we'd sort it out in discussion. I think we probably should add AESGCM since hardware support is pretty pervasive.

 Is there a reason not to line up with TLS 1.3 and make AES required, with a recommendation to also support ChaCha? Seems fairly sensible.

noise/README.md
yusefnapora: @zmanian yes, I think that sums it up well - I'll try to clarify it some more in the write up.

I like the phrase "signed identity payload", that does seem better than "certificate".

noise/README.md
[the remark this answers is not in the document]
yusefnapora: @shahankhatch I'm fine with using Noise names directly instead of wrangling them into paths like this. Are you thinking e.g. `/noise/IK_25519_ChaChaPoly_SHA256/1.0.0`?

noise/README.md
```diff
@@ -0,0 +1,400 @@
...
+
+noise-libp2p supports a [compound protocol][npf-compound-protocols] consisting
+of two fundamental handshake patterns, `XX` and `IK`. A variation of the `XX`
+pattern, `XXfallback` is used to gracefully recover from a failed `IK`
+handshake attempt.
+
+This pattern of using `XX`, `IK` and `XXfallback` is described in the Noise spec
+as the [Noise Pipes compound protocol][npf-noise-pipes].
+
+The`XX` pattern is used when the initiator (or in libp2p terms, the dialing peer)
```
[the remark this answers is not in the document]
yusefnapora: Yeah, I think you're right that we should encourage static keys to have a long lifetime. How about implementations MUST support saving and loading static Noise keys, and we add a recommendation to only change the Noise key if you suspect it's been compromised?

yusefnapora:
Thanks for all the great feedback so far everyone.

Here's what I'll plan to focus on for today / tomorrow:

- add a section on PSK / private networks
- add AESGCM as a required DH function.
  - maybe make ChaChaPoly optional?
- clarify identity payload section & add domain separation for signature.
- more detail about transport encryption keys, message framing, and encryption / decryption process
- write up user-facing API

@shahankhatch about the chaining key, I think that it gets deleted after the handshake, since it's part of the [SymmetricState object](https://noiseprotocol.org/noise.html#processing-rules). The two CipherState objects are what we keep around and use for encryption of transport messages. There's also a handshake hash for [channel binding](https://noiseprotocol.org/noise.html#channel-binding) after the handshake completes, which could be interesting.

@raulk 
> Consider adding support for the IX pattern. XX avoids dumping the initiator's public key in plaintext (initiator privacy), but it adds 0.5 RTT in contrast to IX.

I think this is a good idea. Users / applications that don't care about initiator privacy can skip the whole Noise Pipes dance entirely and just use IX. I'll also add an overview of the security and identity hiding properties of each handshake pattern, so it's more clear what the tradeoffs are.

noise/README.md
yusefnapora: @marten-seemann I think you're right that the IK handshake is susceptible to replay. It has this destination [security property](https://noiseprotocol.org/noise.html#payload-security-properties):

> Encryption to a known recipient, forward secrecy for sender compromise only, vulnerable to replay. This payload is encrypted based only on DHs involving the recipient's static key pair. If the recipient's static private key is compromised, even at a later date, this payload can be decrypted. This message can also be replayed, since there's no ephemeral contribution from the recipient.

marten-seemann: Thanks @yusefnapora. I think we should document these security properties carefully.
This is something that an implementation will need to account for in the API in one way or the other.

marten-seemann:
> One aspect worth considering is rekeying, but coordinating on this could require a meta-message of some kind on a dedicated stream (à la CRYPTO frame in QUIC). Might not be worth it for v1 of this secure channel.

Most AEADs should be safe up to several TB of data, according to today's research. However, we never know how a connection is used, and  further research will lower this bound in the future. 

> EDIT -- continuous rekeying is an option, but could be expensive depending on the cipher suite.

I benchmarked this when implementing it for QUIC. The cost of computing the new (AES) key is about 20x the cost of opening / sealing a full-size packet. So it should be negligible as long as you don't do it for every single message you send.

yusefnapora:
Quick RFC: I'm not sure I've really thought through the replay attack scenarios, or if I'm thinking about it correctly. In the bit about the signed identity payload, I wrote

> Encrypting the handshake payload is required to avoid replay attacks, as there is no timestamp or other validity criteria in the payload itself apart from the signature.

But after some more reading of the Noise spec, I think that you are only protected against replay attacks if you've performed a DH with an ephemeral key belonging to the recipient.

The IK handshake's first message is "vulnerable to replay" according to the spec, since there's no ephemeral DH contributing to the encryption. And in IX, there's no encryption of the first message at all, so the initiator's identity payload would have to be sent in the clear as well.

How much of a problem is this? Is it acceptable to have the identity payload in plaintext for the IX handshake? There's no identity hiding in that handshake pattern anyway, although it seems like if you compromised the Noise static key you could impersonate a node without needing their libp2p identity key by replaying a handshake message.

One way to add a bit of extra assurance that you actually possess the libp2p identity key would be to sign the handshake hash with the identity key after the handshake completes, as described in the [channel binding section of the noise spec](https://noiseprotocol.org/noise.html#channel-binding). We could then require the token to be present in each encrypted transport message.

Other thoughts?

noise/README.md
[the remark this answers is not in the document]
shahankhatch: The Protocol Names and Modifiers section of the Noise protocol specification (linked above) says that the protocol name is something prefixed with "Noise_", like "Noise_IK_25519_ChaChaPoly_SHA256".
I think it's fine to have a path like `/noise/Noise_IX_25519_ChaChaPoly_SHA256/1.0.0` even if `Noise_` seems repetitive since it meets libp2p and Noise specs.

Kubuxu:
> We could then require the token to be present in each encrypted transport message.

The signature of the token could be sent just once.

noise/README.md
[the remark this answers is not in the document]
arnetheduck: mandating it seems overly restrictive and unenforceable (on the wire). a strong recommendation? since the advantage is obvious it seems that clients have little incentive to avoid storing on a whim - but those with a legit reason (tailored for read-only devices for example) get locked out.

noise/README.md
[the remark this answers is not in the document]
arnetheduck: In the #eth2 protocol we've decided not to go with semver for protocols due to the difficulties of interpreting what patch/minor upgrades mean and how clients should actually behave when they receive a not fully matching version - except maybe that if a client supports two versions, it might prioritise the newer protocol. use a single digit here also to avoid this confusion?

noise/README.md
raulk: Let’s be careful not to replicate the Noise spec here, but rather reference it.

raulk: Note that lack of encryption of 0-RTT data or replay ability, in practice, do not pose a threat for us.

Remember this handshake takes place _in the context of a libp2p connection establishment_, 0-RTT data will presumably be used by multiselect 2.0 to preemptively negotiate the multiplexer. We should not expose 0-RTT facilities to the application itself, IMO.

Notes:

1. Replaying the handshake itself is fruitless. Even in the case of IX or IK (1-RTT, both susceptible to replay), because the attacker does not know the initiator's private keys, unless compromised (in which case, it's a different attack).
2. The 0-RTT data we'll presumably send has no side-effects, unlike in TLS 1.3 with HTTP, where you _could_ be sending an HTTP POST.

noise/README.md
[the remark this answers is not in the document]
raulk: A strong SHOULD with an overview of the implications of not complying would work for me.

noise/README.md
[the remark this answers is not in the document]
raulk: @arnetheduck we haven't settled on the change yet in the libp2p community. The point has come up in several discussions, but it was lacking a dedicated one (I think). Until now. https://github.com/libp2p/specs/issues/203.

noise/README.md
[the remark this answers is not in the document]
romanb: I think the choice of lifetime and storage of static DH keys should generally be left to the implementation and ultimately the concrete application. Given that in the context of libp2p a peer already has to (securely) store the identity keypair, I don't think that additionally persisting the DH static keypair somewhere is strictly advantageous, considering that it increases exposure of that keypair. In this context, only keeping static DH keypairs around in-memory, e.g. for the lifetime of a process, seems quite attractive. I am thus in favor of a wording that states that support and use of the Noise Pipes compound protocol (and support for `XXfallback`) is strictly optional but may be recommended for increased efficiency. If necessary for protocol negotiation, support for `XXfallback` could be bound to support for `IK`, so that an implementation must either support both of these or none, leaving support for `XX` and possibly `IX` as mandatory.

Kubuxu:
Regarding rekeying: In case of Salsa family, it is not needed, you can transfer up to: `2^(64+6) bytes = 1ZiB` over a stream without rekeying. I'm not sure about AES, it very much depends on the scheme used for it.

noise/README.md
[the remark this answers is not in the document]
yusefnapora: @romanb makes a good point that persisting the keys comes with increased exposure. BTW, I wasn't saying that persisted keys MUST be used, just that the API should allow you to use persisted keys... but even that seems a bit too prescriptive.

I think you're right that making Noise Pipes optional is a good call; it's a fair bit of complexity for something that not everyone will want. If you're fine with `IX`, for example, there's no efficiency benefit to using Noise Pipes, since you get 1-RTT every time with IX.

I'd be fine with removing Noise Pipes (and `XXfallback`) from the spec for this iteration, actually. It would simplify things quite a bit, and we could mention in the design consideration section that it's a possible future enhancement. For now I'll put it into its own section and describe it as an optional optimization.

noise/README.md
Kubuxu: From my runs on i5-4200M (only one I have available right now), I'mne  getting:
```
BenchmarkCacha20Poly1305-4   	    3000	    460213 ns/op	1139.23 MB/s
BenchmarkAES128GCM-4         	    5000	    301244 ns/op	1740.41 MB/s
BenchmarkAES256GCM-4         	    5000	    347137 ns/op	1510.32 MB/s
```

Here is Go's benchmark: https://github.com/Kubuxu/go-crypto-bench/
It would best to compare with more powerful Intel and AMD processor and also some ARMs.

yusefnapora:
@Kubuxu good call on just sending the token once. 

Here's a thought... if we end up sending the channel binding token in the first transport message because the signed identity payloads are replay-able (or in plaintext), why not just send the identity payload in the first transport message instead of the handshake and include the channel binding token in the payload?

The problem with that is that it requires both sides to send a transport message before the connection is authenticated. But if we're don't trust the connection until we've received the channel binding token, that's the case anyway...

Kubuxu:
Hmm, in theory if we are storying a static key for some peer, this static key was proven in the past to correspond to given peer id (either through signing the key or signing the session token to that peer).

So in case of 0-rtt you don't really need reconfirm that session is bound to given peer.

noise/README.md
marten-seemann: @Kubuxu Thanks for the benchmark! I ran it on my Macbook, and it looks like the difference is a lot bigger here:
```
goos: darwin
goarch: amd64
BenchmarkChacha20Poly1305-12    	   30000	    321145 ns/op	1632.56 MB/s
BenchmarkAES128GCM-12           	  100000	    145071 ns/op	3614.00 MB/s
BenchmarkAES256GCM-12           	   50000	    156157 ns/op	3357.44 MB/s
```

raulk:
Thoughts from a sync discussion with @yusefnapora:

1. 0-RTT or message data facilities should NOT be exposed to the application layer. They should stay internal in libp2p. 0-RTT data should be purely informational and should not lead to side effects or mutation.
    * **We plan to use 0-RTT to expedite multiplexer selection.** That way, two peers can establish a capable libp2p connection (encrypted, multiplexed) within a single RTT, by announcing and intersecting supported multiplexers in message data. This may be subject to downgrade attacks, though, but I think it's unavoidable. It's definitely censorable though in IX, but not in XX (plaintext).
    * Unreplayability is not a requirement if data is purely informational.
    * At the very least, implementations MUST ensure that 0-RTT data is signed to prevent MITM.

2. Assuming the motivation for peer A to open a connection to peer B is to perform an RPC, `IX` and `XX` are equivalent in terms of round trips, whilst `XX` offers better security properties. Rationale: `A` can pipeline message 3 of `XX` + the initial message (encrypted).

3. Exposing `IK` separately may be useless. A use case would be bootstrappers or other nodes whose identity we know beforehand, but the Noise static key is separate from the libp2p identity key, and we don't know it beforehand.

4. I'm really _for_ Noise Pipes.

yusefnapora:
Thanks for capturing that raúl!

Talking through the replay scenarios helped clarify my thinking a lot. Like you mention earlier in the thread, replaying the first message doesn't let you complete a handshake without the private keys, in which case the game is over anyway. So long as accepting an initial handshake message has no potential side-effects, replays aren't a big concern.

Using the handshake messages to advertise multiplexers seems really useful, and it suggests that the internal API should be able to accept some "early data" to include in the handshake payloads. We should think about how that ought to look in practice, since it feels pretty intertwined with the multiselect 2 design. We might just want to note the possibility now, and note that the signed handshake payload might be extended in the future to include muxer negotiation data.

noise/README.md
kirushik: I'm not a real cryptographer (and I hope @burges can correct me where I'm wrong), but I have a strong preference of ChaCha-Poly constructions over AES-GCM, for the following reasons:

1. I think we should also at least benchmark Chacha-Poly vs AES on some smartphone CPUs. Those usually don't have AES-NI or comparable hardware acceleration, and I expect the results to be much in favor of Chacha (since it was benchmarked "in software" in your tests as well, and performs at least comparable to AES "in hardware"). Chacha is based on add-rotate-xor (ARX) PRNG, which can be relatively efficiently implemented in most hardware infrastructures.
I would pretty much expect that mobile (or even embedded and IoT) devices should be in scope for libp2p design choices?

2. Continuation of the previous point, we would require AES implementations "in software" which don't suck, and, surprise-surprise, those are rare — there're a lot of potential footguns with timing attacks and AES in general not being very _copiler-_ or _developer-friendly_. It's much easier to achieve constant-time execution for Chacha than for AES, and the implementation will be much easier (taking in account you probably shouldn't trust your compiler for the critical parts anyway, and should roll out your own platform-specific assembly to substitute for that; I would expect ARX in assembly to be much much simpler).

3. AFAIU there're a couple of cryptography-related footguns in AES-GCM as well, so AFAIR NIST doesn't recommend using it for more than 2³² messages with the same key, even if you use cryptographically-strong nonces, and do everything else correctly.
Not sure how much of a practical limitation it is for libp2p, but at least it worth pointing out.

yusefnapora:
I just pushed a big commit that restructures things a fair bit - I ended up ripping things up too much for a tidy commit history, and there are still some missing pieces, but I wanted to put it up and see what people think.

The biggest change is separating out the negotiation of Noise protocols from the rest of libp2p's protocol negotiation stack. multistream-select is still used to decide _whether to use Noise at all_, but the noise-libp2p implementation is in charge of selecting which concrete Noise protocol to use. 

Choosing a Noise protocol happens using the Noise Socket protocol, which is a pretty simple message framing protocol that includes negotiation data. The Noise Pipes 0-RTT with fallback pattern is pretty simple to implement using Noise Sockets, which is nice.

I personally like this approach a bit better than defining a libp2p protocol ID for each Noise protocol and letting multistream-select figure it out. It seems a bit cleaner, and the negotiation data is fed into the Noise prologue to prevent tampering by a MITM. I put some bits of rationale at the end, but I'm very happy to hear counterpoints. It definitely pushes some complexity into the noise-libp2p implementation compared to letting multistream-select figure things out.

Other changes:

- AESGCM is now required, ChaChaPoly is a SHOULD
- the handshake payload can contain an opaque "early data" byte string (signed by the identity key)
- the message framing is more clearly specified, largely thanks to being stolen from Noise Socket
- Noise Pipes is communicated as optional, with the IK and XXfallback patterns dependent on it

I haven't addressed PSK at all yet but have been thinking about it. It seems like you should be able to initialize the module with a PSK at boot, which would convert all the handshake patterns you support into a `psk` variant, probably `psk0`. Anyone attempting a Noise handshake with you would then require the key to decrypt the initial message.

Kubuxu:
It makes sense to use `psk0` patterns. There is also a bit of a problem with how it would work with current private networks implementation.

Private Network in libp2p right now is implemented with an additional layer of encryption with the PSK key (before anything else). This means that initial multiselect for crypto types is already encrypted and inclusion of PSK in Noise is meaningless (additional overhead already exists).

We could transition off the private networks as currently implemented, to having every security protocol to allow the inclusion of PSK in the negotiation/shared secret but until then this path would be dormant.

My recommendation is to spec this out for the future when we transition to implementing PKS on a transport security layer and not before it.

@marten-seemann AFAIK TLS 1.3 allows for the inclusion of PSK, does QUIC also supports it?

burdges:
> AESGCM is now required, ChaChaPoly is a SHOULD

As noted in https://github.com/libp2p/specs/pull/202#discussion_r313470731 I think AES-GCM should be required on platforms with AES-NI, and recommend when some degree of hardware acceleration and timing protections exist, ala https://blog.mozilla.org/security/2017/09/29/improving-aes-gcm-performance/ but recommend against AES-GCM when no hardware or timing protections exist.  And ChaCha-Poly1305 should be required on all platforms.

That said, individual libp2p based networks should set their own policies.  If you only run validator nodes on high end hardware, then sure require AES-GCM.

We should never negotiate about the handshakes or ciphersuites when using Noise. 

TLS 1.3 makes negotiation secure.  We lack the resources and buy-in by cryptographers to do negotiation correctly.  Just use TLS 1.3 anytime you want negotiation. 

Noise provides fast niche handshakes with precise control over when you reveal static keys.  You only profit form Noise when you know exactly what you're doing, which any negotiation breaks.

burdges:
Also, Noise supports certificates, etc. with early lesser protected protocol messages, so one must expose precise control over these messages, and never send data with the wrong message. 

In polkadot's distant future, there are some situations where some node might contact another node, and prove it possesses some credential, likely by winning some ring VRF contest.  In this, the ring VRF would act as a certificate while we'd authenticate the connection with some ring KEX or ring signature, or maybe use another ring implicit certificate.  

We'd select Noise for anything like this because we first select some handshake that mostly reveals handles everything exactly when we like, and then insert the ring blabla into the appropriate early messages.

yusefnapora:
> We should never negotiate about the handshakes or ciphersuites when using Noise.

This would make me pretty happy, tbh, since it would simplify this spec & implementations enormously. 

I feel like there are two motivations for negotiation that we should examine; maybe we can address or dismiss them and get rid of a lot of potential pitfalls. 

The first motivation is runtime flexibility - if we want to support multiple handshake patterns and ciphers we need a way to distinguish between them. That doesn't necessarily mean that we need to negotiate protocols though; we just need to identify them. Instead of allowing the responder to propose a fallback protocol, we could just have them accept / reject. I'm pretty sure we could still adopt Noise Pipes without requiring the ability to negotiate arbitrary protocols by using trial decryption.

The second motivation is "future proofing", or the ability to evolve the protocol after it's been deployed in the wild. There seems to be a trend in modern security protocols against this goal, which I'm pretty sympathetic to. Allowing "upgrades" also allows downgrades and introduces potentially exploitable complexity in a sensitive area. We could take the position that future-proofing is a non-goal; breaking changes to the security protocol _should_ break the deployed network in this mindset.

I think the runtime flexibility motivation is less compelling than future proofing, but the problem with future proofing is that you don't know what the future holds, so there's no guarantee that the complexity you add now will actually pay off. And in the meantime, it has a concrete cost in the form of edge cases and increased attack surface.

I'm not a Real Cryptographer by any means, so I mostly view my role in this process as flushing out discussions like this and writing up the outcome. Despite writing up how negotiation could work, I'm not tied to the idea at all and would be happy to scrap it.

Thanks for promoting that viewpoint @burdges - I think this is a really important point and I'm glad you bring it up. What does everyone think?

kirushik:
I really don't think that having cypher agility is desired in this particular case — it goes against the major principle of robust crypto design, "have a single joint and keep it well-oiled".
Sometimes agility inevitable (like in TLS case, but even there they become pretty aggressive in reducing amount of supported variants in the latest versions) — but if we can get away without it, we definitely should. Otherwise we will end up with _a poor, incomplete and buggy_ reimplementation of TLS — and maybe we should've picked the actual TLS instead then?

This also applies to the future-proofing; can't we just specify this as `noise-v1` or something, hardcode the cypher choice, and then upgrade to something different in `noise-v2` when we feel like it's needed?
Luckily, crypto protocols (decent ones, I mean) are almost never broken suddenly, and instead we observe a gradually shifting state-of-the-art, weakening the security guarantees of the partiucular algorithm (something like [this](https://valerieaurora.org/hash.html)).
This will give us enough time (like, a couple of years) to specify and implement the replacement cyphersuite selection to be deployed in any maintained software.

We can then avoid this weird negotiation-inside-of-negotiation setup, and completely rely on multistream protocol selection (both current and redesigned, no additional requirements on `multiselect2` imposed); our multiaddresses will still be readable (`/ip4/1.2.3.4/tcp/42/noise2` is not any worse than just `/ip4/1.2.3.4/tcp/42/noise`); the resulting code will be much simpler (resulting both in more diverse and less buggy implementations); there will be less real-life problems with cypher selection incompatibilities (I've seen weird problems even in SSH cypher negotiation in real life).

**PS.** I still think that Chacha-Poly should be the cypher to go for the current revision of noise in libp2p; at least its speed shouldn't be considered a limiting factor, at all.
For example, Wireguard uses combination noise and Chacha-poly — and it easily saturates a 1GB link on an average hardware.
And to add another datapoint to my "hardcoded cyphersuites make software easier to implement and deploy", Wireguard is incredibly easy to set up (in contrast to, say, OpenVPN, which supports cypher agility and in general tends to delegate crypto-related choices to the end-user).

yusefnapora:
@kirushik that sounds like a very reasonable take to me - defining a new multiaddr code or libp2p protocol id for infrequent breaking changes to the spec seems acceptable, and removing agility makes things so much simpler.

ChaChaPoly does seem like the cipher with the most "reach", in the sense that a solid implementation should be available on all platforms, whereas AES more-or-less needs hardware support. If we define a single cipher, that seems like the sensible choice.

The more I think about it, the more I dislike the "weird negotiation in negotiation" scheme I proposed :) I felt compelled to write it up because using multistream for Noise protocol negotiation seemed like it wasn't quite right, but maybe that's because we really shouldn't be doing Noise protocol negotiation at all...

yusefnapora:
BTW, just want to quickly thank everyone for being so involved in this thread and helping me "think out loud" about the design. I don't trust myself to do it well by myself, but I do trust _us_ to figure it out. 💯

burdges:
We should support TLS 1.3 as a transport and TLS 1.3 addresses for numerous reasons, including that people might want runtime flexibility and future proofing TLS-style @yusefnapora 

I think Noise should handle both flexibility and future proofing at the public key infrastruture level.  In other words, there is an implicit negotiation of sorts when the code using libp2p first learns some node's public key and decides to ask libp2p to open a connection, because the requesting code must tell libp2p what exactly which Noise handshake and ciphersuite.  

You might fold relevant information into some multi-address format, so any libp2p network changing ciphersuite preferences happens in the defaults their code supplies to the multi-address writer and reader code.  

Yes, we'll witness some networks making mistakes like distributing builds without AES-GCM but later deciding it should be the default, and making the transition too quickly so the nodes without AES-GCM were still running while nodes started decoding multi-address with AES-GCM as the default.  Yet honestly, if they choose Noise over TLS 1.3 then they signed up for being careful about upgrade paths.  

It's also possible the multi-address could express the protocol version for the code invoking libp2p, so that users only need to maintain a mapping from protocol versions and "roles" to noise configurations.  In other words, a versioned-address instead of a multi-address.

Also, all developers understand versioning, so actually the more restrictive form designed and explained via versioning might prove more usable than any fancy magic negotiation schemes.

tomaka:
Since that we agree that being flexible over the cipher suit is a bad idea, and that the cipher suit should be known ahead of time, why aren't we doing the same for the format of the public key by restricting the use of Noise with X25519 keys?

yusefnapora:
My objection to forcing Ed25519 keys isn't really about wanting flexible key types; it's about avoiding static key reuse with other security protocols. If we use the identity key as the static Noise key, I think we also have to tell users that if they use Noise, they can _only_ use Noise and shouldn't allow other protocols like SECIO or TLS to use the same key.

I can't personally think of a scenario where reusing the key with SECIO would be a problem, but I also don't really think I'm qualified to evaluate that and so am taking a very conservative view.

yusefnapora:
BTW, when I said a moment ago

> I think we also have to tell users that if they use Noise, they can only use Noise and shouldn't allow other protocols like SECIO or TLS to use the same key.

Maybe this is okay? If the people that want to use Noise really don't care about using other protocols and are fine with generating new identities if they ever change their mind, then maybe this isn't a problem. In that case, I don't really have a problem with making this spec as "rigid" as possible, including the key type.

If we restrict to Ed25519 keys, we should still send the original public key in the handshake payload, since conversion to X25519 loses the sign bit of the key. But we don't need to send any signatures, which is a big simplification.

raulk:
@tomaka the Noise spec advises _against_ reusing the static key beyond Noise handshaking (e.g. for identity).

I concur with all the above comments that choosing a specific cipher suite and tagging that as noise-libp2p v1 is the way forward.

We can choose Curve25519, however restricting the libp2p host identity key to the same curve on the grounds of reusing it as the Noise static key is discouraged by the Noise spec itself.

So I’d rather follow that advise and link/assert the static key with the (polymorphic) libp2p identity key via a signature during handshake.

raulk:
In terms of message data, the proposal was to NOT make this feature available to the application layer. Rather, it would stay “confined” inside libp2p-land, and we’d use it to advertise and intersect multiplexers. This allows us to complete the entire connection bootstrapping process in a single round trip.

If we want to make it available to the application, we’d need to craft that API very carefully because the details of the Noise handshake are abstracted away, yet the application needs to be made a participant in the process.

I have ideas, but they create complexity which I’m not convinced it’s necessary.

@kirushik @burdges could you walk us through scenarios where you’d use this feature, and how you’d vary your usage depending on the available guarantees?

yusefnapora:
> the Noise spec advises against reusing the static key beyond Noise handshaking (e.g. for identity).

The way you phrase that reminds me that users might well be using their identity keypair for other purposes (e.g. signing pubsub messages, etc) and we can't assume that restricting the security protocol to Noise is enough to be safe. 

So with that in mind I think using the identity key is a non-starter, and we might as well support all identity key types.

yusefnapora:
About restricting the early data to "libp2p-land", since the set of multiplexers should be static, we can have the API accept the early data when you init the noise module. So there won't be an opportunity for any connection-specific data to ever end up in the payload.

If you do change your supported multiplexers at runtime, you'd have to re-init the module, but that seems fine.

yusefnapora:
The latest commit removes explicit negotiation of Noise protocols, and defines a "cipher suite" of 
`25519_ChaChaPoly_SHA256`, and also fills in the libp2p api section and adds some more color around treatment of early handshake data.

I also removed the optional `IX` handshake, since Noise Pipes can provide similar efficiency without sacrificing initiator privacy, and in many cases the final handshake message can be "pipelined" with the initiator's initial transport message, effectively removing the cost of the final 0.5 RTT in XX. Also, without explicit protocol negotiation, Noise Pipes is much simpler to implement without `IX` in the mix, since it's easier to distinguish between handshakes.

Still no PSK - I think we should defer this to the next round, when we promote to Candidate Recommendation. I'll add a "future work" section in a bit to cover this and other things like potential integration with QUIC.

I'll be off on vacation next week, but will be around reading feedback and chiming in.

PR #204, opened 2019-08-14
[Working Draft] Fetch protocol spec

aschmahmann:
This is an initial stab at putting together a spec for the `Fetch` protocol. As explained in the spec this protocol is basically just a simple network request for `Fetch(key) (value, statusCode)`.

The document contains both the answers to questions about a Working Draft and put together the spec outline (as detailed in the [lifecycle document](https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md)).

^ @vyzo @raulk @Stebalien

### DIFF
--- a/fetch/README.md
+++ b/fetch/README.md
@@ -91,14 +91,14 @@ The libp2p protocol ID for this protocol is `/libp2p/fetch/0.0.1`
 The messages in the Fetch protocol use on of the following protobufs (proto3 syntax):
 
 ```
-message RequestLatest {
+message FetchRequest {
 	string identifier = 1;
 }
 
-message RespondLatest {
+message FetchResponse {
 	StatusCode status = 1;
 	enum StatusCode {
-		SUCCESS = 0;
+		OK = 0;
 		NOT_FOUND = 1;
 	}
 	bytes data = 2;
--- a/fetch/README.md
+++ b/fetch/README.md
@@ -100,6 +100,7 @@ message FetchResponse {
 	enum StatusCode {
 		OK = 0;
 		NOT_FOUND = 1;
+		ERROR = 2;
 	}
 	bytes data = 2;
 }
--- a/fetch/README.md
+++ b/fetch/README.md
@@ -26,9 +26,6 @@ and spec status.
 - [Overview](#overview)
 - [Spec Proposal Info](#spec-proposal-info)
 - [Wire protocol](#wire-protocol)
-- [Interfaces](#interfaces)
-- [Implementation Details](#implementation-details)
-- [Removing existing relay protocol](#removing-existing-relay-protocol)
 
 ## Overview
 
@@ -81,7 +78,7 @@ The `go-libp2p-pubsub-router` implementation requires:
  
  `value`: At least as generic as a byte array
  
- `statusCode`: Supports at least `SUCCESS` and `NOT_FOUND`
+ `statusCode`: Supports at least `OK`, `NOT_FOUND`, and `ERROR`
 
 ## Wire protocol
 
--- a/fetch/README.md
+++ b/fetch/README.md
@@ -1,0 +1,134 @@
+# Fetch v0.0.1
+
+> The Fetch protocol is used for performing a direct peer-to-peer key-value lookup 
+
+| Lifecycle Stage | Maturity       | Status | Latest Revision |
+|-----------------|----------------|--------|-----------------|
+| 1A              | Working Draft  | Active | r1, 2019-08-13  |
+
+Authors: [@aschmahmann]
+
+Interest Group: [Insert Your Name Here]
+
+[@aschmahmann]: https://github.com/aschmahmann
+
+See the [lifecycle document][lifecycle-spec] for context about maturity level
+and spec status.
+
+[lifecycle-spec]: https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md
+
+## Implementations
+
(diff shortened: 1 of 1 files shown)


fetch/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+```
+message RequestLatest {
+	string identifier = 1;
+}
+
+message RespondLatest {
+	StatusCode status = 1;
+	enum StatusCode {
+		SUCCESS = 0;
```
vyzo: let's call this `OK`.

fetch/README.md
```diff
@@ -0,0 +1,134 @@
...
+The messages in the Fetch protocol use on of the following protobufs (proto3 syntax):
+
+```
+message RequestLatest {
+	string identifier = 1;
+}
+
+message RespondLatest {
+	StatusCode status = 1;
+	enum StatusCode {
```
vyzo: we also need an error code for internal server errors.

fetch/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Wire protocol
+
+The libp2p protocol ID for this protocol is `/libp2p/fetch/0.0.1`
+
+### Message Format
+The messages in the Fetch protocol use on of the following protobufs (proto3 syntax):
+
+```
+message RequestLatest {
```
vyzo: `RequestLatest` doesn't rhyme with the fetch protocol, we should call the request just `Request` and the response `Response`.

vyzo: Or `FetchRequest` and `FetchResponse`.

fetch/README.md
aschmahmann: Sure

fetch/README.md
aschmahmann: As discussed related to the go implementation PR on this spec (https://github.com/libp2p/go-libp2p-pubsub-router/pull/33#discussion_r312675975) adding new protobuf enums is really easy and it's not clear anyone would care about actually sending errors over the wire (as opposed to something like just resetting the stream).

If you'd really me to put in an unused `ERR` enum though, that's not particularly bothersome.

vyzo: An `ERROR` is useful for debugging, streams may be closed for other reasons, so please add it.

aschmahmann: I don't have serious issues with reserving a potentially unused `ERROR` enum in the spec (although we don't have any elements in the spec for when/how to use it, and I don't think we have a use for that yet).

However, if you want to discuss whether the go implementation should utilize the Error, please check out the comment on the PR I linked and voice your opinion there. Both options were explored and implemented, and there are some thoughts there as to why I settled on the no ERROR approach.

vyzo: This is a more general protocol than just your patch! 
There can be situations where you try to fetch data from a datastore, and it fails, and the appropriate response is to return an `ERROR`.

aschmahmann: There might be. There also might be situations where it's more appropriate to return `[][]byte` then `[]byte`.

Given that this is literally just a lookup you could choose to encode any information you want in the returned `[]byte`, including error information. There's nothing in the protocol, as it currently is, that prevents returning useful errors.

However, if all you want is a reserved enum option, with no detailing what types of errors it's for, we can do that.

Stebalien: Let's add an error code but not use it. We can treat all codes >= 2 to mean "error" so we can add more if we need them.

aschmahmann: Done. There is now `ERROR = 2` in the status code enum. (btw it is also in the go implementation protobuf in case anyone wants to copy the file, but remains unused)

fetch/README.md
```diff
@@ -0,0 +1,135 @@
...
+Should support: `Fetch(key) (value, statusCode)`
+
+However, the level of specificity in the types of the above variables has wiggle room if people are interested.
+The `go-libp2p-pubsub-router` implementation requires:
+ 
+ `key`: At least as generic as a UTF-8 string
+ 
+ `value`: At least as generic as a byte array
+ 
+ `statusCode`: Supports at least `SUCCESS` and `NOT_FOUND`
```
vyzo: it's called `OK` now.

fetch/README.md
```diff
@@ -0,0 +1,135 @@
...
+## Wire protocol
+
+The libp2p protocol ID for this protocol is `/libp2p/fetch/0.0.1`
+
+### Message Format
+The messages in the Fetch protocol use on of the following protobufs (proto3 syntax):
+
+```
+message FetchRequest {
+	string identifier = 1;
```
vyzo: why not call this either `key` or `id`?

fetch/README.md
```diff
@@ -0,0 +1,135 @@
...
+See the [lifecycle document][lifecycle-spec] for context about maturity level
+and spec status.
+
+[lifecycle-spec]: https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md
+
+## Implementations
+
+- [go-libp2p-pubsub-router](https://github.com/libp2p/go-libp2p-pubsub-router/pull/33)
+
+## Table of Contents
```
vyzo: the TOC needs to be consistent with the document.

fetch/README.md
vyzo: and we have `ERROR` as well.

fetch/README.md
vyzo: it's more consistent with the document, but it's a nit really.

PR #207, opened 2019-08-18
demote mDNS spec to Working Draft.

raulk:
A Recommendation has at least TWO implementations demonstrating interoperability.

A Candidate Recommendation has ONE implementation serving as the Reference Implementation.

This mDNS spec currently has no known libp2p implementations.

mDNS support in go-libp2p and js-libp2p are non-compliant because they predate this spec, and this spec introduces breakages.

Hence, as per above, we demote this specification to the maturity stage it belongs to: Working Draft.

### DIFF
--- a/discovery/mdns.md
+++ b/discovery/mdns.md
@@ -4,7 +4,7 @@
 
 | Lifecycle Stage | Maturity       | Status | Latest Revision |
 |-----------------|----------------|--------|-----------------|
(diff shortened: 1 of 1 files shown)


richardschneider:
I've also implemented in .Net (c#) see [peer-talk](https://github.com/richardschneider/peer-talk/blob/master/src/Discovery/MdnsNext.cs)

PR #213, opened 2019-09-13
Change pubsub Message.seqno protobuf type to uint64

Nashatyrev:
This is to enforce the `seqno` field 8-bytes boundary on the protobuf type level 

**CAUTION:** this change is not backward compatible


vyzo:
agreed with @Stebalien

Nashatyrev:
@raulk what do you think?

Issue #194, opened 2019-07-26
Pubsub: Specify how streams should be handled


AgeManning:
Currently go-libp2p uses long-lived streams, whereas rust-libp2p initializes a new stream per message.  These implementations are currently incompatible. 

This issue is to add specification of how streams should be handled in pubsub/gossipsub.

raulk:
See https://github.com/libp2p/specs/issues/199 for more context and explanation.

PR #214, opened 2019-09-23
revise gossipsub spec

yusefnapora:
This is some very early revision & restructuring of the gossipsub spec to try to make it a bit more normative and clarify a few things. So far it's mostly a skeleton - I've been writing in a separate markdown doc and copying things over and tweaking as I go. I've kept most of @vyzo's writing but have made some changes here and there.

@vyzo there's one change that I want to get your opinion on.  I personally like how you build up a progression from floodsub -> randomsub -> meshsub -> gossipsub, but I think it might be confusing for readers looking for a straightforward normative description of what gossipsub is and how to implement it. 

In what I've got so far, I jump directly from floodsub to gossipsub - to me this feels a bit clearer than naming randomsub and meshsub and describing them separately. Is this a change you're alright with?

Also would love any input from @yiannisbot, as I know you've been thinking about gossipsub a lot lately as well.

Things todo:

- [x] fill in the outline (message processing, mesh maintenance, etc)
- [ ] add guidance on connection management (address #194)
- [ ] clarify some terms that may be unfamiliar to people new to pubsub/p2p in general (overlay mesh, fanout, etc)

### DIFF
--- a/pubsub/gossipsub-wip.md
+++ b/pubsub/gossipsub-wip.md
@@ -235,9 +235,11 @@ There are three main pieces of peering state:
   `fanout` map contains topics to which we _are not_ subscribed.
 
 In addition to the gossipsub-specific state listed above, the libp2p pubsub
-implementation maintains some "router-agnostic" state. This includes the set of
+framework maintains some "router-agnostic" state. This includes the set of
 topics to which we are subscribed, as well as the set of topics to which each of
-our peers is subscribed.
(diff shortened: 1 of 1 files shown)


vyzo:
So now we have two specs for gossipsub in the repo; one is the extant one in the README and the other is this one in the gossipsub-wip.md file. What gives? We want just one!

daviddias:
@yusefnapora can you look into coalescing both?
image

Issue #198, opened 2019-07-30
Framework specification: RFC process


raulk:
_RFC = Request For Change._

An RFC is a formal request for something to change. Even though change is arguably a fractal concept (specifying an inexistent subsystem afresh can be construed as change too), this issue is strictly concerned with the process of proposing **an alteration of an existing spec document, recording the change in a traceable, structured, archivable manner so it can be understood decades later.**

For a case study, imagine we want to modify the [peer ID spec](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md) to alter the string representation, as is precisely the case of https://github.com/libp2p/go-libp2p-core/pull/41.

The Request For Change needs to be documented in a form that captures, at least:

* Context/status quo.
* Rationale/motivation for change.
* Proposed change.
* Backwards-compatibility analysis.
* Best-effort evaluation of impact on existing implementations.
* Proposed rollout strategy.

## Options

Other approaches exist, but these are the ones that I want to weigh one against the other now:

1. Submit a PR that modifies the specification document(s), explaining the above points in the PR's description.

    This has the upside of simplicity, but has many downsides, namely:

    * the source of truth and archive thereof is spread across a proprietary technology (GitHub PRs) and the repository.
    * changes in PR descriptions are versioned (they are traceable via GitHub's UI, but there are no guarantees of history preservation, and one can't associate a commit message to a change in a PR description).

2. Create an **RFC template** and a top-level `RFC` folder in this repo to archive documents of this kind, including a numbering system and a naming scheme. 

    Each RFC PR carries (at least) two artifacts:

    1. An instantiation of the RFC template explaining the change.
    2. The changes on the specification(s) themselves that are affected.

    This way, the commit history of this repo will contain, archive and index of history of a spec, as it changed over time.

    A spec document itself can thus be regarded as a **consolidated text** (like in [EU legislation](https://eur-lex.europa.eu/collection/eu-law/consleg.html)) that integrates the initial text + all successive amendments by way of RFCs.

---

I lean towards solution 2.

yusefnapora:
I agree that having the decision record is important, and in general will always pick the solution in which important data is in version control. So I'm also leaning towards option 2. 

However, the exchange of commentary on a PR is often as valuable as the initial description, which is one of the benefits of the first option, since some of the context from the discussion may not make it into the RFC document. Arguably though, being forced to distill the commentary into the RFC itself should end up being a good thing.

jacobheun:
[...]

PR #217, opened 2019-10-04
RFC: Signed Address Records

yusefnapora:
Hey all, this introduces an RFC for a self-certified address record that could be published to a DHT or gossiped about in a pubsub topic without worrying about it being altered in-flight.

This is motivated by the discussion in https://github.com/libp2p/libp2p/issues/47 and https://github.com/libp2p/go-libp2p/issues/436.

I wrote up the record as an IPLD schema, but we could go with another format if we'd rather not have a dependency on IPLD.

I also wasn't sure whether we might want to issue records about peers other than ourselves - this draft does not, but we could have records with distinct `subject` and `issuer` peers.

There are some TODOs at the bottom that are better tracked here:

- [x] how to store signed records 
  - should be separate from "working set" that's optimized for retrieval
  - need to store unaltered bytes
- [x] figure out if IPLD is the way to go here. If not, what serialization format, etc.
- [x] extend identify protocol to include signed records?

@bigs @raulk @mgoelzer @Stebalien @vyzo


mgoelzer:
cc @whyrusleeping

yusefnapora:
What do people think about IPLD? I mostly just thought the schema notation is pretty cool, but I think if we want to use it for real we'd need a block store (at least an in-memory one), which doesn't feel like it's really a libp2p-layer concern. Maybe I'm overthinking that though. 

Other options that I can think of are protobuf, which plays nice with most other libp2p things, or CBOR. Raw CBOR is as annoying to work with as raw JSON, but it does have a canonical encoding that is deterministic, which protobuf lacks. I'm not sure that a deterministic encoding is strictly necessary, and as far as I can tell, IPLD doesn't enforce one either. Anyway, protobuf is likely the simplest to use, so that seems like a decent way to go.

yusefnapora:
I should note that I haven't tried actually using IPLD yet :) I'm going to play with it a bit and see how it works in practice. If anyone has experience with it, how is it?

yusefnapora:
I've been thinking about this some more and just wanted to write up where my head is at before I start revising.

Ultimately I think IPLD is overkill for this, and we should just use protobufs. Earlier I was leaning away from this because there's no guarantees of deterministic encoding for protobufs, which was an issue with the peer id spec. However, as long as we sign, transmit and store _serialized_ protobufs, we can still validate the record before deserializing it, and we don't really have to worry about whether our encoder has the same behavior as the one that produced the record.

I think I want to tease this apart into two RFCs:

1. An all-purpose "signed envelope" record containing a peer id, byte string payload, and a signature of the payload. May optionally include public key (which must be consistent with peer id). Will require a "signature service" that can sign and validate.
2. An address record with routability & confidence scores. This will get serialized to a byte string, and the encoded record will be wrapped in a signed envelope. The peer store will accept these directly, or inside a signed envelope. If inside an envelope, peer store will reject if signature is invalid. If signature is valid, mark the record as certified & surface that via peer store APIs.

This is basically the same idea as the original draft, except instead of an `AddressEnvelope` struct that stores a specific structured data type, the envelope only stores opaque binary blobs.

e.g.:


```protobuf
message SignedEnvelope {
  bytes peerId = 1;
  bytes contents = 2;
  bytes signature = 3;

  optional PublicKey pubkey = 4;
}

message AddressInfo {
  bytes addr = 1;
  Routability routability = 2;
  Confidence confidence = 3;
}

message AddressState {
  bytes subject = 1;
  repeated AddressInfo addresses = 2;
  // etc...
}
```

Storing serialized protobufs as a byte string inside another protobuf feels weird, but it seems necessary, given the lack of a deterministic encoder.

Does this make sense?

Stebalien:
1. Instead of requiring the peer ID and optionally having a key, how about we just always have the key and then always derive the peer ID? That way it's impossible to just assume that the key matches the peer ID?

2. Yeah, a byte string within a protobuf seems like the right tactic if we're using protobufs.

Also note: we need to tag the data with some kind of "purpose" string so we can't reuse these records as messages in other protocols with signed messages. We do this in pubsub by prepending `libp2p-pubsub:` before signing/verifying the signature (but _not_ when serializing) but I wonder if we should have a better solution. Maybe prepend a length-prefixed string? Null terminated string?

yusefnapora:
> how about we just always have the key and then always derive the peer ID

That is better 👍

Good question about the "domain separation" string. Maybe we could have a fixed string, e.g. "libp2p-signed-message:" but also allow the user to add an arbitrary "purpose" string that gets appended to the fixed string. This could be included in plain text as a field in the `SignedEnvelope` message, which might be a nice way to let users indicate how to process the contents of the message.

Something like


```javascript
{
  publicKey: { 
    //...
  },
  contents: Buffer(),
  purpose: "AddressState",
  signature: Buffer()
}
```

And you prepend `"libp2p-signed-message:" ++ purpose` to the contents to get the signature ("libp2p-signed-message:AddressState" in the example).

Stebalien:
> This could be included in plain text as a field in the SignedEnvelope message, which might be a nice way to let users indicate how to process the contents of the message.

That will take extra space in the message itself which may discourage high entropy names.

I actually _like_ forcing users to _explicitly_ specify the purpose/domain. If we include it in the message, users can be lazy and ignore it. If we don't, users will have to call some function like:

```go
func Open(domain string, envelope []byte) (message []byte, author PublicKey, err error) { ... }
```

> And you prepend "libp2p-signed-message:"

I'm not sure if `libp2p-signed-message` is _all_ that important as long as users specify purpose strings. However, I guess it can't hurt.

> Maybe prepend a length-prefixed string? Null terminated string?

BTW, my concern here is that `foo` is a prefix of `foobar` which would allow an attacker to add `bar` to the signed data and pass it off as a `foo` message. This isn't an issue if we require this prefix to end in `:` but it may be better to just length-prefix it.

yusefnapora:
Ah, I see where you're coming from with the out-of-band domain string, that makes sense.

Using only user-controlled data for the domain string makes me nervous, although I'm not sure I can really come up with a plausible attack. Still, there's not really any cost to using the fixed string plus the user-provided one, so might be worth doing just in case.

mgoelzer:
From my experience with security auditors, letting arbitrary or user-defined input control the interpretation of a message is going to raise a yellow flag for them _even if it's not exploitable_.  (A plain text field in the protobuf does not seem as likely to raise those concerns.)  Perhaps something to consider since we've fixed a lot of other stuff that was not technically broken (but is a nonstandard approach) to satisfy auditor concerns.

Stebalien:
This is _less_ of an arbitrary user-defined input than the bytes being signed as the purpose string is fixed at compile time.

My only concerns are:

1. Ideally, we should always combine the purpose with the signed data the same way.
2. We need to _structure_ the signed data such that we can split `$purpose$signedData` into the `$purpose` and $`$ignedData` without external information. Otherwise, as noted above, messages for protocol `foobar` could be passed to protocol `foo`.

`purpose:[data]` makes (1) easy as we _already_ do this in pubsub, TLS, and QUIC (once we cut a release). However, that _does_ mean that `:` can't appear in the purpose.

yusefnapora:
Good point - the "user" that controls the domain string is another libp2p subsystem, and there's no external input at runtime.

I guess I lean towards using a `:` separator and prohibiting `:` in the purpose string, since it's consistent with what we're doing elsewhere.

RFC/0003-address-records.md
```diff
@@ -0,0 +1,267 @@
...
+  GLOBAL = 4;
+
+  // reserved for future use
+  INTERPLANETARY = 100;
+}
+
+
+// Confidence indicates how much we believe in the validity of the
+// address.
+enum Confidence {
```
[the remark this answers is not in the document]
raulk: If it's to signal which dials to prioritise, we can simplify this by conveying a 1-byte precedence value in the range (-128, 128).

raulk:
@yusefnapora aims to have a final draft of incorporating comments by today, in order to be reviewed and merged on Monday, to have a minimal implementation ready by the end of the week in Go.

Stebalien:
> Another extension I'd like to see is a "service bloom filter" that allows peers to advertise their protocols compactly and efficiently (without enumerating them, therefore with a degree of privacy preservation),

Hm, not really. If these bloom filters are _at all_ useful, peers will be able to ask "does peer X likely speak protocol Y". I'd expect an _attacker_ to ask "which peers speak protocol Y" (e.g., trying to find all the IPFS, Filecoin, etc. peers).

However, shrinking the _size_ of the protocol list is likely worth it (but multicodecs!).

Stebalien:
> We should think about "record extensions", where confidence and routability would be two examples of future extensions.

YES!

Stebalien:
> Hoist the sequence number from the peer routing record to the envelope.

Why? Are we planning on _duplicating_ these? I thought the idea was to make the envelope general-purpose.

> Add a type field: not sure if this should be a string, an enumeration, a multicodec.

Multicodec prefixed byte string? That allows:

* /path/thing... (/ is a valid multicodec :grin:)
* Multicodec.
* Cids if we merge https://github.com/multiformats/multicodec/issues/49.

This is also consistent with ENS records.

yusefnapora:
@Stebalien about moving the sequence number to the envelope, we were thinking it might be broadly useful for other kinds of payload, but as I was writing it up, I realized that each system that uses envelopes would need to have their own rules about how to interpret the sequence number (e.g. whether to invalidate & replace old records, merge them, etc). So it may be best to just leave it in the routing record and keep the envelope as simple as we can make it.

For the type field, I think your suggestion makes sense and gives us a lot of flexibility. 👍

Since we're talking about prefixing multiple fields to the content to sign instead of just the domain string, I'm now thinking that length-prefixing the fields is better than using a `:` separator. Especially if we want the type field to contain a CID, since a valid CID might randomly contain the UTF-8 code point for `:`.

yusefnapora:
Just wanted to update this, since I've been a bit quiet this week. I've been working on an initial implementation in Go that I'm hoping to make PRs for today. 

I diverged from what's written up here in one place. It seemed a bit silly to use varints for length-prefixing when building up the buffer to sign, since we don't need to save space on the wire or anything. So I just used uint64, which seems like way more than enough to fit any payload we might want to use envelopes for. I'll update this RFC to reflect that in a second.

Also, how does everyone feel about the name "routing records"? In the implementation I ended up calling them RoutingStateRecords to emphasize that they contain transient state, but I don't know if that's really any better. I'm also wondering if an even more generic name like "peer records" makes sense, since we're thinking about putting things like the protocol bloom filter in here, which isn't strictly a routing concern. Maybe "availability records"? IDK 🚲 🏠 

Anyway, my plan for today is to fix up a few things in the go implementation and push those branches. Then I was thinking I'd merge this RFC in and start new PRs to turn the RFCs into Working Draft specs and move the conversation over to those PRs.

RFC/0002-signed-envelopes.md
```diff
@@ -0,0 +1,79 @@
...
+Since we already have a [protobuf definition for public keys][peer-id-spec], we
+can use protobuf for this as well and easily embed the key in the envelope:
+
+
+```protobuf
+message SignedEnvelope {
+  PublicKey publicKey = 1; // see peer id spec for definition
+  bytes contents = 2;      // payload
+  bytes signature = 3;     // signature of domain string + contents
+}
```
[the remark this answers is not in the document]
Stebalien: This is _basically_ what that was trying to be. However, that was trying to be even more flexible and allow for alternative validator functions.

RFC/0002-signed-envelopes.md
```diff
@@ -0,0 +1,108 @@
...
+- Related RFC: [0003 Address Records][addr-records-rfc]
+
+## Abstract
+
+This RFC proposes a "signed envelope" structure that contains an arbitray byte
+string payload, a signature of the payload, and the public key that can be used
+to verify the signature.
+
+This was spun out of an earlier draft of the [address records
+RFC][addr-records-rfc], since it's generically useful.
```
aschmahmann: It may be worth considering how generically useful this structure is given that the payload must be kept exactly as it is received (instead of allowing it to be deserailized and then reserialized).

If we chose to use a deterministic encoding scheme (e.g. Canonical CBOR or IPLD) instead of Protobufs this would be less of a problem. However, if we'd like to keep using Protobufs then it'd be great to have some documentation letting people know.

Thanks @yusefnapora for the great work putting this together

RFC/0002-signed-envelopes.md
```diff
@@ -0,0 +1,108 @@
...
+## Wire Format
+
+Since we already have a [protobuf definition for public keys][peer-id-spec], we
+can use protobuf for this as well and easily embed the key in the envelope:
+
+
+```protobuf
+message SignedEnvelope {
+  PublicKey publicKey = 1; // see peer id spec for definition
+  bytes typeHint = 2;      // type hint
```
vyzo: shouldn't this be a string?

RFC/0002-signed-envelopes.md
```diff
@@ -0,0 +1,108 @@
...
+
+- The length of the [domain separation string](#domain-separation) string in
+  bytes
+- The domain separation string, encoded as UTF-8
+- The length of the `typeHint` field in bytes
+- The value of the `typeHint` field
+- The length of the `contents` field in bytes
+- The value of the `contents` field
+
+The length values for each field are encoded as 64-bit unsigned integers in
```
vyzo: 64-bit is rather large; we can get away with u32 here and perhaps save a few precious bytes.

RFC/0003-routing-records.md
```diff
@@ -0,0 +1,243 @@
...
+All libp2p peers keep a "peer store", which maps [peer ids][peer-id-spec] to a
+set of known addresses for each peer. When the application layer wants to
+contact a peer, the dialer will pull addresses from the peer store and try to
+initiate a connection on one or more addresses.
+
+Addresses for a peer can come from a variety of sources. If we have already made
+a connection to a peer, the libp2p [identify protocol][identify-spec] will
+inform us of other addresses that they are listening on. We may also discover
+their address by querying the DHT, checking a fixed "bootstrap list", or perhaps
+through a pubsub message or an application-specific protocol.
```
vyzo: probably worth mentioning rendezvous here.

vyzo: also Peer eXchange.

RFC/0003-routing-records.md
```diff
@@ -0,0 +1,243 @@
...
+dial self-certified addresses, so we should include some configuration options
+to control whether non-certified addresses are acceptable.
+
+## Changes to core libp2p protocols
+
+How to publish these to the DHT? Are there backward compatibility issues with
+older unsigned address records? Maybe we just publish these to a different key
+prefix...
+
+Should we update identify and mDNS discovery to use signed records?
```
vyzo: rendezvous needs to be updated to support signed address records as well.

RFC/0002-signed-envelopes.md
raulk: The envelope contains both the byte payload, and the signature over that byte payload. The serialisation scheme is irrelevant at this layer. The recipient of this payload validates that the signature matches the plaintext and the key, *then* deserialises the payload with the serialisation format mandated for the payload type to process it (e.g. to consume the multiaddrs).

If the recipient intends to relay this payload (as is the case of p2p discovery mechanisms), it does not send a re-serialised form, but rather it forwards the original envelope. In general, it's bad and fragile practice to reconstitute a payload in the hope that it'll continue matching the original signature that was annexed to it.

aschmahmann: > The serialisation scheme is irrelevant at this layer....
> In general, it's bad and fragile practice to reconstitute a payload in the hope that it'll continue matching the original signature that was annexed to it.

These two statements are tied together and are following a rule set that you may think is correct, but is not obvious. Not obvious restrictions dictating how the data may be interacted with should be documented. Additionally, this restriction does not have to exist it's just something that's been decided is ok/insufficiently problematic to bother dealing with.

I also disagree with it being "bad" to allow consistent serialization/deserialization of objects. If I have data which I need to propagate frequently and access infrequently then I'll just store the message bytes and deserialize every time I need to access the data. If I frequently propagate and access the data I'll store the data both serialized and deserialized. If however, I infrequently propagate the data and frequently access it I'm now forced to waste space by storing both the serialized and deserialized versions for no reason other than we like Protobufs.

raulk: > These two statements are tied together and are following a rule set that you may think is correct, but is not obvious.

They are not. You are mixing up the concerns of a cryptographic wrapper, with the details of how the inner payload is constructed. These two layers are decoupled, and @yusefnapora has done a good job of modelling that in this spec.

> I also disagree with it being "bad" to allow consistent serialization/deserialization of objects.

That's not what I said.

> If however, I infrequently propagate the data and frequently access it I'm now forced to waste space by storing both the serialized and deserialized versions...

Yes, and it's a cost you assume to preserve the integrity of a signature.

> for no reason other than we like Protobufs.

Incorrect. Systems preserve the original data along with the signature for many reasons including reducing the surface for bugs, traceability/auditability, and others.

I insist it is a terrible idea to assume that, even with canonical serialisation, your system will be perennially capable of reconstituting a payload out of its constituents, in a way that it matches the original signature. Developers introduce bugs, systems change, schemas change, and maintaining such hypothetical logic is error-prone, brittle and convoluted.

RFC/0002-signed-envelopes.md
raulk: Few points:

* Are we intending to use multicodecs? In that case it should be a bytes.

* The term "type hints" tends to be a term used in reflection/dynamic typing. In this case I would just call this field `payload_class`, or `payload_type`, and the actual content `payload`.

* Convention is to use snake case for protobufs, e.g. from the Kubernetes project: https://github.com/kubernetes/kubernetes/blob/master/third_party/protobuf/google/protobuf/descriptor.proto.

RFC/0002-signed-envelopes.md
raulk: Yeah, or even a uvarint, unless we're worried about alignment.

RFC/0003-routing-records.md
raulk: Yes, and mDNS. Linking to the specs would be ideal.

RFC/0002-signed-envelopes.md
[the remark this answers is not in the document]
raulk: 👍

RFC/0002-signed-envelopes.md
aschmahmann: Could you please explain what you think the downsides are of utilizing a format with canonical serialization?

I've already given a use case that would be helped by enabling canonical serialization, a record type that is infrequently propagated but frequently used would benefit from reduced memory and storage consumption.

Are you suggesting that we are **intentionally** using a format that has non-canonical serialization to dissuade other people from making design decisions you think are "terrible"? Are there other reasons you feel using IPLD or Canonical CBOR would be bad?

The point I'm trying to make here is that you seem to think "it's a terrible idea" for people to assume de+re serializing data will keep it identical, I think that in some situations it could be useful. Could you please list some of the negatives of utilizing a canonical serialization format and enabling developers to make their own decisions about whether to rely on its ability to de+re serialize accurately?

raulk: @aschmahmann Woah, that's some remarkable twisting of my words and contrived reading into this discussion. Funny because I'm trying to say precisely the opposite. This format is _generic and not opinionated_, precisely so that "developers can make their own decisions", which may as well be: "we don't need canonical serialisation, and prefer to use a amply supported, performant, well-vetted format like protobuf".

All I've been trying to explain (patiently) is that the spec you are commenting on holds no opinions on how the inner payload gets generated, consumed and used. Its only responsibility is to transmit a signed sequence of bytes. Full stop.

Users that rely on this spec to transmit certified data (e.g. address records) are free to select any serialisation format to satisfy the guarantees/properties that are relevant for them.

Address records – the first data type embeddable in this envelope – require no canonical serialisation guarantees. The whole record is an atomic unit that we store and forward as-is.

aschmahmann: @raulk this "generic and not opinionated" wrapper cannot be used if someone wanted to reference by CID a collection of shared objects and still access them efficiently.

Concrete example. If IPNS was being created today it could easily use one of these signed envelopes to contain its data. However, if I wanted to share over IPFS a set of IPNS records I could not just take the IPNS records and stuff them into an IPLD object without compromising on storing two copies of the envelope.

I'm not saying the above example is common or something we should definitely do, but it shows a use case that your approach blocks. If we have a justification for ignoring this use case (e.g. you think protobuf is a more "amply supported, performant, well-vetted format" then the alternatives that support canonical serialization) then that's fine rationale.

RFC/0002-signed-envelopes.md
```diff
@@ -0,0 +1,108 @@
...
+generated as described below.
+
+## Signature Production / Verification
+
+When signing, a peer will prepare a buffer by concatenating the following:
+
+- The length of the [domain separation string](#domain-separation) string in
+  bytes
+- The domain separation string, encoded as UTF-8
+- The length of the `typeHint` field in bytes
```
Stebalien: I'd feel better if we _also_ signed the key. I'm concerned users will use this in a "commitment" and/or "ticket" protocol as follows:

1. Victim publishes a signed envelop stating "I (unspecified) claim X" to a blockchain.
2. Attacker generates a key such that the key validates the signed envelope.
3. Attacker claims X.

Basically, "an attacker can't make a public key that validates a target signature S" is _not_ a guarantee of public key cryptography. However, it's one that's often assumed.

RFC/0002-signed-envelopes.md
yusefnapora: Thanks for all the discussion :) I definitely feel @aschmahmann about signing structured data that doesn't have a deterministic encoding - it just feels kind of wrong. Serializing to bytes before signing side-steps the issue, but it's also a bit awkward.

My first pass at this did use IPLD, mostly because of this issue of deterministic encoding, and also because I think the IPLD schema DSL is pretty cool. I ended up backing away from that, but I don't think I explained my thought process very well.

IPLD is attractive because you can get deterministic output with the CBOR encoding, but I was hesitant to rely on that, mostly because IPLD is still pretty new. If we start assuming that we can always serialize IPLD to the same bytes, that seems like it kind of limits the future evolution of the IPLD CBOR format. If we ever need to change how IPLD gets serialized to CBOR, any signatures made with the older implementation will be invalid.

The other problem with IPLD is just that we seem to be in the middle of a Cambrian explosion of libp2p implementations, and it seems like a tough ask to make libp2p implementers also implement IPLD.

I don't think either of those arguments really apply to just using plain CBOR & requiring the canonical encoding (sorted map keys, etc). CBOR has broad language support, and the canonical encoding is (hopefully) stable. And of course, if we did use CBOR, you could embed our records into an IPLD graph as-is without having to treat them as opaque blobs, since a valid CBOR map will presumably always be valid IPLD.

Honestly, I ended up going with protobuf instead simply because it seemed easier to define a protobuf schema than to specify the map keys, value types, etc that we'd need to define for a CBOR-based format. Also, since we need to include the public key & there's already a protobuf definition for that. That's mostly just me being lazy though, and I'd rather revisit this now than after we've baked it into a bunch of implementations.

I do like the idea of having a standard way to ship signed byte arrays around, but it's also possible that because I'm focused on this one use case of routing records that it's not actually as generic or broadly useful as I'm hoping. 

You could certainly argue that it would be even more useful to have a standard way of shipping signed structured data around. We could potentially define an envelope as something like

```javascript
{
  publicKey: {
    // cbor map containing key
  },
  contents: {
    // cbor map containing whatever you want
  },
  signature: "byte blob containing sig of canonically-serialized contents map"
}
```

RFC/0002-signed-envelopes.md
```diff
@@ -0,0 +1,108 @@
...
+- The value of the `typeHint` field
+- The length of the `contents` field in bytes
+- The value of the `contents` field
```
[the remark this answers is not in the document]
yusefnapora: Sorting the protobuf fields seems possible but awkward in go. There's a [Properties type](https://godoc.org/github.com/golang/protobuf/proto#Properties) that will give you the field names, but you need to use reflection on the message struct and use the reflected Type of each struct field to get the properties.

RFC/0002-signed-envelopes.md
yusefnapora: +1 on the field naming - I like `payload_type` better than "type hint" & it looks like snake case is the protobuf way.

I did make the type field `bytes` based on @Stebalien's suggestion to use multicodec-prefixed byte strings. Question about that: do we have a go library that validates / parses multicodecs? I found go-multicodec-packed, but it was deprecated a while ago.

RFC/0002-signed-envelopes.md
yusefnapora: Does space matter here? The length fields aren't stored or sent on the wire, just used when preparing the buffer to sign / validate. I guess all that time writing javascript has warped my memory efficiency sensibilities 😆

RFC/0002-signed-envelopes.md
aschmahmann: Not sure if this is what you're looking for, but go-cid has references to the codecs in the multicodec table at https://github.com/ipfs/go-cid/blob/9bb7ea69202c6c9553479eb355ab8a8a97d43a2e/_rsrch/cidiface/enums.go

RFC/0002-signed-envelopes.md
yusefnapora: I realized I didn't address @raulk's point in my last comment

> I insist it is a terrible idea to assume that, even with canonical serialisation, your system will be perennially capable of reconstituting a payload out of its constituents, in a way that it matches the original signature

That's the other reason I "gave up" on IPLD / CBOR and just went with the signed binary blob, although I don't know if I feel as strongly as Raúl does about it. We could potentially try to guard against differences in encoder implementations by having a ton of test vectors, but of course there's no way to guarantee we'd catch everything.

aschmahmann: @yusefnapora thanks for the detailed explanation here. I get IPLD being a big ask here, although it's probably worth thinking about (for the future) if there's a minimal subset of IPLD that it would be useful for libp2p to have access to.

It being easier to implement and wanting to get this shipped are totally reasonable reasons for us to want to go with protobufs. I guess I just wanted to clarify why the decision was made.

Also, I'm not sure if this is what @raulk was trying to explain but after speaking with @Stebalien I see that if a new format came along that wanted to have a consistent hash for a set of envelopes that we could just define the encoding for that new format. It's unfortunate, from a developer perspective, that we'd have to define and implement a canonical protobuf encoding instead of just using a pre-standardized and packaged encoder but it's still achievable within the spec. Given that IPLD defines codecs for each serialization format we import if we're not going with a pre-supported IPLD format then we'd have to define a new codec anyway.

@yusefnapora your suggestion would certainly do the job.

RFC/0002-signed-envelopes.md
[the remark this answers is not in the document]
Stebalien: Sort by tag:

1. Split into raw fields.
2. Sort the fields.
3. Reserialize.

However, this _won't_ handle protobufs within protobufs...

(we can probably punt on this)

RFC/0002-signed-envelopes.md
Stebalien: You know, this is probably more work than necessary.

RFC/0003-routing-records.md
```diff
@@ -0,0 +1,278 @@
...
+2. Peers may not know their own addresses. It's often impossible to automatically
+   infer one's own public address, and peers may need to rely on third party
+   peers to inform them of their observed public addresses.
+3. A peer may inadvertently or maliciously sign an address that they do not
+   control. In other words, a signature isn't a guarantee that a given address is
+   valid.
+4. Some addresses may be ambiguous. For example, addresses on a private subnet
+   are valid within that subnet but are useless on the public internet.
+
+The first point can be addressed by having records contain a sequence number
```
vasco-santos:
```suggestion
The first point can be addressed by having records containing a sequence number
```

Stebalien: "contain" is correct.

RFC/0003-routing-records.md
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  // the peer id of the subject of the record (who these addresses belong to).
+  bytes peer_id = 1;
+  
+  // A monotonically increasing sequence number, used for record ordering.
+  uint64 seq = 2;
+  
+  // All current listen addresses
+  repeated AddressInfo addresses = 4;
```
jacobheun, line 100: go-libp2p has this marked as 3 and it should be.

```suggestion
  repeated AddressInfo addresses = 3;
```

RFC/0002-signed-envelopes.md
```diff
@@ -0,0 +1,112 @@
...
+without invalidating the envelope.
+
+## Wire Format
+
+Since we already have a [protobuf definition for public keys][peer-id-spec], we
+can use protobuf for this as well and easily embed the key in the envelope:
+
+
+```protobuf
+message SignedEnvelope {
```
jacobheun, line 66: `Signed` is redundant, this also brings this inline with the Go language. (although the Go signature index does match, it's 5 instead of 4).
```suggestion
message Envelope {
```

RFC/0002-signed-envelopes.md
```diff
@@ -0,0 +1,112 @@
...
+integers as defined in the [multiformats uvarint spec][uvarint].
+
+Then they will sign the buffer according to the rules in the [peer id
+spec][peer-id-spec] and set the `signature` field accordingly.
+
+To verify, a peer will "inflate" the `public_key` into a domain object that can
+verify signatures, prepare a buffer as above and verify the `signature` field
+against it.
+
+[addr-records-rfc]: ./0003-address-records.md
```
tmakarios, line 109:
```suggestion
[addr-records-rfc]: ./0003-routing-records.md
```
The 0003 file got renamed; the link should point to where the file is now.

jacobheun:
Now that this has been implemented in Go and JS I intend to merge this later today unless there is a reasonable objection. Ideally we would couple the approved RFC in a merge with its spec per https://github.com/libp2p/specs/issues/198, but I'd prefer this PR not sit open, completed, until someone can dedicate time to completing it.

jacobheun (approved):

aarshkshah1992:
@jacobheun Should we go ahead and merge this ?

PR #191, opened 2019-07-22
Add spec on addressing & use of multiaddr

yusefnapora:
This is a spec on addressing I've been chipping away at for a while now. So far it covers how multiaddrs basically work, the rules for DNS resolution, and the address formats for (most) transports.

TODO:

- [ ] add missing transport address formats (webrtc-direct, ws-star, etc)

Also TODO, but probably in a second pass:

- Add section about address discovery and exchange. This could potentially be where we spell out the abstract "peer routing" interface, which would be useful to xref from e.g. the DHT spec.

@Stebalien, the DNS resolution described here aligns with where you're going with this PR, I think: https://github.com/multiformats/go-multiaddr-dns/pull/17

PR #227, opened 2019-11-04
add a multiselect 2.0 spec

marten-seemann:
I tried to condense everything down to make the new protocol as simple as possible, while providing an extension point that will allow us to deploy all of the optimizations that people have already suggested in the future.

The document here describes **only**  the stream-based use case, and doesn't cover the packet-based use case yet. I believe that we can use a very similar format for unreliable connections, but there seem to be some subtle differences that would prevent us from reusing the exact wire format of the `Use` and `Offer` message defined in this document:
* `Offer` can offer multiple protocols. This makes sense when establishing communication on a reliable channel, but it doesn't really work out in an unreliable protocol. For packet-based protocols, it probably makes sense to restrict `Offer` such that it's only possible to offer a single protocol.
* `Use` uses a `oneof` of `name` and `id`. This works well when a stream correlates the `Use` to an `Offer` sent before, but isn't enough in the packet-based use case, where you might have multiple `Offer`s in flight. Here, it would make sense to define a `Use` message such that it includes both a `name` and an `id`.
* A packet-based protocol needs a `Reject` message to communicate that a certain protocol is not supported.


connections/multiselect2.md
```diff
@@ -0,0 +1,103 @@
...
+
+**TODO**: Do we need to describe the format here? I guess we don't, but we will probably need another document for that change, and we can link to it from here.
+
+Peers advertising a multiaddr that includes a handshake protocol MUST support Multiselect 2.0 as described in this document.
+
+#### TCP Simultaneous Open
+
+TCP allows the establishment of a connection if two endpoints start initiating a connection at the same time. This is called TCP Simultaneous Open. For libp2p, this is problematic, since most stream multiplexers assign stream IDs based on the role (client or server) of and endpoint.
+
+It is therefore desirable to fail a connection attempt as early as possible, if a TCP Simultaneous Open occurs. TLS 1.3 and Noise provide this guarantee: For example, the TLS handshake will fail if an endpoint receives a ClientHello instead of a ServerHello as a response to its ClientHello.
```
yusefnapora: @vyzo, is it true that we want to fail the connection completely if a TCP simultaneous open occurs? I remember you were thinking though how to do NAT hole-punching with TCP and that simultaneous open was part of that story. 

It seems like for hole punching, we want one of the simultaneous connection attempts to succeed, but we need a deterministic way to figure out which party has the initiator role. I don't know how that fits in to this multiselect design, since multiselect now only kicks in _after_  the crypto handshake, and it seems like TLS and Noise would both just fail the handshake if both parties believe they are the initiator.

marten-seemann:
@vyzo If I understand #196 correctly, it works by adding (yet) one more round trip to multistream/1.0. I'd like to avoid adding round trips to multiselect 2.0 in the common case (where no simultaneous open is needed).

What do you think of this strawman proposal: If a handshake fails due to an *unexpected message* error (e.g. in the case of TLS, when the client receives a ClientHello in response to a ClientHello), both peers clear their handshake state and start over again. The roles are determined by the following rule: Both peers calculate the hash of the message(s) they sent and they received during the last handshake attempt. The peer that sent the message that hashes to the smaller value will then act as a client, the other one as the server in the subsequent handshake.

A protocol along those lines would make sure that peers that don't use simultaneous connect don't suffer any round-trip penalty, while peers that do simultaneous connect would pay the cost of a single additional round-trip.

vyzo:
#196 does not add a round-trip to the protocol; it is pipelined together with the protocol selection.

marten-seemann:
> #196 does not add a round-trip to the protocol; it is pipelined together with the protocol selection.

@vyzo The crypto protocol selection, which we're getting rid of here. This negotiation made (makes) us vulnerable to a trivial MITM attack, which was one the motivations to develop a new version of this protocol that fixes this vulnerability (not to mention that removing it also speeds up the handshake by one round-trip).
So we clearly need something different than #196 for this new protocol. I'd really like to hear what you think of https://github.com/libp2p/specs/pull/227#issuecomment-549913845. Is that something we can build on to make it a workable specification for simultaneous open?

vyzo:
@marten-seemann the strawman proposal is a hack, the TLS handshake can fail for all sorts of reasons, and the "unexpected message error" is not necessarily indicative of an unexpected "ClientHello".

Why can't we add an `iamclient` bit to the opening handshake?
If both parties set it, then they enter an initiator selection round, similar to #196 and after deciding who is the initiator they can proceed with TLS.

marten-seemann:
@vyzo The problem with a `iamclient` message is that it's sent in cleartext, before the handshake even started, and it would make libp2p traffic as easily identifiable as it is today, which was a non-goal in the design of the multiselect 2 (see #205):
> Current libp2p traffic can be blocked by using deep packet inspection. Handshakes should be indistinguishable from ordinary HTTP/2 (in the case of TCP) and HTTP/3 (in the case of QUIC) traffic, or other popular Internet applications.

connections/multiselect2.md
```diff
@@ -0,0 +1,103 @@
...
+2. Zero-rountrip stream multiplexer negotiation for handshake protocols that take advantage of early data mechanisms (one-roundtrip negotation for protocols / implementations that don't).
+3. Compression for the protocol identifiers of frequently used protocols.
+
+By using protobufs for all control messages, Multiselect 2.0 provides an easy path for future protocol upgrades. The protobuf format guarantees that unknown field in a message will be skipped, thus future version of the protocol can add new fields that signal support for new protocol features.
+
+## High-Level Overview
+
+### Handshake Protocol Selection
+
+Handshake protocols are not being negotiated, but are announced in the peers' multiaddrs.
```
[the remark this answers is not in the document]
marten-seemann: What is that corner case you're referring to? I'm not sure if we should even build that flexibility into our stack, even if we could come up with a situation where a negotiation wouldn't be susceptible to a MITM attack. If it's not there, it can't be exploited.

connections/multiselect2.md
raulk: One of the requirements of ms2.0 is to be deep-packet inspection censorship resistant. This means not exchanging easily recognisable traffic in plaintext (e.g. secure channel negotiation, or tokens like `iamclient`).

We now advertise secure channels in the multiaddr, and the initiator jumps right into the handshake.

IMO we need a conflict resolution mechanism that sits above multiselect 2.0, detects TCP simultaneous connects, and resolves the conflict via some deterministic or probabilistic heuristic.

An example of a deterministic heuristic could be:

> If we are the initiator and we receive an invalid payload from the responder (plus points if we can detect it was a ClientHello – TLS 1.3 – or the wrong Noise message), both sides hash the outgoing bytes and incoming bytes exchanged until then, and the highest value becomes the initiator. The handshake is tried again with each side taking their new role.

That scheme would require us to "tee" the `io.Reader` and `io.Writer` until the handshake is over, which is probably fine.

An example of a probabilistic heuristic could be:

> If we are the initiator and receive an invalid payload from the responder, we roll a dice to determine if we're initiator or responder, and try again, up to a maximum of 3 times.

connections/multiselect2.md
```diff
@@ -0,0 +1,103 @@
...
+
+It is therefore desirable to fail a connection attempt as early as possible, if a TCP Simultaneous Open occurs. TLS 1.3 and Noise provide this guarantee: For example, the TLS handshake will fail if an endpoint receives a ClientHello instead of a ServerHello as a response to its ClientHello.
+
+Since secio doesn't provide this property, secio cannot be used with Multiselect 2.0.
+
+### Stream Multiplexer Selection
+
+This section only applies if Multiselect 2 is run over a transport that is not natively multipexed. Transports that provide stream multiplexing on the transport layer (e.g. QUIC) don't need to do anything described in this section.
+
+Some handshake protocols (TLS 1.3, some variants of Noise (**TODO**: specify which)) support sending of *Early Data*. Early Data can be sent by the server after receiving the first handshake message from the client. It is encrypted, however, at that point of the handshake the client's identity is not yet verified.
```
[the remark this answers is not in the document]
raulk: Clarification: all Noise messages can carry application payloads. The security guarantees of the payload vary depending on the stage of the handshake we send it in.

connections/multiselect2.md
```diff
@@ -0,0 +1,103 @@
...
+
+**TODO**: Do we need to define a way to send an error code / error string? Or do we have something like that in libp2p already?
+
+![](handshake.png)
+
+Handshake protocols (or implementations of handshake protocols) that don't support sending of Early Data will have to run the stream multiplexer selection after the handshake completes.
+
+#### 0-RTT
+
+When using 0-RTT session resumption as offered by TLS 1.3 and some variants of Noise (**TODO**: specify which), the endpoints MUST remember the negotiated stream multiplexer used on the original connection. This ensures that the client can send application data in the first flight when resuming a connection.
```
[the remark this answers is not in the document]
raulk: Take into account that multiplexers can change across restarts.

connections/multiselect2.md
```diff
@@ -0,0 +1,103 @@
...
+
+1. Downgrade protection for the security protocol negotiation.
+2. Zero-rountrip stream multiplexer negotiation for handshake protocols that take advantage of early data mechanisms (one-roundtrip negotation for protocols / implementations that don't).
+3. Compression for the protocol identifiers of frequently used protocols.
+
+By using protobufs for all control messages, Multiselect 2.0 provides an easy path for future protocol upgrades. The protobuf format guarantees that unknown field in a message will be skipped, thus future version of the protocol can add new fields that signal support for new protocol features.
+
+## High-Level Overview
+
+### Handshake Protocol Selection
```
raulk: Let's call this Secure Channel Selection, to keep it consistent with the rest of the spec corpus.

connections/multiselect2.md
raulk: Proposal for prose:

> Conversely to multistream-select 1.0, secure channel protocols are not dynamically negotiated in multiselect 2.0. Instead, they are announced upfront in the peer multiaddrs. It is up to the implementers to decide whether each secure channel is exposed over a different port, or if a single port handles all secure channels, and a trial-and-error strategy is used to deserialise and identify which protocol is being used.

marten-seemann:
I just rewrote the Simultaneous Open section and specified how the hashing of handshake messages works to determine the roles in a subsequent handshake attempt. PTAL.

connections/multiselect2.md
```diff
@@ -22,11 +22,14 @@ Peers advertising a multiaddr that includes a handshake protocol MUST support Mu
...
 
-It is therefore desirable to fail a connection attempt as early as possible, if a TCP Simultaneous Open occurs. TLS 1.3 and Noise provide this guarantee: For example, the TLS handshake will fail if an endpoint receives a ClientHello instead of a ServerHello as a response to its ClientHello.
+TLS as well as Noise will fail the handshake if both endpoints act as clients. In case of such a handshake failure, the two endpoints need to restart the handshake. Endpoints MUST NOT close the underlying TCP connection in this case. Implementations SHOULD specifically test for this type of handshake failure, and not treat any handshake failure as a potential Simultaneous Open.
 
-Since secio doesn't provide this property, secio cannot be used with Multiselect 2.0.
+To determine the roles in the second handshake attempt, endpoints calculate the SHA-256 hash of the handshake messages that were sent and received (including any error message(s) that the handshake protocol might have sent) during the failed handshake attempt.
```
vyzo: Are these handshake messages available to the application? It seems to me that these will be internal to the TLS implementation and not actually available to hash.

marten-seemann: You can just wrap the `Read` and `Write` of the `net.Conn`.

vyzo: This is true in go, but what about other languages?
Also, does the "wrap the Read/Write" imply that we hash all the bytes sent/received to make the decision?
It's still now clear how an implementor would implement this part of the spec.

connections/multiselect2.md
```diff
@@ -0,0 +1,106 @@
...
+
+TLS as well as Noise will fail the handshake if both endpoints act as clients. In case of such a handshake failure, the two endpoints need to restart the handshake. Endpoints MUST NOT close the underlying TCP connection in this case. Implementations SHOULD specifically test for this type of handshake failure, and not treat any handshake failure as a potential Simultaneous Open.
+
+To determine the roles in the second handshake attempt, endpoints calculate the SHA-256 hash of the handshake messages that were sent and received (including any error message(s) that the handshake protocol might have sent) during the failed handshake attempt.
+The peer that sent the messages resulting in the numerically smaller hash value acts a client in the second handshake attempt, the peer that sent the messages resulting in the numerically larger hash value acts as a server.
+
+Since secio assign roles during the handshake, it is not possible to detect a Simultaneous Open in this case. Therefore, secio MUST NOT be used with Multiselect 2.0.
```
vyzo: ok, this is not quite right; we need something for secio too, unless we are actively deprecating it.

connections/multiselect2.md
marten-seemann: These are the bytes sent in the clear over the wire. I'm not familiar with the TLS / Noise implementations in other languages, but I'd be surprised if this posed a major challenge in any language.
@tomaka, @yusefnapora Can you shed some light on this?

connections/multiselect2.md
marten-seemann: Which part of it is not right?
We've been talking about deprecating it for a long time, and the main reason we've been sticking to it is for backwards compatibility. Since Multiselect 2.0 is not backwards compatible anyways, this would be a good opportunity to finally phase it out.

connections/multiselect2.md
vyzo: Any language that uses a TLS library that reads/writes to the file descriptor will have problems with this.

connections/multiselect2.md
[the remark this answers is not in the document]
marten-seemann: @raulk That's on us to define. We could say that if you resume a connection, you have to use the same multiplexer. This would save us a roundtrip for negotiation the stream multiplexer in the common case.
In the rare case when a peer dropped support for a muxer between two connections, you would just reject 0-RTT and fall back to the 1-RTT handshake.

connections/multiselect2.md
marten-seemann: Applied with minor rewording.

connections/multiselect2.md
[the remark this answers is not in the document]
marten-seemann: Thanks. I updated the text about Noise, prohibiting the use of unencrypted data for announcing the stream multiplexers, but allowing it as soon as data can be encrypted *somehow* (even if the peer's identity is not verified at that moment). This should give us roughly the same security guarantees as we're getting with TLS 1.3.

connections/multiselect2.md
vyzo: We discussed this with @Stebalien and we both think this (hashing the plaintext messages) is the wrong way to approach it.
We totally don't have to do that: we have the peer IDs for both endpoints here, because we need them for the TLS handshake. 
So what we should do is simply pick the peer with the highest (or lowest) numeric peer id to act as the initiator if the handshake fails with unexpected client hello.

marten-seemann: @vyzo Thank you. Using peer IDs makes a lot of sense to me, and is definitely much simpler than recording messages sent on the wire.
I updated the spec using SHA256(peer ID). I believe that makes numerical comparisons of peer IDs easier, especially if they have different lengths.

marten-seemann: There's an additional detail we need to specify: When receiving the unexpected handshake message, TLS does two things:
1. It sends a TLS alert to the peer, and
2. it aborts the handshake and returns an error message.

If we just restart the handshake when the handshake message is returned, the peer's alert will still be in flight, and typically will be received after we initiated the second handshake, leading to a handshake failure. There are three ways we can solve this:
1. Modify the TLS stack to not send the alert,
2. Filter out the alert before it reaches the TLS stack
3. Wait for the alert to arrive, and then start the second connection attempt afterwards.

I don't think 1. is very practical, since it requires us to modify the TLS stack. 2. would be doable, but requires wrapping of the connection. 3. costs us half a roundtrip, but could be implemented without modifications to the TLS stack or wrapping of the connection.

marten-seemann: Update: I just implemented option 3 in https://github.com/libp2p/go-libp2p-tls/compare/tcp-sim-open-option3, and unfortunately, it's inherently racy: When `tls.Conn().Handshake()` returns, we have no idea if this is before or after the TLS alert from the peer was received. We therefore can't reliably catch the TLS alert afterwards, since it might never arrive.

I think there's an option 4:
4. Filter out the `ClientHello` before it arrives at the TLS stack. If we receive one, abort the current handshake (taking care that no abort alert is sent out), and start anew.

Unfortunately, this option requires wrapping of the TLS connection, but at least it should be race-free. Furthermore, it doesn't cost us any additional (half) round-trips.

marten-seemann: Here's an implementation of option 4, which seems to work: https://github.com/libp2p/go-libp2p-tls/pull/38.
The downside is that we're duplicating a bit of the record layer parsing code of TLS, since we need to make sure that we read the whole `ClientHello` from the connection before we can start the second connection attempt.

vyzo: The concern here is again other language implementations and whether they can do that in reasonable fashion.
Other than that, I am fine with it.

connections/multiselect2.md
bigs: I agree with the push towards deprecation, as seems to be the growing consensus. Multiselect 2.0 presents a convenient opportunity to upgrade our secure channels.

connections/multiselect2.md
```diff
@@ -0,0 +1,108 @@
...
+
+#### TCP Simultaneous Open
+
+TCP allows the establishment of a connection if two endpoints start initiating a connection at the same time. This is called TCP Simultaneous Open. For libp2p, on the one hand, this poses a problem, since most stream multiplexers assign stream IDs based on the role (client or server) of an endpoint. On the other hand, it can be used as a hole punching technique to facilitate NAT traversal.
+
+TLS as well as Noise will fail the handshake if both endpoints act as clients. In case of such a handshake failure, the two endpoints need to restart the handshake. Endpoints MUST NOT close the underlying TCP connection in this case. Implementations SHOULD specifically test for this type of handshake failure, and not treat any handshake failure as a potential Simultaneous Open.
+
+To determine the roles in the second handshake attempt, endpoints compare the SHA-256 hashes of their peer IDs. The peer with the numerically smaller hash value acts as a client in the second handshake attempt, the peer with the numerically larger hash value acts as a server.
```
tomaka: This forces us to transfer our peer ID in the initial message, right?
While this is normally good, it also means we can't for example fool proxies by pretending that our traffic in regular HTTP3.

tomaka: Handling TCP simultaneous connections seems a bit off-topic for multistream-select, and could be specific to each transport and/or encryption protocol.

marten-seemann: When you dial a peer, you already know it’s peer ID. And of course you also know your own peer ID, so we don’t have to send anything extra on the wire.

connections/multiselect2.md
```diff
@@ -0,0 +1,108 @@
...
+
+```protobuf
+# Wraps every message
+message Multiselect {
+    oneof message {
+        Offer offer = 1;
+        Use use = 2;
+    }
+}
+```
```
tomaka: Nit: for readability I'd put the definition of `Offer` and `Use` here and not below.

connections/multiselect2.md
marten-seemann: @tomaka I tend to agree with this being slightly off-topic. The reason it's in here is because ms-2.0 also changes the connection bootstrapping, and #196 won't work any more (at least not if we want to be able to mask our traffic).

What do you think of saying in this document 
> Secure Channels MUST define how to handle TCP simultaneous open, if they can be used over TCP.

and then moving the text here to the TLS document?

connections/multiselect2.md
yusefnapora: I haven't verified this at all, so we shouldn't assume this is correct... but the [docs for node's tls.connect()](https://nodejs.org/api/tls.html#tls_tls_connect_options_callback) show that you can pass in any duplex Stream as the underlying socket which will be used for the connection. 

So it seems like we should be able to wrap a `net.Socket` with some code similar to @marten-seemann's go code which parses the first handshake message and aborts with a custom error if the server sends a `ClientHello`. Then we can do the peer id comparison and retry the connection when we get the special error type.

This won't work in browsers, but I don't think we have a clear path to supporting TLS 1.3 at all there, and we don't have access to raw TCP in browsers anyway, so it's kind of a moot point.

connections/multiselect2.md
tomaka: > When you dial a peer, you already know it’s peer ID. And of course you also know your own peer ID, so we don’t have to send anything extra on the wire.

The receiving side doesn't know the dialer's peer ID, though!

While maybe not very important, there's also the use-case where you just want to connect to a node you know is at a given IP address without bothering fetching its peer ID.

raulk: @marten-seemann 

> When you dial a peer, you already know it’s peer ID. And of course you also know your own peer ID, so we don’t have to send anything extra on the wire.

I may be missing something, but how does the responder learn the peer ID without a hard requirement for the crypto handshake to transmit it on the first message?

raulk: I think there are two different specs here:

* one for connection bootstrapping: crypto channel, multiplexer intersection, connection-scoped feature flags, role-deduplication (TCP simultaneous open is just one instance of a case where such deduplication is needed).
* one for stream-wise protocol negotiation: proper multiselect 2.0 in my head.

The hypothesis is that having a holistic discussion is easier, because this spec is the spiritual successor of multistream-select 1.0 which participates in both, but materially land it as two different specs.

vyzo: It's a simultaneous dial, you have both peer IDs.

connections/multiselect2.md
```diff
@@ -0,0 +1,108 @@
...
+2. Zero-rountrip stream multiplexer negotiation for handshake protocols that take advantage of early data mechanisms (one-roundtrip negotation for protocols / implementations that don't).
+3. Compression for the protocol identifiers of frequently used protocols.
+
+By using protobufs for all control messages, Multiselect 2.0 provides an easy path for future protocol upgrades. The protobuf format guarantees that unknown field in a message will be skipped, thus future version of the protocol can add new fields that signal support for new protocol features.
+
+## High-Level Overview
+
+### Secure Channel Selection
+
+Conversely to multistream-select 1.0, secure channel protocols are not dynamically negotiated in multiselect 2.0. Instead, they are announced upfront in the peer multiaddrs. It is up to the implementers to decide whether each secure channel is exposed over a different port, or if a single port handles all secure channels, and a demultiplexing strategy is used to identify which protocol is being used.
```
raulk:
```suggestion
Conversely to multistream-select 1.0, secure channel protocols are not dynamically negotiated in-band. Instead, they are announced upfront in the peer multiaddrs (<add link to multiaddr spec>). This way, implementations can jump straight into a cryptographic handshake, thus curtailing the possibility of packet-inspection-based censorship and dynamic downgrade attacks.

It is up to the implementers to decide whether each secure channel is exposed over a different port, or if a single port handles all secure channels, and a demultiplexing strategy is used to identify which protocol is being used.
```

connections/multiselect2.md
```diff
@@ -0,0 +1,103 @@
...
+
+By using protobufs for all control messages, Multiselect 2.0 provides an easy path for future protocol upgrades. The protobuf format guarantees that unknown field in a message will be skipped, thus future version of the protocol can add new fields that signal support for new protocol features.
+
+## High-Level Overview
+
+### Handshake Protocol Selection
+
+Handshake protocols are not being negotiated, but are announced in the peers' multiaddrs. 
+
+**TODO**: Do we need to describe the format here? I guess we don't, but we will probably need another document for that change, and we can link to it from here.
```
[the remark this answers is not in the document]
raulk: We need an extension to the multiaddr spec + multicodec table to add secure channels as an atom. Also, @yusefnapora was writing this document to specify the semantics of multiaddrs: https://github.com/libp2p/specs/pull/191.

connections/multiselect2.md
```diff
@@ -0,0 +1,108 @@
...
+
+## High-Level Overview
+
+### Secure Channel Selection
+
+Conversely to multistream-select 1.0, secure channel protocols are not dynamically negotiated in multiselect 2.0. Instead, they are announced upfront in the peer multiaddrs. It is up to the implementers to decide whether each secure channel is exposed over a different port, or if a single port handles all secure channels, and a demultiplexing strategy is used to identify which protocol is being used.
+
+**TODO**: Do we need to describe the format here? I guess we don't, but we will probably need another document for that change, and we can link to it from here.
+
+Peers advertising a multiaddr that includes a handshake protocol MUST support Multiselect 2.0 as described in this document.
```
raulk: I know we had discussed this assumption as a way to simplify things, but I do think it'll end up being short-sighted, and in some ways, a regression compared to multistream-select 1.0, which does announce its version (although admittedly the implementations are not ready to support multiple versions, but the protocol is).

Something I've been thinking about is to create an extension to multistream-select 1.0 / upgraders that would allow us to go straight into a cryptographic handshake, as a way to deliver censorship resistance to downstream users that require it before we realistically ship ms2.0.

connections/multiselect2.md
```diff
@@ -0,0 +1,108 @@
...
+
+#### TCP Simultaneous Open
```
raulk: From the perspective of this spec, we should not talk about specific transports, but rather about "role deduplication" or something like that. The TCP simultaneous open is a specific instance where this deduplication is necessary, and at the connection bootstrapping layer, the transport mechanics are irrelevant.

connections/multiselect2.md
```diff
@@ -0,0 +1,108 @@
...
+### Stream Multiplexer Selection
+
+This section only applies if Multiselect 2 is run over a transport that is not natively multipexed. Transports that provide stream multiplexing on the transport layer (e.g. QUIC) don't need to do anything described in this section.
+
+Some handshake protocols (TLS 1.3, Noise) support sending of *Early Data*. 
+
+In Multiselect 2 the server makes use of Early Data by sending a list of stream multiplexers. This ensures that the client can choose a stream multiplexer as soon as the handshake completes (or fail the connection if it doesn't support any stream multiplexer offered by the server).
+
+When using TLS 1.3, the server can send Early Data after it receives the ClientHello. Early Data is encrypted, but at this point of the handshake the client's identity is not yet verified.
+While Noise in principle allows sending of unencrypted data, endpoints MUST NOT use this to send their list of stream multiplexers. An endpoint MAY send it as soon it is possible to send encrypted data, even if the peers' identity is not verified at that point.
```
raulk: A `MAY` here suggests that multiplexers may never get sent in Noise. How do we handle that case?

connections/multiselect2.md
```diff
@@ -0,0 +1,108 @@
...
+
+Since secio assign roles during the handshake, it is not possible to detect a Simultaneous Open in this case. Therefore, secio MUST NOT be used with Multiselect 2.0.
+
+### Stream Multiplexer Selection
+
+This section only applies if Multiselect 2 is run over a transport that is not natively multipexed. Transports that provide stream multiplexing on the transport layer (e.g. QUIC) don't need to do anything described in this section.
+
+Some handshake protocols (TLS 1.3, Noise) support sending of *Early Data*. 
+
+In Multiselect 2 the server makes use of Early Data by sending a list of stream multiplexers. This ensures that the client can choose a stream multiplexer as soon as the handshake completes (or fail the connection if it doesn't support any stream multiplexer offered by the server).
```
raulk: I'd word similar as follows:

> Both parties share the list of supported stream multiplexers in the earliest handshake message that supports encrypted early data, listed in order of precedence. Both parties intersect the list and choose the highest-ranking shared multiplexer, or abort the connection by sending a `DISCONNECT` message (we need to specify it) if no mutually supported multiplexer exists.

connections/multiselect2.md
```diff
@@ -0,0 +1,108 @@
...
+
+TLS as well as Noise will fail the handshake if both endpoints act as clients. In case of such a handshake failure, the two endpoints need to restart the handshake. Endpoints MUST NOT close the underlying TCP connection in this case. Implementations SHOULD specifically test for this type of handshake failure, and not treat any handshake failure as a potential Simultaneous Open.
+
+To determine the roles in the second handshake attempt, endpoints compare the SHA-256 hashes of their peer IDs. The peer with the numerically smaller hash value acts as a client in the second handshake attempt, the peer with the numerically larger hash value acts as a server.
+
+Since secio assign roles during the handshake, it is not possible to detect a Simultaneous Open in this case. Therefore, secio MUST NOT be used with Multiselect 2.0.
+
+### Stream Multiplexer Selection
```
raulk: I would make this section about the pattern we're adopting:

1. use early message data.
2. share multiplexers in order of precedence, in the first handshake message that supports encrypted early data.
3. intersect the result.

In a separate section, we can norm on the concrete implementation with TLS 1.3 and Noise (and those specs needs to be enhanced accordingly).

Question: can we assume that all handshakes now, and in the future, will support one encrypted early data per peer (A -> B, B -> A)? I don't think we can count on that, so we need a safe fallback.

connections/multiselect2.md
```diff
@@ -0,0 +1,108 @@
...
+When using TLS 1.3, the server can send Early Data after it receives the ClientHello. Early Data is encrypted, but at this point of the handshake the client's identity is not yet verified.
+While Noise in principle allows sending of unencrypted data, endpoints MUST NOT use this to send their list of stream multiplexers. An endpoint MAY send it as soon it is possible to send encrypted data, even if the peers' identity is not verified at that point.
+
+Note that this negotiation scheme allows peers to negotiate a "monoplexed" connection, i.e. a connection that doesn't use any stream multiplexer. Endpoints can offer support for monoplexed connections by offering the `/monoplex` stream multiplexer.
+
+**TODO**: Do we need to define a way to send an error code / error string? Or do we have something like that in libp2p already?
+
+![](handshake.png)
+
+Handshake protocols (or implementations of handshake protocols) that don't support sending of Early Data will have to run the stream multiplexer selection after the handshake completes.
```
raulk: Oh ok, I see that you added the fallback. We need to define how that'll work.

connections/multiselect2.md
```diff
@@ -0,0 +1,108 @@
...
+
+#### 0-RTT
+
+When using 0-RTT session resumption as offered by TLS 1.3 and Noise, the endpoints MUST remember the negotiated stream multiplexer used on the original connection. This ensures that the client can send application data in the first flight when resuming a connection.
```
raulk: 1. Stream multiplexers can change over time.
2. Peers may forget other peers' protocols.

We may need some form of opaque token with which we can verify that our assertions about the other party still remain valid, upon reconnecting. If the other party NACKs, we fall back to full connection bootstrapping.

connections/multiselect2.md
raulk: Yes, we need the definitions and protocol overview before the wire format.

raulk: And can we have the entire schema dumped at once after the overview. It's difficult to reason when types definitions are split across prose.

connections/multiselect2.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Protocol Speficiation
```
raulk:
```suggestion
## Protocol Specification
```

connections/multiselect2.md
marten-seemann: The TLS session ticket is exactly that opaque token.

connections/multiselect2.md
marten-seemann: I'm not sure how this would work without either defining another multicoded or requiring an additional round-trip to negotiate. I'd prefer to keep ms2.0 as simple as possible, which will allow us to ship it faster.

connections/multiselect2.md
raulk: At this layer, you only know you have established a connection and have exchanged some bytes. You *think* you know the other side’s peer ID, but it’s not authenticated. After exchanging those initial bytes, you notice the other party sends you a message you didn’t expect. Does each party work on their assumption of the peer ID of the other party? That leaves the system open to a series of attacks where a peer advertises a very large/small peer ID for itself, then responds to all SYN packets with another SYN + the initiator message, to force a conflict resolution pathway it knows it’ll always win. I much rather introduce a vector of randomness per-session.

connections/multiselect2.md
marten-seemann: @tomaka already raised that point in https://github.com/libp2p/specs/pull/227#discussion_r350673105, and we agreed to move the actual logic to the transport / secure channel specification, while requiring these to define *some way* to deal with TCP sim open in this document.

connections/multiselect2.md
```diff
@@ -0,0 +1,108 @@
...
+
+TLS as well as Noise will fail the handshake if both endpoints act as clients. In case of such a handshake failure, the two endpoints need to restart the handshake. Endpoints MUST NOT close the underlying TCP connection in this case. Implementations SHOULD specifically test for this type of handshake failure, and not treat any handshake failure as a potential Simultaneous Open.
+
+To determine the roles in the second handshake attempt, endpoints compare the SHA-256 hashes of their peer IDs. The peer with the numerically smaller hash value acts as a client in the second handshake attempt, the peer with the numerically larger hash value acts as a server.
+
+Since secio assign roles during the handshake, it is not possible to detect a Simultaneous Open in this case. Therefore, secio MUST NOT be used with Multiselect 2.0.
+
+### Stream Multiplexer Selection
+
+This section only applies if Multiselect 2 is run over a transport that is not natively multipexed. Transports that provide stream multiplexing on the transport layer (e.g. QUIC) don't need to do anything described in this section.
```
jacobheun:
```suggestion
This section only applies if Multiselect 2 is run over a transport that is not natively multiplexed. Transports that provide stream multiplexing on the transport layer (e.g. QUIC) don't need to do anything described in this section.
```

connections/multiselect2.md
marten-seemann: @raulk What's the attack here? Sure, an attacker can go through the hassle you're describing to make sure it ends up in the client role (from the viewpoint of the cryptographic handshake). During the cryptographic handshake both peer validate each other's peer IDs, so it seems to me that the attacker gains nothing from this attack at all.

connections/multiselect2.md
jacobheun:
```suggestion
While Noise in principle allows sending of unencrypted data, endpoints MUST NOT use this to send their list of stream multiplexers. An endpoint MAY send it as soon as it is possible to send encrypted data, even if the peers' identity is not verified at that point.
```

marten-seemann: It's MAY send as soon ..., otherwise it will send it later.

connections/multiselect2.md
marten-seemann: That's a new proposal how stream multiplexer negotiation should work. The spec doesn't define any intersecting of lists, and therefore doesn't need to specify an order of preference.

connections/multiselect2.md
marten-seemann: I think this is pretty much guaranteed from the way a key exchange works, that the server will end up with the keys half a round-trip before the client.

If we want to do the intersection of stream multiplexers you're suggesting, we'll need to specify a new protobuf message that's *only* used for that purpose. I'm not opposed to it, but I remember that @Stebalien had a strong preference for using the same mechanism for all protocol selections (i.e. not making a difference between muxer and application protocol negotiation).

connections/multiselect2.md
```diff
@@ -0,0 +1,108 @@
...
+        oneof protocol {
+            string name = 1;
+            uint64 id = 2;
+        }
+    }
+    repeated Protocol protocols = 1;
+}
+```
+
+The `Use` message is sent in response to the `Offer`. And endpoint MUST treat the receipt of a `Use` message before having sent a `Offer` message on the stream as a connection error.
```
jacobheun:
```suggestion
The `Use` message is sent in response to the `Offer`. An endpoint MUST treat the receipt of a `Use` message before having sent an `Offer` message on the stream as a connection error.
```

connections/multiselect2.md
marten-seemann: I'm not sure if there's anything we need to define. "Early data" is not a separate byte stream, it's the same byte stream as the rest of the connection. The only difference is that the data is sent earlier (and, depending on the handshake protocol, might use a different set of keys).
To the application Early Data and Late (?) Data is not distinguishable at all.

connections/multiselect2.md
```diff
@@ -0,0 +1,108 @@
...
+If none of the protocol(s) listed in the `Offer` message are acceptable, and endpoint resets both the send- and the receive-side of the stream.
```
jacobheun: Should the reset be required?
```suggestion
If none of the protocol(s) listed in the `Offer` message are acceptable, an endpoint MUST reset both the send- and the receive-side of the stream.
```

connections/multiselect2.md
```diff
@@ -0,0 +1,108 @@
...
+
+If an endpoint receives an  `Offer` message that only offers a single protocol, it accepts this protocol by sending an empty `Use` message (i.e. a message that doesn't list any `protocol`), or a `Use` message that assigns a protocol id (see below). Sending an empty `Use` message in response to a`Offer` message that offers multiple protocols is not permitted, and MUST be treated as a connection error by an endpoint.
```
jacobheun:
```suggestion
If an endpoint receives an  `Offer` message that only offers a single protocol, it accepts this protocol by sending an empty `Use` message (i.e. a message that doesn't list any `protocol`), or a `Use` message that assigns a protocol id (see below). Sending an empty `Use` message in response to an `Offer` message that offers multiple protocols is not permitted, and MUST be treated as a connection error by an endpoint.
```

connections/multiselect2.md
```diff
@@ -0,0 +1,108 @@
...
+
+If an endpoint receives a `Offer` message that offers multiple protocols, it chooses an application protocol that it would like to speak on this stream. It informs the peer about its choice by sending its selection in the `protocol` field of the `Use` message.
```
jacobheun:
```suggestion
If an endpoint receives an `Offer` message that offers multiple protocols, it chooses an application protocol that it would like to speak on this stream. It informs the peer about its choice by sending its selection in the `protocol` field of the `Use` message.
```

connections/multiselect2.md
marten-seemann: Agreed. I'll do another editorial pass over the document soon, and will apply this suggestion.

connections/multiselect2.md
jacobheun: There is no path for backwards compatibility here with multistream 1 correct? Pairing the secio deprecation with multiselect 2 is going to put a lot of stress on limited environments (browsers) to upgrade, or segregate them from the rest of the network. I don't think it warrants blocking the spec, especially if providing that compatibility hampers the performance/feature gains, but we should be cognizant and clear of the network rollout time table and its impact on the network.

connections/multiselect2.md
raulk: But multiselect 2.0 doesn't know about concrete secure channels. If there's a handshake-specific token that we can bind to, it needs to percolate up to this layer.

connections/multiselect2.md
raulk: I'm not worried about peers _lying_ about their identity. I'm worried about making the conflict resolution 100% deterministic, and the transitive attack surface that might expose going forward. We're specifying against our future selves, not against the situations we foresee now.

If there's a bug exploitable in this circumstance, an attack can be crafted that works 100% of the time, by precomputing a large/small peer identity.

I admit I may be thinking too far. But if we want to make this conflict resolution probabilistic (which I think is the right way from first principle), when we can make the protocol can send a random nonce and have peers XOR their nonces to calculate who wins, or something like that.

connections/multiselect2.md
marten-seemann: I did another editorial pass over the document and moved the protobuf definitions to a separate section.

connections/multiselect2.md
marten-seemann: See #230.

connections/multiselect2.md
marten-seemann: See #230.

connections/multiselect2.md
marten-seemann: I’m not sure I understand the argument here. The security of our handshake relies on the TLS handshake. If TLS is broken, we don’t gain a lot from reducing the cost of an attack from succeeding in 100% of the cases to 50% of the cases.

connections/multiselect2.md
marten-seemann: Resolved by merging #230.

raulk:
I have not had time to review all the commentary here and the resulting changes. We should move with this rapidly, knowing that this spec is going to continue change. I'm happy to lock in a baseline so we can start a PoC, but not before we see the following. If these are implemented/rejected, please synthesise and explain (reiterating I have no time to go through the commentary and doc, but I have strong ideas):

* Sharing session capabilities during connection bootstrapping. I'd like peers to advertise things like compressors, erasure coding, etc. protocols, that can apply to an entire connection or to individual streams.
  * IMO this addresses the rationale behind @Stebalien's initial recursive/tunneling proposal.
* If peer A has learnt and memorised protocol mappings of peer B in the peerstore, the next time A and B connect, they should be able to verify that the same protocol table still stands. We had discussed incorporating some opaque ID (that is fundamentally different to session resumption, which may or may not be supported by secure channels).
* Not sure if the XOR pathway has been implemented ("I want to open a stream for either protocol Xv1, or Xv2, or Xv3, each accompanied with their initial message; it should be possible to have a single message for multiple protocol proposals; otherwise we'd be wasteful in some scenarios").

marten-seemann:
@raulk 

> Sharing session capabilities during connection bootstrapping. I'd like peers to advertise things like compressors, erasure coding, etc. protocols, that can apply to an entire connection or to individual streams. 

Compressors came up on #230. It was not part of our requirement document, and to be honest, I don't really understand the use case well enough to even write down the requirements for this, let alone a specification.

> If peer A has learnt and memorised protocol mappings of peer B in the peerstore, the next time A and B connect, they should be able to verify that the same protocol table still stands. We had discussed incorporating some opaque ID (that is fundamentally different to session resumption, which may or may not be supported by secure channels).
 
We discussed this, and decided to remove this from the specification. The reason is that the cost of sending the string for one round-trip (and learning a new ID for this protocol) is minor, and doesn't justify the cost of indexing a mapping and verifying that this index is still valid.

> Not sure if the XOR pathway has been implemented ("I want to open a stream for either protocol Xv1, or Xv2, or Xv3, each accompanied with their initial message; it should be possible to have a single message for multiple protocol proposals; otherwise we'd be wasteful in some scenarios").

It's possible to `Offer` multiple protocols in the same message, but it's not possible to send application data for each protocol. Therefore, it will take a single round-trip (once per connection) to learn which protocol version the peer supports. This leads to a much simpler wire format of ms2.0.

raulk:
> We discussed this, and decided to remove this from the specification. The reason is that the cost of sending the string for one round-trip (and learning a new ID for this protocol) is minor, and doesn't justify the cost of indexing a mapping and verifying that this index is still valid.

I disagree here. This assertion can be sent in early data, therefore amortised in terms of segmentation. A normal IPFS peer will need to agree on identify, dht, pubsub, bitswap, etc. Imagine the network is spotty and this peer keeps connecting and reconnecting frequently. Why would we want to renegotiate protocol indices every time, when we can exchange opaque table IDs upfront and just assert that we both maintain the same table?

> It's possible to Offer multiple protocols in the same message, but it's not possible to send application data for each protocol. Therefore, it will take a single round-trip (once per connection) to learn which protocol version the peer supports. This leads to a much simpler wire format of ms2.0.

[This was captured in the design doc.](https://github.com/libp2p/specs/pull/205/files#diff-d08ee123ae67f37b6337d8131c7577ceR162) "A simpler wire format" is not a convincing answer for a feature that's required; of course things are going to be simple if you remove features.

Aren't we talking about an `optional bytes` field? I don't understand where the complication is.

raulk:
> Compressors came up on #230.

Well, you merged that PR without consensus. And now we're having the discussion again, which was predictable ;-)

It's a mistake to make the early data a multiselect 2.0 message. Early data forms part of connection bootstrapping, which is really a process that's decoupled from ms2.0 -- that's the way we're designing it.

Therefore, connection bootstrapping needs its own specific payload, which IMO should include:

```protobuf
enum CapabilityType {
  COMPRESSOR
  ERASURE_CODER
}

message ConnectionBootstrap {
  message SessionCapGroup {
    CapabilityType type = 1;
    repeated string supported = 2;
  }
  fixed32 protocol_table_id = 1;   // an opaque string that identifies the version of the protocol table we are using.
  repeated SessionCapGroup session_capabilities = 2;
}
```

marten-seemann:
> I disagree here. This assertion can be sent in early data, therefore amortised in terms of segmentation. A normal IPFS peer will need to agree on identify, dht, pubsub, bitswap, etc. Imagine the network is spotty and this peer keeps connecting and reconnecting frequently. Why would we want to renegotiate protocol indices every time, when we can exchange opaque table IDs upfront and just assert that we both maintain the same table?

Because the overhead of it is a few bytes, for a single roundtrip. I suggest we first show that this is actually a problem before designing a complicated solution. Synchronizing state across connections is not trivial, and will add significant complexity, both in terms of specification as well as implementation.
**If** we can show that this complexity is actually justified by real-world benefits, the protobuf-based format of ms2.0 will allow us to define a new message, which could (for example) announce the ID(s) of a remembered mapping. Peers that are still on the old ms2.0 will ignore this message, and peers that already support this update can then send a confirmation protobuf. This is a very smooth upgrade path.

> > It's possible to Offer multiple protocols in the same message, but it's not possible to send application data for each protocol. Therefore, it will take a single round-trip (once per connection) to learn which protocol version the peer supports. This leads to a much simpler wire format of ms2.0.
>
> This was captured in the design doc. "A simpler wire format" is not a convincing answer for a feature that's required; of course things are going to be simple if you remove features.

It was marked "tentative" in the design doc, so it's hard to argue that this is a feature that was a hard requirement in the protocol design. [The only comment I received](https://github.com/libp2p/specs/pull/205#issuecomment-544529917) about this tentative point was that it would be complex. Considering that we don't have the concept of protocol upgrades at all in multistream 1 at all, I'm not sure how we know that improving the current situation to a one round-trip per connection wait for protocol upgrades would be prohibitively expensive.

> Aren't we talking about an optional bytes field? I don't understand where the complication is.

I don't think it's that easy. First of all, it breaks with the concept of ms2.0 to use protobufs for protocol messages, but not application data. Furthermore, it's unclear how this interacts with stream flow control. We might also want to think about an API that would generate this kind of data first.

> >    Compressors came up on #230.
>
> Well, you merged that PR without consensus. And now we're having the discussion again, which was predictable ;-)

I merged #230 into this PR, not into master. Consensus would have been nice, but difficult to achieve if so few people actually review the PR. Compressors were unrelated to #230 anyway, so this is the better place to discuss this anyway. Or at least as far as I understand them. This is a totally new requirement to me, I don't know how they're supposed to work, so I have no idea how I should argue about a specific wire format at this point.

connections/multiselect2.md
jacobheun: secio is actively being deprecated and removed from the network, so this no longer needs to be a consideration.

aarshkshah1992: @jacobheun @raulk 

IIUC, even Noise needs an initiator and a responder. So, even after deprecating SecIO, we still need to assign roles.

marten-seemann: I'm happy to update the document and remove any mention of secio, provided there's interest in moving forward with this document. It's been quiet for long time...

PR #234, opened 2019-12-04
noise-libp2p: introduce "handshake seal"; more.

raulk:
_This PR supersedes #210, and implements the outcome of that discussion._

The current construction is vulnerable to replay attacks by a MITM agent that records previous handshakes, and transplants old user data from those handshakes into new ones, as long as the static key hasn't changed.

We fix this vulnerability by leveraging Noise channel binding, in order to "seal" the handshake and assert that the exchanged data was witnessed equally by both parties.

This commit also simplifies protobuf field naming.

Finally, we formalise in which Noise messages of IK and XX the message payload is to be shared, to guarantee secrecy, integrity and authentication.

ping: @shahankhatch @Mikerah @djrtwo @dryajov @mpetrunic @morrigan @araskachoi -- please review too (can't assign you as a reviewer as you're not part of the libp2p org yet).

### DIFF
--- a/noise/README.md
+++ b/noise/README.md
@@ -178,7 +178,7 @@ exposure.
 
 To authenticate the static Noise key used in a handshake, noise-libp2p includes
 a signature of the static Noise public key in a [handshake
-payload](#the-libp2p-signed-handshake-payload). This signature is produced with
+payload](#the-libp2p-handshake-payload). This signature is produced with
 the private libp2p identity key, which proves that the sender was in possession
 of the private identity key at the time the payload was generated.
 
@@ -203,8 +203,8 @@ libp2p components after the handshake is complete and the payload signature has
 been validated. If the handshake fails for any reason, the early data payload
 MUST be discarded immediately.
 
-Any early data provided to noise-libp2p MUST be included in the [signed
-handshake payload](#the-libp2p-signed-handshake-payload) as a byte string
+Any early data provided to noise-libp2p MUST be included in a [handshake 
+payload](#the-libp2p-handshake-payload) as a byte string
 without alteration by the noise-libp2p implementation, and a valid signature of
 the early data MUST be included as described below.
 
@@ -288,8 +288,8 @@ each party performs the following actions:
 3. Writes the result on the encrypted channel, prefixed with a Protobuf varint
    length prefix. We call this payload the "handshake seal".
 4. Awaits to receive the handshake seal from its peer.
-5. Verifies the signature against the libp2p public key of its peer, and its
-   local value of `HandshakeHash`.
+5. Verifies the peer's seal signature against the peer's libp2p public key, and
+   its local value of `HandshakeHash`.
 6. If the signature verification passes, the Noise state machine is destroyed
    and the encrypted channel is handed over to the libp2p stack. If the
    verification fails, the party terminates the network connection immediately.
@@ -333,8 +333,8 @@ to the other party.
 The first handshake message contains the initiator's ephemeral public key, which
 allows subsequent key exchanges and message payloads to be encrypted.
 
-The second and third handshake messages include a [signed handshake
-payload](#the-libp2p-signed-handshake-payload), which contains a signature
+The second and third handshake messages include a 
+[payload](#the-libp2p-handshake-payload), which contains a signature
 authenticating the sender's static Noise key as described in the [Static Key
 Authentication section](#static-key-authentication) and may include other
 internal libp2p data.
@@ -398,9 +398,8 @@ key has changed, they may initiate an [`XXfallback`](#xxfallback) handshake,
 using the ephemeral public key from the failed `IK` handshake message as
 pre-message knowledge.
 
-Each handshake message will include a [libp2p signed handshake
-payload](#the-libp2p-signed-handshake-payload) that identifies the sender and
-authenticates the static Noise key.
+Each handshake message will include a [payload](#the-libp2p-handshake-payload)
+that identifies the sender and authenticates the static Noise key.
 
 #### XXfallback
 
@@ -429,9 +428,8 @@ key is obtained from her initial `IK` message, moving it to the pre-message
 section of the handshake pattern. Essentially, the failed `IK` message serves
 the same role as the first handshake message in the standard `XX` pattern.
 
-Each handshake message will include a [libp2p signed handshake
-payload](#the-libp2p-signed-handshake-payload) that identifies the sender and
-authenticates the static Noise key.
+Each handshake message will include a [payload](#the-libp2p-handshake-payload)
+that identifies the sender and authenticates the static Noise key.
 
 #### Noise Pipes Message Flow
 
--- a/noise/README.md
+++ b/noise/README.md
@@ -4,7 +4,7 @@
 
 | Lifecycle Stage | Maturity      | Status | Latest Revision |
 |-----------------|---------------|--------|-----------------|
-| 1A              | Working Draft | Active | r0, 2019-08-05  |
+| 1A              | Working Draft | Active | r1, 2019-12-04  |
 
 Authors: [@yusefnapora]
 
@@ -23,6 +23,7 @@ Interest Group: [@raulk], [@tomaka], [@romanb], [@shahankhatch], [@Mikerah], [@d
 [@morrigan]: https://github.com/morrigan
 [@araskachoi]: https://github.com/araskachoi
 
+
 See the [lifecycle document][lifecycle-spec] for context about maturity level
 and spec status.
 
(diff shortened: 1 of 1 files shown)


raulk:
@yusefnapora awesome! Let's wait for a few more approvals. @burges -- would appreciate thoughts, if any.

noise/README.md
```diff
@@ -207,56 +208,95 @@ handshake payload](#the-libp2p-signed-handshake-payload) as a byte string
...
+* In IK-initiated handshakes, the initiator will optimistically send its payload
+  in message 1 (as it satisfies the guarantee). Next, this case bifurcates:
+    * If the responder continues the IK handshake, it will send its payload in
+      message 2. The handshake ends.
+    * If the responder fall backs to `XXfallback`, it will have failed to
+      decrypt the payload in message 1. A retransmission from the initiator with
+      the fresh cryptographic material is necessary. This is performed in
+      message 3.
+
+When decrypted, the signed payload has the structure described in [Encrypted
```
romanb: But there is no "signed payload", is there? Only the public identity key within the payload is signed.

noise/README.md
```diff
@@ -207,56 +208,95 @@ handshake payload](#the-libp2p-signed-handshake-payload) as a byte string
...
-validated against the supplied `libp2p_data`. If the signature is invalid, the
-connection MUST be terminated immediately.
+`identity_key` field into a usable form. The key MUST then be used to validate
+the `identity_sig` field against the static Noise key received in the handshake.
+If the signature is invalid, the connection MUST be terminated immediately.
 
 If a noise-libp2p implementation does not expose an API for early data, they
-MUST still validate the signature upon receiving a non-empty `libp2p_data`
-field and abort the connection if it is invalid.
+MUST still validate the signature upon receiving a non-empty `data` field and
```
romanb: What signature? There is no signature on `data`?

noise/README.md
```diff
@@ -207,56 +208,95 @@ handshake payload](#the-libp2p-signed-handshake-payload) as a byte string
...
-noise-libp2p and is expected to be defined in a future iteration of the
-connection establishment spec.
-
-If `libp2p_data` is non-empty, the `libp2p_data_signature` field MUST contain a
-signature produced with the libp2p identity key. The data to be signed is the
-UTF-8 string `noise-libp2p-early-data:` followed by the contents of the
-`libp2p_data` field.
+The `data` field contains the "early data" provided to the Noise module when
+initiating the handshake, if any. The structure of this data is opaque to
+noise-libp2p and is defined in the connection establishment specs.
```
romanb: Do I understand correctly that supporting the sending and receiving of this extra `data` is optional (and thus handshake sealing is optional)? Consequently the only use it may have is for data that is transmitted for some optimisation purposes by other libp2p components but that is never critical for interoperability between different implementations, so if one end sends such data that the other discards, that has no consequence other than the recipient who discarded it missing out on a possible optimisation?

noise/README.md
raulk: Remnants of a previous version. Thanks for catching this.

noise/README.md
raulk: Remnants of a previous version. Thanks for catching this.

noise/README.md
```diff
@@ -207,56 +208,95 @@ handshake payload](#the-libp2p-signed-handshake-payload) as a byte string
 without alteration by the noise-libp2p implementation, and a valid signature of
 the early data MUST be included as described below.
 
-#### The libp2p Signed Handshake Payload
+#### The libp2p Handshake Payload
+
+The Noise Protocol Framework caters for sending early data alongside handshake
```
jacobheun:
```suggestion
The Noise Protocol Framework caters to sending early data alongside handshake
```

noise/README.md
```diff
@@ -207,56 +208,95 @@ handshake payload](#the-libp2p-signed-handshake-payload) as a byte string
...
+messages. We leverage this construct to transmit:
+
+1. the libp2p identity key along with a signature, to authenticate each party to
+   the other.
+2. arbitrary data private to the libp2p stack. This facility is not exposed to
+   userland. Examples of usage include streamlining muxer selection.
+
+These payloads MUST be inserted the first message of the handshake pattern
```
jacobheun: > MUST be inserted the first message

Inserted when? `into`, `after`, `before`?
```suggestion
These payloads MUST be inserted into the first message of the handshake pattern
```

noise/README.md
```diff
@@ -207,56 +208,95 @@ handshake payload](#the-libp2p-signed-handshake-payload) as a byte string
...
+The `body` of the payload contains a serialized [protobuf][protobuf]
+`NoiseHandshakePayload` message with the following schema:
 
 ``` protobuf
-	bytes libp2p_key = 1;
-	bytes noise_static_key_signature = 2;
-    bytes libp2p_data = 3;
-    bytes libp2p_data_signature = 4;
+	bytes identity_key = 1;
```
jacobheun: consistent spacing fix
```suggestion
  bytes identity_key = 1;
```

noise/README.md
```diff
@@ -207,56 +208,95 @@ handshake payload](#the-libp2p-signed-handshake-payload) as a byte string
...
+	bytes identity_sig = 2;
```
jacobheun: consistent spacing fix
```suggestion
  bytes identity_sig = 2;
```

noise/README.md
raulk: We're planning for early handshake data to become key in the new connection establishment flow. We will likely be sharing some form of "connection capabilities" in early data. It's being discussed in the scope of multiselect 2.0:

https://github.com/libp2p/specs/pull/227
https://github.com/libp2p/specs/pull/205

In my head, leveraging early data will be highly encouraged, but ultimately optional, as you say. Not leveraging early data means incurring an extra round trip post-handshake.

Regardless, I would make handshake sealing compulsory for simplicity purposes.

noise/README.md
```diff
@@ -202,61 +203,96 @@ libp2p components after the handshake is complete and the payload signature has
...
+1. Obtains the `HandshakeHash` from the Noise state machine.
+2. Signs the `HandshakeHash` with its libp2p identity key.
+3. Writes the result on the encrypted channel, prefixed with a Protobuf varint
+   length prefix. We call this payload the "handshake seal".
+4. Awaits to receive the handshake seal from its peer.
+5. Verifies the peer's seal signature against the peer's libp2p public key, and
+   its local value of `HandshakeHash`.
+6. If the signature verification passes, the Noise state machine is destroyed
+   and the encrypted channel is handed over to the libp2p stack. If the
+   verification fails, the party terminates the network connection immediately.
```
dshulyak: have you considered noise signatures extension - https://github.com/noiseprotocol/noise_sig_spec/blob/master/output/noise_sig.pdf ? 

it replaces dh used for auth (es, se) with signatures over session hash. for example in XX, after ee is mixed into the symmetric state - responder will sign session hash and append signature with sig-key to an output buffer. initiator performs the same in step 3.

extension specifies support only for ed25519, but i guess you can even do RSA, drop noise static key and reuse libp2p static keypair for signatures. though protocol will slightly deviate from the formed specification :)

noise/README.md
```diff
@@ -202,61 +203,96 @@ libp2p components after the handshake is complete and the payload signature has
...
+
+In the above scheme, the `data` field is not guarded against tampering.
+Consequently, a MITM agent could alter the data in transit, or even record the
+handshake to replay it at a later time with mutated or transplanted data.
+
+An option to counteract such attacks is to sign the data too. However, for
+efficiency and enhanced security purposes, we choose to adopt the [channel
+binding][npf-channel-binding] technique defined in the Noise Protocol Framework.
+
+The Noise Protocol state machine tracks the handshake hash throughout the entire
```
morrigan: Can you be more precise which data we are hashing? Also, do you want to specify this to be SHA256?

raulk: The handshake hash is defined in the Noise spec: https://noiseprotocol.org/noise.html#overview-of-handshake-state-machine. The noise library you use should be computing the hash -- you shouldn't need to do it in the noise-libp2p implementation.

raulk:
🚃🚃🚃🚃🚃 Merge train leaving tomorrow morning. Please get your remarks in today!

romanb:
Could someone elaborate (e.g. with concrete examples) on

> The current construction is vulnerable to replay attacks by a MITM agent that records previous handshakes, and transplants old user data from those handshakes into new ones, as long as the static key hasn't changed.

as well as

> [..] In the above scheme, the data field is not guarded against tampering. Consequently, a MITM agent could alter the data in transit, or even record the handshake to replay it at a later time with mutated or transplanted data.

Given that all "early data" is sent as part of encrypted handshake payloads using an AEAD scheme and the current handshake hash `h` as associated data, how is such transplantation or mutation performed exactly?

I'm probably missing the obvious, but I currently don't see what the signing of the handshake hash is supposed to accomplish on top of signing the static DH public key as far as authentication is concerned. I mean, I can understand why one would rather want to sign the handshake hash _instead of_  the static DH public key (e.g. to prevent signature replays when a static DH private key is compromised, allowing impersonating the owner of the private identity key) but I fail to see why you would want to do both. If someone could enlighten me with more details on the attack vectors addressed here and why both, signing the static DH key as well as the handshake hash is necessary, I would be very grateful.

raulk:
@romanb I think you're right. I missed the fact that `EncryptAndHash` and `DecryptAndHash` also call `MixHash` with the ciphertexts, and that `WriteMessage(payload)` and `ReadMessage(into_payload_buffer)` call `EncryptAndHash` and `DecryptAndHash` respectively.

For some reason, I thought only the keys were mixed in, but I was wrong.

Consequently, the AD (authentication data) already includes the payload ciphertexts we've witnessed, and by introducing this handshake seal we are not proposing any novel authentication that is not already covered by the AEAD scheme that encompasses the payloads.

If an intermediary constructs a transplant or tamper attack, I'm confident it'll be caught by an authentication tag mismatch.

I'm going to retract this RFC, and open a new PR with the other changes.

PR #236, opened 2019-12-09
Add a limit of 8MiB to secio messages size

tomaka:
This is apparently already the case in practice, but we should add it to the specs.

### DIFF
--- a/secio/README.md
+++ b/secio/README.md
@@ -124,7 +124,8 @@ negotiation.
 ### Message framing
 
 All messages sent over the wire are prefixed with the message length in bytes,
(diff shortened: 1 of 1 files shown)


tomaka:
They should be unified to either value. If the Go code sends a 6MiB frame to the JS, it will get refused, and that's an incompatibility.

jacobheun:
> They should be unified to either value. If the Go code sends a 6MiB frame to the JS, it will get refused, and that's an incompatibility.

Agreed. If no one has issues with this, JS can easily make the switch for the upcoming refactor release.

PR #243, opened 2020-01-20
noise spec: fix anchor link

mhchia:


raulk:
Thank you, @mhchia! We should probably use [doctoc](https://www.npmjs.com/package/doctoc) to generate the TOC. WDYT, @yusefnapora?

PR #250, opened 2020-03-02
define how private networks work with QUIC v1

marten-seemann:


mxinden:
@marten-seemann what is the current status of this pull request? Is this awaiting review?

marten-seemann:
We just closed the implementation of this https://github.com/libp2p/go-libp2p-quic-transport/pull/119.
Unfortunately, everything but implementing [RFC 8773](https://tools.ietf.org/html/rfc8773) will be a hack.

mxinden:
Thanks @marten-seemann. With your reply in mind I am assuming that there is no intention to merge this pull request, thus I am closing it. Feel free to re-open in case I am missing something. This could still act as a reference in case we get around to implementing RFC 8773.

PR #253, opened 2020-03-03
identify: make identify delta use a new protobuf message

Stebalien:


Stebalien:
Ah, I thought this hadn't been released yet. You're right, this will require a protocol version bump.

Stebalien:
This ship has sailed.

PR #255, opened 2020-03-16
Add additional restrictions on the TLS certificate

DemiMarie-parity:
* Require the `NamedCurve` encoding of elliptic curve parameters to prevent [“Whose Curve Is It Anyway”](https://whosecurve.org) attacks.
* Forbid the `subjectUniqueId` and `issuerUniqueId` fields in certificates.  These fields are essentially unused, and the Rust implementation rejects certificates that have them.
* Require that optional NULL parameters of RSA-PSS AlgorithmIds be omitted.  This corresponds to a SHOULD in [RFC8017](//tools.ietf.org/html/8017) and simplifies verification.
* Forbid hash algorithms with output length less than 256 bits to prevent collision attacks.


tls/tls.md
```diff
@@ -54,15 +54,19 @@ In order to be able use arbitrary key types, peers don’t use their host key to
...
 
-Endpoints MUST NOT send a certificate chain that contains more than one certificate. The certificate MUST have `NotBefore` and `NotAfter` fields set such that the certificate is valid at the time it is received by the peer. When receiving the certificate chain, an endpoint MUST check these conditions and abort the connection attempt if (a) the presented certificate is not yet valid, OR (b) if it is expired.
+Endpoints MUST NOT send a certificate chain that contains more than one certificate. The certificate MUST have `NotBefore` and `NotAfter` fields set such that the certificate is valid at the time it is received by the peer. When receiving the certificate chain, an endpoint MUST check these conditions and abort the connection attempt if (a) the presented certificate is not yet valid, OR (b) if it is expired. Endpoints MUST abort the connection attempt if more than one certificate is received.
```
Stebalien: @marten-seemann do we do this?

marten-seemann: Yes: https://github.com/libp2p/go-libp2p-tls/blob/6d5ce188a8033d057deb2553572142e9eb685ce8/crypto.go#L105-L108.

tls/tls.md
```diff
@@ -54,15 +54,19 @@ In order to be able use arbitrary key types, peers don’t use their host key to
...
 
-The certificate MUST contain the [libp2p Public Key Extension](#libp2p-public-key-extension). If this extension is missing, endpoints MUST abort the connection attempt. The certificate MAY contain other extensions; implementations MUST ignore extensions with unknown OIDs.
+The certificate MUST contain the [libp2p Public Key Extension](#libp2p-public-key-extension). If this extension is missing, endpoints MUST abort the connection attempt. This extension MAY be marked critical. The certificate MAY contain other extensions; implementations MUST ignore extensions with unknown OIDs. Except as permitted by a future version of this specification, other extensions MUST NOT be critical.
```
marten-seemann: Why? This is unenforceable anyway.

tls/tls.md
```diff
@@ -54,15 +54,19 @@ In order to be able use arbitrary key types, peers don’t use their host key to
...
 
-Note for clients: Since clients complete the TLS handshake immediately after sending the certificate (and the TLS `ClientFinished` message), the handshake will appear as having succeeded before the server had the chance to verify the certificate. In this state, the client can already send application data. If certificate verification fails on the server side, the server will close the connection without processing any data that the client sent.
+Certificates MUST omit the `subjectUniqueId` or `issuerUniqueId` fields in X.509 certificates. Endpoints MAY abort the connection attempt if these are present.
```
marten-seemann: Why?

tls/tls.md
```diff
@@ -54,15 +54,19 @@ In order to be able use arbitrary key types, peers don’t use their host key to
...
+
+Certificates MUST omit the optional ASN.1 NULL parameters in RSA-PSS AlgorithmIds. Endpoints MAY abort the connection attempt if these parameters are not omitted.
 
+Certificates MUST use the `NamedCurve` encoding for elliptic curve parameters. Endpoints MUST abort the connection attempt if is not used. Failure to enforce this restriction allows [“Whose Curve Is It Anyway”](https://whosecurve.com) attacks, which completely compromise the security of the connection. Similarly, hash functions with an output length less than 256 bits MUST NOT be used, due to the possibility of collision attacks.
```
marten-seemann: Re `NamedCurve`: Does Go do that? How would I check?
Re hashes: Which hash functions do you want to forbid? Just using the hash length seems like a crude way to ensure that it's safe.

tls/tls.md
DemiMarie-parity: Both fields are deprecated, and the Rust implementation won’t support them.  Furthermore, CAs are forbidden from issuing certificates containing them.

tls/tls.md
DemiMarie-parity: This is to improve interoperability.  Certificates with unknown critical extensions will be rejected as per the X.509 specification.

tls/tls.md
DemiMarie-parity: Not sure.  If it allows arbitrarily-chosen curves, it has an exploitable security vulnerability.  I mostly want to rule out MD5 and SHA-1, but any hash function with output length less than 256 is vulnerable.  No implementation that I am aware of supports weak hashes with longer outputs.

DemiMarie-parity:
> > Require that optional NULL parameters of RSA-PSS AlgorithmIds be omitted. This corresponds to a SHOULD in RFC8017 and simplifies verification.
> 
> Why do we make this a MUST? I'm not sure how I would implement a check for that in Go.

It’s a MUST on the generation side, not on the use side.  The behavior on the use side is unspecified.  `libp2p-go-tls` doesn’t use RSA-PSS for generating certificates, so it isn’t affected.

tls/tls.md
Stebalien: To be specific, the hash function needs 128 bits of security against collision attacks. Given the birthday paradox, the minimum hash size with this security is 256bits _regardless of the algorithm_.

Note, SHA1 (even MD5) is actually safe in our case. Not that we should _use_ it for compatibility with existing TLS implementations, but the "no SHA1" requirement in browsers is to prevent an attacker from:

1. Creating a certificate signing request for a domain they control with a hash that collides with a domain they _don't_ control.
2. Getting a certificate for the domain they control.
3. Using it for the domain they don't control.

MD5/SHA1 are still preimage resistant.

DemiMarie-parity:
@marten-seemann can you re-review?

DemiMarie-parity:
@marten-seemann ping

Stebalien:
Thank you both! :heart:

PR #257, opened 2020-03-20
Small edits to gossipsub v1.1  & list of new parameters

yusefnapora:
This adds a few typo edits, links and formatting, plus a bit more detail around parameter decay. I also added some clarification that flood publishing applies even if you're subscribed to a topic (hope that's correct @vyzo).

The big thing is a section that lists all the new configurable parameters in one place, which I think will be nice for implementors.


vyzo:
I would put the parameters in the end, I think it's a mistake to start with the parameters when there is no context.

PR #262, opened 2020-04-11
gossipsub v1.1 draft r2

vyzo:
Updates the gossipsub v1.1 spec to document/specify new behaviours introduced and tested in the reference implementation.

### DIFF
--- a/pubsub/gossipsub/gossipsub-v1.1.md
+++ b/pubsub/gossipsub/gossipsub-v1.1.md
@@ -28,7 +28,7 @@ See the [lifecycle document][lifecycle-spec] for context about maturity level an
 
 - [Overview](#overview)
 - [Protocol extensions](#protocol-extensions)
-  - [Direct Peering Agreements](#direct-peering-agreements)
+  - [Explicit Peering Agreements](#explicit-peering-agreements)
   - [Peer Exchange on PRUNE](#peer-exchange-on-prune)
     - [Protobuf](#protobuf)
   - [Flood Publishing](#flood-publishing)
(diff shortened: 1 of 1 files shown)


pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -58,6 +60,22 @@ as the protocol string.
 
 ## Protocol extensions
 
+### Direct Peering Agreements
```
daviddias, line 63:
```suggestion
### Explicit Peering Agreements
```

When preparing the presentation for the Matrix, I found it simpler to explain by calling these "Explicit" to contrast with the "Peering Agreements" that are created through GRAFTing

vyzo: Well, the explicit is slightly different -- they are direct peers in the protocol, hence direct peering agreements.

daviddias: But they are set by the user/app, hence explicit.

All the peering agreements are always between two nodes, hence one could call all of them direct.

vyzo: Precisely! These are agreements between pairs of peers, so they are called direct peering agreements in the protocol.

vyzo: So the contrast is between "Mesh peering agreements" and "direct peering agreements".

vyzo: At any rate I don't feel all that strongly about the wording in the section title; if you insist I will bulge :)

daviddias: we resolved async, discussed the merits of both terms and ended up with "explicit. Next commit updates the spec, including doctoc :)

PR #265, opened 2020-04-28
extended validators for gossipsub v1.1

vyzo:
This adds extended validators to the specification of gossipsub v1.1
see https://github.com/ethereum/eth2.0-specs/issues/1761

### DIFF
--- a/pubsub/gossipsub/gossipsub-v1.1.md
+++ b/pubsub/gossipsub/gossipsub-v1.1.md
@@ -2,7 +2,7 @@
 
 | Lifecycle Stage | Maturity       | Status | Latest Revision |
 |-----------------|----------------|--------|-----------------|
-| 1A              | Draft          | Active | r2, 2020-04-11  |
+| 1A              | Draft          | Active | r3, 2028-04-11  |
 
 
 Authors: [@vyzo]
@@ -45,6 +45,7 @@ See the [lifecycle document][lifecycle-spec] for context about maturity level an
       - [P₄: Invalid Messages](#p%E2%82%84-invalid-messages)
       - [Parameter Decay](#parameter-decay)
     - [Guidelines for Tuning the Scoring Function](#guidelines-for-tuning-the-scoring-function)
+    - [Extended Validators](#extended-validators)
   - [Overview of New Parameters](#overview-of-new-parameters)
   - [Spam Protection Measures](#spam-protection-measures)
 
@@ -469,6 +470,20 @@ if invalidMessageDeliveries < DecayToZero {
 #### Guidelines for Tuning the Scoring Function
 
 `TBD`: We are currently developing multiple types of simulations that will inform us on how to best recommend tunning the Scoring function. We will update this section once that work is complete
+
+#### Extended Validators
+
+The pubsub subsystem incorporates application-specific message validators so that the application can
+signal invalid message delivery, and trigger the P₄ penalty.
+However, It is possible to have circumstances where a message should not be delivered to the application
+or forwarded to the network, but without triggering the P₄ penalty.
+This might arise in the case of duplicate beacon messages or while an application is syncing its blockchain, in which case it would be unable to ascertain the validity of new messages.
+
+In order to address this situation, all gossipsub v1.1 implementations _must_ support extended validators with a ternary decision interface.
+The outcome of extended validation can be one of three things:
+- Accept message; in this case the message is considered valid, and it should be delivered and forwarded to the network.
+- Reject message; in this case the message is considered invalid, and it should be rejected and trigger the P₄ penalty.
+- Drop message; in this case the message is neither delivered nor forwarded to the network, but the router does not trigger the P₄ penalty.
 
 ### Overview of New Parameters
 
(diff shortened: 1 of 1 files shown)


pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -2,7 +2,7 @@
 
 | Lifecycle Stage | Maturity       | Status | Latest Revision |
 |-----------------|----------------|--------|-----------------|
-| 1A              | Draft          | Active | r2, 2020-04-11  |
+| 1A              | Draft          | Active | r3, 2028-04-11  |
```
daviddias, line 5:
```suggestion
| 1A              | Draft          | Active | r3, 2020-04-28  |
```

Not a "that far in the future" change 😅

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -470,6 +471,20 @@ if invalidMessageDeliveries < DecayToZero {
 
 `TBD`: We are currently developing multiple types of simulations that will inform us on how to best recommend tunning the Scoring function. We will update this section once that work is complete
 
+#### Extended Validators
+
+The pubsub subsystem incorporates application-specific message validators so that the application can
+signal invalid message delivery, and trigger the P₄ penalty.
+However, It is possible to have circumstances where a message should not be delivered to the application
+or forwarded to the network, but without triggering the P₄ penalty.
+This might arise in the case of duplicate beacon messages or while an application is syncing its blockchain, in which case it would be unable to ascertain the validity of new messages.
```
daviddias, line 480:
```suggestion
A known use-case where this need exists is in the case of duplicate beacon messages or while an application is syncing its blockchain, in which case it would be unable to ascertain the validity of new messages.
```

pubsub/gossipsub/gossipsub-v1.1.md
vyzo: lol indeed!

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -470,6 +471,20 @@ if invalidMessageDeliveries < DecayToZero {
...
 
+#### Extended Validators
+
+The pubsub subsystem incorporates application-specific message validators so that the application can
+signal invalid message delivery, and trigger the P₄ penalty.
+However, It is possible to have circumstances where a message should not be delivered to the application
+or forwarded to the network, but without triggering the P₄ penalty.
+A known use-case where this need exists is in the case of duplicate beacon messages or while an application is syncing its blockchain, in which case it would be unable to ascertain the validity of new messages.
+
+In order to address this situation, all gossipsub v1.1 implementations _must_ support extended validators with a ternary decision interface.
```
raulk, line 482: I'd prefer to keep the current cardinality out of the spec -- because it makes this a bit bespoke. I'd support making this an enum, out of which three values are supported in this version of the spec.

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -470,6 +471,20 @@ if invalidMessageDeliveries < DecayToZero {
...
+The outcome of extended validation can be one of three things:
+- Accept message; in this case the message is considered valid, and it should be delivered and forwarded to the network.
+- Reject message; in this case the message is considered invalid, and it should be rejected and trigger the P₄ penalty.
+- Drop message; in this case the message is neither delivered nor forwarded to the network, but the router does not trigger the P₄ penalty.
```
raulk, line 486: The "reject" and "drop" terminology is confusing to me, because reject implies a drop too 🙃 

Maybe a more accurate term for the "drop" behaviour could be "ignore"?

pubsub/gossipsub/gossipsub-v1.1.md
vyzo: Yes, indeed this is the intention as well, will update.

pubsub/gossipsub/gossipsub-v1.1.md
vyzo: ok, I will call it `ignore`.

pubsub/gossipsub/gossipsub-v1.1.md
vyzo: done, I called it "enumerated interface".

pubsub/gossipsub/gossipsub-v1.1.md
vyzo: done, called it `Ignore`.

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -470,6 +471,20 @@ if invalidMessageDeliveries < DecayToZero {
 
 `TBD`: We are currently developing multiple types of simulations that will inform us on how to best recommend tunning the Scoring function. We will update this section once that work is complete
 
+#### Extended Validators
+
+The pubsub subsystem incorporates application-specific message validators so that the application can
+signal invalid message delivery, and trigger the P₄ penalty.
+However, It is possible to have circumstances where a message should not be delivered to the application
```
yusefnapora, line 478:
```suggestion
However, it is possible to have circumstances where a message should not be delivered to the application
```

PR #277, opened 2020-06-09
docs: add logo!

daviddias:
All versions of the logo at http://ipfs.io/ipfs/QmYv9CCtcvUZQmwnRYiNQ8UqTDRpL3ARe7h6e4qHESi47C


raulk:
Note: we can totally include a link to the design kit for implementors and others who may want to use it (maybe we should add a license to the kit), but for the spec doc we should try to optimise transfer size.

PR #288, opened 2020-08-07
Rework connections/readme.md to be more concise

### LINKED ISSUE #238: connections/README.md too specific to golang implimentation
Currently, `connections/README.md` contains language that's specific to an *implimentation* of a libp2p library, go-libp2p. In general, to my view, the document also fails to address that Streams are essentially still "connections" to protocols.

The document also fails to address that multistream-select is a *protocol* that's used to allow for a more efficient protocol selection process, beyond the default behaviour (without multistream), which to my memory is; send/receive a null-terminated protocol byte-string, and when the other node repeats this string, the protocol is agreed upon, if not, the connection is closed. This is something that the document also fails to address, or at least fails to mention.

"Upgrading" connections is a *strategy* that a client can apply to the connection that has been established in between two nodes, however, this process is completely optional and *non-binding* per the definition of libp2p: modularity. The fact that this is indeed optional, or non-binding, is something that the document fails to address.

Lastly, at the end there are still some leftover concerns for the golang implimentation of libp2p. In my personal opinion, *these issues/concerns have no home in a general spec document, and should instead be raised in their respective libraries (go-libp2p, in this example)*

**Please note:** I am new to the libp2p ecosystem in regards to discussion and history of its development, but however I saw some issues with this document when it was handed to me to read it, on how to impliment it in py-libp2p. These are all personal thoughts and issues that I've gotten during reading of that document, please correct me if any of my assumptions in these comments are simply false, or if my opinions are unfounded.

ShadowJonathan:
Fixes #238

Here I have some reworking and rewriting of the original connections spec, I personally find it too much of a core specification to be left ambiguous, thus I have done a rough cleanup of the document to more precisely document the core definitions and leave less ambiguity.

My methodology of moving and changing up the document was thus:

- Put exacting definitions first

- Describe the rough process of establishing a connection, be agnostic to implementations
  - Provide *exact* and *clear* definition of how bootstrapping a connection to a process works
    (this is most critical, and the main motivation behind reworking this document)

- Reuse as much from the original document, place in the rough footer of these definitions.

(pinging @yusefnapora, @JustMaier, @vasco-santos, @bigs, @mgoelzer)

(Please provide feedback to my draft, I expect this document to not be final as-is, but to at least put a step in the right direction)


connections/README.md
```diff
@@ -21,122 +22,164 @@ and spec status.
...
 
-This document does not cover the establishment of "transport level" connections,
+This document does not cover the establishment of "transport level" connections, 
 for example opening "raw" TCP sockets, as those semantics are specific to each
 transport.
 
 What is covered here is the process that occurs after making the initial
-transport level connection, up to the point where "application level" streams
-are opened, and their protocols are identified and data is routed appropriately
+transport level connection, up to the point where "application level" connections
```
raulk, line 60: Streams is more correct. There are no transport-level connections and application connections. There is one connection, and multiple streams over it.

connections/README.md
```diff
@@ -21,122 +22,164 @@ and spec status.
...
 ## Definitions
 
-A **connection** is a reliable, bidirectional communication channel between two
-libp2p peers that provides **security** and the ability to open multiple
-logically independent **streams**.
+### Protocol
+
+Protocols lay at the heart of libp2p, providing modularity by availability on both parties.
+
+Protocols are abritrary processes on both ends of the libp2p [session] that write/read to the underlying [connection].
```
raulk, line 70: A protocol doesn't need to be a process.
Tiny typo:

```suggestion
Protocols are arbitrary processes on both ends of the libp2p [session] that write/read to the underlying [connection].
```

raulk: They do not read/write to the underlying connection, but to a stream.

connections/README.md
```diff
@@ -21,122 +22,164 @@ and spec status.
...
+
+Protocols can be pretty much anything, from HTTP to Identity exchange, from [multiplexing] to security.
+
+Protocols are identified by their **Protocol ID**. This can be any string, but convention is that protocols identify themselves by forward slashes ( `/` )
```
raulk, line 74: A more precise statement would be:

```suggestion
Protocols are identified by their **Protocol ID**. This can be any string, but it's conventional for protocols to follow hierarchical naming, using forward slashes (`/`) for namespacing and/or to separate levels.
```

connections/README.md
```diff
@@ -21,122 +22,164 @@ and spec status.
...
+
+For example, `/mplex/1.0.0` is the Protocol ID for the [ `mplex` stream-multiplexing protocol][mplex].
+
+*A recommended markup of Protocol IDs is: `/<corresponding propery>/<protocol name>/<semantic version>` , with `corresponding property` often dropped with core libp2p-backed protocols.*
```
raulk, line 78: More often than not, the first path element will be the name of the application/system, e.g. eth2 uses `eth2` IIRC, filecoin uses `fil`, libp2p uses `p2p` sometimes.

connections/README.md
```diff
@@ -21,122 +22,164 @@ and spec status.
...
+
+An unique trait of protocols is that they're programmatically able to produce further agnostic [connections][connection] or [transports][transport], making it possible to stack features of protocols on top of eachother (see [Securing Connections](#securing-connections) and [Multiplexing Streams](#multiplexing-streams)).
```
raulk, line 80: I'm not sure what you mean with a protocol being able to produce a "agnostic" connection or transport? This is mixing various concepts in a way that is not truthful.

connections/README.md
```diff
@@ -21,122 +22,164 @@ and spec status.
...
+
+### Connection
+
+A connection is an abstract bi-directional pipe that carries a [protocol].
```
raulk, line 84: This is not correct. A connection does not carry a protocol. A connection is a transport-dependent pipe that connects two peers. A connection can hold streams, each of which does associate with a protocol.

connections/conn-upgrade.plantuml
```diff
@@ -1,36 +1,35 @@
...
-entity Responder
+entity Receiver
 
-== Initial multistream handshake ==
+== Initial protocol handshake ==
 
-Responder -> Initiator: /multistream/1.0.0
-Initiator -> Responder: /multistream/1.0.0
+Initiator -> Receiver: /multistream/1.0.0
+Receiver -> Initiator: /multistream/1.0.0
```
raulk, line 9: We borrowed this terminology from the Noise Spec: https://noiseprotocol.org/noise.html

raulk (changes requested):
This is a large rewriting. I'm ok revisiting this spec, but unfortunately of the prose that's proposed is factually incorrect.

connections/README.md
ShadowJonathan: Connections is more correct, "streams" is a concept born from multiplexing, this is concerning itself with the lowest levels of the protocol, where multiplexing is a higher concept which simply provides more connections (streams) over one connection.

connections/README.md
ShadowJonathan: > They do not read/write to the underlying connection, but to a stream.

They do, see my comment above, this document should not be specific to application framework structure, only explain the fundamentals of connection establishment.

connections/README.md
ShadowJonathan: Is that a detail that should be included in this fundamental document? It looks unnecessary to me to specify those next to explaining the basics of protocol ID naming convention.

connections/README.md
ShadowJonathan: The `proxy` protocol produces transports to other libp2p nodes, these don't inherit any connection properties (security, identity, etc.), and should instead be treated as a new opened transport to other nodes.

The `multiplex` protocol produces more connections from one connection, these connections inherit any underlying properties (security, etc.) from underlying connections, stopping at the transport.

connections/README.md
ShadowJonathan: See my above comment, this is an abstraction. Btw, a connection *does* associate itself with a protocol technically, with the default stack of `security -> mplex`, a connection technically carries a TLS protocol, which produces another connection for `mplex` to use to produce multiple connections (streams) from.

connections/conn-upgrade.plantuml
ShadowJonathan: To what should that be changed, then?

ShadowJonathan:
@raulk i want to get one thing down: to this document, streams are an application-level concept, implemented by a multiplexer protocol, while it is indeed convention to establish these at the start of each connection, i think this should explicitly be separated from connection-establishment as a method in further chapters to address application-level concerns (easily create connections for protocols to live on), not to mingle amongst fundamental definitions, as i think that would only complicate and obfuscate this document.

Specifically, streams are an IPFS concern and solution, libp2p is supposed to be modular, thus this document should reflect that, and effectively allow any and all forms of connection-establishment by mutual agreement on protocols, and to have that form itself from a baseline (the Bootstrap Protocol Negotiation).

ShadowJonathan:
I will be closing this as there has been no progression, response, or any substantial change to the libp2p methodology lately. You can take these changes and work with them, or you can contact me to reopen it, but i will be closing this as to remove this from my todo list.

PR #292, opened 2020-09-05
gossipsub v1.1: Validation queue protection with Random Early Drop

vyzo:
Updates the specification for the circuit breaker implemented in https://github.com/libp2p/go-libp2p-pubsub/pull/384

### DIFF
--- a/pubsub/gossipsub/gossipsub-v1.1.md
+++ b/pubsub/gossipsub/gossipsub-v1.1.md
@@ -2,7 +2,7 @@
 
 | Lifecycle Stage | Maturity                  | Status | Latest Revision |
 |-----------------|---------------------------|--------|-----------------|
-| 2A              | Candidate Recommendation  | Active | r8, 2020-09-05  |
+| 2A              | Candidate Recommendation  | Active | r7, 2020-05-30  |
 
 Authors: [@vyzo]
 
@@ -55,9 +55,6 @@ See the [lifecycle document][lifecycle-spec] for context about maturity level an
     - [Extended Validators](#extended-validators)
   - [Overview of New Parameters](#overview-of-new-parameters)
   - [Spam Protection Measures](#spam-protection-measures)
-  - [Validation Queue Protection](#validation-queue-protection)
-    - [Random Early Drop Algorithm](#random-early-drop-algorithm)
-    - [RED Parameters](#red-parameters)
   - [Recommendations for Network Operators](#recommendations-for-network-operators)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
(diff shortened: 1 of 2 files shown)


raulk (commented):
Reviewing as I comment, but my first thought is that we should make this an optional extension, because: (a) it doesn't affect the protocol, and (b) we want to avoid unilaterally updating a very well adopted spec. It's best if this extension has its own lifecycle (let's file it as a working draft), and implementations can specify if they support it or not.

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -634,6 +637,83 @@ In order counter spam that elicits responses and consumes resources, some measur
...
+by the validation front end, resulting in denial of service.
+
+In order to protect the system from this class of attacks, gossipsub v1.1 incorporates a circuit
+breaker that sits before the validation queue and can make informed decisions on whether to
+push a message into the validation queue. This defensive mechanism kicks in when the system detects
+an elevated rate of dropped messages, and makes decisions on whether to accept incoming messages for
+validation based on the statistical performance of peers in the origin IP address. The decision is
+probabilistic and implements a Random Early Drop (RED) strategy that drops messages with a probability
+that depends on the acceptance rates for messages from the origin IP. This strategy can neuter
+attacks on the validation queue, because messages are no longer dropped indiscrimantly in a drop-tail
```
raulk: editorial nit

```suggestion
attacks on the validation queue, because messages are no longer dropped indiscriminately in a drop-tail
```

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -634,6 +637,83 @@ In order counter spam that elicits responses and consumes resources, some measur
...
+
+#### Random Early Drop Algorithm
+
+The algorithm has two aspects:
+- The decision on whether to trigger RED.
+- The decision on whether to drop a message from an origin IP address.
+
+In order to trigger RED, the circuit breaker maintains the following queue statistics:
+- a (decaying) counter for the number of message validations.
+- a (decaying) counter for the number of dropped messages.
```
raulk: Is the _decaying_ attribute a clarification? Seems like a requirement that we want to highlight, in which case the parenthesis are a bit dodgy. Would make it italic to avoid confusion.

vyzo:
@raulk done -- moved it to an extension spec and left the original spec unmolested.

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -634,6 +637,83 @@ In order counter spam that elicits responses and consumes resources, some measur
...
+In order to make the actual RED decision, the circuit breaker maintains the following statistics per
+IP:
+- a (decaying) counter for the number of accepted messages.
+- a (decaying) counter for the number of duplicate messages, mixed with a weight `W_duplicate`.
+- a (decaying) counter for the number of ignored messages, mixed with a weight `W_ignored`.
+- a (decaying) counter for the number of rejected messages, mixed with a weight `W_rejected`.
+
+The router generates a random float `r` and accepts the message if and only if
+```
+r < (1 + accepted) / (1 + accepted + W_duplicate * duplicate + W_ignored * ignored + W_rejected * rejected)
```
raulk: It would be great to add recommendations or suggestions of concrete values and reasoning to choose those values (heuristics), so that's it not entirely left to the reader to figure out what a sane set of values would be here.

I wouldn't just plug in the Lotus values here, but maybe the similar weights and explain why. For example:

"With a recipe of W_duplicate=1, W_ignored=2, W_rejected=3, we are rapidly penalising rejections, mildly penalising ignored messages, and (relatively) softly penalising duplicate messages. The result is that clearly misbehaving peers (whose messages lead to outright rejection) will make up for 50% of the decision to break the circuit, while underperforming peers will also factor in, but with less force."

pubsub/gossipsub/red.md
```diff
@@ -0,0 +1,122 @@
+# gossipsub v1.1: Validation Queue Protection
```
raulk, line 1:
```suggestion
# gossipsub v1.1: Functional Extension for Validation Queue Protection
```

pubsub/gossipsub/red.md
```diff
@@ -0,0 +1,122 @@
...
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Overview
+
+This document specifies an extension to [gossipsub v1.1](gossipsub-v1.1.md) intended to
+provide a circuit breaker so that routers can withstand concerted attacks targetting the
+validation queue with a flood of spam.
+This extension does not modify the protocol in any way and works in conjuction with the defensive
+mechanism of gossipsub v1.1
```
raulk, line 46: 👍

raulk (approved):

pubsub/gossipsub/gossipsub-v1.1.md
vyzo: I have a defaults section, but yeah we could add some explanation for them.

pubsub/gossipsub/gossipsub-v1.1.md
vyzo: sure, will do.

pubsub/gossipsub/gossipsub-v1.1.md
vyzo: ugh, thanks for being a good spell checker!

pubsub/gossipsub/gossipsub-v1.1.md
vyzo: done.

pubsub/gossipsub/gossipsub-v1.1.md
vyzo: done.

pubsub/gossipsub/gossipsub-v1.1.md
vyzo: fixed!

raulk:
cc Interest Group: @yusefnapora, @raulk, @whyrusleeping, @Stebalien, @daviddias, @protolambda, @djrtwo, @dryajov, @mpetrunic, @AgeManning, @Nashatyrev, @mhchia

This is a functional extension to gossipsub 1.1 (does not affect the protocol) to introduce a circuit-breaker-like feature to defend against spam attacks.

It's a working draft, and we're testing it out in the [Filecoin Space Race network](https://spacerace.filecoin.io/). Might be worth to adopt in Eth2 too.

PR #294, opened 2020-09-14
gossipsub v1.1: introduce message signing policy

protolambda:
See https://github.com/libp2p/go-libp2p-pubsub/pull/359 for motivation and related discussion.

In short: Eth2 uses content-addressing and does not use these fields. If not content-addressed, signature data should be verified more strictly.

Changes:
- Pubsub readme: 
  - Describe the optionality of singing in the relevant signing section of pubsub, and refer to new gossipsub v1.1 configuration option
  - Clarify that signatures are singing over all fields (incl unrecognized fields)
- Gossipsub v1.1:
  - Briefly describe the general two cases when the fields are desired or not.
  - Document legacy "lax" behavior may be implemented as additional signing policies
  - Document two main (strict) signing policies
    - Feedback wanted: best way to describe the rejection of messages with bad signature or unexpected/missing fields?

Out of scope:
- We had a related discussion about rejecting unrecognized protobuf fields altogether, to avoid any unexpected privacy issues or encoding edge cases. (will open an issue for this)

See https://github.com/ethereum/eth2.0-specs/issues/1981 for Eth2 progress on this work, and clients implementing this functionality in gossipsub implementations.

PS. This is my first contribution directly to the libp2p specs, let me know if you have feedback to details such as formatting, I am happy to make changes.

cc @raulk, implemented the spec changes, review/edits welcome.

### DIFF
--- a/pubsub/README.md
+++ b/pubsub/README.md
@@ -171,7 +171,7 @@ It was also proposed in [#116](https://github.com/libp2p/specs/issues/116)
 to use a `message_hash`, however, it was noted:
 > a potential caveat with using hashes instead of seqnos:
 the peer won't be able to send identical messages (e.g. keepalives) within the
-timecache interval, as they will get rejected as duplicates.
+timecache interval, as they will get treated as duplicates.
 
 Some applications may not need keepalives, or choose to implement something more specific than a message hash. In those cases where duplicate payloads are not desirable, a `content-based` message ID function may be more appropriate.
 
--- a/pubsub/README.md
+++ b/pubsub/README.md
@@ -32,6 +32,7 @@ and spec status.
     - [The RPC](#the-rpc)
     - [The Message](#the-message)
     - [Message Signing](#message-signing)
+    - [Message Identification](#message-identification)
     - [The Topic Descriptor](#the-topic-descriptor)
         - [AuthOpts](#authopts)
             - [AuthMode 'NONE'](#authmode-none)
@@ -145,7 +146,7 @@ economics (see e.g.
 and
 [here](https://ethresear.ch/t/improving-the-ux-of-rent-with-a-sleeping-waking-mechanism/1480)).
 
-## Message identification
+## Message Identification
 
 To uniquely identify a message in a set of topics, a `message_id` is computed based on the message.
 This can be configured on the application layer, as `message_id_fn(*Message) => message_id`, 
--- a/pubsub/README.md
+++ b/pubsub/README.md
@@ -153,6 +153,10 @@ and
 
 Messages can be optionally signed, and it is up to the peer whether to accept and forward
 unsigned messages.
+When the receiver expects unsigned content-based messages, and thus does not expect
+the `from`, `seqno`, `signature`, or `key` fields, it may reject the messages (`StrictNoSign`).
+And if not, the receiver may choose to enforce signatures strictly (`StrictSign`).
+This optionality is configurable with the signing policy options starting from `v1.1`. 
 
 For signing purposes, the `signature` and `key` fields are used:
 - The `signature` field contains the signature.
@@ -160,6 +164,7 @@ For signing purposes, the `signature` and `key` fields are used:
   When present, it must match the peer ID.
 
 The signature is computed over the marshalled message protobuf _excluding_ the key field.
+This includes any fields that are not recognized, but still included in the marshalled data.
 The protobuf blob is prefixed by the string `libp2p-pubsub:` before signing.
 
 When signature validation fails for a signed message, the implementation must
--- a/pubsub/README.md
+++ b/pubsub/README.md
@@ -153,17 +153,18 @@ and
 
 Messages can be optionally signed, and it is up to the peer whether to accept and forward
 unsigned messages.
-When the receiver expects unsigned content-based messages, and thus does not expect
+The default choice of origin-stamped messaging, the receiver should enforce signatures strictly (`StrictSign`).
+When the receiver expects unsigned content-stamped messages, and thus does not expect
 the `from`, `seqno`, `signature`, or `key` fields, it may reject the messages (`StrictNoSign`).
-And if not, the receiver may choose to enforce signatures strictly (`StrictSign`).
-This optionality is configurable with the signing policy options starting from `v1.1`. 
+
+This optionality is configurable with the signature policy options starting from gossipsub v1.1. 
 
 For signing purposes, the `signature` and `key` fields are used:
 - The `signature` field contains the signature.
 - The `key` field contains the signing key when it cannot be inlined in the source peer ID.
   When present, it must match the peer ID.
 
-The signature is computed over the marshalled message protobuf _excluding_ the key field.
+The signature is computed over the marshalled message protobuf _excluding_ the `signature` field itself.
 This includes any fields that are not recognized, but still included in the marshalled data.
 The protobuf blob is prefixed by the string `libp2p-pubsub:` before signing.
 
--- a/pubsub/README.md
+++ b/pubsub/README.md
@@ -148,23 +148,24 @@ and
 
 ## Message Identification
 
-To uniquely identify a message in a set of topics, a `message_id` is computed based on the message.
-This can be configured on the application layer, as `message_id_fn(*Message) => message_id`.
-A `message_id_fn` may conditionally call different `message_id_fn` implementations per topic (or group thereof).
+To uniquely identify a message in a set of topics (for de-duplication, tracking, scoring and other purposes), a `message_id` is calculated based on the message.
+How the calculated happens can be configured on the application layer by supplying a function `message_id_fn`, such that `message_id_fn(*Message) => message_id`.
 
-The message ID approach generally fits in two flavors:
+> [[ Implementation note ]]: At the time of writing this section, go-libp2p-pubsub (reference implementation of this spec) only allows configuring a single top-level `message_id_fn`. This function may, however, vary its behaviour based on the topic (contained inside its `*Message`) argument. Thus, it's feasible to implement a per-topic policy using branch selection control flow logic. go-libp2p-pubsub plans to push down the configuration of the `message_id_fn` to the topic level. Other implementations are encouraged to do the same.
+
+The message ID calculation approach generally fits in two flavors:
 - **origin-stamped** messaging: the combination of the `seqno` and `from` fields
   uniquely identifies a message based on the *author*.
-- **content-stamped** messaging: a message ID derived from the `data` field
+- **content-addressed** messaging: a message ID derived from the `data` field
   uniquely identifies a message based on the *data*.
 
-The default `message_id_fn` is origin-stamped, and defined as the string concatenation of `from` and `seqno`.
+**The default `message_id_fn` is origin-stamped,** and defined as the string concatenation of `from` and `seqno`.
 
 If fabricated collisions are not a concern, or difficult enough within the window the message is relevant in,
-a `message_id` based on a short digest of inputs may benefit performance.
+a `message_id` based on a short digest of inputs may benefit performance. Whichever the choice, it is crucial that **all peers** participating in a topic implement the same message ID calculation logic, or the topic may function suboptimally.
 
-Note that different specialized pubsub components, such as the 'timecache' used in the Go implementation,
-may use the `message_id` to key messages.
+Note that different specialized pubsub components, such as the 'timecache' used in the Go implementation, scoring functions or circuit-breakers 
+may use the `message_id` to key and track messages.
 
 It was also proposed in [#116](https://github.com/libp2p/specs/issues/116)
 to use a `message_hash`, however, it was noted:
@@ -172,7 +173,7 @@ to use a `message_hash`, however, it was noted:
 the peer won't be able to send identical messages (e.g. keepalives) within the
 timecache interval, as they will get rejected as duplicates.
 
-Some applications may not need keepalives, or choose to implement something more specific than a message hash. 
+Some applications may not need keepalives, or choose to implement something more specific than a message hash. In those cases where duplicate payloads are not desirable, a `content-based` message ID function may be more appropriate.
 
 ## Message Signing
 
--- a/pubsub/README.md
+++ b/pubsub/README.md
@@ -112,6 +112,9 @@ message Message {
 }
 ```
 
+The `optional` fields may be omitted, depending on the
+[signature policy](#message-signing) and [message ID function](#message-identification)
+
 The `from` field denotes the author of the message, note that this is not
 necessarily the peer who sent the RPC this message is contained in. This is
 done to allow content to be routed through a swarm of pubsubbing peers.
@@ -123,14 +126,7 @@ The `seqno` field is a 64-bit big-endian uint that is a linearly increasing
 number that is unique among messages originating from each given peer. No two
 messages on a pubsub topic from the same peer should have the same `seqno`
 value, however messages from different peers may have the same sequence number,
-so this number alone cannot be used to address messages. Notably the
-'timecache' in use by the go implementation contains a `message_id`,
-which is constructed from the concatenation of the `seqno` and the `from`
-fields. This `message_id` is then unique among messages. It was also proposed
-in [#116](https://github.com/libp2p/specs/issues/116) to use a `message_hash`,
-however, it was noted: "a potential caveat with using hashes instead of seqnos:
-the peer won't be able to send identical messages (e.g. keepalives) within the
-timecache interval, as they will get rejected as duplicates."
+so this number alone cannot be used to address messages by origin-stamping.
 
 The `topicIDs` field specifies a set of topics that this message is being
 published to.
@@ -149,6 +145,30 @@ economics (see e.g.
 and
 [here](https://ethresear.ch/t/improving-the-ux-of-rent-with-a-sleeping-waking-mechanism/1480)).
 
+## Message identification
+
+To uniquely identify a message in a set of topics, a `message_id` is computed based on the message.
+This can be configured on the application layer, as `message_id_fn(*Message) => message_id`, 
+which generally fits in two flavors:
+- **origin-stamped** messaging: the concatenation of the `seqno` and `from` fields
+  uniquely identifies a message based on the *author*.
+- **content-stamped** messaging: a message ID derived from the `data` field
+  uniquely identifies a message based on the *data*.
+
+If fabricated collisions are not a concern, or difficult enough within the window the message is relevant in,
+a `message_id` based on a short digest of inputs may benefit performance.
+
+Note that different specialized pubsub components, such as the 'timecache' used in the Go implementation,
+may use the `message_id` to key messages.
+
+It was also proposed in [#116](https://github.com/libp2p/specs/issues/116)
+to use a `message_hash`, however, it was noted:
+> a potential caveat with using hashes instead of seqnos:
+the peer won't be able to send identical messages (e.g. keepalives) within the
+timecache interval, as they will get rejected as duplicates.
+
+Some applications may not need keepalives, or choose to implement something more specific than a message hash. 
+
 ## Message Signing
 
 Messages can be optionally signed, and it is up to the peer whether to accept and forward
@@ -161,7 +181,7 @@ This optionality is configurable with the signature policy options starting from
 
 For signing purposes, the `signature` and `key` fields are used:
 - The `signature` field contains the signature.
-- The `key` field contains the signing key when it cannot be inlined in the source peer ID.
+- The `key` field contains the signing key when it cannot be inlined in the source peer ID (`from`).
   When present, it must match the peer ID.
 
 The signature is computed over the marshalled message protobuf _excluding_ the `signature` field itself.
--- a/pubsub/README.md
+++ b/pubsub/README.md
@@ -149,12 +149,16 @@ and
 ## Message Identification
 
 To uniquely identify a message in a set of topics, a `message_id` is computed based on the message.
-This can be configured on the application layer, as `message_id_fn(*Message) => message_id`, 
-which generally fits in two flavors:
-- **origin-stamped** messaging: the concatenation of the `seqno` and `from` fields
+This can be configured on the application layer, as `message_id_fn(*Message) => message_id`.
+A `message_id_fn` may conditionally call different `message_id_fn` implementations per topic (or group thereof).
+
+The message ID approach generally fits in two flavors:
+- **origin-stamped** messaging: the combination of the `seqno` and `from` fields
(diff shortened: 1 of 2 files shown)


pubsub/README.md
```diff
@@ -153,13 +153,18 @@ and
...
+And if not, the receiver may choose to enforce signatures strictly (`StrictSign`).
+This optionality is configurable with the signing policy options starting from `v1.1`. 
 
 For signing purposes, the `signature` and `key` fields are used:
 - The `signature` field contains the signature.
 - The `key` field contains the signing key when it cannot be inlined in the source peer ID.
   When present, it must match the peer ID.
 
 The signature is computed over the marshalled message protobuf _excluding_ the key field.
+This includes any fields that are not recognized, but still included in the marshalled data.
```
arnetheduck, line 167: this is problematic because protobuf is non-deterministic in its marshalling  - the way to write this specification would be to remove mentions of protobuf and instead explicitly outline the marshalling in terms of key-value pairs and field types - it would have to be done carefully such that the result is readable by a compliant protobuf parser - it might also happen by chance that some protobuf encoders produce a valid byte stream, but with unrecognised fields in particular, the likelihood is low.

raulk: Let's hoist the protobuf deterministic serialisation discussion to a new issue.

raulk: It's been recurrent enough that it needs to be a first-class discussion with a solid outcome.

pubsub/README.md
```diff
@@ -153,13 +153,18 @@ and
 
 Messages can be optionally signed, and it is up to the peer whether to accept and forward
 unsigned messages.
+When the receiver expects unsigned content-based messages, and thus does not expect
```
raulk, line 156: Mind elaborating what does `content-based messages` means here?

raulk: I think the two paragraphs that follow are better expressed as a bullet. Why don't we introduce the Lax modes here too? We anyway restrict all of this to v1.1, so we might as well introduce all options here.

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -134,6 +135,29 @@ message PeerInfo {
 }
 ```
 
+### Signature policy
+
+The usage of the `signature`, `key`, `from`, and `seqno` fields in `Message` is now configurable.
+These fields may negatively affect privacy in content-addressed messaging,
+and may need to be strictly enforced in author-addressed messaging.
```
raulk, line 142: I think the content-addressed and author-addressed word choice might be a bit confusing. How about "anonymous" and "origin-stamped"?

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -134,6 +135,29 @@ message PeerInfo {
...
+
+In gossipsub v1.0, a "lax" signing policy is effective: verify signatures, and if not, only when present.
+In gossipsub v1.1, these fields are strictly present and verified, or completely omitted altogether.
+An implementation may choose to support the legacy v1.0 "lax" signing policy,
+ along with an explicit message authoring option.
```
raulk, line 147: "origin-stamping" option.

raulk: Move this paragraph below the specification of both options. Otherwise we're speaking about v1.0, then v1.1, then v1.0 support in v1.1, back to v1.1.

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -134,6 +135,29 @@ message PeerInfo {
...
+
+Gossipsub v1.1 has two policies to choose from:
+- `StrictSign`:
+  - Produces the `signature`, `key` (`from` may be enough), `from` and `seqno` fields.
+  - Enforce the fields to be present, reject otherwise.
+  - Verify the signature, reject otherwise.
+  - Propagate the fields if valid.
+- `StrictNoSign`:
+  - Produces messages without the `signature`, `key`, `from` and `seqno` fields.
+    The corresponding protobuf key-value pairs are absent from the marshalled message, not just empty.
```
raulk, line 157: Separate these in two sub-bullets. "On the message producer side:" / "On the message receiver side:"

raulk (commented):
Few comments; I'm going to take the liberty to push to your branch to make a few editorial changes, if you don't mind.

pubsub/README.md
protolambda: Should have called it `content-addressed`, or maybe you can think of better wording, to describe messages that identify themselves without the author, seqnoi and signature approach, like in eth2. These messages do not need to be signed, as for these it depends on the application layer, which checks contents etc.

pubsub/gossipsub/gossipsub-v1.1.md
protolambda: Sounds good, can you include this in the changes you wanted to make? I'm including this line mostly for awareness of backwards compatibility options, but it does not need to be part of the spec otherwise.

arnetheduck:
One important point to raise here is that one reason to use gossipsub and not a custom protocol is that it's a generic message transport - we can expect that the gossipsub networks will cross-pollinate - say between ipfs and eth2 for example - and it's plausible that one might wish to be a participant of both. In fact, it's one of the main advantages of using gossipsub: peer diversity is increased. 

The strict policies here somewhat limit that benefit - it creates, in effect, two different networks. The issue could be mitigated by making sure that implementations support different policies per topic - the way to achieve this would be to move some of the validation up to the application layer which easier can handle the per-topic nuances and instead only have a lax mode in gossipsub where:

* _iff_ a field is present, it must be valid for its type
  * `seqno` must be 8 bytes
  * public keys and peerid's must parse as valid for their type
  * signatures, if present, must validate according to rules that will be lifted to a different PR
* this is congruent with the fields being optional in the protobuf description of the messages

on top of that policy, applications could then choose to restrict things further within the topic namespace they occupy - ie ETH2 might choose to not propagate messages with gossipsub signatures _on eth2 topics_ (because we already do a mostly equivalent check / validation by verifying a payload signature). 

Status is developing a similar protocol on top of gossipsub that also doesn't use the gossipsub signatures (because we negotiate a different scheme which is more appropriate for the status use case - the gossipsub signatures themselves live at the wrong level of abstraction which is why their usefulness is limited, often)

djrtwo:
> In fact, it's one of the main advantages of using gossipsub: peer diversity is increased.

Is this one of the main benefits? That is a clear benefit of using a cross-pollinated DHT but nodes of different networks are not going to generally subscribe to (or be able to execute validations of) gossipsub topics on an alternate network.

arnetheduck:
well, I'd say so, yes - we're paying the protobuf tax, the "optional-seqno-and-signature-tax", we namespace the topics etc in order to achieve this interoperability where multiple applications _can_ share the infrastructure - and it's not so much about validation specifically but alternatives that might have different needs but still be related (L2, a future ETC fork, client extensions etc). 

If it was only ETH2, a much simpler thing to do would have been to take the research behind gossipsub, even forking it, and run it as a separate libp2p protocol (`eth2sub/1.0.0`) that would use SSZ and pure content-hashing - content-hashing is already a fairly significant departure from "standard" gossipsub - allowing per-topic optionality in the gossipsub fields in a way achieves the same goal but in a softer, more enabling way - ie it's still possible to create a restricted software where only a certain topic space with specific validations rules are applied, but it's also possible to create a mixed network with different rules applied to separate namespaces, without polluting the libp2p spec with application details.

raulk:
I think @arnetheduck makes a valid point. If we make these signing policies configurable _at the entire gossipsub stack level_, it's no longer possible to participate in pubsub topics that adopt different signing policies. That is, you could never have a single node subscribed to both Filecoin (lax) and Eth2 (strict no sign) topics, which does harm interoperability and sabotages the feasibility of very desirable cross-chain applications and use cases.

I'm inclined to not merge this PR and iterate on the implementation submitted in libp2p/go-libp2p-pubsub#359 to push these options down to the topic level. That way, when you join to a topic, you specify the policy that applies to that topic.

@protolambda I know you are probably backlogged. Maybe someone in the libp2p or IPFS teams could take this on.

protolambda:
@raulk pushing these options down to a topic level would be great, sounds like a good solution. And yes, quite backlogged with work. I think it is fair to add a notice in this PR that the options can be implemented on a per-topic basis. Then expand on that once the go-libp2p implementation has decided on the best approach.

raulk:
@protolambda makes sense. I'll rejig the text, and add an implementation margin note that we can remove once this is pushed down to the topic level in go-libp2p.

arnetheduck:
While working on this, we also noticed that this breaks the default message id generation, which relies on sequence number and peer id, is somewhat broken when the fields are not included - ie migrating fully to a content-based hash method when the anonymous mode is enabled is necessary.

protolambda:
@arnetheduck yes, good point. And so message ID should be per-topic as well (it's kind of already, if you switch based on the topic info within the message). Will update PR now to incorporate formatting + per-topic behavior recommendation.

protolambda:
@raulk update:
- Implemented your PR suggestions (or at least my interpretation of them, I wanted to clarify and format some more details).
- Added a section about message identification, and clarify the two common flavors, as well as moving the extra details from the fields section there.
- Message signature policy now states that it can initially be a global configuration, but per-topic policies facilitate mixed protocols better
- Optionality of fields is clarified, and states to depend on message ID and signature policy choices.

Again, feel free to make editorial choices, squash/rename commits, or any feedback. I would like to keep this PR moving, so we can continue with https://github.com/ethereum/eth2.0-specs/pull/2060 before the next Eth2 specs release.

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -134,6 +136,50 @@ message PeerInfo {
 }
 ```
 
+### Signature Policy
+
+The usage of the `signature`, `key`, `from`, and `seqno` fields in `Message` is now configurable.
+Initially this could be configured globally, however, configuration on a per-topic basis will facilitate mixed protocols better.
```
arnetheduck, line 142:
```suggestion
Initially this could be configured globally, however, because the policies are mutually incompatible, configuration on a per-topic basis will facilitate mixed protocols better.
```

raulk:
Apologies. Will get to this later today.

pubsub/README.md
```diff
@@ -149,17 +146,51 @@ economics (see e.g.
 and
 [here](https://ethresear.ch/t/improving-the-ux-of-rent-with-a-sleeping-waking-mechanism/1480)).
 
+## Message Identification
+
+To uniquely identify a message in a set of topics, a `message_id` is computed based on the message.
+This can be configured on the application layer, as `message_id_fn(*Message) => message_id`.
+A `message_id_fn` may conditionally call different `message_id_fn` implementations per topic (or group thereof).
```
raulk, line 153:
```suggestion
To uniquely identify a message in a set of topics (for de-duplication, tracking, scoring and other purposes), a `message_id` is calculated based on the message.
How the calculated happens can be configured on the application layer by supplying a function `message_id_fn`, such that `message_id_fn(*Message) => message_id`.

> [[ Implementation note ]]: At the time of writing this section, go-libp2p-pubsub (reference implementation of this spec) only allows configuring a single top-level `message_id_fn`. This function may, however, vary its behaviour based on the topic (contained inside its `*Message`) argument. Thus, it's feasible to implement a per-topic policy using branch selection control flow logic. go-libp2p-pubsub plans to push down the configuration of the `message_id_fn` to the topic level. Other implementations are encouraged to do the same.
```

pubsub/README.md
```diff
@@ -149,17 +146,51 @@ economics (see e.g.
...
+
+The message ID approach generally fits in two flavors:
```
raulk, line 155:
```suggestion
The message ID calculation approach generally fits in two flavors:
```

pubsub/README.md
```diff
@@ -149,17 +146,51 @@ economics (see e.g.
...
+- **origin-stamped** messaging: the combination of the `seqno` and `from` fields
+  uniquely identifies a message based on the *author*.
+- **content-stamped** messaging: a message ID derived from the `data` field
```
raulk, line 158:
```suggestion
- **content-addressed** messaging: a message ID derived from the `data` field
```

pubsub/README.md
```diff
@@ -149,17 +146,51 @@ economics (see e.g.
...
+  uniquely identifies a message based on the *data*.
+
+The default `message_id_fn` is origin-stamped, and defined as the string concatenation of `from` and `seqno`.
```
raulk, line 161:
```suggestion
**The default `message_id_fn` is origin-stamped,** and defined as the string concatenation of `from` and `seqno`.
```

pubsub/README.md
```diff
@@ -149,17 +146,51 @@ economics (see e.g.
...
+
+If fabricated collisions are not a concern, or difficult enough within the window the message is relevant in,
+a `message_id` based on a short digest of inputs may benefit performance.
```
raulk, line 164:
```suggestion
If fabricated collisions are not a concern, or difficult enough within the window the message is relevant in,
a `message_id` based on a short digest of inputs may benefit performance. Whichever the choice, it is crucial that **all peers** participating in a topic implement the same message ID calculation logic, or the topic may function suboptimally.
```

pubsub/README.md
```diff
@@ -149,17 +146,51 @@ economics (see e.g.
...
+
+Note that different specialized pubsub components, such as the 'timecache' used in the Go implementation,
+may use the `message_id` to key messages.
```
raulk, line 167:
```suggestion
Note that different specialized pubsub components, such as the 'timecache' used in the Go implementation, scoring functions or circuit-breakers 
may use the `message_id` to key and track messages.
```

pubsub/README.md
```diff
@@ -149,17 +146,51 @@ economics (see e.g.
...
+
+It was also proposed in [#116](https://github.com/libp2p/specs/issues/116)
+to use a `message_hash`, however, it was noted:
+> a potential caveat with using hashes instead of seqnos:
+the peer won't be able to send identical messages (e.g. keepalives) within the
+timecache interval, as they will get rejected as duplicates.
+
+Some applications may not need keepalives, or choose to implement something more specific than a message hash.
```
raulk, line 175:
```suggestion
Some applications may not need keepalives, or choose to implement something more specific than a message hash. In those cases where duplicate payloads are not desirable, a `content-based` message ID function may be more appropriate.
```

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -134,6 +136,50 @@ message PeerInfo {
...
+
+In content-stamped messaging, the fields may negatively affect privacy:
+revealing the relationship between `data` and `from`/`seqno`.
+
+#### Signature Policy Options
+
+In gossipsub v1.1, these fields are strictly present and verified, or completely omitted altogether:
+- `StrictSign`:
+  - On the producing side:
+    - Build messages with the `signature`, `key` (`from` may be enough), `from` and `seqno` fields.
```
raulk, line 155:
```suggestion
    - Build messages with the `signature`, `key` (`from` may be enough for certain inlineable public key types), `from` and `seqno` fields.
```

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -134,6 +136,50 @@ message PeerInfo {
...
+    - Enforce the fields to be present, reject otherwise.
+    - Propagate only if the fields are valid and signature can be verified, reject otherwise.
+- `StrictNoSign`:
+  - On the producing side:
+    - Build messages without the `signature`, `key`, `from` and `seqno` fields.
+    - The corresponding protobuf key-value pairs are absent from the marshalled message, not just empty.
+  - On the consuming side:
+    - Enforce the fields to be absent, reject otherwise.
+    - Propagate only if the fields are absent, reject otherwise.
+  - A `message_id` function will not be able to use the above fields, and may instead rely on the `data` field.
```
raulk, line 166:
```suggestion
  - A `message_id` function will not be able to use the above fields, and should instead rely on the `data` field. A commonplace strategy is to calculate a hash.
```

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -134,6 +136,50 @@ message PeerInfo {
...
+
+In gossipsub v1.0, a legacy "lax" signing policy could be configured, to not verify signatures except when present:
```
raulk, line 168:
```suggestion
In gossipsub v1.0, a legacy "lax" signing policy could be configured, to only verify signatures when present. For security reasons, this is strategy is discarded in subsequent versions, but MAY still be supported for backwards-compatibility. If so, its use should be discouraged through prominent deprecation warnings. These strategies will be entirely dropped in the future.
```

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -134,6 +136,50 @@ message PeerInfo {
...
+- `LaxSign`: *Defined for completeness, insecure*. Also known as authoring but not verifying.
```
raulk, line 169:
```suggestion
- `LaxSign`: *this was never an original gossipsub 1.0 option, but it's defined here for completeness, and considered insecure*. Always sign, and verify incoming signatures, and but accept unsigned messages.
```

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -134,6 +136,50 @@ message PeerInfo {
...
+  - On the producing side:
+    - Build messages with the `signature`, `key` (`from` may be enough), `from` and `seqno` fields.
+  - On the consuming side:
+    - `signature` may be absent, and not verified.
+    - Verify `signature`, iff the `signature` is present, then reject if `signature` is invalid.
+- `LaxNoSign`: *Previous default for no-verification*
```
raulk, line 175:
```suggestion
- `LaxNoSign`: *Previous default for no-verification*. Do not sign nor origin-stamp, but verify incoming signatures, and accept unsigned messages.
```

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -134,6 +136,50 @@ message PeerInfo {
 }
 ```
 
+### Signature Policy
+
+The usage of the `signature`, `key`, `from`, and `seqno` fields in `Message` is now configurable.
```
raulk, line 141:
```suggestion
The usage of the `signature`, `key`, `from`, and `seqno` fields in `Message` is now configurable per topic, in the manners specified in this section.
```

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -134,6 +136,50 @@ message PeerInfo {
...
+Initially this could be configured globally, however, because the policies are mutually incompatible, configuration on a per-topic basis will facilitate mixed protocols better.
```
raulk, line 142:
```suggestion
> [[ Implementation note ]]: At the time of writing this section, go-libp2p-pubsub (reference implementation of this spec) allows for configuring the signature policy at a global pubsub instance level. This needs to be pushed down to topic-level configuration. Other implementations are encouraged to support topic-level configuration, as the spec mandates.
```

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -134,6 +136,50 @@ message PeerInfo {
...
+
+In the default origin-stamped messaging, the fields need to be strictly enforced:
+the `seqno` and `from` fields form the `message_id`, and should be verified to avoid `message_id` collisions.
+
+In content-stamped messaging, the fields may negatively affect privacy:
+revealing the relationship between `data` and `from`/`seqno`.
```
raulk, line 148: I'm not sure what the goal of this is. This might be better moved to after "Signature Policy Options", and put under a "Usage indications/notes" section.

I would reword to something like this:

* Origin-stamped messaging: choose `StrictSign`, and use in combination with the default `message_id_fn`.
* Content-addressed messaging: enable `StrictNoSign`, with a custom hash-based `message_id_fn`. Any signatures would now be part of the payload, and would need to be validated through a custom message validator.

raulk:
I'm going to commit my suggestions, and I'll do another pass tomorrow. Also requesting @vyzo's review here.

pubsub/gossipsub/gossipsub-v1.1.md
protolambda: The idea was to give some context/intuition why the fields are handled like they are. It could be worded directly as `message id approach -> signing policy` directions, like you suggest. Either option works for us.

pubsub/README.md
protolambda: I think you mean `content-addressed` here, to keep it consistent with other changes.

pubsub/README.md
```diff
@@ -149,17 +146,52 @@ economics (see e.g.
...
+a `message_id` based on a short digest of inputs may benefit performance. Whichever the choice, it is crucial that **all peers** participating in a topic implement the same message ID calculation logic, or the topic may function suboptimally.
+
+Note that different specialized pubsub components, such as the 'timecache' used in the Go implementation, scoring functions or circuit-breakers 
+may use the `message_id` to key and track messages.
+
+It was also proposed in [#116](https://github.com/libp2p/specs/issues/116)
+to use a `message_hash`, however, it was noted:
+> a potential caveat with using hashes instead of seqnos:
+the peer won't be able to send identical messages (e.g. keepalives) within the
+timecache interval, as they will get rejected as duplicates.
```
vyzo, line 174: ignored, not rejected.

pubsub/README.md
vyzo: I think we should specify canonical encoding here to remove all ambiguity.

pubsub/gossipsub/gossipsub-v1.1.md
```diff
@@ -134,6 +136,50 @@ message PeerInfo {
 }
 ```
 
+### Signature Policy
```
vyzo, line 139: I am not sure that this should be in the gossipsub spec -- it is a generic pubsub policy that applies to all routing algorithms.

vyzo (commented):
Thank you, this is a very welcome addition.

However, I think that the signature policy should be part of the pubsub spec and not the gossipsub spec.
There is nothing gossipsub specific about signature policies, and the policy applies to all routing algorithms as signing is handled outside the router -- at least in the reference implementation.

pubsub/gossipsub/gossipsub-v1.1.md
raulk: Agreed, will move it.

protolambda: Was looking for a place to version it, but since it's not breaking anything necessarily, and applies to pubsub, moving it to the pubsub doc sounds good.

PR #298, opened 2020-09-24
Make the name-change to `PeerRecord` more consistent

tmakarios:
An earlier commit (377f05abe37fca3bfe392b5e07c134da94306c02) changed `RoutingState` to `PeerRecord` in the protobuf definition, but left `RoutingState` throughout the rest of the document.  This patch makes the document a bit more consistent, but I've been fairly conservative with my changes, so while there are no `RoutingState`s left, there is still a `SignedRoutingState`, and a couple of slightly odd things like:
> When wrapping a `PeerRecord` in a signed envelope, the domain string MUST be
`libp2p-routing-state`.

### DIFF
--- a/RFC/0003-routing-records.md
+++ b/RFC/0003-routing-records.md
@@ -142,30 +142,30 @@ This structure can be serialized and contained in a [signed
 envelope][envelope-rfc], which lets us issue "self-certified" address records
 that are signed by the peer that the addresses belong to.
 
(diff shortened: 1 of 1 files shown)


tmakarios:
@jacobheun The bug still remains (that `PeerRecord` is the name when it's defined in this document, but the old name `RoutingState` is still used everywhere else in the same document).  Are there plans to resolve this bug some other way?

PR #299, opened 2020-09-25
pubsub: signing policy -- #294 + editorial changes

raulk:
This introduces editorial changes on top of #294. We now concentrate all our additions in `pubsub.md`.

---

@protolambda, I wanted to target your branch, but it's in a fork.

cc @arnetheduck

### DIFF
--- a/pubsub/README.md
+++ b/pubsub/README.md
@@ -171,7 +171,7 @@ It was also proposed in [#116](https://github.com/libp2p/specs/issues/116)
 to use a `message_hash`, however, it was noted:
 > a potential caveat with using hashes instead of seqnos:
 the peer won't be able to send identical messages (e.g. keepalives) within the
-timecache interval, as they will get rejected as duplicates.
+timecache interval, as they will get treated as duplicates.
 
 Some applications may not need keepalives, or choose to implement something more specific than a message hash. In those cases where duplicate payloads are not desirable, a `content-based` message ID function may be more appropriate.
 
--- a/pubsub/README.md
+++ b/pubsub/README.md
@@ -32,6 +32,7 @@ and spec status.
     - [The RPC](#the-rpc)
     - [The Message](#the-message)
     - [Message Signing](#message-signing)
+    - [Message Identification](#message-identification)
     - [The Topic Descriptor](#the-topic-descriptor)
         - [AuthOpts](#authopts)
             - [AuthMode 'NONE'](#authmode-none)
@@ -145,7 +146,7 @@ economics (see e.g.
 and
 [here](https://ethresear.ch/t/improving-the-ux-of-rent-with-a-sleeping-waking-mechanism/1480)).
 
-## Message identification
+## Message Identification
 
 To uniquely identify a message in a set of topics, a `message_id` is computed based on the message.
 This can be configured on the application layer, as `message_id_fn(*Message) => message_id`, 
--- a/pubsub/README.md
+++ b/pubsub/README.md
@@ -153,6 +153,10 @@ and
 
 Messages can be optionally signed, and it is up to the peer whether to accept and forward
 unsigned messages.
+When the receiver expects unsigned content-based messages, and thus does not expect
+the `from`, `seqno`, `signature`, or `key` fields, it may reject the messages (`StrictNoSign`).
+And if not, the receiver may choose to enforce signatures strictly (`StrictSign`).
+This optionality is configurable with the signing policy options starting from `v1.1`. 
 
 For signing purposes, the `signature` and `key` fields are used:
 - The `signature` field contains the signature.
@@ -160,6 +164,7 @@ For signing purposes, the `signature` and `key` fields are used:
   When present, it must match the peer ID.
 
 The signature is computed over the marshalled message protobuf _excluding_ the key field.
+This includes any fields that are not recognized, but still included in the marshalled data.
 The protobuf blob is prefixed by the string `libp2p-pubsub:` before signing.
 
 When signature validation fails for a signed message, the implementation must
--- a/pubsub/README.md
+++ b/pubsub/README.md
@@ -153,17 +153,18 @@ and
 
 Messages can be optionally signed, and it is up to the peer whether to accept and forward
 unsigned messages.
-When the receiver expects unsigned content-based messages, and thus does not expect
+The default choice of origin-stamped messaging, the receiver should enforce signatures strictly (`StrictSign`).
+When the receiver expects unsigned content-stamped messages, and thus does not expect
 the `from`, `seqno`, `signature`, or `key` fields, it may reject the messages (`StrictNoSign`).
-And if not, the receiver may choose to enforce signatures strictly (`StrictSign`).
-This optionality is configurable with the signing policy options starting from `v1.1`. 
+
+This optionality is configurable with the signature policy options starting from gossipsub v1.1. 
 
 For signing purposes, the `signature` and `key` fields are used:
 - The `signature` field contains the signature.
 - The `key` field contains the signing key when it cannot be inlined in the source peer ID.
   When present, it must match the peer ID.
 
-The signature is computed over the marshalled message protobuf _excluding_ the key field.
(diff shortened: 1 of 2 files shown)


pubsub/README.md
```diff
@@ -149,22 +156,172 @@ economics (see e.g.
 and
 [here](https://ethresear.ch/t/improving-the-ux-of-rent-with-a-sleeping-waking-mechanism/1480)).
 
+## Message Identification
+
+Pubsub requires to uniquely identify messages via a message ID. This enables
+a wide range of processes like for de-duplication, tracking, scoring,
```
raulk, line 162:
```suggestion
a wide range of processes like de-duplication, tracking, scoring,
```

protolambda (approved):
Looks good, and thanks for making the necessary changes @raulk

pubsub/README.md
```diff
@@ -112,30 +115,34 @@ message Message {
...
+The `optional` fields may be omitted, depending on the
+[signature policy](#message-signing) and
+[message ID function](#message-identification).
 
-The `data` field is an opaque blob of data, it can contain any data that the
-publisher wants it to.
+The `from` field denotes the author of the message. This is the peer who
+initially authored the message, and NOT the peer who propagated it. Thus, as
+the message is routed through a swarm of pubsubbing peers, the original
+authorship is preserved.
```
vyzo, line 125: we should say it is optional.

pubsub/README.md
```diff
@@ -112,30 +115,34 @@ message Message {
...
-so this number alone cannot be used to address messages. Notably the
-'timecache' in use by the go implementation contains a `message_id`,
-which is constructed from the concatenation of the `seqno` and the `from`
-fields. This `message_id` is then unique among messages. It was also proposed
-in [#116](https://github.com/libp2p/specs/issues/116) to use a `message_hash`,
-however, it was noted: "a potential caveat with using hashes instead of seqnos:
-the peer won't be able to send identical messages (e.g. keepalives) within the
-timecache interval, as they will get rejected as duplicates."
+value, however messages from different peers may (and likely will) have the same
+sequence number. In other words, this number is not globally unique. It is used
```
vyzo, line 131: Better to say "It _may_ be used..."

vyzo: Also let's specify it is optional.

pubsub/README.md
```diff
@@ -149,22 +156,172 @@ economics (see e.g.
...
+    and `seqno` fields.
+  
+On the consuming side:
+  - `signature` may be absent, and not verified.
+  - Verify `signature`, iff the `signature` is present, then reject if
+    `signature` is invalid.
+
+**`LaxNoSign` option**
+
+_Previous default_.
```
vyzo, line 305: that's not quite true -- `StrictSign` has been the default in the reference implementation for some time now.

pubsub/README.md
```diff
@@ -112,30 +115,34 @@ message Message {
...
-done to allow content to be routed through a swarm of pubsubbing peers.
+The `optional` fields may be omitted, depending on the
+[signature policy](#message-signing) and
+[message ID function](#message-identification).
 
-The `data` field is an opaque blob of data, it can contain any data that the
-publisher wants it to.
+The `from` field denotes the author of the message. This is the peer who
+initially authored the message, and NOT the peer who propagated it. Thus, as
+the message is routed through a swarm of pubsubbing peers, the original
```
vyzo, line 124: pubsubbing? I think one `b` might be more correct.

pubsub/README.md
```diff
@@ -149,22 +156,172 @@ economics (see e.g.
...
+> within the timecache interval, as they will get treated as duplicates. This
+> consequence may or may not be relevant to the application at hand.
+> Reference: [#116](https://github.com/libp2p/specs/issues/116).
+
+**Note that the availability of these fields on the `Message` object will depend
+on the [signature policy](#signature-policy) configured for the topic.**
+
+Whichever the choice, it is crucial that **all peers** participating in a topic
+implement identical message ID calculation logic, or the topic may function
+suboptimally.
```
vyzo, line 191: Maybe use stronger wording here? It's not just suboptimal, it's nearly a _malfunction_: divergent message id functions may cause the gossipsub router to use excessive bandwidth and penalize peers unduly for not following up on advertisements.

vyzo (commented):
Overall looks good; some minor comments, probably worth addressing.

pubsub/README.md
protolambda: Previous default for no-verification mode. But yes, this is confusing.

pubsub/README.md
raulk: Leaving aside that this verb doesn't exist in any English dictionary (😄), -ed and -ing forms of verbs that end in consonant + vowel + consonant duplicate the last consonant (stopped, dubbed, clubbed, batted, zapping, etc.), so this is correct.

pubsub/README.md
raulk: Adjusting.

pubsub/README.md
raulk: Adjusted text. The spec already says it's optional, but it wouldn't hurt to put the optionality attribute in the prose for each field. Will do that.

pubsub/README.md
raulk: It's already implied by the context. And all fields are optional except for topic ids, which makes this a bit pedantic, but will do.

pubsub/README.md
```diff
@@ -112,30 +115,34 @@ message Message {
 }
 ```
 
-The `from` field denotes the author of the message, note that this is not
-necessarily the peer who sent the RPC this message is contained in. This is
-done to allow content to be routed through a swarm of pubsubbing peers.
+The `optional` fields may be omitted, depending on the
+[signature policy](#message-signing) and
+[message ID function](#message-identification).
```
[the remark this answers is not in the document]
raulk: It's already implied by the protobuf:

image

pubsub/README.md
raulk: Done.

raulk:
@vyzo all comments addressed. Was your review an approval, subject to the edits you suggested? Or do you want to do another full review?

pubsub/README.md
```diff
@@ -112,30 +115,34 @@ message Message {
...
+
+The `from` field (optional) denotes the author of the message. This is the peer
+who initially authored the message, and NOT the peer who propagated it. Thus, as
+the message is routed through a swarm of pubsubbing peers, the original
+authorship is preserved.
+
+The `seqno` field (optional) is a 64-bit big-endian uint that is a linearly
+increasing number that is unique among messages originating from each given
```
arnetheduck, line 128: afair rust-libp2p puts a random number here? is linearly increasing a "hard" requirement?

vyzo: not really, it just has to be unique within the time cache window.

vyzo (approved):

pubsub/README.md
```diff
@@ -149,22 +156,171 @@ economics (see e.g.
...
 
+> **[[ Margin note: ]]** Protobuf serialization is non-deterministic/canonical,
+> and the same data structure may result in different, valid serialised bytes
+> across implementations, as well as other issues. In the near future, the
+> signature creation and verification algorithm will be made deterministic.
+
+**Signature verification.** There are two configurations possible:
+
+* `Strict`: either expect or not expect a signature.
+* `Lax` (legacy, insecure, underterministic, to be deprecated): accept a signed
```
arnetheduck, line 234: It would be nice to export some of this language to a separate section which simply says that all fields that are set must be valid for their type - `seqno` must be 8 bytes, if the signature is set it *must* be valid - these would be pre-conditions to even consider the message for propagation before the "modes" are considered - the "lax" vs "strict" distinction then becomes a requirement for a signature or not.

raulk: Sure, that would be a welcome improvement. All protobuf types are `bytes` and `string`s, which express no constrains in and of themselves.

raulk: Mind submitting a PR against master when this lands (i.e. now)?

PR #302, opened 2020-11-20
chore: add rendezvous recommendations

vasco-santos:
This PR adds recommendations to the rendezvous spec regarding how a rendezvous point should be configured. These were based on the discussions on the go implementation PR and current code. The discover limit seems a bit higher, but if we have smaller defaults on the client, we can keep it as 1000.

It also fixes an error response status to `E_INVALID_SIGNED_PEER_RECORD` since we are using signed peer records now and this is now yet implemented.

### DIFF
--- a/rendezvous/README.md
+++ b/rendezvous/README.md
@@ -21,6 +21,19 @@ and spec status.
 
 [lifecycle-spec]: https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md
 
+## Table of Contents
+
+- [Overview](#overview)
+- [Use Cases](#use-cases)
+   - [Replacing ws-star-rendezvous](#replacing-ws-star-rendezvous)
+   - [Rendezvous and pubsub](#rendezvous-and-pubsub)
+- [The Protocol](#the-protocol)
+   - [Registration Lifetime](#registration-lifetime)
+   - [Interaction](#interaction)
+   - [Proof of Work](#proof-of-work)
+   - [Protobuf](#protobuf)
+- [Recommendations for Rendezvous Points configurations](#recommendations-for-rendezvous-points-configurations)
+
 ## Overview
 
 The protocol described in this specification is intended to provide a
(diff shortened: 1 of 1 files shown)


rendezvous/README.md
```diff
@@ -94,6 +107,8 @@ greatly simplifies real time discovery. It also allows for pagination
 of query responses, so that large numbers of peer registrations can be
 managed.
 
+The rendezvous protocol runs over libp2p streams using `/rendezvous/1.0.0`.
```
mxinden, line 110:
```suggestion
The rendezvous protocol runs over libp2p streams using the protocol id `/rendezvous/1.0.0`.
```

rendezvous/README.md
```diff
@@ -234,3 +249,17 @@ message Message {
   optional DiscoverResponse discoverResponse = 6;
 }
 ```
+
+## Recommendations for Rendezvous Points configurations
+
+Rendezvous points should have well defined configurations to enable libp2p nodes running the rendezvous protocol to have friendly defaults, as well as to guarantee the security and efficiency of a Rendezvous point. This will be particularly important in a federation, where rendezvous points should share the same expectations.
```
mxinden, line 255: Nit pick: The rest of the specification is wrapped at 80 characters. Can you do the same here for consistency?

mxinden (commented):
:+1: on documenting recommendations. I am not familiar enough with the Rendezvous protocol to approve here. Who is familiar enough to do the review @vasco-santos?

rendezvous/README.md
vasco-santos: fixed

vasco-santos:
@vyzo is the author, but I would say that anyone in the interest group can also weight in

mxinden:
`@ interest-group - @ vasco-santos` (@daviddias, @whyrusleeping, @Stebalien, @jacobheun, @yusefnapora) can one of you review the suggested changes here?

mxinden (approved):
Merging here since there have been no concerns raised since the last round of pings. Thanks @vasco-santos.

PR #310, opened 2021-04-01
pubsub/gossipsub/README: Mark rust gossipsub v1.0 as complete

mxinden:
With https://github.com/libp2p/rust-libp2p/pull/898 and https://github.com/libp2p/rust-libp2p/pull/1720 I think it is safe to say that the Rust GossipSub v1.0 implementation is complete.

@AgeManning please correct me in case I am missing something.


AgeManning:
Yes, it is complete up to what is written in the spec. 

We don't have signed peer records, but that is not gossipsub-specific. 

Anyone wanting further features into our gossipsub, they can make issues to the repo and we can add them. I think we should mark it as complete in specs.

BigLep (approved):

PR #316, opened 2021-04-21
feat: drop RSA to SHOULD

### LINKED ISSUE #313: PeerId spec mandates supporting RSA
Can the spec be loosened to making RSA optional? It is not clear why there is such a wild zoo of supported identity algorithms, what amount of usage they have. Personally I think algorithms other than ed25519 should probably be deprecated at some point, since I don't see a good justification for supporting them.

Stebalien:
While its important for interop with IPFS, it's not really a _requirement_ for libp2p. This makes Ed25519 the "standard".

fixes #313

### DIFF
--- a/peer-ids/peer-ids.md
+++ b/peer-ids/peer-ids.md
@@ -117,9 +117,10 @@ Four key types are supported:
  - Secp256k1
  - ECDSA
 
-Implementations MUST support RSA and Ed25519. Implementations MAY support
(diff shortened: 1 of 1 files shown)


mxinden (approved):

aschmahmann (approved):

Stebalien:
Let's wait for @raulk but I expect we'll merge this at the end of next week.

raulk (approved):

PR #317, opened 2021-04-21
fix: specify max mplex stream ID size

### LINKED ISSUE #259: mplex: size of `id` unspecified
https://github.com/libp2p/specs/tree/master/mplex#message-format

the `id` field of the message is unlimited in size as specified - base128 does not impose a max size per se - practically though, if a `uint64` is used, there's 61 bits of id to play with and clients should support this (or 29, if we cap on 32-bits).

Stebalien:
NOTE: go-mplex _technically_ supports 61bit IDs, but those violate our unsigned-varint spec and may not work on all implementations.

fixes #259

### DIFF
--- a/mplex/README.md
+++ b/mplex/README.md
@@ -62,7 +62,8 @@ flag = header & 0x07
 id = header >> 3
 ```
 
-The maximum header length is 9 bytes (per the unsigned-varint spec). With 9 continuation bits and 3 message flag bits  the maximum stream ID is 60 bits (maximum value of `2^60 - 1`).
+- Implementations MUST support up to 28bit stream IDs (32bit headers pre-encoding, 5 bytes after varint encoding).
+- Implementations SHOULD support up to 60bit stream IDs (64bit headers pre-encoding, 9 bytes after varint encoding).
 
 ### Flag Values
 
--- a/mplex/README.md
+++ b/mplex/README.md
@@ -62,7 +62,7 @@ flag = header & 0x07
 id = header >> 3
 ```
 
-The maximum length is 9 bytes (per the unsigned-varint spec) and the maximum stream ID is 60 bytes (`2^60 - 1`).
+The maximum header length is 9 bytes (per the unsigned-varint spec). With 9 continuation bits and 3 message flag bits  the maximum stream ID is 60 bits (maximum value of `2^60 - 1`).
 
 ### Flag Values
 
--- a/mplex/README.md
+++ b/mplex/README.md
@@ -54,13 +54,15 @@ Implementations in:
 
 Every communication in mplex consists of a header, and a length prefixed data segment.
 
-The header is an unsigned base128 varint, as defined in the [protocol buffers spec](https://developers.google.com/protocol-buffers/docs/encoding#varints). The lower three bits are the message flags, and the rest of the bits (shifted down by three bits) are the stream ID this message pertains to:
+The header is an [unsigned base128 varint](https://github.com/multiformats/unsigned-varint). The lower three bits are the message flags, and the rest of the bits (shifted down by three bits) are the stream ID this message pertains to:
 
 ```
 header = readUvarint()
 flag = header & 0x07
 id = header >> 3
 ```
+
+The maximum length is 9 bytes (per the unsigned-varint spec) and the maximum stream ID is 60 bytes (`2^60 - 1`).
 
 ### Flag Values
 


mplex/README.md
```diff
@@ -54,14 +54,16 @@ Implementations in:
...
-The header is an unsigned base128 varint, as defined in the [protocol buffers spec](https://developers.google.com/protocol-buffers/docs/encoding#varints). The lower three bits are the message flags, and the rest of the bits (shifted down by three bits) are the stream ID this message pertains to:
+The header is an [unsigned base128 varint](https://github.com/multiformats/unsigned-varint). The lower three bits are the message flags, and the rest of the bits (shifted down by three bits) are the stream ID this message pertains to:
 
 ```
 header = readUvarint()
 flag = header & 0x07
 id = header >> 3
 ```
 
+The maximum length is 9 bytes (per the unsigned-varint spec) and the maximum stream ID is 60 bytes (`2^60 - 1`).
```
mxinden, line 65:
```suggestion
The maximum length is 9 bytes (per the unsigned-varint spec) and the maximum stream ID is 60 bits (`2^60 - 1`).
```

mxinden, line 65:
```suggestion
The maximum header length is 9 bytes (per the unsigned-varint spec). With 9 continuation bits and 3 message flag bits  the maximum stream ID is 60 bits (maximum value of `2^60 - 1`).
```

Might make it easier for people to understand where `60 bits` is coming from.

mxinden (commented):
For the reference, today rust-libp2p restricts the whole mplex header to an unsigned base32 varint with 1 varint continuation bit, 3 bits for the flags and 28 bits for the stream ID. 

https://github.com/libp2p/rust-libp2p/blob/29d969be340b610e19bf05661b731883f25e6821/muxers/mplex/src/codec.rs#L213-L214

Do I understand correctly that we deem `2^28` bit stream ID space to small for long running nodes?

mxinden:
> > Do I understand correctly that we deem 2^28 bit stream ID space to small for long running nodes?
> 
> It _should_ be enough. But I don't see any reason to have a restriction like that given that we already need to decode up to 9 byte varints.

That depends on the implementation of the varint decoder right? E.g. in [unsigned-varint it will at most decode 4 bytes when decoding for a `u32`](https://github.com/paritytech/unsigned-varint/blob/8dda965bd4e7b95c484d5f4a6bef6156b376cc6b/src/decode.rs#L101-L107).

Either way, I am fine with allowing up to `2^60` stream IDs, especially since in most cases (`< 2^28`) less bytes are needed on the wire.

mxinden (approved):
Great. Thanks!

Stebalien:
> That depends on the implementation of the varint decoder right?

Yeah... it just seems like there's no good reason not to use 64bits. If you have a 32bit decoder, you probably have a 64bit one.

How about this:

1. Implementations MUST support up to 32bit headers.
2. Implementations MAY support up to 64bit headers.

Worst case, the connection will break after a few years...

Stebalien:
Changed. Does that work for you?

mxinden (approved):
> Yeah... it just seems like there's no good reason not to use 64bits. If you have a 32bit decoder, you probably have a 64bit one.

Agreed.

> Changed. Does that work for you?

Using (A) 32 bit, (B) 64 bit or (C) requiring 32 bit and recommending 64 bit all works fine for me.

arnetheduck:
I'm reading this PR and I'm not sure what it brings - introducing `SHOULD` for 64-bit values is of no value basically, because the whole purpose of libp2p is interoperability and in order to be interoperable, you must be able to depend on what goes on at the _other_ end - which with a `SHOULD` in the middle is.. 32 bits.

mxinden:
I understand your reasoning @arnetheduck, though I am having difficulties coming up with a probable use-case of `> 2^28` streams.

Again, I am fine with either (A) 32 bit, (B) 64 bit or (C) requiring 32 bit and recommending 64 bit. If you feel strongly about a specific option @arnetheduck I am fine with altering the wording once more, as long as @Stebalien agrees as well.

tomaka:
I think @arnetheduck's point is that option (C) doesn't make sense.
If a node A supports 64 bits substreams and another node B only supports 32 bits substreams, then node A has no choice but to only generate 32 bits substreams just in case.

Stebalien:
In almost all cases, the connection will break before we form `2**28` connections. My assumption is that a peer will break the connection after it exhausts all stream IDs.

From an interop standpoint, this should be fine. If at least one peer supports `2**28` stream IDs, the peers will be able to interop up to the point where they open `2**28` streams. If both happen to support `2**60` streams, they'll just be able to keep the connection open longer.

Also note: SHOULD specifically means "all implementations SHOULD do this unless they know the repercussions of not doing so". REQUIRED/MUST means "you need to do this to interop", which isn't the case here.

tomaka:
> In almost all cases, the connection will break before we form 2**28 connections. My assumption is that a peer will break the connection after it exhausts all stream IDs.

You seem to also assume that stream IDs start at 0 and increase linearly, while I don't think the specs mention this.

> Also note: SHOULD specifically means "all implementations SHOULD do this unless they know the repercussions of not doing so". REQUIRED/MUST means "you need to do this to interop", which isn't the case here.

What I don't understand is: why not settle on either 32 or 64? Why leave the door open for uncertainties?

Stebalien:
> You seem to also assume that stream IDs start at 0 or 1 and increase linearly, while I don't think the specs mention this.

Fair enough.

> What I don't understand is: why not settle on either 32 or 64? Why leave the door open for uncertainties?

I want implementations to be _able_ to support up to 64 bit headers. Otherwise, connections will break after a few hundred thousand streams.

But yeah, there's no reason not to just require 64bit headers (other than the fact that rust will need to make a change).

mxinden:
> But yeah, there's no reason not to just require 64bit headers (other than the fact that rust will need to make a change).

I don't think the quirks of a single implementation should influence our choice here (said as one of the maintainers of rust-libp2p). At this point, to simplify the spec, I am in favor of just requiring implementations to support a 64bit header.

Stebalien:
https://github.com/libp2p/specs/pull/322

PR #308, opened 2021-03-26
secio: Transition SECIO spec from Active to Deprecated

mxinden:
The SECIO security transport has been deprecated [1]. This commit
updates the specification accordingly.

> The transition from the `Active` status to the `Deprecated` status is
performed via a *Pull Request* that is reviewed by the _Interest Group_,
allowing 5 working days to elapse to collect feedback from the libp2p
community at large.

As per the specification life cycle specification I am requesting a review
from @Stebalien, @richardschneider, @tomaka and @raulk.

[1] https://blog.ipfs.io/2020-08-07-deprecating-secio/

PR #320, opened 2021-04-27
connections: Replace SECIO with Noise as baseline sec proto

mxinden:
See https://github.com/libp2p/specs/pull/308 for SECIO deprecation status.


vasco-santos (approved):
LGTM

Issue #324, opened 2021-05-04
[identify] Handling of asymetrical protocols


D4nte:
Given an asymmetrical protocol `foo/bar` where Alice only supports initiating a query to Bob and Bob only supports responding to a request of said protocol.

Currently, the identify protocol does not include valuable information. Indeed, the specs reads:

> protocols
> This is a list of protocols supported by the peer.

In the case of Carole, who only support initiating query of `foo/bar` protocol. Once she finishes the identify protocol with Alice, both Carole and Alice may think they can initiate the query with the other, when it is not the fact.

One idea I had was simply for Alice and Carole to not include `foo/bar` in the `Identify` message they send. I am not sure of the ramifications of such design.

What are your thoughts on the matter?

mxinden:
> One idea I had was simply for Alice and Carole to not include foo/bar in the Identify message they send. I am not sure of the ramifications of such design.

Yes, this is the libp2p best practice. Implementations doing this today:

- The (go-)libp2p Kademlia implementation differentiates in client mode and server mode. In client mode a node sends queries but does not answer them. In client mode the node would not include `/ipfs/kad/1.0.0` in its _identify_ protocol list nor its [`ls` multistream-select responses](https://github.com/multiformats/multistream-select#listing). See also https://github.com/libp2p/rust-libp2p/issues/2032.

- The [circuit relay v2](https://github.com/libp2p/specs/issues/314) protocol differentiates nodes in two roles: _clients_ which are at  either end of a relayed connection and _relays_ which are in-between relaying the data between two _clients_. The two _clients_ would advertise the `/libp2p/circuit/relay/0.2.0/stop` protocol, the _relay_ would advertise the `/libp2p/circuit/relay/0.2.0/hop` protocol in the _identify_ protocol.

Does the above make sense to you @D4nte? If so, would you mind sending in a patch to update the _identify_ protocol specification in this repository?

D4nte:
Great answer, thank you very much.

D4nte:
Closing, will reopen if further questions arise.

PR #327, opened 2021-05-14
identify: Document identify behavior for asymetrical protocols

mxinden:
> Note on asymmetrical protocols: Assume an asymmetrical
request-response style protocol `foo` where some clients only support
initiating requests while some servers (only) support responding to
requests. To prevent clients from initiating requests to other clients,
which given them being clients they fail to respond, clients should not
list `foo` in their `protocols` list.

Question emerged in https://github.com/libp2p/specs/issues/324.

### DIFF
--- a/identify/README.md
+++ b/identify/README.md
@@ -120,3 +120,9 @@ observable source address.
 ### protocols
 
 This is a list of protocols supported by the peer.
+
+Note on asymmetrical protocols: Assume an asymmetrical request-response style
+protocol `foo` where some clients only support initiating requests while some
+servers (only) support responding to requests. To prevent clients from
+initiating requests to other clients, which given them being clients they fail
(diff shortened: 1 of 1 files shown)


mxinden:
@marten-seemann do you have some time to review this pull request?

Stebalien (approved):
This works, but it's a bit confusing. I'd talk about this in terms of inbound streams. E.g.,

> A node should only advertise a protocol if it's willing to _receive_ inbound streams on that protocol.

mxinden:
> This works, but it's a bit confusing. I'd talk about this in terms of inbound streams. E.g.,
> 
> > A node should only advertise a protocol if it's willing to _receive_ inbound streams on that protocol.

:+1: Adjusted the paragraph with your excerpt at first, followed by a small example.

Stebalien (approved):

PR #333, opened 2021-06-15
Update Envelope and PeerRecord protobufs to latest version

### LINKED ISSUE #329: [Signed Envelopes] Protobuf description not valid
I've attempted to implement signed envelopes in the Rust libp2p implementation and noticed the following:

1. The `signature` field is given the number 5 but there is no number 4. Is this an overlook from an earlier iteration where there used to be another field?
2. The fields are missing an `optional` or `required` specifier. Without that, the protobuf doesn't actually compile.

Happy to submit a patch if there is consensus on what the protobuf should be changed to!

thomaseizinger:
These protobuf definitions are taken from the go-libp2p repo.

Fixes #329.

cc @mxinden

### DIFF
--- a/RFC/0002-signed-envelopes.md
+++ b/RFC/0002-signed-envelopes.md
@@ -63,11 +63,33 @@ can use protobuf for this as well and easily embed the key in the envelope:
 
 
 ```protobuf
+syntax = "proto3";
+
+package record.pb;
+
+// Envelope encloses a signed payload produced by a peer, along with the public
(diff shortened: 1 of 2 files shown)


mxinden (approved):
Thank you Thomas! This looks good to me. I want to give @vasco-santos a chance to review as well to ensure js-libp2p is aligned too.

vasco-santos (approved):
All good from JS side 👌🏼 thanks for pinging

PR #338, opened 2021-06-21
[Rendezvous] Change `ttl` and `limit` to `uint64`

thomaseizinger:
To be meaningful, TTL has to be a positive integer hence the use of
an signed integer does not bring any value and only introduces
unnecessary error handling.

### DIFF
--- a/rendezvous/README.md
+++ b/rendezvous/README.md
@@ -2,7 +2,7 @@
 
 | Lifecycle Stage | Maturity      | Status | Latest Revision |
 |-----------------|---------------|--------|-----------------|
-| 1A              | Working Draft | Active | r1, 2019-01-18  |
+| 1A              | Working Draft | Active | r2, 2021-06-24  |
 
 Authors: [@vyzo]
 
--- a/rendezvous/README.md
+++ b/rendezvous/README.md
@@ -230,7 +230,7 @@ message Message {
 
   message Discover {
     optional string ns = 1;
-    optional int64 limit = 2;
+    optional uint64 limit = 2;
     optional bytes cookie = 3;
   }
 
--- a/rendezvous/README.md
+++ b/rendezvous/README.md
@@ -214,13 +214,13 @@ message Message {
   message Register {
     optional string ns = 1;
     optional bytes signedPeerRecord = 2;
-    optional int64 ttl = 3; // in seconds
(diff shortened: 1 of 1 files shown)


mxinden:
I am guessing you meant to write:

```diff
- hence the use of an unsigned integer does not bring any value
+ hence the use of a signed integer does not bring any value
```

I agree, though I don't think we can just alter the type as that would be a breaking protobuf change while this protocol is already used in the wild. @vyzo @vasco-santos can you comment here?

thomaseizinger:
> I am guessing you meant to write:
> 
> ```diff
> - hence the use of an unsigned integer does not bring any value
> + hence the use of a signed integer does not bring any value
> ```
> 
> I agree, though I don't think we can just alter the type as that would be a breaking protobuf change while this protocol is already used in the wild. @vyzo @vasco-santos can you comment here?
> 

Yep will fix that!

Re breaking change: How does the protocol being a Draft affect this? I would have assumed that prior to being finalized, we can make breaking changes. It is not particularly nice for implementations but the way I see it, early adoptions of protocols have to accept that?

thomaseizinger:
@mxinden I've changed the commit message to say "signed integer".

thomaseizinger:
In the same spirit, I changed `limit` of `Discover` to a `uint64`.

mxinden (approved):
After a bit of research it seems like the rendezvous protocol isn't used anywhere yet, thus the breaking changes here are fine by me. Thanks for cleaning these types up Thomas!

Would you mind bumping the spec revision?

```diff
diff --git a/rendezvous/README.md b/rendezvous/README.md
index 4d98b2f..9277d40 100644
--- a/rendezvous/README.md
+++ b/rendezvous/README.md
@@ -2,7 +2,7 @@
 
 | Lifecycle Stage | Maturity      | Status | Latest Revision |
 |-----------------|---------------|--------|-----------------|
-| 1A              | Working Draft | Active | r1, 2019-01-18  |
+| 1A              | Working Draft | Active | r2, 2021-06-24  |
 
 Authors: [@vyzo]
```

thomaseizinger:
> Would you mind bumping the spec revision?

Done!

mxinden:
Thanks @thomaseizinger!

Issue #222, opened 2019-10-25
Simple, minimal peer exchange protocol


raulk:
This issue proposes a general-use peer exchange protocol, that is not embedded in any specific protocol like gossipsub/episub.

The goal of PEX is to enable peers to share records about other peers they're connected to in a 1:1, ad-hoc fashion. It does not intend to produce deterministic results like DHTs, nor does it rely on a structured network or shared heuristic. [BitTorrent uses PEX](https://en.wikipedia.org/wiki/Peer_exchange) to streamline tracker-less peer discovery.

In the context of gossipsub, PEX proves useful to find additional peers in a topic we subscribe to, as a way of strengthening our topic mesh. Through subscription beaconing (ie. peers gossiping about which topics they're subscribed to), it can even be possible to bootstrap a topic subscription without hitting the DHT, or other structured discovery mechanisms at all.

I'm thinking we should spec out a minimal PEX protocol, consisting of a simple advertisement schema, and two operations: advertise, lookup.

## Advertisement schema

An _advertisement struct_ consists of a [peer address record](https://github.com/libp2p/specs/pull/217) and a set of CIDs we are advertising, signed by the peer's key to prevent MITM attacks.

## Local advertisement record maintenance

Our advertisement record is kept in memory and updated at runtime. It is populated with:

* our peer ID.
* own addresses (which may change over time; the PEX protocol can subscribe to updates via the eventbus).
* advertised CIDs.

The `Host` API would expose methods so that downstream components (e.g. protocols) can manage advertised CIDs, e.g.:

```go
// We don't want to add an accessor for PEX in the Host interface. The host-service refactor is a prerequisite to be able to do this.
svc, ok := host.GetService(&PEX{})
if !ok {
    return nil
}
pexsvc := svc.(PEXService)
ad := pex.NewAd("gossipsub:topic_name")
cancelFn, err := pexsvc.Advertise(ad)
if err != nil {
    return err
}

// ... store the cancelFn in state ...

// atomically replace the advertised value, possibly not useful for gossipsub, but it will be for other protocols. Helps mitigate add/remove noise when sending deltas.
ad.Replace("gossipsub:topic_name_b")

// when done / closing down
err := cancelFn()
if err != nil {
    return err
}
```

## Advertise operation

Upon establishing a libp2p connection:

1. We open a stream for protocol ID `/libp2p/pex/v01`.
2. If successful, we push our local advertisement record.
3. When receiving a record, we store it in memory.

[...]

PR #339, opened 2021-06-22
ROADMAP.md: Add high level project roadmap

mxinden:
This documents the libp2p short and long term roadmap based on
previous work (e.g. see [libp2p/libp2p/60], [ipfs/roadmap/18],
[libp2p/specs/archive]).

#### Status

Feedback from the whole community is welcome.

Fixes https://github.com/libp2p/libp2p/issues/68.

[libp2p/libp2p/60]: https://github.com/libp2p/libp2p/issues/60
[ipfs/roadmap/18]: https://github.com/ipfs/roadmap/issues/18
[libp2p/specs/archive]: https://github.com/libp2p/specs/tree/master/_archive

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -19,12 +19,12 @@ third-party ownership of data.
         - [🚑 Self-healing networks](#🚑-self-healing-networks)
         - [📮 Offline message queue / postbox](#📮-offline-message-queue--postbox)
         - [🤖 libp2p as a WASM library](#🤖-libp2p-as-a-wasm-library)
-        - [⚙️ libp2p as a kernel module](#⚙️-libp2p-as-a-kernel-module)
         - [♻️ Packet switching](#♻️-packet-switching)
     - [Evolve](#evolve)
         - [🤝 Low latency, efficient connection handshake](#🤝-low-latency-efficient-connection-handshake)
         - [🕸 Unprecedented global connectivity](#🕸-unprecedented-global-connectivity)
         - [Standardized interaction of different protocols via Smart Records](#standardized-interaction-of-different-protocols-via-smart-records)
+        - [Peer Routing Records](#peer-routing-records)
         - [💣 Attack resistance, threat models and security](#💣-attack-resistance-threat-models-and-security)
         - [📈 Proving we are scalable and interoperable](#📈-proving-we-are-scalable-and-interoperable)
         - [🌐 Browser use cases](#🌐-browser-use-cases)
@@ -217,24 +217,6 @@ model.
 **Links:**
 
 - [WASM support in rust-libp2p](https://github.com/libp2p/rust-libp2p/issues/23).
-
-### ⚙️ libp2p as a kernel module
-
-**What?** Native libp2p implementations are proliferating, maturing and
-gaining adoption in different languages. This is arguably a sign that
-developers accept the concepts, semantics and APIs that libp2p
-introduces. If we stay on track, libp2p may become the de-facto standard
-library for P2P networking.
-
-When that happens, it'll make sense to push our constructions down one level
-into OS kernels. This can entail first incubating external modules, to
-eventually propose merging them into the main codebases of kernels like Linux.
-An intermediary step would be to run libp2p as a daemon, providing peer-to-peer
-network capabilities to all processes on a machine.
-
-**Why?** To ship libp2p as the default P2P network API inside future
-operating systems. Actually, taking it one step further, make the libp2p
-stack *THE default* networking stack.
 
 ### ♻️ Packet switching
 
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -19,7 +19,6 @@ third-party ownership of data.
         - [🚑 Self-healing networks](#🚑-self-healing-networks)
         - [📮 Offline message queue / postbox](#📮-offline-message-queue--postbox)
         - [🤖 libp2p as a WASM library](#🤖-libp2p-as-a-wasm-library)
-        - [♻️ Packet switching](#♻️-packet-switching)
     - [Evolve](#evolve)
         - [🤝 Low latency, efficient connection handshake](#🤝-low-latency-efficient-connection-handshake)
         - [🕸 Unprecedented global connectivity](#🕸-unprecedented-global-connectivity)
@@ -218,34 +217,6 @@ model.
 
 - [WASM support in rust-libp2p](https://github.com/libp2p/rust-libp2p/issues/23).
 
-### ♻️ Packet switching
-
-**What?** A packet-switched libp2p is on the pathway to achieving
-decentralisation and seamless connectivity across heterogeneous networks
-across the globe, that are running on different transports and not
-necessarily connected to the Internet directly. Essentially it's a
-*global transit system* for libp2p traffic.
-
-For illustration, a TCP packet sent via an IP-based local LAN in
-the US connected to the Internet backbone (via a router) could travel
-through the libp2p network, taking multiple hops along the way based on
-routing decisions, and traversing various networks over different
-transports. Eventually it'll reach its destination in Nepal: a node in a
-mesh network, that's in turn connected to a community network with
-intermittent Internet connection via satellite.
-
-To build a reality like this entails doing away with the concept of
-Autonomous Systems (AS), and rethinking the Internet protocols and
-structures from the viewpoint of peer-to-peer: BGP, OSPF, peering,
-transit networks, multihoming, IANA, etc.
-
-**Why?** It's a crucial requirement in our visionary goal of
-decentralising networks. WIP.
-
-**Links:**
-
--   [On \"DialMe\", NAT traversal, Relay, line switching, packet switching
-    (libp2p protocols)](https://github.com/ipfs/notes/issues/110)
 
 ## Evolve
 
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -22,7 +22,6 @@ third-party ownership of data.
     - [Evolve](#evolve)
         - [🤝 Low latency, efficient connection handshake](#🤝-low-latency-efficient-connection-handshake)
         - [🕸 Unprecedented global connectivity](#🕸-unprecedented-global-connectivity)
-        - [Standardized interaction of different protocols via Smart Records](#standardized-interaction-of-different-protocols-via-smart-records)
         - [Peer Routing Records](#peer-routing-records)
         - [🗣️ Polite peering](#🗣️-polite-peering)
         - [🧱 Composable routing](#🧱-composable-routing)
@@ -275,19 +274,6 @@ rest of the system.
   vision](https://github.com/mxinden/specs/blob/hole-punching/connections/hole-punching.md).
 
 - [NAT traversal tracking issue](https://github.com/libp2p/specs/issues/312).
-
-### Standardized interaction of different protocols via Smart Records
-
-TODO: Extend
-
-Copied and adapted from: https://github.com/libp2p/go-smart-record/blob/4a1b33c6d7cb2fb38d166abc1023fbbbab33b60f/README.md
-
-**What?** With Smart Records (SRs) we generalize DHT's key/value put/get as a
-separate protocol that can be leveraged by any other protocol (including DHT
-protocols) to store arbitrary data.
-
-**Why?** We currently don't have a standardized, shared, public medium decoupled
-from the transport protocol for the interaction of different protocols.
 
 ### Peer Routing Records
 
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -308,7 +308,7 @@ have the highest probability of being able to establish a direct connections.
 Mechanisms we wish to add include:
 
 - Project Flare stack (via *Circuit Relay v2*, *Direct Connection Upgrade
-  through Relay*, *AutoNAT*, ...)
+  through Relay*, *AutoNAT*, *Stream Migration*, ...)
 
 - WebRTC
 
@@ -468,7 +468,7 @@ protocol](https://github.com/multiformats/multigram/tree/draft1).
 
 Progressing with this work entails designing the abstractions, and
 implementing, at least, a UDP transport, [as the community has
-demanded]](https://github.com/libp2p/go-libp2p/issues/353).
+demanded](https://github.com/libp2p/go-libp2p/issues/353).
 
 **Why?** Current the libp2p API precludes us from modelling
 message-oriented transports like UDP or Bluetooth.
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -578,3 +578,10 @@ like RPC to developers they should be able to more efficiently develop
 their projects. Additionally, easy to use high level communication
 frameworks should allow developers not very familiar with low-level
 networking to get started with libp2p.
+
+**Links**
+
+- See Rust's [libp2p-request-response] as one possible communication pattern
+  abstraction.
+
+[libp2p-request-response]: https://docs.rs/libp2p-request-response/0.11.0/libp2p_request_response/
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -324,10 +324,11 @@ identity, etc. may simplify things. An open question is where
 Multistream would fit into this model: could we conflate it into the
 Wire protocol?
 
-**Why?** Peers act haphazardly when observed from the outside, as they
-do not have the means to act more collaboratively. We are lacking a wire
-protocol that governs the connection between two peers. Negotiating lots
-of piecemeal protocols is inefficient, especially for essential stuff.
+**Why?** Peers act haphazardly when observed from the outside, leading e.g. to
+unnecessary connection churn. They do not have the means to act more
+collaboratively. We are lacking a wire protocol that governs the connection
+between two peers. Negotiating lots of piecemeal protocols is inefficient,
+especially for essential stuff.
 
 **Links:**
 
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -24,7 +24,8 @@ third-party ownership of data.
         - [🕸 Unprecedented global connectivity](#🕸-unprecedented-global-connectivity)
         - [Standardized interaction of different protocols via Smart Records](#standardized-interaction-of-different-protocols-via-smart-records)
         - [Peer Routing Records](#peer-routing-records)
-        - [🗣️ Polite peering and wire protocol](#🗣️-polite-peering-and-wire-protocol)
+        - [🗣️ Polite peering](#🗣️-polite-peering)
+        - [🧱 Composable routing](#🧱-composable-routing)
         - [💣 Attack resistance, threat models and security](#💣-attack-resistance-threat-models-and-security)
         - [📈 Proving we are scalable and interoperable](#📈-proving-we-are-scalable-and-interoperable)
         - [🌐 Browser use cases](#🌐-browser-use-cases)
@@ -318,6 +319,49 @@ chance to keep important connections open, etc.
 unnecessary connection churn. They do not have the means to act more
 collaboratively. We are lacking a wire protocol that governs the connection
 between two peers.
+
+### 🧱 Composable routing
+
+**What?** The composable routing framework has four "pillars**:
+
+* Routing syntax and language
+
+  The routing syntax is a common data model (with a user-facing syntactic
+  representation) for communicating routing-related information (requests and
+  responses). The [implementation of the routing
+  syntax](https://github.com/libp2p/go-routing-language/tree/master/syntax)
+  provides serialization and pretty printing of routing expressions.
+
+* Routing interface
+
+  A routing system processes routing requests and returns routing responses.
+  Routing requests and responses have semantics, which are unlike those of a
+  traditional request/response protocols, that implement function call arguments
+  and return values.
+
+* Smart Records
+
+  Smart Records (SR) is a technology that enables multiple decentralized parties
+  and protocols to randevouz and share information on a topic. Specifically, SR
+  is a key-value store service, which facilitates conflict-free updates to
+  values by multiple writers.
+
+* Subnets
+
+  At present, there is a single DHT network which spans across all IPFS nodes.
+  We would like to introduce a general standard for the co-existence of multiple
+  network instances with individual discovery, membership and content routing
+  semantics.
+
+**Why?** Enable interoperability of protocols and nodes with different versions
+and capabilities, developed by independent teams or organizations. As a
+consequence, it also enables middleware components (like caching, batching,
+predictive fetching, and so on** and the co-existence of different content
+finding protocols and subnetworks.
+
+**Links:**
+
+- [Tracking issue](https://github.com/libp2p/specs/issues/343)
 
 ### 💣 Attack resistance, threat models and security
 
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -97,7 +97,7 @@ use cases we are targeting at Protocol Labs.
 
 This implies building/supporting Java/Kotlin and Swift implementations
 -- either fully-fledged or trimmed-down versions that do the minimum
-necessary to enable other visionary goals.
+necessary or wrap existing implementations like rust-libp2p.
 
 **Why?** Bringing libp2p natively to mobile devices is an enabler for
 offline-first use cases.
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -24,12 +24,12 @@ third-party ownership of data.
         - [🕸 Unprecedented global connectivity](#🕸-unprecedented-global-connectivity)
         - [Standardized interaction of different protocols via Smart Records](#standardized-interaction-of-different-protocols-via-smart-records)
         - [Peer Routing Records](#peer-routing-records)
+        - [🗣️ Polite peering and wire protocol](#🗣️-polite-peering-and-wire-protocol)
         - [💣 Attack resistance, threat models and security](#💣-attack-resistance-threat-models-and-security)
         - [📈 Proving we are scalable and interoperable](#📈-proving-we-are-scalable-and-interoperable)
         - [🌐 Browser use cases](#🌐-browser-use-cases)
         - [🌡️ Opt-in telemetry protocol](#🌡️-opt-in-telemetry-protocol)
         - [📩 Message-oriented transports](#📩-message-oriented-transports)
-        - [🗣️ Polite peering and wire protocol](#🗣️-polite-peering-and-wire-protocol)
         - [☎️ Reducing the dial fail rate](#️-reducing-the-dial-fail-rate)
         - [🔀 Peer exchange protocol](#🔀-peer-exchange-protocol)
         - [🏹 RPC and other common node communication patterns](#🏹-rpc-and-other-common-node-communication-patterns)
@@ -217,7 +217,6 @@ model.
 
 - [WASM support in rust-libp2p](https://github.com/libp2p/rust-libp2p/issues/23).
 
-
 ## Evolve
 
 **Our short-term roadmap**.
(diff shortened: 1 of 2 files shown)


marten-seemann (commented):
Thank you @mxinden for this **fantastic** document!

mxinden:
> Thank you @mxinden for this **fantastic** document!

Thanks and also thanks for the quick feedback!

I want to stress once more that this is based on previous work, thus directing this praise to people like @mgoelzer and @raulk.

ROADMAP.md
```diff
@@ -0,0 +1,577 @@
...
+implementations across different languages in sync, i.e. create more
+value and impact. WASM also presents a huge vector for making libp2p the
+de-facto networking stack, for cases where WASM is a suitable deployment
+model.
+
+**Links:**
+
+- [WASM support in rust-libp2p](https://github.com/libp2p/rust-libp2p/issues/23).
+
+### ⚙️ libp2p as a kernel module
```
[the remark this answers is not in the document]
mxinden: Agreed.

> An intermediary step would be to run libp2p as a daemon, providing peer-to-peer
network capabilities to all processes on a machine.

I do see value in the proposed intermediary step of providing libp2p as a daemon, though purely out of practical reasons, e.g. de-duplication by only running a single Kademlia instance per machine.

That said, I am in favor of removing the whole "libp2p as a kernel module" section. @marten-seemann what do you think?

ROADMAP.md
```diff
@@ -0,0 +1,577 @@
...
+sneaker nets, and making IPFS work in Mars where general
+connectedness to the Internet does not exist. *The network should pave
+the way for IPFS to evolve in that direction.*
+
+### 🤫 A spyproof libp2p
+
+**What? Why?** Supporting transports like cjdns, I2P and Tor (see
+[OpenBazaar/go-onion-transport](https://github.com/OpenBazaar/go-onion-transport))
+helps create resilient, privacy-centric communications, and moves the
+needle forward towards censorship-resistant networks.
```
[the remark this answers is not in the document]
mxinden: :+1: see e6cb366b72b2ac8a4e1aab6604d7195a447d74e8. Happy for additional suggestions.

ROADMAP.md
```diff
@@ -0,0 +1,577 @@
...
+**What?** A DHT crawl measurements (Nov 22nd 2019) showed that out
+of 4344 peers, 2754 were undialable (\~63%). This evidence correlates
+with feedback from the IPFS and Filecoin teams.
+
+We need to implement additional mechanisms for Firewall and NAT traversal to
+have the highest probability of being able to establish a direct connections.
+Mechanisms we wish to add include:
+
+- Project Flare stack (via *Circuit Relay v2*, *Direct Connection Upgrade
+  through Relay*, *AutoNAT*, ...)
```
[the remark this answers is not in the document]
mxinden: See 10f347917b4b1e21642ad6d6283ab636d3e6601c.

ROADMAP.md
```diff
@@ -0,0 +1,579 @@
...
+when securing connections.
+
+### 📱 libp2p in mobile devices
+
+**What?** While js-libp2p technically works in mobile browsers, it is
+constrained by the web sandbox. First-class integration of native libp2p
+in mobile devices could be a prerequisite for the class of offline-first
+use cases we are targeting at Protocol Labs.
+
+This implies building/supporting Java/Kotlin and Swift implementations
```
BigLep, line 98: Are any customers asking for this?  Have any contributors created these?  If not, how come?

ROADMAP.md
```diff
@@ -0,0 +1,579 @@
...
+### 💡 libp2p in IoT
+
+**What?** Distinct from consumer mobile devices like smartphones, IoT
+devices like embedded sensors are a potential use case for libp2p. These
+devices are typically more resource constrained and may have
+intermittent connectivity. They also tend to do minimal processing
+locally, but transmit frequently to a cloud where data analysis occurs.
+
+**Why?** IoT can greatly benefit from peer-to-peer networking. Supporting IoT
+environments with libp2p greatly expands libp2p's userbase on the one side and
```
BigLep, line 114: Are we aware of certain customers asking for this?  What has held them back from developing it?

ROADMAP.md
```diff
@@ -0,0 +1,579 @@
...
+should be as cheap and fast as possible. Supporting things like
+*selective* stream opening, *preemption*, *speculative* negotiation,
+*upfront* negotiation, protocol table *pinning*, etc. may enable us to
+achieve lower latencies when establishing connections, and even the
+0-RTT holy grail in some cases. These features are being discussed in
+the *Protocol Select* protocol design.
+
+**Why?** Multistream 1.0 is chatty and naïve. Streams are essential to
+libp2p, and negotiating them is currently inefficient in a number of
+scenarios. Also, bootstrapping a multiplexed connection is currently
```
BigLep, line 288: I assume some of this information is in the docs below, but I think it would be good to call out a couple of numbers here.  For example, (this is me making something up), "The p50 to create a bootstrapped multiplex connect between IPFS nodes currently takes Xms from Y round trips.  With improvements to this space, we believe we can reduce this to Ams and B round trips.  This is important because IPFS nodes establish these connections each time connecting to a peer for doing a lookup or initiating a bitswap session.  The p50 of time to first byte of content download for a cold query will reduce from Cms to Dms."

ROADMAP.md
```diff
@@ -0,0 +1,579 @@
...
+
+- [Multiselect 2.0 design doc](https://github.com/libp2p/specs/pull/205).
+
+- [Multiselect 2.0 specification](https://github.com/libp2p/specs/pull/227)
+
+- [RFC Multistream 2.0](https://github.com/libp2p/specs/pull/95).
+
+### 🕸 Unprecedented global connectivity
+
+**What?** A DHT crawl measurements (Nov 22nd 2019) showed that out
```
BigLep, line 302: Can we link to where this data was collected?
Also, can this be an automated process that runs periodically (e.g., weekly, monthly) so we can see improvements to this over time?

ROADMAP.md
```diff
@@ -0,0 +1,579 @@
...
+TODO: Extend
+
+Copied and adapted from: https://github.com/libp2p/go-smart-record/blob/4a1b33c6d7cb2fb38d166abc1023fbbbab33b60f/README.md
+
+**What?** With Smart Records (SRs) we generalize DHT's key/value put/get as a
+separate protocol that can be leveraged by any other protocol (including DHT
+protocols) to store arbitrary data.
+
+**Why?** We currently don't have a standardized, shared, public medium decoupled
+from the transport protocol for the interaction of different protocols.
```
BigLep, line 337: Likely newbie question here, but what will this unlock if we have this?

ROADMAP.md
```diff
@@ -0,0 +1,579 @@
...
+
+### 💣 Attack resistance, threat models and security
+
+**What?** This is an overarching element of the roadmap, and it affects
+everything else. Activities under this endeavour include:
+
+1. *Past and present.* Pragmatically analysing and profiling the
+   exposure of each component against a taxonomy of attacks and
+   threat models, e.g. DDoS, byzantine, sybil, poisoning, eclipse,
+   flooding, etc.
```
BigLep, line 347: I assume this is something we could contract out right?

ROADMAP.md
```diff
@@ -0,0 +1,579 @@
...
+been discussed](https://github.com/libp2p/specs/issues/71)
+in the past, and there is even a [draft proposal for a multigram
+protocol](https://github.com/multiformats/multigram/tree/draft1).
+
+Progressing with this work entails designing the abstractions, and
+implementing, at least, a UDP transport, [as the community has
+demanded](https://github.com/libp2p/go-libp2p/issues/353).
+
+**Why?** Current the libp2p API precludes us from modelling
+message-oriented transports like UDP or Bluetooth.
```
BigLep, line 474: Newbie here but I'm not following what kind of end uses cases this would unlock.

ROADMAP.md
```diff
@@ -0,0 +1,579 @@
...
+special status. Similarly, PING is a housekeeping protocol.
+
+Grouping all these essential messages into a solid *Wire protocol* that
+covers connection lifecycle, heartbeats, health, status, diagnostics,
+identity, etc. may simplify things. An open question is where
+Multistream would fit into this model: could we conflate it into the
+Wire protocol?
+
+**Why?** Peers act haphazardly when observed from the outside, as they
+do not have the means to act more collaboratively. We are lacking a wire
```
BigLep, line 494: I'm assuming this is obvious to folks in the P2P world, but can we state the ramifications of peers acting haphazardly?  What will be the benefit if this work gets done?

ROADMAP.md
```diff
@@ -0,0 +1,579 @@
...
+
+   - Users should also have the capability to customise how addresses
+     are resolved, how dials are throttled, how connections are
+     selected if multiple dials succeeded, etc. In a nutshell, the
+     dialing behaviour should be fully customisable per host.
+
+-  **Context-sensitive advertisement of self-addresses:** private
+   addresses should not leak into public DHTs, both for connectivity
+   and security reasons (cf. IPFS DHT pollution in 2016). Similarly,
+   local discovery mechanisms should only transmit private addresses.
```
BigLep, line 523: While this may lead to dial failures, this also seems like a separable topic on its own.

ROADMAP.md
```diff
@@ -0,0 +1,579 @@
...
+-   [Discuss: batch vs. on-demand connection
+    pruning](https://github.com/libp2p/go-libp2p-connmgr/issues/19).
+
+
+### ☎️ Reducing the dial fail rate
+
+**What?** Reducing the dial fail rate implies improvements in several
+areas.
+
+-  **Outbound dials:** We should strategise dials based on context, or based on
```
BigLep, line 509: It would be good to discuss what metrics we expect to be improved by this work.  If we don't have those metrics, then that should get added to the project.

ROADMAP.md
```diff
@@ -0,0 +1,579 @@
...
+  [Improve NAT traversal](#improve-nat-traversal).
+
+**Why?** Dialing is a crucial function of our system and using a
+non-deterministic, scattershot approach is inefficient.
+
+**Links:**
+
+- [Reliable dialing](https://github.com/libp2p/libp2p/issues/27).
+
+### 🔀 Peer exchange protocol
```
BigLep, line 541: Maybe move this one up since it's actively being worked on?

ROADMAP.md
```diff
@@ -0,0 +1,579 @@
...
+
+Copied and adapted from:
+[https://github.com/libp2p/notes/issues/3](https://github.com/libp2p/notes/issues/3)
+
+**What?** We currently use mDNS to discover local peers but it would
+also be nice if we could have a peer exchange protocol for discovering
+even more local peers.
+
+**Why?** (1) relying on the global DHT to discover local peers is
+problematic; (2) ideally, we want to connect to all peers on the local
```
BigLep, line 551: Lets spell out why more.  Is it because it's slow, gives incomplete results, etc.?

ROADMAP.md
```diff
@@ -0,0 +1,579 @@
...
+network for UX reasons (if we\'re in the same room, we should be able to
+share files with each-other). [This ties into offline and decentralised
+networks.](#kix.xlvu7ikvo1jg)
+
+### 🏹 RPC and other common node communication patterns
```
BigLep, line 556: This one seems interesting, particularly from it's ability to unlock others.  

I think it could benefit from something like:

```
This is what a customer (developer) has to do today to satisfy usecase Y.
...List the various steps/commands...

If we do this work, they will instead only need to do:
...List the various steps/commands...
```

BigLep (changes requested):
Lots of good stuff here.  Thanks @mxinden and @marten-seemann for the efforts so far (even if a good chunk of it was bringing forward work done by others in the past).

A couple of thoughts:

## 1
This document right now is a good list of themes (I think the visionary items particular fall in this camp) and items.  

Using the map analogy, we ultimately need to be able to create/show our route on the map.  Right now we have a list of possible destinations, but no indication of costs to get there, impact, and ordering.  

We can talk offline (or here) on how we get there, but we're going to need something "spreadsheet-like" where we can order, sort, etc.

## 2
This document is covering scope that would expect PM's to be heavily involved in (totally fine - thanks for wading into these waters).  That said, I also want to make sure we capture if there are any more tactical things that Stewards/project teams should help with in the next 6 months.  This could be automating releases, establishing clearer contribution guidelines, closing out old PRs/issues, etc.  I know those things aren't free, and if they help us move faster/safer in the future, I want to make sure we're accounting for them.  (They don't need to be in this document, but I do want them for the "roadmap" we look at internally for where we're investing our efforts in 2021H2).

ROADMAP.md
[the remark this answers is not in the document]
marten-seemann: We used to have https://github.com/libp2p/go-libp2p-daemon, but we stopped working on it once implementations in other languages become stable enough.

If I understand the go-libp2p-daemon API correctly, it only provided the option to use a single peer identity. If multiple apps (as opposed to just one app written in a different language) are supposed to use the daemon, we'd probably need a way for different identities to interact with it.

ROADMAP.md
marten-seemann: It creates a lot of connection churn. For example, a peer might close the connection to you just because they reached their watermark and you got unlucky. 
If we had a collaboration protocol, you could say "please, please don't close this connection" or "I don't really care about this connection, go ahead and close it", resulting in the less important connections getting pruned and the more important connections staying alive.

ROADMAP.md
marten-seemann: The mDNS part, yes.
Furthermore, I believe we're already running two DHTs in parallel, one for the public internet and one for local networks.
@mxinden Is it possible this issue is already resolved (assuming we can land my mdns changes), and we can remove this section entirely?

ROADMAP.md
marten-seemann: Enabling Bluetooth connectivity would be nice. Imagine your node discovering and connecting to all devices in your proximity, even if those devices didn't join the same network (in that case, they'd discover each other via mDNS). That said, I'm not sure if message orientation is a prerequisite for that, or if there's a way to establish a Bluetooth 
"connection". I'm a total Bluetooth newbie, so don't trust me on this ;)

In general, the appeal of message-oriented protocols is that they enable "fire and forget". I could send DHT insertion queries to peers without establishing connections to them.

ROADMAP.md
```diff
@@ -0,0 +1,598 @@
...
+TODO: Extend
+
+Copied and adapted from: https://github.com/libp2p/go-smart-record/blob/4a1b33c6d7cb2fb38d166abc1023fbbbab33b60f/README.md
+
+**What?** With Smart Records (SRs) we generalize DHT's key/value put/get as a
+separate protocol that can be leveraged by any other protocol (including DHT
+protocols) to store arbitrary data.
+
+**Why?** We currently don't have a standardized, shared, public medium decoupled
+from the transport protocol for the interaction of different protocols.
```
mxinden, line 337: I think Smart Records should be part of the libp2p road map. Objections?

@petar @adlrocha I am sure you are better at defining the *What* and *Why* for Smart Records. Would you mind suggestion two paragraphs to be included in the libp2p roadmap? I adapted the current two paragraphs from [the go-smart-record README](https://github.com/libp2p/go-smart-record/blob/4a1b33c6d7cb2fb38d166abc1023fbbbab33b60f/README.md).

mxinden: Also, should this section relate to the section below, namely "Peer Routing Records"?

ROADMAP.md
mxinden: I am not particularly familiar with this part of the libp2p space. I know there is a [JVM libp2p implementation](https://github.com/libp2p/jvm-libp2p).

I don't think we, as in Protocol Labs have the bandwidth to maintain either of these. That said I do think we can support other teams building these, either from scratch or wrapping existing implementations like rust-libp2p (added via 5aed005cf4085b0c5a86af843d3533a682cd0885).

ROADMAP.md
mxinden: > Are we aware of certain customers asking for this?

Even better. There is https://github.com/Actyx built on top of rust-libp2p using libp2p in the IoT / factory space today.

// @dvc94ch @rklaehn

ROADMAP.md
mxinden: Good call. I don't have these numbers (yet). See https://github.com/protocol/web3-dev-team/pull/119 for a qualitative analysis.

ROADMAP.md
mxinden: Unfortunately I don't know the source of these crawl measurements. Maybe @vyzo or @aschmahmann know more?

>  Also, can this be an automated process that runs periodically (e.g., weekly, monthly) so we can see improvements to this over time?

I would expect there is already such tool. If not, we might be able to extend the periodically run [ipfs-crawler](https://github.com/wiberlin/ipfs-crawler) project.

:+1: on running this periodically to see our improvements over time.

ROADMAP.md
mxinden: What other major optimizations other than reducing dial failure would this unlock? It might reduce the amount of storage needed on the DHT, though I am not sure that is very relevant.

ROADMAP.md
mxinden: This goes beyond mDNS and beyond two (local and global) DHTs.

See https://github.com/libp2p/notes/issues/3 for local network peer exchange and https://github.com/libp2p/notes/issues/7 and https://github.com/libp2p/specs/issues/222 on a peer exchange protocol in general.

Also related is the [rendezvous protocol](https://github.com/libp2p/specs/blob/master/rendezvous/README.md) solving a similar problem with a smaller number of (rendezvous) nodes instead of a fully decentralized peer exchange protocol.

> Is it possible this issue is already resolved (assuming we can land my mdns changes), and we can remove this section entirely?

I don't think this is resolved by providing mDNS and two DHTs. That said (a) I don't feel strongly about it and (b) I am not sure it requires a whole section in our roadmap on its own. What do you all think? Should we remove it? Should we move it up?

ROADMAP.md
dvc94ch: interesting ROADMAP. but it seems a little broad in general to be actionable and make progress. maybe picking some clear goals for the next quarter would be more productive?

ROADMAP.md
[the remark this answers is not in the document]
mxinden: Removed with https://github.com/libp2p/specs/pull/339/commits/e63b8b3f08a4450c4f3a8defc8cde8ed71331bbb.

ROADMAP.md
```diff
@@ -0,0 +1,587 @@
...
+methods, we should keep in mind aspects like replication, storage,
+liveness, byzantine fault tolerance, etc. Depending on the mechanism,
+the challenges may even align with those of Filecoin.
+
+**Why?** It's an enabler for offline-first scenarios, low quality connections,
+roaming, etc.
+
+**Links:**
+
+-   https://github.com/libp2p/notes/issues/2
```
dvc94ch, line 191: Did some experiments with performant authenticated replicated streams this week. Achieving speeds of 220MB/s was fairly easy, but I think this could be pushed even further. Not sure if this is of interest here.

- https://github.com/dvc94ch/blake-streams
- https://github.com/ipfs-rust/ipfs-embed/pull/81

ROADMAP.md
petar: I'll write a new section that captures the big picture: composable routing, routing language and smart records.

ROADMAP.md
arnetheduck: Native support for a lot of android devices at least is possible with https://github.com/status-im/nim-libp2p - I imagine that can be extended to OSX as well

dvc94ch: rust-libp2p already works on android and iOS and has been for years. Both sunshine and actyx actually build Android apps and sunshine had an iOS app.

dvc94ch: Not that it is a huge accomplishment, rust literally just runs everywhere. This also doesn't mean that there isn't room for improvement. Adding ci to rust-libp2p for example to prevent regressions like: https://github.com/libp2p/rust-libp2p/issues/2064

arnetheduck: At the risk of creating a language shootout, the main point of my comment was that a Java/Kotlin fork is not a prerequisite ;)

ROADMAP.md
```diff
@@ -0,0 +1,595 @@
...
+
+Copied and adapted from: https://github.com/libp2p/go-smart-record/blob/4a1b33c6d7cb2fb38d166abc1023fbbbab33b60f/README.md
+
+**What?** With Smart Records (SRs) we generalize DHT's key/value put/get as a
+separate protocol that can be leveraged by any other protocol (including DHT
+protocols) to store arbitrary data.
+
+**Why?** We currently don't have a standardized, shared, public medium decoupled
+from the transport protocol for the interaction of different protocols.
+
```
petar, line 320:
```suggestion
### Composable routing

Composable Routing is a framework that was designed to enable the
interoperability of protocols and nodes with different versions and capabilities,
developed by independent teams or organizations. As a consequence, it also
enables middleware components (like caching, batching, predictive fetching, and so on)
and the co-existence of different content finding protocols and subnetworks.

The framework has four "pillars":
* Routing syntax and language
* Routing interface
* Smart Records
* Subnets

#### Routing syntax and language

The routing syntax is a common data model (with a user-facing syntactic representation)
for communicating routing-related information (requests and responses).
The [implementation of the routing syntax](https://github.com/libp2p/go-routing-language/tree/master/syntax)
provides serialization and pretty printing of routing expressions.

The routing language is a semantic that specifies how to interpret
routing expressions, and specifically how to compose
complex routing statements out of simpler ones.

Finally, the routing grammar is a set of rules that define the "vocabulary"
of the language. In other words, the grammar defines the set of 
valid and recognizable routing statements. The grammar is, by design,
a continuously evolving specification of the recognizable language vocabulary.

The grammar is reflected in [code](https://github.com/libp2p/go-routing-language/tree/master/patterns),
as a collection of parsing functions for the valid statements in the language. When new rules
are coined (and added to the grammar specification), their respective parsers are codified manually. Going forward,
the project plans to have a grammar compiler, which generates parsers
and checks for ambiguities.

For intuition, expressions in the routing language can represent statements, like:
* I am providing CID via bitswap.
* I am interested in having the content of CID.
* Fetch the content corresponding to IPFS_PATH.

They can also express compositions of statements, like:
* Find providers for CID, then fetch its content,
then extract the field "foo" and view it as a CID,
then fetch its content.

#### Routing interface

A routing system processes routing requests and returns routing responses.
Routing requests and responses have semantics, which are unlike those of a
traditional request/response protocols, that implement function call arguments
and return values.

A routing request is a routing expression which represents a program to be evaluated
by the routing system. In the general case, a routing system may not be able to
evaluate all steps in a program, since it may have limited capabilities or a legacy
version which does not understand newer routing vocabulary. For this reason,
routing responses have to follow a format that allows the routing system to
communicate back the results of partially evaluating the routing request.

Specifically, a routing response is definitionally a copy of the routing request
with the evaluated sub-expressions substituted by the result of their evaluation
or by equivalent expressions (expressions that would evaluate to the same result).

In other words, a routing request and a response are both (usually different)
programs which, if evaluated fully, would produce the same result. In the 
realm of lambda calculus and formal verification, this is known as beta-equivalence.

#### Smart records

Smart Records (SR) is a technology that enables multiple decentralized parties and protocols
to randevouz and share information on a topic. Specifically, SR is a key-value store
service, which facilitates conflict-free updates to values by multiple writers.

SR keys are opaque strings, while SR values are hierarchical documents in the
data model of the routing syntax. An SR service provides a value update operation,
which enables a writer to update a subexpression of a value without interfering with
the rest of the value's contents. Update operations are commutative by design,
ensuring that (i) operation re-ordering due to asynchronous phenomena has no
effect on semantics, and (ii) multiple physical replicas of a value can be merged
unambiguously. In other words, SR values are CRDTs.

The quintessential application of SR is to use it as the store for DHT put/get operations.
This enables any third-party protocol to use the DHT for its own purposes, respectively
unlocking a vast diversity of applications that can be developed by the community.
Some simple examples are email-over-DHT and chat-over-DHT.

More generally, a DHT with flexible value semantics is effectively a giant 
general-purpose, random-access storage layer. And, therefore, highly complex
applications, like a crypto exchange-over-DHT, could be built on this substrate.

The current [implementation](https://github.com/libp2p/go-smart-record) of the SR
virtual machine supports in-memory storage only. We hope to add support for persistence.
The SR virtual machine can also be used programmatically as a Go library that implements a
key-value store for routing expressions. This use comes up, for instance, when implementing
a filecoin content indexer, which must store multiple different provide records for every cid.

#### Subnets

At present, there is a single DHT network which spans across all IPFS nodes.
We would like to introduce a general standard for the co-existence of
multiple network instances with individual discovery, membership and content routing semantics.

This would enable communities of users to configure custom networks
that meet their physical, business, logistics, privacy, security and other
application needs.

Subnets address this goal. A subnet is effectively a set of peers that run
an instance of a routing protocol (e.g. DHT or pubsub) amongst themselves.
Subnets can have custom membership and authentication semantics.
Optionally, a subnet can have an externally-facing routing interface, allowing
others to use it as a routing service.

```

ROADMAP.md
mxinden: > Adding ci to rust-libp2p for example to prevent regressions like: libp2p/rust-libp2p#2064

As an aside, feel free to add a CI step to rust-libp2p to prevent regressions going forward @dvc94ch and @wngr.

ROADMAP.md
mxinden: Added connection churn in 14c67e8220b895d9f0eda8da3259e0f162d1071d.

mxinden:
https://github.com/libp2p/specs/pull/339/commits/486682b17b6a4ae78f60a7048c9d29f23d7b1d01 reduces the #polite-peering section to a connection status protocol only. I don't think there is much use in combining identify, ping and potentially even multistream select into a single protocol. //CC @marten-seemann

ROADMAP.md
mxinden: Thanks @petar for the detailed suggestion. Given that each section is only giving a short overview, I added a reduced version in https://github.com/libp2p/specs/pull/339/commits/64a7be3fe1cd9de39fe38845f031596b24424959 linking to the full description in https://github.com/libp2p/specs/issues/343. 

Let me know what you think.

ROADMAP.md
mxinden: Resolving here since it is replaced by the _Composable Routing_ section.

ROADMAP.md
mxinden: Resolving here. Feel free to un-resolve in case you want to continue the conversation.

ROADMAP.md
mxinden: Resolving here since I haven't heard back. Feel free to un-resolve to continue the discussion.

ROADMAP.md
mxinden: Done in 7fbbd4c9eb2482fdf3f65a635d5c6b088514aac6. Thanks for the suggestion.

ROADMAP.md
mxinden: Marking this conversation as resolved. Feel free to un-resolve in case you want to continue the stalled discussion.

ROADMAP.md
mxinden: Resolving here as "Composable Routing" section has been included.

ROADMAP.md
mxinden: Thanks for the input. 38290e731ae461bbeb11a0caf83c74414858088f (not yet pushed) includes the suggestions above. I am resolving here, feel free to un-resolve or start a follow up conversation.

ROADMAP.md
mxinden: Happening with 4dc18207bfc38fe76fe013ae2e833958deaf463a. Resolving here for now. Feel free to un-resolve.

marten-seemann (approved):

ROADMAP.md
mxinden: For now, I would prefer to keep this section on a high level, especially as each of the detailed steps needed by users today is very implementation specific.

mxinden:
@BigLep given that you are still tracked as "requesting changes", would you mind taking another look?

mxinden:
> @BigLep given that you are still tracked as "requesting changes", would you mind taking another look?

Friendly ping @BigLep.

mxinden:
Thanks everyone for the input. I will merge here.

Note to anyone out there, this is a living document, thus please feel free to suggest additions / post comments / ... going forward.

Looking forward to tackling the tasks ahead of us!

mxinden:
_"Dismissing" @BigLep request for change, as discussed offline._

Issue #337, opened 2021-06-21
[Rendezvous] Is a PoW requirement useful?


thomaseizinger:
Originally raised by @burdges in https://github.com/libp2p/rust-libp2p/pull/2107#issuecomment-864468604.

> I think DDoS adversaries are botnets, so they often have massive untapped CPU, GPU, and even hard drive. I'd think honest peers cannot quickly solve a challenge within a difficulty level that presents an obstacle for adversaries.
> 
> You should probably make applications choose their rendezvous points and simply document that they should do so carefully using their own deterministic randomness. We'll eventually use a rendezvous-like protocol in polkadot, but we'll dictate the rendezvous point by VRF. Tor also dictates introduction points also via deterministic randomness, not a VRF per se but is satisfies their security properties, including being anonymous to the rendezvous point.

thomaseizinger:
According to https://github.com/libp2p/specs/pull/334#issuecomment-865171882, we don't actually want PoW in the spec.

I'll leave this ticket open to track that the spec needs to be changed to clarify that.

burdges:
Any deterministic/verifiable randomness sounds application specific, but that's fine given the applications being build on libp2p.  I only gave examples with a radomness beacon, but anyone without one could consider H("app" ++ date ++ dest_pk) if dest_pk stays hidden from the network, or somehow VRF.Sign(sk, "app" ++ date) and VRF.Verify(pk, "app" ++ date) if pk is public.

burdges:
Any deterministic/verifiable randomness sounds application specific, but that's fine given the applications being build on libp2p.  I only gave examples with a radomness beacon, but anyone without one could consider H("app" ++ date ++ dest_pk) if dest_pk stays hidden from the network, or somehow VRF.Sign(sk, "app" ++ date) and VRF.Verify(pk, "app" ++ date) if pk is public.

PR #342, opened 2021-06-29
Remove PoW from the spec

thomaseizinger:
This is the follow-up PR mentioned in https://github.com/libp2p/specs/pull/334#issuecomment-865456270.

Relates to https://github.com/libp2p/specs/pull/334.
Fixes https://github.com/libp2p/specs/issues/337.


mxinden (approved):
Thanks for the follow-up!

Looks good to me. Giving others a chance to review as well. I will merge this end of this week in case no one objects.

PR #344, opened 2021-06-30
remove link to secio spec from the README

marten-seemann:


mxinden (commented):
I would deem the spec still helpful to gain a historical perspective. With that in mind, and given that the specification header clearly marks it as deprecated, I would suggest keeping the link, though maybe postfixing it with an additional deprecation warning.

PR #348, opened 2021-07-11
[Rendezvous] Clarify `UNREGISTER` behaviour and remove `id` field

### LINKED ISSUE #335: [Rendezvous] How exactly does `Unregister` work?
The spec is fairly thin on the behaviour of `Unregister`.

I am opening this issue to discuss primarily two aspects:

1. According to https://github.com/libp2p/specs/commit/af5ba414c285e73b33843214be8c3a4c61ae69c1, the `id` in the `Unregister` message is the PeerId. Why do we need to send this? We know who is sending the message because our connection is authenticated.
2. Is there a response to an `Unregister` message? I don't think there needs to be because there is no useful information the server can share with the client.

Tagging interest group + author: @daviddias, @whyrusleeping, @Stebalien, @jacobheun, @yusefnapora, @vasco-santos, @vyzo (Please let me know if one shouldn't do this)

thomaseizinger:
All connections in libp2p are authenticated. As such, we don't have
to include the PeerId in the `Unregister` message. We only allow
peers to unregister themselves and therefore, this `id` would always
be equal to the one we are learn from the authentication layer. Having
to compare those and ensure they are equal is an unnecessary error
path.

This also clarifies the absence of an `UNREGISTER` response.

Resolves #335.

### DIFF
--- a/rendezvous/README.md
+++ b/rendezvous/README.md
@@ -122,7 +122,9 @@ an `E_INVALID_TTL` status.
 Peers can refresh their registrations at any time with a new
 `REGISTER` message; the TTL of the new message supersedes previous
 registrations. Peers can also cancel existing registrations at any
-time with an explicit `UNREGISTER` message.
+time with an explicit `UNREGISTER` message. An `UNREGISTER` message does
+**not** have an explicit response. `UNREGISTER` messages for a namespace
+that a client is currently not registered for should be treated as a no-op.
 
 The registration response includes the actual TTL of the registration,
 so that peers know when to refresh.
--- a/rendezvous/README.md
+++ b/rendezvous/README.md
@@ -2,7 +2,7 @@
 
 | Lifecycle Stage | Maturity      | Status | Latest Revision |
 |-----------------|---------------|--------|-----------------|
-| 1A              | Working Draft | Active | r2, 2021-06-24  |
+| 1A              | Working Draft | Active | r3, 2021-07-12  |
 
 Authors: [@vyzo]
 
--- a/rendezvous/README.md
+++ b/rendezvous/README.md
@@ -225,7 +225,7 @@ message Message {
 
   message Unregister {
     optional string ns = 1;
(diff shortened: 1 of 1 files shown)


thomaseizinger:
If we are worried about breaking changes, we could also simply deprecate this field instead or at least make sure the next field that gets added is id `3`.

mxinden (commented):
> Having to compare those and ensure they are equal is an unnecessary error
path.

Agreed.

As far as I can tell, the go-libp2p (not yet merged) implementation [ignores the field if missing](https://github.com/libp2p/go-libp2p-rendezvous/pull/1/files#diff-2a3751147c1b77f13f47c875942089eea84adbedefd45b67064d4cc38c35b601R182-R192). //CC @vyzo as the author.

> at least make sure the next field that gets added is id 3.

Yes, please let's make sure the next field starts at `3` e.g. by commenting out the `id` field and adding a comment.

//CC @moul as a user and @vasco-santos from the js-libp2p side.

thomaseizinger:
> Yes, please let's make sure the next field starts at `3` e.g. by commenting out the `id` field and adding a comment.

Done. Also clarified the commit message to provide more context.

mxinden:
@vyzo can you comment on why you added the `id` field to the `Unregister` message in https://github.com/libp2p/specs/commit/af5ba414c285e73b33843214be8c3a4c61ae69c1 ?

vyzo:
honestly i dont remember ;)

mxinden:
> honestly i dont remember ;)

Thanks for the quick reply @vyzo.

Unless there are any objections, I will merge this pull request early next week.

@thomaseizinger would you mind bumping the specification revision number to help other implementors track the change?

thomaseizinger:
> @thomaseizinger would you mind bumping the specification revision number to help other implementors track the change?

Done!

mxinden (approved):
Thank you @thomaseizinger!

PR #349, opened 2021-07-12
protocol-select/: Add Protocol Select specification

mxinden:
This pull request adds the specification for the _Protocol Select_ protocol.

> _Protocol Select_ is a protocol negotiation protocol. It is aimed at negotiating libp2p protocols on connections and streams. It replaces the _Multistream Select_ protocol.

**Status**

For now, this is a draft only. I am opening this pull request to (a) ease collaboration on the draft and (b) allow others to subscribe to the progress.

See also https://github.com/protocol/web3-dev-team/pull/119.

### DIFF
--- a/protocol-select/README.md
+++ b/protocol-select/README.md
@@ -1,0 +1,536 @@
+<!-- mxinden TODO: Update
+https://github.com/libp2p/specs/tree/master/connections#protocol-negotiation in
+Protocol Select pull request. -->
+
+<!-- mxinden TODO: Hole Punching is not only concerned with punching holes into
+NATs, but Firewalls and NATs. Should this fact be stressed in this document? -->
+
+<!-- mxinden TODO: Consistently use either dialer/listener or client/server -->
+
+# Protocol Select
+
+| Lifecycle Stage | Maturity       | Status | Latest Revision |
+|-----------------|----------------|--------|-----------------|
+| 1A              | Working Draft  | Active | r0, 2021-XX-XX  |
+
+Authors: [@marten-seemann], [@mxinden]
+
+Interest Group:
+
+[@marten-seemann]: https://github.com/marten-seemann
+[@mxinden]: https://github.com/mxinden
+
+See the [lifecycle document][lifecycle-spec] for context about maturity level
+and spec status.
+
+[lifecycle-spec]: https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md
+
+## Table of Contents
+
+- [Protocol Select](#protocol-select)
+    - [Table of Contents](#table-of-contents)
+    - [Introduction](#introduction)
+        - [Improvements over _[Multistream Select]_](#improvements-over-_multistream-select_)
+    - [High-Level Overview](#high-level-overview)
+        - [Secure Channel Selection](#secure-channel-selection)
+        - [TCP Simultaneous Open](#tcp-simultaneous-open)
+            - [Coordinated TCP Simultaneous Open](#coordinated-tcp-simultaneous-open)
+            - [Uncoordinated TCP Simultaneous Open](#uncoordinated-tcp-simultaneous-open)
+        - [Stream Multiplexer Selection](#stream-multiplexer-selection)
+            - [Process](#process)
+            - [Early data optimization](#early-data-optimization)
+            - [Monoplexed connections](#monoplexed-connections)
+            - [0-RTT](#0-rtt)
+    - [Transitioning from [Multistream Select]](#transitioning-from-multistream-select)
+        - [Multiphase Rollout](#multiphase-rollout)
+            - [Phase 1](#phase-1)
+            - [Phase 2](#phase-2)
+            - [Phase 3](#phase-3)
+        - [Additional Rollout Mechanisms](#additional-rollout-mechanisms)
+        - [Heuristics](#heuristics)
+            - [TCP](#tcp)
+            - [QUIC](#quic)
+            - [Protocol Differentiation](#protocol-differentiation)
+    - [Protocol Specification](#protocol-specification)
+        - [Multiplexer Protocol Negotiation](#multiplexer-protocol-negotiation)
+        - [Stream Protocol Negotiation](#stream-protocol-negotiation)
+            - [Initiator](#initiator)
+            - [Listener](#listener)
+        - [Protocol Names vs. Protocol IDs](#protocol-names-vs-protocol-ids)
+            - [Migration from _Protocol Name_ to _Protocol ID_](#migration-from-_protocol-name_-to-_protocol-id_)
+    - [FAQ](#faq)
+
+## Introduction
+
+_Protocol Select_ is a protocol negotiation protocol. It is aimed at negotiating
+libp2p protocols on connections and streams. It replaces the _[Multistream
+Select]_ protocol.
+
+### Improvements over _[Multistream Select]_
+
+- **Downgrade attacks** and **censorship resistance**
+
+  Given that **[Multistream Select]** negotiates a connection's security
+  protocol unencrypted and unauthenticated it is prone to [downgrade attack]s.
+  In addition, a man-in-the-middle can detect that a given connection is used
+  to carry libp2p traffic, allowing attackers to censor such connections.
+
+  **Protocol Select** is combined with a change to the Multiaddr format,
+  advertising the secure channel protocol through the latter instead of
+  negotiating them in-band. Thus [Downgrade attack]s are no longer possible at
+  the protocol negotiation level and a man-in-the-middle can no longer detect
+  a connection being used for libp2p traffic through the negotation process.
+
+- **Connection establishment**
+
+  In addition to making us vulnerable to downgrade attacks, negotiating the
+  security protocol takes one round-trip in the common case with **[Multistream
+  Select]**. On top of that negotiating a stream multiplexer (on TCP) takes
+  another round-trip.
+
+  **Protocol Select** on the other hand depends on security protocols being
+  advertised, thereby eliminating the need for negotiating them. For optimized
+  implementations, stream muxer negotiation will take zero round-trips for the
+  client (depending on the details of the cryptographic handshake protocol). In
+  that case, the client will be able to immediately open a stream after
+  completing the cryptographic handshake. In addition the protocol supports
+  zero-round-trip optimistic stream protocol negotiation when proposing a single
+  protocol.
+
+- **Data schema**
+
+  The **[Multistream Select]** protocol is defined as a plaintext protocol
+  with no strict schema definition, making both implementation and protocol
+  evolution time consuming and error-prone. See [rust-libp2p/1795] showcasing
+  complexity for implementors and [specs/196] to showcase difficulty evolving
+  protocol.
+
+  The **Protocol Select** protocol will use a binary data format defined in a
+  machine parseable schema language allowing protocol evolution at the schema
+  level.
+
+- **Bandwidth**
+
+  **Multistream Select** is not as bandwidth efficient as it could be. For
+  example negotiating a protocol requires sending the protocol name back and
+  forth. For human readability protocol names are usually long strings (e.g.
+  `/ipfs/kad/1.0.0`).
+
+  **Protocol Select** will include the option to improve bandwidth efficiency
+  e.g. around protocol names in the future. While _Protocol Select_ will not
+  solve this in the first iteration, the protocol should be designed with this
+  optimization in mind, and allow for a smooth upgrade in a future iteration.
+
+## High-Level Overview
+
+### Secure Channel Selection
+
+Conversely to [Multistream Select], secure channel protocols are not dynamically
+negotiated in-band. Instead, they are announced upfront in the peer multiaddrs
+(**TODO**: add link to multiaddr spec). This way, implementations can jump
+straight into a cryptographic handshake, thus curtailing the possibility of
+packet-inspection-based censorship and dynamic downgrade attacks.
+
+Given that there is no in-band security protocol negotiation, nodes have to
+listen on different ports for each offered security protocol. As an example a
+node supporting both [Noise] and [TLS] over TCP will need to listen on two TCP
+ports e.g. `/ip6/2001:DB8::/tcp/9090/noise` and `/ip6/2001:DB8::/tcp/443/tls`.
+
+Advertising the secure channel protocol through the peer's Multiaddr instead of
+negotiating the protocol in-band forces users to advertise an updated Multiaddr
+when changing the secure channel protocol in use. This is especially cumbersome
+when using hardcoded Multiaddresses. Users may leverage the [dnsaddr] Multiaddr
+protocol as well as using a new UDP or TCP port for the new protocol to ease the
+transition.
+
+Note: A peer MAY advertise a Multiaddr that includes a secure channel handshake
+protocol like `/noise` even if it doesn't support Protocol Select. See
+[Heuristic section](#heuristic) below for details on how listeners can
+differentiate the negotiation protocol spoken by the dialer on incoming
+connections.
+
+### TCP Simultaneous Open
+
+TCP allows the establishment of a single connection if two endpoints start
+initiating a connection at the same time. This is called _TCP Simultaneous
+Open_. Since many application protocols running on top of a connection (most
+notably the secure channel protocols e.g. TLS) assume their role (client /
+server) based on who initiated the connection, TCP Simultaneous Open connections
+need special handling. This special handling is described below, differentiating
+between two cases of TCP Simultaneous Open: coordinated and uncoordinated.
+
+#### Coordinated TCP Simultaneous Open
+
+When doing Hole Punching over TCP, the [_Direct Connection Upgrade through
+Relay_][DCUTR] protocol coordinates the two nodes to _simultaneously_ dial each
+other, thus, when successful, resulting in a TCP Simultaneous Open connection.
+The two nodes are assigned their role (client / server) out-of-band by the
+[_Direct Connection Upgrade through Relay_][DCUTR] protocol.
+
+#### Uncoordinated TCP Simultaneous Open
+
+In the uncoordinated case, where two nodes coincidentally simultaneously dial
+each other, resulting in a TCP Simultaneous Open connection, the secure channel
+protocol (e.g. TLS) will fail, given that both nodes assume to be in the
+initiating / client role. Nodes SHOULD close the connection and back off for a
+random amount of time before trying to reconnect.
+
+<!--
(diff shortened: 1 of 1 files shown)


protocol-select/README.md
```diff
@@ -0,0 +1,541 @@
...
+Conversely to [Multistream Select], secure channel protocols are not dynamically
+negotiated in-band. Instead, they are announced upfront in the peer multiaddrs
+(**TODO**: add link to multiaddr spec). This way, implementations can jump
+straight into a cryptographic handshake, thus curtailing the possibility of
+packet-inspection-based censorship and dynamic downgrade attacks.
+
+Given that there is no in-band security protocol negotiation, nodes have to
+listen on different ports for each offered security protocol. As an example a
+node supporting both [Noise] and [TLS] over TCP will need to listen on two TCP
+ports e.g. `/ip6/2001:DB8::/tcp/9090/noise` and `/ip6/2001:DB8::/tcp/443/tls`.
```
vyzo, line 155: have we given any thought on how this affect relay connection establishment?

vyzo: Or will that stay in multistream-select?

vyzo (commented):
thank you for putting this together, it is looking quite good.

protocol-select/README.md
mxinden: Thanks for taking a look! Can you expand on the problem you are foreseeing with the Multiaddress extension and relay connection establishment?

vyzo: we'd have to advertise `..../p2p-circuit/tls/p2p/...` or else we won't be able to negotiate a security protocol.

marten-seemann: I think @vyzo is right. Good catch!
Do you see any reason why `..../p2p-circuit/tls/p2p/...` wouldn't be a good idea?

vyzo: In principle it's fine, but I am concerned that a lot of code that handles relay addresses is going to break. So we'd need to be careful with that.

protocol-select/README.md
```diff
@@ -0,0 +1,541 @@
...
+#### Breaking Changes
+
+When making breaking changes to the _Protocol Select_ protocol,
+implementations need to be able to differentiate the old and the new version on
+the wire. This is done via the `version` field in the `ProtoSelect` message,
+treated as an ordinal monotonically increasing number, with each increase
+identifying a new breaking version of the protocol.
+
+As an example for a made-up breaking change, say we would like the listed
+protocols in the `Offer` message to enumerate the protocols that the local node
```
thomaseizinger, line 433: Is this meant to say `ProtoSelect` instead? `Offer` hasn't been mentioned anywhere up till here.

```suggestion
protocols in the `ProtoSelect` message to enumerate the protocols that the local node
```

lanzafame (commented):
Minor typo

protocol-select/README.md
mxinden: Good catch. Thanks @thomaseizinger.

protocol-select/README.md
```diff
@@ -0,0 +1,541 @@
...
+address that contains the security handshake protocol *does not* imply that
+we'll speak Protocol Select.
+
+The first message received on a freshly established and secured TCP connection
+will be a message trying to negotiate the stream muxer using either Protocol
+Select or [Multistream Select].
+
+#### QUIC
+
+Since QUIC neither negotiates a security nor a stream muxer protocol, we'll have
```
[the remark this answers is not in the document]
mxinden: Thanks @lanzafame!

protocol-select/README.md
```diff
@@ -0,0 +1,541 @@
...
+_Protocol ID_. A _Protocol ID_ is a [Multicodec] or a combination of
+[Multicodec]s. Implementations can specify a protocol either via a _Protocol
+Name_ or a _Protocol ID_ by extending the `Protocol` message type definition as
+follows:
+
+```diff
+message Protocol {
+    oneof protocol {
+        string name = 1;
++       uint64 id = 2;
```
Stebalien, line 460:
```suggestion
+       bytes id = 2;
```

protocol-select/README.md
```diff
@@ -0,0 +1,541 @@
...
+mechanisms:
+
+- Extending the libp2p identify protocol, allowing nodes to announce their
+  supported protocols both by _Protocol Name_ and _Protocol ID_, thus signaling
+  the support for the _Protocol ID_ extension for the concrete protocols.
+
+- Including a protocol both by its _Protocol Name_ and _Protocol ID_ in the list
+  of supported protocols.
+
+  Note, when optimistically negotiating a stream protocol as an initiator, with a
```
Stebalien, line 477: Let's note _why_ one might want this. Specifically, you'll _learn_ that they support the protocol ID.

Also note: Figuring out how to implement this internally will be... a bit interesting. Currently we just listen on a protocol. Now, we'll need some way to specify the mapping.

protocol-select/README.md
```diff
@@ -0,0 +1,541 @@
...
+### Improvements over _[Multistream Select]_
+
+- **Downgrade attacks** and **censorship resistance**
+
+  Given that **[Multistream Select]** negotiates a connection's security
+  protocol unencrypted and unauthenticated it is prone to [downgrade attack]s.
+  In addition, a man-in-the-middle can detect that a given connection is used
+  to carry libp2p traffic, allowing attackers to censor such connections.
+
+  **Protocol Select** is combined with a change to the Multiaddr format,
```
Stebalien, line 79: (slightly confusing because we're not actually changing the multiaddr _format_)

protocol-select/README.md
```diff
@@ -0,0 +1,541 @@
...
+  advertising the secure channel protocol through the latter instead of
+  negotiating them in-band. Thus [Downgrade attack]s are no longer possible at
+  the protocol negotiation level and a man-in-the-middle can no longer detect
```
Stebalien, line 82:
```suggestion
  the protocol negotiation level and a man-in-the-middle can no longer passively detect
```

protocol-select/README.md
```diff
@@ -0,0 +1,541 @@
...
+  implementations, stream muxer negotiation will take zero round-trips for the
+  client (depending on the details of the cryptographic handshake protocol). In
+  that case, the client will be able to immediately open a stream after
+  completing the cryptographic handshake. In addition the protocol supports
+  zero-round-trip optimistic stream protocol negotiation when proposing a single
+  protocol.
+
+- **Data schema**
+
+  The **[Multistream Select]** protocol is defined as a plaintext protocol
```
Stebalien, line 103: This isn't really correct. I'd instead say that multistream used a bespoke format that didn't clearly delineate between different atomic "messages".

That was the issue here, as far as I can tell.

protocol-select/README.md
```diff
@@ -0,0 +1,541 @@
...
+
+  **Multistream Select** is not as bandwidth efficient as it could be. For
+  example negotiating a protocol requires sending the protocol name back and
+  forth. For human readability protocol names are usually long strings (e.g.
+  `/ipfs/kad/1.0.0`).
+
+  **Protocol Select** will include the option to improve bandwidth efficiency
+  e.g. around protocol names in the future. While _Protocol Select_ will not
+  solve this in the first iteration, the protocol is designed with this
+  optimization in mind, and allows for a smooth upgrade in a future iteration.
```
Stebalien, line 123: Hm. So, I gave this as one of the primary motivations for Protocol Select, but it's not actually specific to Protocol Select, in it's current incarnation.

That is, I could use the same trick of aliasing codec-based protocol names in multistream.

Let's think _very carefully_ about our upgrade path here. I'm now less convinced it's something we can punt because, well, it's kind of the main reason we wanted this protocol in the first place.

protocol-select/README.md
Stebalien: Actually... It's even more weird. Currently, `/p2p/QmRelay/p2p-circuit/ip4/1.2.3.4/tcp/1234/p2p/QmTarget` tells the _relay_ to connect to the target at `/ip4/1.2.3.4/tcp/1234`.

In this spec, this would be: `/p2p/QmRelay/p2p-circuit/ip4/1.2.3.4/tcp/1234/tls/p2p/QmTarget`.

However, that _still_ doesn't specify how the local peer should connect to the target. That is, how should we secure the relayed stream? I'm not sure where that would be specified...

Really, we _want_ something like `/p2p/QmRelay/p2p-circuit/{/ip4/1.2.3.4/tcp/1234/tls/,....}/tls/p2p/QmTarget` (where `....` could be additional transport addresses). But the current multiaddr spec doesn't help here.

So, I don't really have a _right_ answer here.

Stebalien: We're going to need to redefine _something_ here. My gut reaction is to say that no, the part after `/p2p-circuit` is _not_ treated as an address to give to the relay. But that makes active relaying a bit... tricky.

vyzo: well yeah, and as much as we don't like active relay it is still a legitimate use case for some scenarios.

protocol-select/README.md
```diff
@@ -0,0 +1,541 @@
...
+
+1. All protocols that aren't supported by both endpoints are removed from the
+   clients' list of protocols.
+
+2. The protocol chosen is the first protocol of the client's list.
+
+If there is no overlap between the two lists, the two endpoints can not
+communicate and thus both endpoints MUST close the connection or stream.
+
+### Secure Channel Selection
```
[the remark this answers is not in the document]
marten-seemann: Fair point. I think it makes sense to split this out, especially considering a reader who might read this document in a year or two, when the way we used to advertise addresses is not relevant any more.
I sent #353 and #354 (both still in draft status).

protocol-select/README.md
```diff
@@ -0,0 +1,541 @@
...
+
+Conversely to [Multistream Select], secure channel protocols are not dynamically
```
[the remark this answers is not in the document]
marten-seemann: This would be fixed by #354.

protocol-select/README.md
marten-seemann: This would be fixed with #354.

protocol-select/README.md
marten-seemann: I applied your suggestion in 4ccea37.

protocol-select/README.md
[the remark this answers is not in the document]
marten-seemann: #354 is merged.

protocol-select/README.md
mxinden: Noted in 843e76097a50f7bad3693f31e24fe64031f0a865.

protocol-select/README.md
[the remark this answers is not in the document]
marten-seemann: Resolving this because #354 was merged.

protocol-select/README.md
mxinden: _The below is based on a discussion with @marten-seemann. While we both agree that it ain't elegant, it might either still hold as a viable solution or inspire a more elegant one_.

Taxonomy:

- `A`: The destination node.
- `B`: The source node.
- `R`: The relay node.
- outer connection: The connection from `R` to `A`.
- relayed (inner) connection: The relayed (inner) connection from `B` to `A`.

Problems we need to solve:

1. `B` and `A` need to know which security protocol to use to upgrade the relayed (inner) connection.
2. In the case of active relaying, `R` needs to know which security protocol to use to upgrade the outer connection to the destination.
3. `A` should be able to offer different security protocols for the relayed (inner) connection. `A` can use different layer 4 ports to demultiplex security protocols for the outer connection. `A` has no demultiplexing mechanism for the security protocols used for the relayed (inner) connection. Thus if `A` advertises different multiaddresses with different security protocols for the relayed (inner) connection, it has no way to determine which one to use on incoming relayed (inner) connections.

We propose two mechanisms to solve the issues above:

1. We have to somehow embed both outer and inner security protocol in the multiaddr. What we could do is introduce another separator, e.g. `p2p-circuit-inner` (name yet to be determined).
    - **Solving (1)**: A multiaddr for a passive relayed connection could then look like: `/ip4/../tcp/../tls/p2p/QmRelay/p2p-circuit/p2p/QmA/p2p-circuit-inner/noise`, with `noise` being used to secure the relayed (inner) connection.
    - **Solving (2)**: A multiaddr for an active relayed connection could then look like: `/ip4/../tcp/../tls/p2p/QmRelay/p2p-circuit/ip6/::1/tls/p2p/QmA/p2p-circuit-inner/noise`, with `tls` being used to secure the outer connection and `noise` being used to secure the relayed (inner) connection.
2. **Solving (3)**: In order for `A` to be able to advertise different multiaddresses with different security protocols for the relayed (inner) connection, while still being able to choose the right security protocol on incoming relayed (inner) connections, we need to include the security protocol for the relayed (inner) connection in the circuit relay v2 CONNECT message from `B` to `A`.

protocol-select/README.md
mxinden: Resolved with split into two specifications.

protocol-select/README.md
mxinden: Given the lack of alternatives, I suggest we integrate the approach above into this specification and https://github.com/libp2p/specs/pull/353. Just to avoid confusion, this should not be interpreted as a final commitment to the suggestion above. Any objections?

mxinden: https://github.com/libp2p/specs/pull/359/ adds the above suggestions to https://github.com/libp2p/specs/pull/353.

mxinden: The above suggestion is now included in https://github.com/libp2p/specs/pull/353. I am resolving this conversation. Please post any follow-ups on https://github.com/libp2p/specs/pull/353.

protocol-select/README.md
```diff
@@ -0,0 +1,511 @@
...
+concluding the transition.
+
+### Additional Rollout Mechanisms
+
+When attempting to upgrade an outbound connection with Protocol Select to a node
+that does not yet support Protocol Select, the connection attempt will fail:
+
+* TCP: when the cryptographic handshake is started
+
+* QUIC: when the first stream is opened
```
kpp, line 281: Do we really need to close a connection? I would say in case of QUIC one can open another substream using Multistream Select.

marten-seemann: In theory that would work, but it might be quite involved implementation-wise.

thomaseizinger (commented):
Thanks for pushing this forward. Excited to see optimizations such as "Early data".

cheako:
Why not take this opportunity to make breaking changes not split the network?  Supporting multi-versions is a simple solution, there must be others:

```protobuf
syntax = "proto2";
message ProtoSelect {
    uint32 version = 1;
    uint32 max_forbid_version = 2;
    message Protocol {
        oneof protocol {
            string name = 1;
        }
    }
    repeated Protocol protocols = 3;
}
```

`max_forbid_version` is set/updated to the oldest/highest offensive version that is hereby banned...  because
1. a change so critical it warrants banning all clients older than
2. years, perhaps decades and it's just time

Allow the other end to reply with, sorry, I speak version X and continue accordingly.

mxinden:
Hi @cheako, I am sorry for the late reply here.

> Why not take this opportunity to make breaking changes not split the network? Supporting multi-versions is a simple solution, there must be others:

I am not quite sure I understand where you are going with this. Could you describe a small scenario where the current versioning scheme would not suffice but your proposal would?

> `max_forbid_version` is set/updated to the oldest/highest offensive version that is hereby banned... because

In particular I am having a hard time understanding why the `max_forbid_version` needs to be exchanged explicitly.

>   1. a change so critical it warrants banning all clients older than

Say client `A` receives a `ProtocolSelect` message with a version pre-cricitical-change from `B`, it could then simply cut off the connection. I expect clients to start with their highest version available, thus there is no need for `A` to retry as `B` likely does not support a higher version.

> Allow the other end to reply with, sorry, I speak version X and continue accordingly.

Note that this would conflict with [_Optimistic Protocol Negotiation_](https://github.com/libp2p/specs/blob/76d2d783531e4c92c62a58a60cdb63297bae6716/protocol-select/README.md#initiator) where the remote might have already sent application data.

cheako:
The point is to avoid disconnection because version.  So...  The lower version should be the one to give up and disconnect when it can't satisfy the required version.

ACID rules should already cover the case where a transaction needs to be rolled back, so it's not a problem that a higher version sends some data that needs to be disregarded.

mxinden:
Again, would you mind describing a concrete example where the current versioning scheme would not suffice? I am still having a hard time understanding which problem you are trying to solve @cheako.

cheako:
The notion of a sudden disconnect is not expressive enough and is just ambiguous.  I don't think it's right to put a lot of effort into the case where a node would be banned.  However, it can be done more than the, simple and to the point, `max_forbid_version` scheme.  Another solution is to have a response along with a disconnect(you can try to pad the TCP RST with data) clients are written to understand.  There are three things I can think of for such a reply:

1. That you're being banned, the packet type(obviously the newer node would be sending a legacy packet) expresses this.
2. The severity of the ban, like should the application close(as in you're a danger to society) after getting a few of these, or is it just this node that it's banned from.
3. Documentation, a web address for the user to get more information as to why.
a. The new software may have updated guidance for where documentation lives.

protocol-select/README.md
```diff
@@ -0,0 +1,511 @@
...
+
+message ProtoSelect {
+    required uint32 version = 1;
+
+    message Protocol {
+        oneof protocol {
+            string name = 1;
+        }
+    }
+    repeated Protocol protocols = 2;
```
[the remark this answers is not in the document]
mxinden: :+1: This came up in previous discussions as well.

I would like to keep the first iteration of the protocol as simple as possible. As far as I can tell, this can be added in a backwards compatible way later on. Let me know in case I am missing something.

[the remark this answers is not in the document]
Menduist: Sure, though two things to keep in mind:

The "Optimistic Protocol Negotiation" from the current spec should probably be removed, to avoid clashes in the future

And to add backward compatibility in the future, we'll need need to add a "early data processed" bool to the listener -> initiator response, otherwise the initiator has no way to know if the early data has been processed. Not too much of a deal, though (I think)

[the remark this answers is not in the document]
mxinden: > And to add backward compatibility in the future, we'll need need to add a "early data processed" bool to the listener -> initiator response, otherwise the initiator has no way to know if the early data has been processed. Not too much of a deal, though (I think)

That would require the listener to only send a Protocol Select message once it received the initiators message, right? If so, that would prevent all early-data optimizations where the listener side is able to send first.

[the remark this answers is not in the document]
Menduist: Right, might be an issue.

In the case where the listener side is sending first, we would have to flip the roles (the listener is becoming initiator)
AFAICT, we always know who should be the "protocol select initiator", even though it may not be the same as the "secure initiator"

But maybe there is a case where we don't know?

Issue #347, opened 2021-07-11
[Identify] How to integrate signed peer records?


thomaseizinger:
The `/identify` protocol was conceived before signed peer records were added.

Is there a plan of integrating signed peer records into `/identify`? I think there are a number of ways:

1. Create `/identify/1.1.0` where `listenAddrs` is deprecated and a new field `listenPeerRecords` is added.
2. Create an entirely new protocol that is built on top of peer records instead of plain addresses and simply deprecate the old identify protocol.

(1) could be implemented in a backwards-compatible way where upgraded peers always populate both fields.
(2) could also be a good chance to do https://github.com/libp2p/specs/issues/45

Given the simplicity of the identify protocol, it might just be easier to design a completely new one. Nodes can always stay backwards compatible by supporting both protocols simultaneously.

mxinden:
> Is there a plan of integrating signed peer records into `/identify`?

Yes, very much, i.e. it is on our roadmap (see `### Peer Routing Records` in https://github.com/libp2p/specs/pull/339/).

As far as I can tell go-libp2p already does via a combination of the two suggestions above, more specifically via:

- A new protocol name: `/p2p/id/1.1.0`.
- A new field in the existing ProtoBuf: `optional bytes signedPeerRecord = 8;`.

See https://github.com/libp2p/go-libp2p/pull/747 for details.

Unfortunately the identify protocol specification has not been updated.

thomaseizinger:
> A new protocol name: `/p2p/id/1.1.0`.

Is there any reasoning why version `1.1.0` was chosen? Is there a `/p2p/id/1.0.0` as well?

> Unfortunately the identify protocol specification has not been updated.

I might set out and try to update the spec then.

mxinden:
> Is there any reasoning why version `1.1.0` was chosen? Is there a `/p2p/id/1.0.0` as well?

I would guess that `/p2p/id/1.1.0` was chosen instead of `/p2p/id/1.0.0` to avoid confusion with `/ipfs/id/1.0.0`.

> I might set out and try to update the spec then.

That would be very much appreciated. Thank you Thomas!

thomaseizinger:
I tried to update the spec and I am in-between two minds.

For one, we are defining a completely new protocol here (i.e. `/p2p/id/1.1.0`) but at the same time, what the go implementation defines as `/p2p/id/1.1.0` carries old legacy stuff around like `listenAddrs` and un-specified functionality like `delta`.

[...]

PR #350, opened 2021-07-16
Introduce identify v2.0.0 based on signed peer records

thomaseizinger:
Best reviewed patch-by-patch.

I plan to amend https://github.com/libp2p/specs/commit/1d715ffcb542a5eab9efdff12fbe89bd93a08715 as I work through feedback. Currently a draft PR because of discussion ongoing [here](https://github.com/libp2p/specs/issues/347#issuecomment-881162882).


Stebalien:
Hm. This is unfortunate. Sending signed peer records has already been implemented in go-libp2p's identify but this was never added to the specs.

The implementation is that we simply added a new field to the identify protobuf that includes the signed peer record. In terms of efficiency, this is less than optimal. But it lets us avoid an extra round-trip for version negotiation.

https://github.com/libp2p/go-libp2p/blob/5d3b015018a34a5875c91c7645ef20f8a91736e3/p2p/protocol/identify/pb/identify.proto#L40-L45

Stebalien:
Ah, I see the discussion. I'll continue there.

PR #353, opened 2021-07-25
addressing: describe multiaddrs containging the security protocol

marten-seemann:
Attempting to split the Protocol Select (#349) into two documents, and decoupling the addressing change from the protocol negotiation part.
I'm sure that the circuit addressing part still needs a lot of work. Suggestions welcome!

### DIFF
--- a/addressing/README.md
+++ b/addressing/README.md
@@ -147,6 +147,33 @@ within it. Using our example above, decapsulating either `/tcp/1234/ws` _or_
 unsurprising if you consider the utility of the `/ip4/7.7.7.7/ws` address that
 would result from simply removing the `tcp` component.
 
+### The security multiaddr
+
+Peers MAY advertise their addresses without the security protocol, e.g.
+`/ip4/6.6.6.6/tcp/1234/` or `/ip4/6.6.6.6/udp/1234/quic`. The security handshake
+protocol is then negotiated using the [multistream](../multistream.md). This is
+the way the libp2p handshake worked until mid 2021.
+This poses a security problem, as the negotiation was not authenticated and
+therefore susceptible to man-in-the-middle attacks. A MITM could modify the list
+of supported handshake protocols, thereby forcing a downgrade to a (potentially)
+less secure handshake protocol. Note that since QUIC is standardized to use
+TLS 1.3, no handshake protocol needs to be negotiated when using QUIC.
+
+Peers SHOULD encapsulate the security protocol in the addresses they advertise,
+for example `/ip4/6.6.6.6/tcp/1234/tls` for a TLS 1.3 server listening on TCP
+port 1234 and `/ip4/6.6.6.6/tcp/1235/noise` for a Noise server listening on TCP
+port 1235. QUIC multiaddrs remain unchanged.
+The nodes jump straight into a cryptographic handshake, thus curtailing the
+possibility of packet-inspection-based censorship and dynamic downgrade attacks.
+This also applies to circuit addresses: the security protocol is encoded in the
+`<destination address>` as defined in [`p2p-circuit` Relay Addresses](#p2p-circuit-relay-addresses).
+
+Implementations using [Protocol Select](https://github.com/libp2p/specs/pull/349/)
+(**TODO**: update link) MUST encapsulate the security protocol in the multiaddr.
+Note that it’s valid to assume that any node that encapsulated the security
+protocol in their multiaddr also supports Protocol Select.
+
+
 ### The p2p multiaddr
 
 libp2p defines the `p2p` multiaddr protocol, whose address component is the


addressing/README.md
```diff
@@ -147,6 +147,33 @@ within it. Using our example above, decapsulating either `/tcp/1234/ws` _or_
 unsurprising if you consider the utility of the `/ip4/7.7.7.7/ws` address that
 would result from simply removing the `tcp` component.
 
+### The security multiaddr
+
+Peers MAY advertise their addresses without the security protocol, e.g.
```
mxinden, line 152:
```suggestion
Peers MAY advertise their addresses without a security protocol, e.g.
```

There are multiple security protocols, e.g. `tls` and `noise`, thus using `the` instead of `a` seems off here. I am not a native speaker, thus my intuition might be wrong.

addressing/README.md
```diff
@@ -147,6 +147,33 @@ within it. Using our example above, decapsulating either `/tcp/1234/ws` _or_
...
+`/ip4/6.6.6.6/tcp/1234/` or `/ip4/6.6.6.6/udp/1234/quic`. The security handshake
+protocol is then negotiated using the [multistream](../multistream.md). This is
```
mxinden, line 154:
```suggestion
protocol is then negotiated using [multistream-select](../connections/README.md#multistream-select). This is
```

addressing/README.md
```diff
@@ -147,6 +147,33 @@ within it. Using our example above, decapsulating either `/tcp/1234/ws` _or_
...
+port 1235. QUIC multiaddrs remain unchanged.
+The nodes jump straight into a cryptographic handshake, thus curtailing the
+possibility of packet-inspection-based censorship and dynamic downgrade attacks.
+This also applies to circuit addresses: the security protocol is encoded in the
+`<destination address>` as defined in [`p2p-circuit` Relay Addresses](#p2p-circuit-relay-addresses).
+
+Implementations using [Protocol Select](https://github.com/libp2p/specs/pull/349/)
+(**TODO**: update link) MUST encapsulate the security protocol in the multiaddr.
+Note that it’s valid to assume that any node that encapsulated the security
+protocol in their multiaddr also supports Protocol Select.
```
mxinden, line 174: I thought last consensus was on not tying the two together. Again, I am fine either way, I only want to be consistent.

https://github.com/libp2p/specs/pull/349 currently includes:

> Note: A peer MAY advertise a Multiaddr that includes a secure channel handshake
protocol like `/noise` even if it doesn't support Protocol Select. See
[Heuristic section](#heuristic) below for details on how listeners can
differentiate the negotiation protocol spoken by the dialer on incoming
connections.

addressing/README.md
```diff
@@ -147,6 +147,33 @@ within it. Using our example above, decapsulating either `/tcp/1234/ws` _or_
 unsurprising if you consider the utility of the `/ip4/7.7.7.7/ws` address that
 would result from simply removing the `tcp` component.
 
+### The security multiaddr
```
mxinden, line 150:
```suggestion
### The multiaddr security component
```

While I know that each multiaddr component (e.g. `/ip6/::1` or `/tcp/1234`) in itself is a valid multiaddr, I don't think there is much value in a single component `/tls` multiaddr. Thus instead of speaking of _the security multiaddr_, I would refer to something along the lines of the _security multiaddr component_.

addressing/README.md
```diff
@@ -147,6 +147,33 @@ within it. Using our example above, decapsulating either `/tcp/1234/ws` _or_
...
+`/ip4/6.6.6.6/tcp/1234/` or `/ip4/6.6.6.6/udp/1234/quic`. The security handshake
+protocol is then negotiated using the [multistream](../multistream.md). This is
+the way the libp2p handshake worked until mid 2021.
+This poses a security problem, as the negotiation was not authenticated and
+therefore susceptible to man-in-the-middle attacks. A MITM could modify the list
+of supported handshake protocols, thereby forcing a downgrade to a (potentially)
+less secure handshake protocol. Note that since QUIC is standardized to use
+TLS 1.3, no handshake protocol needs to be negotiated when using QUIC.
+
+Peers SHOULD encapsulate the security protocol in the addresses they advertise,
```
mxinden, line 162: Nit pick: Given that this is the section about security protocols in multiaddr, wouldn't it make more sense to have this paragraph and the next go first, especially as multiaddr without security protocols will not be used for much longer once this is in place?

addressing/README.md
marten-seemann: Good point.

addressing/README.md
marten-seemann: The intention is to not tie them together, in the sense that advertising a security-enabled multiaddr doesn't mean that you speak Protocol Select. You can advertise a security-enabled multiaddr and still speak multistream.

However, I thought we agreed to require you to advertise a security-enabled multiaddr if you want to speak Protocol Select. Maybe that's not necessary though and you could use the Protocol Select protobufs to negotiate your handshake protocol?

mxinden: ### Protocol Select requiring secure multiaddr component

> However, I thought we agreed to require you to advertise a security-enabled multiaddr if you want to speak Protocol Select.

Yes. We agreed on that.

>  Maybe that's not necessary though and you could use the Protocol Select protobufs to negotiate your handshake protocol?

I don't see a benefit of supporting this. Anything I am missing?

### Secure multiaddr component implying Protocol Select

> in the sense that advertising a security-enabled multiaddr doesn't mean that you speak Protocol Select.

> Note that it’s valid to assume that any node that encapsulated the security protocol in their multiaddr also supports Protocol Select.

Are these two sentences not in direct conflict?

mxinden: ### Protocol Select requiring secure multiaddr component

> However, I thought we agreed to require you to advertise a security-enabled multiaddr if you want to speak Protocol Select.

Yes. We agreed on that.

>  Maybe that's not necessary though and you could use the Protocol Select protobufs to negotiate your handshake protocol?

I don't see a benefit of supporting this. Anything I am missing?

### Secure multiaddr component implying Protocol Select

> in the sense that advertising a security-enabled multiaddr doesn't mean that you speak Protocol Select.

> Note that it’s valid to assume that any node that encapsulated the security protocol in their multiaddr also supports Protocol Select.

Are these two sentences not in direct conflict?

marten-seemann: There was a "not" missing... Fixed.

addressing/README.md
```diff
@@ -147,6 +147,33 @@ within it. Using our example above, decapsulating either `/tcp/1234/ws` _or_
...
+for example `/ip4/6.6.6.6/tcp/1234/tls` for a TLS 1.3 server listening on TCP
+port 1234 and `/ip4/6.6.6.6/tcp/1235/noise` for a Noise server listening on TCP
+port 1235. QUIC multiaddrs remain unchanged.
+The nodes jump straight into a cryptographic handshake, thus curtailing the
+possibility of packet-inspection-based censorship and dynamic downgrade attacks.
+This also applies to circuit addresses: the security protocol is encoded in the
+`<destination address>` as defined in [`p2p-circuit` Relay Addresses](#p2p-circuit-relay-addresses).
+
```
mxinden, line 170:
```suggestion

Advertising the secure channel protocol through the peer's Multiaddr instead of
negotiating the protocol in-band forces users to advertise an updated Multiaddr
when changing the secure channel protocol in use. This is especially cumbersome
when using hardcoded Multiaddresses. Users may leverage the [dnsaddr] Multiaddr
protocol as well as using a new UDP or TCP port for the new protocol to ease the
transition.

```

What do you think of adopting this section from https://github.com/libp2p/specs/pull/354/?

marten-seemann: That's a good idea!

addressing/README.md
marten-seemann: Where exactly? Do you mean before Encapsulation?

mxinden: Does the following diff improve my suggestion?

```diff
diff --git a/addressing/README.md b/addressing/README.md
index abc6bc5..0967407 100644
--- a/addressing/README.md
+++ b/addressing/README.md
@@ -149,16 +149,6 @@ would result from simply removing the `tcp` component.
 
 ### The multiaddr security component
 
-Peers MAY advertise their addresses without a security protocol, e.g.
-`/ip4/6.6.6.6/tcp/1234/` or `/ip4/6.6.6.6/udp/1234/quic`. The security handshake
-protocol is then negotiated using [multistream-select](../connections/README.md#multistream-select). This is
-the way the libp2p handshake worked until mid 2021.
-This poses a security problem, as the negotiation was not authenticated and
-therefore susceptible to man-in-the-middle attacks. A MITM could modify the list
-of supported handshake protocols, thereby forcing a downgrade to a (potentially)
-less secure handshake protocol. Note that since QUIC is standardized to use
-TLS 1.3, no handshake protocol needs to be negotiated when using QUIC.
-
 Peers SHOULD encapsulate the security protocol in the addresses they advertise,
 for example `/ip4/6.6.6.6/tcp/1234/tls` for a TLS 1.3 server listening on TCP
 port 1234 and `/ip4/6.6.6.6/tcp/1235/noise` for a Noise server listening on TCP
@@ -180,6 +170,15 @@ Implementations using [Protocol Select](https://github.com/libp2p/specs/pull/349
 Note that it’s not valid to assume that any node that encapsulated the security
 protocol in their multiaddr also supports Protocol Select.
 
+Peers MAY advertise their addresses without a security protocol, e.g.
+`/ip4/6.6.6.6/tcp/1234/` or `/ip4/6.6.6.6/udp/1234/quic`. The security handshake
+protocol is then negotiated using [multistream-select](../connections/README.md#multistream-select). This is
+the way the libp2p handshake worked until mid 2021.
+This poses a security problem, as the negotiation was not authenticated and
+therefore susceptible to man-in-the-middle attacks. A MITM could modify the list
+of supported handshake protocols, thereby forcing a downgrade to a (potentially)
+less secure handshake protocol. Note that since QUIC is standardized to use
+TLS 1.3, no handshake protocol needs to be negotiated when using QUIC.
 ```

**Comment by mxinden** [comment]
This proposal is ready for review and we welcome **feedback by everyone across the whole libp2p community**. Note that this proposal touches two fundamental building block of libp2p, namely addressing and protocol negotiation. 

For additional context: _Protocol Select_ builds on top of this proposal (https://github.com/libp2p/specs/pull/349).

File: addressing/README.md [hunk 7]
```diff
@@ -147,6 +147,40 @@ within it. Using our example above, decapsulating either `/tcp/1234/ws` _or_
...
+This poses a security problem, as the negotiation was not authenticated and
+therefore susceptible to man-in-the-middle attacks. A MITM could modify the list
+of supported handshake protocols, thereby forcing a downgrade to a (potentially)
+less secure handshake protocol. Note that since QUIC is standardized to use
+TLS 1.3, no handshake protocol needs to be negotiated when using QUIC.
+
+Peers SHOULD encapsulate the security protocol in the addresses they advertise,
+for example `/ip4/6.6.6.6/tcp/1234/tls` for a TLS 1.3 server listening on TCP
+port 1234 and `/ip4/6.6.6.6/tcp/1235/noise` for a Noise server listening on TCP
+port 1235. QUIC multiaddrs remain unchanged.
```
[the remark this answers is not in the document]
mxinden: Resolving here with @marten-seemann recent comment in mind. Feel free to open a new thread in case you want to continue the discussion @kpp.

addressing/README.md
```diff
@@ -329,6 +363,50 @@ Here, the destination peer has the peer id
...
+   suggestions. -->
+
+   Example: `<relay-mulitaddr>/p2p-circuit/<destination-multiaddr>/p2p-circuit-inner/tls`
+
+   Note: One might be tempted to not specify (3) and simply use the security
+   protocol in (2). This would break if the security protocol used for (2) can
+   not be used for (3), e.g. in the case where the relay establishes a QUIC
+   connection to the destination secured via TLS and the source only supports
+   Noise.
+
```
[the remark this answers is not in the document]
mxinden: Good idea. Added via https://github.com/libp2p/specs/pull/353/commits/6d8004e9cd5ec58f8b564443554838a553c02f0e. Thanks @yusefnapora.

addressing/README.md
```diff
@@ -329,9 +363,57 @@ Here, the destination peer has the peer id
...
+3. Upgrading the relayed connection from the source to the destination.
+
+   The security protocol is specified by appending
+   `/p2p-circuit-inner/<relayed-connection-security-protocol>` to the full
+   address.
+
+   <!-- TODO: Is `p2p-circuit-inner` the ideal name? Up for alternative
+   suggestions. -->
+
+   Example: `<relay-mulitaddr>/p2p-circuit/<destination-multiaddr>/p2p-circuit-inner/tls`
```
vyzo, line 402: This will require some changes in our our relay implementations before it can be effected, by I am fine with that.

I agree that using a new inner component is the way to go, but can we call it something else?
Perhaps `p2p-circuit-sec` or `p2p-circuit-termination`.

relay/circuit-v2.md
```diff
@@ -294,6 +294,42 @@ Common failure status codes are:
...
+by the target, the initiator chose to use on the relayed connection.
+
+As an example, let's say the target listens for incoming relayed connections via
+relay `R1` and relay `R2`. In addition it supports both TLS Noise as security
+protocols. It would then advertise the following relayed multiaddresses:
+
+- `<relay-R1-multiaddr>/p2p-circuit/p2p/QmTarget/p2p-circuit-inner/tls`
+- `<relay-R1-multiaddr>/p2p-circuit/p2p/QmTarget/p2p-circuit-inner/noise`
+- `<relay-R2-multiaddr>/p2p-circuit/p2p/QmTarget/p2p-circuit-inner/tls`
+- `<relay-R2-multiaddr>/p2p-circuit/p2p/QmTarget/p2p-circuit-inner/noise`
```
vyzo, line 318: Wait, when we advertise our relay addr we typically omit the `p2p-circit` part; how is this now going to work, should we duplicate the `p2p` part?
Also, how does this parse as an AddrInfo?

vyzo:
It seems to me that we are doing this backwards with the relay addrs.
What's wrong with `/addr.../p2p-circuit/sec-proto`? This allows us to easily compose and create a valid AddrInfo with `/addr.../p2p-circuit/tls/p2p/QmX` and has no ambiguities.

vyzo:
If we must introduce `p2p-circuit-inner` (I really want to call this `p2p-circuit-sec`), then it should come before the `/p2p/QmX` part in the multiaddr.

So the circuit multiaddr for `QmX` would be `/relay-addr/p2p-circuit/p2p-circuit-sec/tls/p2p/QmX` and not `/relay-add/p2p-circuit/p2p/QmX/p2p-circuit-sec/tls`.

vyzo (changes requested):
I think we are handling relay addresses wrong.

The unambiguous form would be `/relay-addr/p2p-circuit/p2p-circuit-sec/tls/p2p/QmX`, which parses as an `AddrInfo` and allows us to advertise just `/relay-addr/p2p-circuit/p2p-circuit-sec/tls` as the address of `QmX`.

mxinden:
@vyzo thank you for taking a look. I am having difficulties following your five comments above. Would you mind summarizing them in a single one?

Do I understand correctly that you would like to:

- Rename `/p2p-circuit-inner` to `/p2p-circuit-sec`?
- Move `p2p-circuit-sec` before the `/p2p/QmX` of the destination node?

vyzo:
yes.  the second part is important for composability and parsing of
multiaddrs; it allows us to omit the p2p part for our own addresses and
parse complete multiaddrs as AddrInfos.

mxinden:
https://github.com/libp2p/specs/pull/353/commits/3c1487a106943ade936c5185def680c67c9496c8 and https://github.com/libp2p/specs/pull/353/commits/90bef6fa4752add96bff7ebf016d247cc120c34d change the two points discussed above. Note that instead of `p2p-circuit-sec` it uses `p2p-circuit-security` to disambiguate SECurity with SECond. 

@vyzo would you mind taking another look?

relay/circuit-v2.md
```diff
@@ -294,6 +294,42 @@ Common failure status codes are:
 
 ***Note: implementations _should not_ accept connection initiations over already relayed connections***
 
+##### Security protocol selection for the relayed connection
```
vyzo, line 297: we need to update the circuit-v2 spec version for this.

relay/circuit-v2.md
```diff
@@ -294,6 +294,42 @@ Common failure status codes are:
...
+
+Instead of negotiating the security protocol in-band, security protocols should
```
vyzo, line 299: I think this should be a `may` instead of a `should`.

relay/circuit-v2.md
```diff
@@ -294,6 +294,42 @@ Common failure status codes are:
...
+- `<relay-R2-multiaddr>/p2p-circuit/p2p-circuit-security/tls/p2p/QmTarget`
+- `<relay-R2-multiaddr>/p2p-circuit/p2p-circuit-security/noise/p2p/QmTarget`
+
+Once the initiator received the above multiaddresses and decides to initiate a
+relayed connection to the target, it needs to decide whether it wants to use
+`R1` or `R2` and whether it wants to secure the relayed connection via TLS or
+Noise. Say it decides for `R1` and Noise it would then include the multiaddress
+below in it `HopMessage` with type `Connect` in the `peer` field:
+
+- `<relay-R1-multiaddr>/p2p-circuit/p2p-circuit-security/noise/p2p/QmTarget`
```
vyzo, line 326: I think we can omit the relay multiaddr prefix here.

relay/circuit-v2.md
```diff
@@ -309,11 +345,13 @@ The relay sends a `StopMessage` with `type = CONNECT` and the following form:
 ```
 StopMessage {
   type = CONNECT
-  peer = Peer { ID = ...}
+  initiator = Peer { ID = ...}
+  target = Peer { addrs = ...}
```
vyzo, line 349: I don't quite like the name `target` here, perhaps use `destination` or just `dest`.

vyzo: also, we can call `initiator` as `src` or `source` to match standard nomenclature.

vyzo (commented):
It looks much better now.

However, i think the changes to circuit v2 spec need a bit more thought; and it also should upgrade the document version!

relay/circuit-v2.md
vyzo: I think it should be a `MAY`, at least for the time being. We can change it to `SHOULD` later, when all is implemented and the spec is ready for candidate recommendation status.

Menduist:
If I understand correctly, we won't be able to support multiple encryption scheme on a single port anymore?
ie, need to do
`/ip6/::1/tcp/1234/tls/`
`/ip6/::1/tcp/4567/noise/`

That seems bad to me, it would become very difficult to, for instance, update noise to a new non-compatible version. (if that seem unlikely to you, https://github.com/libp2p/specs/issues/355 requires a new non-compatible noise version to be fixed, for instance)
The only solution would be to listen on multiple ports for the duration of the transition, which forces end-users to setup multiple ports, multiple redirection in their NAT, etc

I'm also not 100% convinced by the "downgrade" argument. Sure, it applies to mobile phones because you can't update the 3310s to safer protocols, but in our case, if a protocol is less secure, just disable it (like it was done for secio), push an update, and problem solved.


Finally, I'm having a hard time to imagine how active relays work in this scenario.
A: only supports tcp/noise
R: supports quic + tcp/noise
B: supports quic + noise + tls (not TCP)

How does B advertise that it handles noise? Is it supposed to advertise something like
- `/ip6/::1/udp/1234/quic/`
- `/ip6/::1/udp/1234/quic/p2p-circuit-security/noise`
- `/ip6/::1/udp/1234/quic/p2p-circuit-security/tls`
? 

Weird, but ok. Except that now we're back to my first remark, when it will receive a connection coming from a relay, it will not know if it's a noise or a tls one. So again, need to listen on 2 different ports.

Now, let's be crazy and say that B also handles Webrtc, it will need to advertise:
- `/ip6/::1/udp/1234/webrtc/`
- `/ip6/::1/udp/1234/webrtc/p2p-circuit-security/noise`
- `/ip6/::1/udp/1235/webrtc/p2p-circuit-security/tls`
- `/ip6/::1/udp/1236/quic/`
- `/ip6/::1/udp/1236/quic/p2p-circuit-security/noise`
- `/ip6/::1/udp/1237/quic/p2p-circuit-security/tls`


To sum up, it seems to me that we are putting "capabilities" into the MA, and that doesn't seem right. Each MA is already a capability. My understanding is that this grew out of the goal of 0/low-RTT setups, but low RTTs can also be achieved with a negotiation.

A: supports Noise, TLS, Secio
B: supports TLS

A -> B: open connection using Noise, TLS or Secio
If you are compatible with noise, here is some early data: [the noise handshake]
B -> A: only compatible with TLS, please retry
A -> B: [TLS handshake]

You can improve on this in multiple ways (include handshake for both noise & tls in the first message, or in case of incompatibility send a TLS handshake from B to A to save half a RTT)

Menduist:
If I understand correctly, we won't be able to support multiple encryption scheme on a single port anymore?
ie, need to do
`/ip6/::1/tcp/1234/tls/`
`/ip6/::1/tcp/4567/noise/`

That seems bad to me, it would become very difficult to, for instance, update noise to a new non-compatible version. (if that seem unlikely to you, https://github.com/libp2p/specs/issues/355 requires a new non-compatible noise version to be fixed, for instance)
The only solution would be to listen on multiple ports for the duration of the transition, which forces end-users to setup multiple ports, multiple redirection in their NAT, etc

I'm also not 100% convinced by the "downgrade" argument. Sure, it applies to mobile phones because you can't update the 3310s to safer protocols, but in our case, if a protocol is less secure, just disable it (like it was done for secio), push an update, and problem solved.


Finally, I'm having a hard time to imagine how active relays work in this scenario.
A: only supports tcp/noise
R: supports quic + tcp/noise
B: supports quic + noise + tls (not TCP)

How does B advertise that it handles noise? Is it supposed to advertise something like
- `/ip6/::1/udp/1234/quic/`
- `/ip6/::1/udp/1234/quic/p2p-circuit-security/noise`
- `/ip6/::1/udp/1234/quic/p2p-circuit-security/tls`
? 

Weird, but ok. Except that now we're back to my first remark, when it will receive a connection coming from a relay, it will not know if it's a noise or a tls one. So again, need to listen on 2 different ports.

Now, let's be crazy and say that B also handles Webrtc, it will need to advertise:
- `/ip6/::1/udp/1234/webrtc/`
- `/ip6/::1/udp/1234/webrtc/p2p-circuit-security/noise`
- `/ip6/::1/udp/1235/webrtc/p2p-circuit-security/tls`
- `/ip6/::1/udp/1236/quic/`
- `/ip6/::1/udp/1236/quic/p2p-circuit-security/noise`
- `/ip6/::1/udp/1237/quic/p2p-circuit-security/tls`


To sum up, it seems to me that we are putting "capabilities" into the MA, and that doesn't seem right. Each MA is already a capability. My understanding is that this grew out of the goal of 0/low-RTT setups, but low RTTs can also be achieved with a negotiation.

A: supports Noise, TLS, Secio
B: supports TLS

A -> B: open connection using Noise, TLS or Secio
If you are compatible with noise, here is some early data: [the noise handshake]
B -> A: only compatible with TLS, please retry
A -> B: [TLS handshake]

You can improve on this in multiple ways (include handshake for both noise & tls in the first message, or in case of incompatibility send a TLS handshake from B to A to save half a RTT)

marten-seemann:
@Menduist Thank for you for your review.

There are multiple reasons for including the security in the multiaddr:
1. It saves one roundtrip (for a reply regarding your example see below).
2. It prevents downgrade attacks.
3. It makes it possible to disguise libp2p traffic as regular web traffic (by running /tls on port 443).

Note that is not _required_ to run different handshake algorithms on different ports. You could also try to demux on the first (couple of) bytes of the first handshake message. Using different ports is the implementation strategy we will use in go-libp2p though.
The censorship-resistance argument is actually the convincing one for me.

> I'm also not 100% convinced by the "downgrade" argument. Sure, it applies to mobile phones because you can't update the 3310s to safer protocols, but in our case, if a protocol is less secure, just disable it (like it was done for secio), push an update, and problem solved.

Sure, this is clear-cut example. If a protocol is broken, you disable it. But the more realistic option is that you (strongly) prefer one protocol over the other, even if it's not considered broken yet. This was actually the case with TLS / Noise vs. secio for the longest time.

> To sum up, it seems to me that we are putting "capabilities" into the MA, and that doesn't seem right.

Another angle to look is that we're using port number to run multiple incompatible protocols on the same host, which is _exactly_ what port numbers were invented for.

> A -> B: open connection using Noise, TLS or Secio
> If you are compatible with noise, here is some early data: [the noise handshake]
> B -> A: only compatible with TLS, please retry
> A -> B: [TLS handshake]

1. Saving a connection after an invalid message has been sent has proved quite difficult. Realistically, we'd need to introduce special framing for that.
2. You're assuming that Noise is supported by all / most implementations. That's the case today, but may not be the case in the future.

Menduist:
Thanks for the details, let me address the benefits you talked about:

### It prevents downgrade attacks.

I would argue that since different security protocols runs on different ports, it becomes even easier to pull up a downgrade attack.
Let me introduce the Generalized MITM Libp2p Connection Downgrader™:
```
proc newConnectionDetected(connection):
  if ICanDecryptTheConnection:
    # hurray!
  else:
    blockThePort()
```
Since libp2p will try every address, it will eventually try a port/security protocol combo that is unsafe, or the connection won't go through. It's a bit slower than the previous attack, but can be sped up by knowing the "safe" ports before hand, and blocking them

Note that even if you manage to put every security protocol on the same port,
This attack could still work: instead of blocking the port, just block the specific message.


I presented in my first review a solution to downgrade attacks, I'll go in more detail here:
We could, once the connection is safe, restart the security negotiation. It doesn't have to block the upgrade process, and doesn't have to do anything in the end, but _if_ we come out with a different security protocol, we kill the connection because we've been downgraded.

It's slow and stupid, but the fact that this security exist should deter anyone from trying to downgrade the connections.
It also has the added benefit that we could reuse the same system to check security protocols we don't control (eg, in quic, check that we use TLS 1.3 or whatever the case may be)
We could piggy-back by adding it to the identify / SPR, which wouldn't cause any more round trips, and would have other benefits

### It saves one roundtrip

#### Saving a connection after an invalid message has been sent has proved quite difficult
See my proposal here: https://github.com/libp2p/specs/pull/349#discussion_r822768801
It's very straightforward, once `protocol-select` is implemented, I suspect this would represent ~15 LOC in nim-libp2p to add this extension

#### You're assuming that Noise is supported by all / most implementations
I used noise as an example, but it could be any protocol. Here are a few possibilities to improve that:
- Send supported security protocols in the SPR, this way, if you have the SPR of a peer (which will hopefully become the most common way to dial a peer), the round-trip will be saved
- If you don't have the SPR, we could add one "optimistic data" per offered protocol. It's a bit heavy, depending on the number of protocol you offer, but possible

Also, I'm may be biased because of how the nim-libp2p is used, but sometimes, the security protocols are defined by an upper spec, eg [eth2](https://github.com/ethereum/consensus-specs/blob/825a39577e06d312661e08eb461a317635336700/specs/phase0/p2p-interface.md#encryption-and-identification)

### It makes it possible to disguise libp2p traffic as regular web traffic (by running /tls on port 443).
I agree. However, if we take a step back, that is, as far as I know, just a "theoretical" problem for now, ie, no one blocked libp2p using pattern detection so far. If it happens one day*, we already have some transports that can't be detected (WSS, Quic, Webrtc), which would leave us enough time to apply proper countermeasures.

`*` I personally hope that libp2p will spread enough, and "blocking libp2p" won't just imply "I can't access eth2, ipfs and some dApps", but "I can't talk to my friends nor play this cool game". But as they say, hope for the best, plan for the worst


### Other comments
> Another angle to look is that we're using port number to run multiple incompatible protocols on the same host, which is exactly what port numbers were invented for.

Yes, though you don't see a new port being used everytime TLS is updated. I mean, in the 30 years HTTP existed, they only used 2 ports (80, 443), but in my last relay example I already need 4 ports.
imo, more port is just super user-unfriendly, and that applies at every level:
- libp2p users will have to setup who knows how many flags to setup every port, and transform into ma
- users of libp2p applications will have to configure theses ports in their firewall

Let's just take as an example "noise migration from `/noise` to `/noise/2`" from the perspective of a eth2 CL user:
- New release of my client
- Need to configure a new port in my `.service` configuration
- Need to open the port in my firewall
- [Few months later] New release of my client
- Need to notice that the old port is no longer used, and close it in my firewall

You could argue that the port should be transparent to the user, and in some cases that apply. But if you run a node worth >70k$, you _probably_ want drastic security on it, with every unknown port blocked.
That whole upgrade process is kinda painful, and quite error-prone.

### Closing thoughts
With this proposal, the downgrade is not addressed completely, the round trip could be saved in other ways.

The complexity added to relays is imo, unnecessary (if a relay is blocking libp2p.. just use another one)

imo, the only valid improvement is the ability to do `/ip4/6.6.6.6/tcp/1234/tls` in the case where your ISP/whatever actively blocks libp2p traffic. That's a _very_ specific use case, that, AFAIK, no one hit yet.
However, this proposal have multiple downsides: you need more ports, security protocols upgrades becomes extra painful

So it should not become a default, but just something we could do if at some point we start to see traffic-based censorship

addressing/README.md
mxinden: Renamed in https://github.com/libp2p/specs/pull/353/commits/3c1487a106943ade936c5185def680c67c9496c8.

PR #354, opened 2021-07-25
split the multiaddr change out of this spec

marten-seemann:
This PR goes with #353, and aim to split the multiaddr change into a separate change, unrelated to Protocol Select.

### DIFF
--- a/protocol-select/README.md
+++ b/protocol-select/README.md
@@ -69,7 +69,9 @@ Select]_ protocol.
 
 ### Improvements over _[Multistream Select]_
 
-- **Protocol Select** requires security protocols to be advertised, and doesn't 
+- **Connection establishment**
+  
+  **Protocol Select** requires security protocols to be advertised, and doesn't 
   allow negotiation them. For optimized implementations, stream muxer
   negotiation will take zero round-trips for the client (depending on the
   details of the cryptographic handshake protocol).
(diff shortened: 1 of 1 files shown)


protocol-select/README.md
```diff
@@ -69,31 +69,11 @@ Select]_ protocol.
...
-  security protocol takes one round-trip in the common case with **[Multistream
-  Select]**. On top of that negotiating a stream multiplexer (on TCP) takes
-  another round-trip.
-
-  **Protocol Select** on the other hand depends on security protocols being
-  advertised, thereby eliminating the need for negotiating them. For optimized
-  implementations, stream muxer negotiation will take zero round-trips for the
-  client (depending on the details of the cryptographic handshake protocol). In
-  that case, the client will be able to immediately open a stream after
+- **Protocol Select** requires security protocols to be advertised, and doesn't
```
mxinden, line 72:
```suggestion
- **Connection establishment**
  
  **Protocol Select** requires security protocols to be advertised, and doesn't 
```

I would like to be consistent with the other bullet points, namely that each of them has a heading. Any objections to simply keep the previous one?

mxinden (commented):
Only one comment, other than that this looks good to me. Thanks for this pull request and https://github.com/libp2p/specs/pull/353.

mxinden (approved):
Looks good to me.

I think this can be merged, especially since the pull request is not targeting _master_ but _protocol-select.

PR #356, opened 2021-07-27
Clarify properties of the libp2p-noise handshake's second message

dnkolegov:
The libp2p-noise specification states the following related to early data (payloads): 
```
These payloads MUST be inserted into the first message of the handshake pattern that guarantees secrecy.
In practice, this means that the initiator must not send a payload in their first message.
Instead, the initiator will send its payload in message 3 (closing message),
whereas the responder will send theirs in message 2 (their only message)
```

According to Noise XX security properties, the second message with payload provides forward
secrecy, however, the sender has not authenticated the responder, so this payload might be sent to any
party, including an active attacker. So an active attacker can just establish a connection with the
responder host and get the early data. 

So, this MR just clarifies this moment.

### DIFF
--- a/noise/README.md
+++ b/noise/README.md
@@ -218,9 +218,10 @@ These payloads MUST be inserted into the first message of the handshake pattern
 **that guarantees secrecy**. In practice, this means that the initiator must not
 send a payload in their first message. Instead, the initiator will send its
 payload in message 3 (closing message), whereas the responder will send theirs
-in message 2 (their only message). It should be stressed, that the second message of the handshake pattern has forward
-secrecy, however the sender has not authenticated the responder, so this payload might be sent to any
-party, including an active attacker.
+in message 2 (their only message). It should be stressed, that the second
+message of the handshake pattern has forward secrecy, however the sender has not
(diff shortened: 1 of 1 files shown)


noise/README.md
```diff
@@ -218,7 +218,9 @@ These payloads MUST be inserted into the first message of the handshake pattern
 **that guarantees secrecy**. In practice, this means that the initiator must not
 send a payload in their first message. Instead, the initiator will send its
 payload in message 3 (closing message), whereas the responder will send theirs
-in message 2 (their only message).
+in message 2 (their only message). It should be stressed, that the second message of the handshake pattern has forward
+secrecy, however the sender has not authenticated the responder, so this payload might be sent to any
+party, including an active attacker.
```
mxinden, line 223:
```suggestion
in message 2 (their only message). It should be stressed, that the second
message of the handshake pattern has forward secrecy, however the sender has not
authenticated the responder, so this payload might be sent to any party,
including an active attacker.
```

mxinden (approved):
Thanks @dnkolegov.

Leaving this open for a bit longer to give everyone a chance to review.

PR #357, opened 2021-07-30
Fix tls.md

kpp:
The go implementation uses `OCTET STRING` instead of `BIT STRING` for elements of `SignedKey`.

The options are:

1) fix the go implementation
2) update the specs

However with 1) we should also notify every implementors to fix it, e.g. https://github.com/alanshaw/js-libp2p-quic/blob/1f3c1dd6dff3c2e94710a1ae76e815e0805c8a99/src/crypto.js#L44

### DIFF
--- a/tls/tls.md
+++ b/tls/tls.md
@@ -80,8 +80,8 @@ The public host key and the signature are ANS.1-encoded into the SignedKey data 
 
 ```asn1
 SignedKey ::= SEQUENCE {
-  publicKey BIT STRING,
-  signature BIT STRING 
+  publicKey OCTET STRING,
+  signature OCTET STRING 
 }
 ```
 


Stebalien:
Nice catch!

@mxinden what does rust-libp2p do here?

dvc94ch:
Nothing yet, this was found by @kpp testing the upcoming rust implementation.

Stebalien (approved):

Stebalien:
Thanks!

@raulk anyone else we should notify?

@dvc94ch related, are there any interop tests for rust? Either with go-libp2p or go-ipfs? We currently run interop tests with js-ipfs (using the IPFS HTTP API) and js-libp2p (using the [daemon](https://github.com/libp2p/go-libp2p-daemon/), also implemented in javascript).

PR #358, opened 2021-08-03
Define precisely how to marshal SubjectPublicKeyInfo structure in tls.md

kpp:
This is a clarification how to marshal SubjectPublicKeyInfo according to the go implementation (https://github.com/libp2p/go-libp2p-tls/blob/7530faa07acbfc0aa918c072c2cb35d3c8d5d859/crypto.go#L143 and https://pkg.go.dev/crypto/x509#MarshalPKIXPublicKey).

### DIFF
--- a/tls/tls.md
+++ b/tls/tls.md
@@ -74,7 +74,7 @@ In order to prove ownership of its host key, an endpoint sends two values:
 
 The public host key allows the peer to calculate the peer ID of the peer it is connecting to. Clients MUST verify that the peer ID derived from the certificate matches the peer ID they intended to connect to, and MUST abort the connection if there is a mismatch.
 
-The peer signs the concatenation of the string `libp2p-tls-handshake:` and the public key that it used to generate the certificate carrying the libp2p Public Key Extension, using its private host key. The encoded public key is a `SubjectPublicKeyInfo` structure (see RFC 5280, Section 4.1):
+The peer signs the concatenation of the string `libp2p-tls-handshake:` and the encoded public key that it used to generate the certificate carrying the libp2p Public Key Extension, using its private host key. The public key is encoded as a `SubjectPublicKeyInfo` structure (see RFC 5280, Section 4.1):
 
 ```asn1
 SubjectPublicKeyInfo ::= SEQUENCE {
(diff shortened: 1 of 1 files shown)


tls/tls.md
```diff
@@ -74,7 +74,20 @@ In order to prove ownership of its host key, an endpoint sends two values:
 
 The public host key allows the peer to calculate the peer ID of the peer it is connecting to. Clients MUST verify that the peer ID derived from the certificate matches the peer ID they intended to connect to, and MUST abort the connection if there is a mismatch.
 
-The peer signs the concatenation of the string `libp2p-tls-handshake:` and the public key that it used to generate the certificate carrying the libp2p Public Key Extension, using its private host key. This signature provides cryptographic proof that the peer was in possession of the private host key at the time the certificate was signed. Peers MUST verify the signature, and abort the connection attempt if signature verification fails.
+The peer signs the concatenation of the string `libp2p-tls-handshake:` and the public key that it used to generate the certificate carrying the libp2p Public Key Extension, using its private host key. The encoded public key is a `SubjectPublicKeyInfo` structure (see RFC 5280, Section 4.1):
```
mxinden, line 77:
```suggestion
The peer signs the concatenation of the string `libp2p-tls-handshake:` and the encoded public key that it used to generate the certificate carrying the libp2p Public Key Extension, using its private host key. The public key is encoded as a `SubjectPublicKeyInfo` structure (see RFC 5280, Section 4.1):
```

Small suggestion, would be easier for me to read.

mxinden (approved):
This looks good to me. Thanks @kpp for extending the docs. This kind of contribution is very much appreciated!

Any additional comments @marten-seemann? If not, I will merge here.

marten-seemann:
All that we do is marshal the public key: https://github.com/libp2p/go-libp2p-tls/blob/7530faa07acbfc0aa918c072c2cb35d3c8d5d859/crypto.go#L167-L174
I'm not sure if we need to say anything more than https://pkg.go.dev/crypto/x509#MarshalPKIXPublicKey says, as this seems to be the standard way of marshaling the key. I find spelling it out explicitly, as suggested in this PR, confusing, as it suggests that we're doing something non-standard here.

kpp:
@marten-seemann Would it be better if I replace that statement with: `The public key is encoded as a `SubjectPublicKeyInfo` structure as described in RFC 5280, Section 4.1:`?

marten-seemann:
> Would it be better if I replace that statement with: "The public key is encoded as a SubjectPublicKeyInfo structure as described in RFC 5280, Section 4.1:"?

Yes, I think referencing the RFC makes a lot of sense here.

kpp:
Done. Sorry for the delay.

mxinden:
Friendly ping @marten-seemann.

marten-seemann (approved):

PR #361, opened 2021-08-22
assign roles after TCP simultaneous connect, describe how hole punching works on QUIC

marten-seemann:


mxinden (approved):
Thanks for preparing a pull request!

On first read, when using TCP, I thought we should instead support both negotiating the role (client / server) in-band via the _Multistream Select Simultaneous Open_ extension and the out-of-band assignment through the _DCUtR_ protocol, as suggested in this pull request.

Thinking about it some more, I am in favor of your suggestion, assigning the role out-of-band through the _DCUtR_ protocol, thus simplifying the process.

Just to make sure I am not missing something, we would thus only need the _Multistream Select Simultaneous Open_ extension for uncoordinated simultaneous TCP connections, correct?

marten-seemann:
> Thinking about it some more, I am in favor of your suggestion, assigning the role out-of-band through the DCUtR protocol, thus simplifying the process.

This is essential if we don't want to couple Flare with Protocol Select (as Protocol Select doesn't come with a low-latency option to assign roles in the case of TCP Sim Open).

> Just to make sure I am not missing something, we would thus only need the Multistream Select Simultaneous Open extension for uncoordinated simultaneous TCP connections, correct?

That is correct.

Stebalien:
> Thinking about it some more, I am in favor of your suggestion, assigning the role out-of-band through the DCUtR protocol, thus simplifying the process.

What if I perform a simultaneous open _without_ coordination? As I've pointed out before, this is a _very_ common case and it needs to be supported before we start making changes.

marten-seemann:
@Stebalien There’s a section “Uncoordinated TCP Simultaneous Open” in the Protocol Select spec (https://github.com/libp2p/specs/pull/349).

Stebalien:
Thanks. I'm mixing issues.

PR #379, opened 2021-12-08
ROADMAP: Add section on "Automated compatibility testing"

mxinden:
I would like to propose (a) including _automated compatibility testing_ into our roadmap and (b) prioritizing it within the _Evolve_ section.

What do folks think?

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -21,8 +21,9 @@ third-party ownership of data.
         - [🤖 libp2p as a WASM library](#🤖-libp2p-as-a-wasm-library)
     - [Evolve](#evolve)
         - [🤝 Low latency, efficient connection handshake](#🤝-low-latency-efficient-connection-handshake)
+        - [🧪 Automated compatibility testing](#🧪-automated-compatibility-testing)
         - [🕸 Unprecedented global connectivity](#🕸-unprecedented-global-connectivity)
-        - [Peer Routing Records](#peer-routing-records)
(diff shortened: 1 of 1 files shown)


ROADMAP.md
```diff
@@ -243,6 +244,21 @@ ping-pong).
 
 - [Protocol Select specification](https://github.com/libp2p/specs/pull/349)
 
+### 🧪 Automated compatibility testing
+
+**What?** There are more than 6 implementations of the [libp2p specification] in
+different languages. We need to ensure compatibility of all combination of these
```
marten-seemann, line 250:
```suggestion
different languages. We need to ensure compatibility of all combinations of these
```

ROADMAP.md
```diff
@@ -243,6 +244,21 @@ ping-pong).
...
+implementations. Given the number of libp2p implementations and the amount of
+libp2p protocols, verifying compatibility on a continuous basis can only be
+tackled in an automated fashion.
+
+**Why?** Automated compatibility testing allows us to move fast with confidence
+and will increase trust in the libp2p project as a whole.
```
marten-seemann, line 256: And move some tests currently at the IPFS layer to where they actually belong.

marten-seemann (approved):

ROADMAP.md
mxinden: Agreed that we should do that, though I would consider it a detail. Do you feel strongly on mentioning it in our roadmap?

marten-seemann: No, that was just a general comment.

PR #382, opened 2021-12-08
relay/circuit-v2: Do minor fixes

thomaseizinger:
A few things I noticed while reading through the spec.

Not sure if we want to bump the revision for such minor fixes? Can always drop the last commit!

### DIFF
--- a/relay/circuit-v2.md
+++ b/relay/circuit-v2.md
@@ -269,7 +269,8 @@ Peer {
 }
 ```
 
-***Note that active relay functionality is considered deprecated for security reasons, at least in public relays.***
+***Note: Active relay functionality is considered deprecated for security reasons, at least in public relays.***
+
 However, the protocol reserves the field to support the functionality for the rare cases where it is actually desirable to use active relay functionality in a controlled environment.
 
 If the relay has a reservation (and thus an active connection) from the peer, then it opens the second hop of the connection using the `stop` protocol; the details are not relevant for the `hop` protocol and the only thing that matters is whether it succeeds in opening the relay connection or not.
--- a/relay/circuit-v2.md
+++ b/relay/circuit-v2.md
@@ -233,6 +233,7 @@ The struct has the following fields:
 Limit {
   duration = ...
   data = ...
+}
 ```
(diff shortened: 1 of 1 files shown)


relay/circuit-v2.md
```diff
@@ -4,7 +4,7 @@ This is the version 2 of the libp2p Circuit Relay protocol.
 
 | Lifecycle Stage | Maturity       | Status | Latest Revision |
 |-----------------|----------------|--------|-----------------|
-| 1A              | DRAFT          | Active | r0, 2021-05-05  |
+| 1A              | DRAFT          | Active | r1, 2021-12-09  |
```
mxinden, line 7: I don't think a version bump is necessary, given that this does not change any functionality, i.e. given that implementations do not have to change based on this.

mxinden (commented):
Thanks. One comment, otherwise looks good to me.

relay/circuit-v2.md
```diff
@@ -226,7 +226,7 @@ Reservation {
  `p2p-circuit/p2p/QmPeer` where `QmPeer` is its peer ID.
 - the `voucher` is the binary representation of the reservation voucher -- see [Reservation Vouchers](#reservation-vouchers) for details.
 
-The `limit` field, if present, provides information about the limits applied by the relay in relayed connection. When omitted, it indicates that the relay does not apply any limits.
+The `limit` field in `HopMessage`, if present, provides information about the limits applied by the relay in relayed connection. When omitted, it indicates that the relay does not apply any limits.
 
 The struct has the following fields:
 ```
```
yusefnapora, line 232: One more missing `}` on the def for `Limit` here :)

yusefnapora (approved):
Hey all, sorry I slept on reviewing the initial draft of this spec. My belated take is that it's really well done and explains the protocol & its motivations clearly. The changes in this PR seem great and non-controversial 😄

relay/circuit-v2.md
thomaseizinger: Dropped the commit.

mxinden (approved):
Thanks Thomas.

PR #385, opened 2021-12-27
autonat/README: Add TOC and describe stream lifecycle

mxinden:
Note that this pull request does not target `master` but the `autonat` branch.

### DIFF
--- a/autonat/README.md
+++ b/autonat/README.md
@@ -24,7 +24,9 @@ and spec status.
 
 ## Table of Contents
 
-TODO: fill in
+- [Overview](#overview)
+- [AutoNAT Protocol](#autonat-protocol)
+- [Security Considerations](#security-considerations)
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


autonat/README.md
```diff
@@ -73,6 +76,27 @@ If all dials fail, the receiver sends a `DialResponse` message with the
...
+changing their set of addresses they're listening on.
+
+## RPC messages
+
+Messages are exchanged by:
+
+1. Opening a new stream.
+2. Sending the RPC request message.
+3. Listening for the RPC response message.
+4. Closing the stream.
```
marten-seemann, line 94: I don't think we need to specify stream closing. Really, after sending the request message, you can already close the send side of the stream.

marten-seemann (approved):

autonat/README.md
mxinden, line 94:
```suggestion
```

PR #392, opened 2022-02-11
kad-dht/README: Document client mode

mxinden:
This commit documents the DHT client mode, keeping routing tables clear
of clients (e.g. laptops), while still allowing clients to utilize the
functionality of the DHT.

A note on support across the implementations: Implemented in Go, pending in Rust (https://github.com/libp2p/rust-libp2p/issues/2032), uncertain about other languages.

//CC @aschmahmann and @MarcoPolo

### DIFF
--- a/kad-dht/README.md
+++ b/kad-dht/README.md
@@ -68,6 +68,31 @@ The concurrency of node and value lookups are limited by parameter `α`, with a
 default value of 3. This implies that each lookup process can perform no more
 than 3 inflight requests, at any given time.
 
+## Client and server mode
+
+When the libp2p Kademlia protocol is run on top of a network of heterogeneous
+nodes, unrestricted nodes should operate in _server mode_ and restricted nodes,
+e.g. those with intermittent availability, high latency, low bandwidth, low
+CPU/RAM/Storage, etc., should operate in _client mode_.
+
+As an example, running the libp2p Kademlia protocol on top of the Internet,
+publicly routable nodes, e.g. servers in a datacenter, would operate in _server
+mode_ and non-publicly routable nodes, e.g. laptops behind a NAT and firewall,
+would operate in _client mode_.
+
+Nodes, both those operating in _client_ and _server mode_, add another node to
+their routing table if and only if that node operates in _server mode_. This
+distinction allows restricted nodes to utilize the DHT, i.e. query the DHT,
+without decreasing the quality of the distributed hash table, i.e. without
+polluting the routing tables.
+
+Nodes operating in _server mode_ advertise the libp2p Kademlia protocol
+identifier via the [identify protocol](../identify/README.md). In addition
+_server mode_ nodes accept incoming streams using the Kademlia protocol
+identifier. Nodes operating in _client mode_ do not advertise support for the
+libp2p Kademlia protocol identifier. In addition they do not offer the Kademlia
+protocol identifier for incoming streams.
+
 ## DHT operations
 
 The libp2p Kademlia DHT offers the following types of operations:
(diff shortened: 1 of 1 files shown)


kad-dht/README.md
```diff
@@ -290,6 +315,7 @@ implementations should include a lookup for their own peer ID.
 Every repetition is subject to a `QueryTimeout` (default: 10 seconds), which
 upon firing, aborts the run.
 
+
```
aschmahmann, line 318: extra space intentional?

kad-dht/README.md
```diff
@@ -68,6 +68,31 @@ The concurrency of node and value lookups are limited by parameter `α`, with a
...
+When the libp2p Kademlia protocol is run on top of a network of heterogeneous
+nodes, unrestricted nodes should operate in _server mode_ and restricted nodes,
+e.g. those with intermittent availability, high latency, low bandwidth, low
+CPU/RAM/Storage, etc., should operate in _client mode_.
+
+As an example, running the libp2p Kademlia protocol on top of the Internet,
+publicly routable nodes, e.g. servers in a datacenter, would operate in _server
+mode_ and non-publicly routable nodes, e.g. laptops behind a NAT and firewall,
+would operate in _client mode_.
```
aschmahmann, line 81: This seems pretty good, although the advice here is all sort of fuzzy and situational. I'm adding some context in case you think it's reasonable to rephrase.

The definition of intermittent availability depends on the parameters of a given kad-dht network (replication factor, republishing periods, record expiration times, etc.). For example, having a laptop with UPnP support behind a NAT that's open for a few hours consecutively is probably a fine DHT server for networks with replication factors high enough to accommodate that (e.g. tuned to assume many nodes disappear after an hour).

The key is basically that nodes should be in client mode if they think they'll ultimately be a drag on the network (e.g. everyone behind a symmetric NAT or firewall), but what that means is up to them. Most implementations should probably do their best to figure this out probabilistically/automatically so that normal users don't have to consider this.

aschmahmann (approved):
LGTM. Thanks 🙏

kad-dht/README.md
aschmahmann: I'm not sure if this is covered elsewhere in the spec, but networks may also want to add their own rules for routing table membership to exclude bad peers (e.g. those without public IP addresses) so they're not just relying on users correctly setting client mode. However, if using rules they should probably be as symmetric/uniformly applicable as possible to maintain the coherence of the network's routing tables.

mxinden: https://github.com/libp2p/specs/pull/392/commits/710c4998afa02a07aa871492d01b5f5f9f2f70c6 adds the factors above as examples to take into consideration. I am not sure how to make it less fuzzy without loosing generality and thus not being applicable to all use-cases. Suggestions welcome.

mxinden:
Thanks for the super quick review @aschmahmann!

kad-dht/README.md
mxinden: :pray: removed with https://github.com/libp2p/specs/pull/392/commits/710c4998afa02a07aa871492d01b5f5f9f2f70c6.

kad-dht/README.md
mxinden: > I'm not sure if this is covered elsewhere in the spec, but networks may also want to add their own rules for routing table membership to exclude bad peers (e.g. those without public IP addresses) so they're not just relying on users correctly setting client mode. However, if using rules they should probably be as symmetric/uniformly applicable as possible to maintain the coherence of the network's routing tables.

Oh, just saw this. I don't think this is covered anywhere. Agreed. I am not sure it needs to be documented though.

PR #393, opened 2022-02-14
Delete the outdated link in peer-ids.md

vogdb:


peer-ids/peer-ids.md
```diff
@@ -47,7 +47,7 @@ format used to store keys on disk is also included as a reference for libp2p
 implementors who would like to import existing libp2p key pairs.
 
 Key encodings and message signing semantics are
-[covered below](#how-keys-are-encoded-and-messages-signed).
+[covered below](#key-types).
```
marten-seemann, line 50: This should not link to #key-types, but just to `Keys`. Given that that's the next section anyway, this doesn't need to be a link.

marten-seemann (changes requested):

peer-ids/peer-ids.md
vogdb: Done, also changed the PR title.

marten-seemann (commented):
Thanks!

marten-seemann (approved):

PR #396, opened 2022-03-04
Fix two minor gossipsub v1.1 spec problems

jtraglia:
1. Update `PublishThreshold` constraint in the table. This does not accurately reflect what the text says.
> `publishThreshold`: when a peer's score drops below this threshold, self published messages are not propagated towards this peer when (flood) publishing. This threshold should be negative, and **less than or equal** to the gossip threshold.

2. Change `IPColocationFactorThreshold` from a `Float` to an `Integer`. It does not make sense for this value to be a float. You cannot have a fractional "number of IPs a peer may have before being penalized."


vyzo (approved):
thank you

PR #397, opened 2022-03-04
Make gossip score vars start with uppercase letter

jtraglia:
Noticed that these variables did not start with a capital letter like the other instances in the document.


vyzo (approved):
thank you

Issue #399, opened 2022-03-07
pubsub: consider single required topic hash


wemeetagain:
Currently, rust-libp2p assumes a single required topic hash (https://github.com/libp2p/rust-libp2p/blob/a168410dbed0d0941f2e5a14543206044ccb2260/protocols/gossipsub/src/types.rs#L123)

This simplifies the code a lot and seems to be the standard use-case for pubsub protocols.

Perhaps this can be applied at the spec level and rolled out to other implementations.

cc @dapplion

vyzo:
No, we've had this discussion before -- we are not changing the spec to require hashes.

Note that messages have a single topic for a while now, if that's what you mean; the spec might need to be updated for that.

vyzo:
we changed the protobuf too in go pubsub, but backwards compatible (same field id).

Mind opening a pr to update the spec?

wemeetagain:
> Mind opening a pr to update the spec?

Sure thing

mxinden:
If I am not mistaken this can be closed with https://github.com/libp2p/specs/pull/400.

PR #400, opened 2022-03-07
pubsub: use single topic in Message protobuf

wemeetagain:
Align the Message protobuf with the rust-libp2p definition

See #399


vyzo:
thank you!

PR #402, opened 2022-03-11
connection/README: Recommend yamux over mplex

mxinden:
The mplex multiplexer does not provide backpressure on a stream. Yamux does.
Given that this is a fundamental feature of a stream multiplexer, this commit
changes the spec to recommend yamux over mplex.

While yamux is not ideal, e.g. lacks backpressure on the connection level and
the number of streams, it is superior in all cases to mplex.

---

I don't think we should recommend mplex over yamux any longer. Thoughts?

### DIFF
--- a/connections/README.md
+++ b/connections/README.md
@@ -112,8 +112,8 @@ functions.
 Each protocol supported by a peer is identified using a unique string called a
 **protocol id**. While any string can be used, the conventional format is a
 path-like structure containing a short name and a version number, separated by
-`/` characters. For example: `/mplex/1.0.0` identifies version 1.0.0 of the
-[`mplex` stream multiplexing protocol][mplex]. multistream-select itself has a
+`/` characters. For example: `/yamux/1.0.0` identifies version 1.0.0 of the
+[`yamux` stream multiplexing protocol][yamux]. multistream-select itself has a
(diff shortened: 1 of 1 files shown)


connections/README.md
```diff
@@ -112,8 +112,8 @@ functions.
 Each protocol supported by a peer is identified using a unique string called a
 **protocol id**. While any string can be used, the conventional format is a
 path-like structure containing a short name and a version number, separated by
-`/` characters. For example: `/mplex/1.0.0` identifies version 1.0.0 of the
-[`mplex` stream multiplexing protocol][mplex]. multistream-select itself has a
+`/` characters. For example: `/yamux/1.0.0` identifies version 1.0.0 of the
```
marten-seemann, line 115: That's not the version number we use for yamux.

mxinden: Are you sure @marten-seemann? It matches the one referenced in `go-libp2p/defaults.go`:

https://github.com/libp2p/go-libp2p/blob/228c6297685f2d95577680128df982e827abf694/defaults.go#L31-L38

marten-seemann: Oh sorry, I mixed this up with mplex!

marten-seemann (approved):

PR #403, opened 2022-03-16
add our metrics effort to the short-term roadmap

marten-seemann:


BigLep (approved):
Looks good to me!   I do think we'll get some basic "automated compatibility testing" as part of "Unprecedented global connectivity".  As long as we get some basic automated compatibility testing going before we spin up new efforts like metrics, I'm good.

ROADMAP.md
```diff
@@ -248,6 +249,24 @@ rest of the system.
...
+
+### ⏱ Full Observability
+
+**What?** libp2p should expose a wide set of metrics, making it easy to
+monitor the system.
+
+Metrics should include:
+- Transport metrics (TCP, QUIC, security protocols, stream multiplexers)
+- Swarm metrics
+- Resource Manager metrics
```
mxinden, line 261:
```suggestion
```

I would categorize the "Resource Manager" a go-libp2p implementation detail. While rust-libp2p (partially) supports the same mechanism, it does not manifest as a separate component.

mxinden (commented):
Sounds good to me.

I would prioritize it below the "Automated compatibility testing", though that is not a strong opinion.

Contrary to what I suggested in past discussions, I do now wonder whether we should standardize on a set of base metric names across the implementations. E.g. each implementation SHOULD expose `libp2p_connection_established{transport="tcp", security="noise", muxer="yamux"}`. We would only be consistent across implementations on a subset / base-set of metrics. What do you think @marten-seemann?

ROADMAP.md
marten-seemann: I'm ok with removing it here, but this is a very important metric to track.

marten-seemann:
> Contrary to what I suggested in past discussions, I do now wonder whether we should standardize on a set of base metric names across the implementations. E.g. each implementation SHOULD expose libp2p_connection_established{transport="tcp", security="noise", muxer="yamux"}. We would only be consistent across implementations on a subset / base-set of metrics. What do you think @marten-seemann?

I think I've suggested that a while ago :) Fully agree.

mxinden (approved):

mxinden:
> > Contrary to what I suggested in past discussions, I do now wonder whether we should standardize on a set of base metric names across the implementations. E.g. each implementation SHOULD expose libp2p_connection_established{transport="tcp", security="noise", muxer="yamux"}. We would only be consistent across implementations on a subset / base-set of metrics. What do you think @marten-seemann?
> 
> I think I've suggested that a while ago :) Fully agree.

Yes, you did! Cool. I would suggest we define a set of base metrics here in the specs repository.

PR #404, opened 2022-04-04
add a WebTransport spec

marten-seemann:
This PR adds a basic spec for WebTransport, based on https://pl-strflt.notion.site/Enabling-WebTransport-across-the-libp2p-network-c6849c7252a1469a828a6c3e4b6abcad.

Most important things to watch out for when reviewing:
1. The multiaddr format: The multiaddr contains the hashes of the certificates used, e.g. `/ip4/1.2.3.4/udp/443/quic/webtransport/<hash1><hash2>`. Is this a sane construction?
2. We need to run a libp2p handshake on the WebTransport connection to verify peer IDs.

UPDATE 2022-07-26: Updated the link to the Notion page. I was mistaken in believing that the previous link was already public.


webtransport/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+When using self-signed certificates, nodes need to take care to regularly renew their certificate. In the following, the RECOMMENDED logic for rolling certificates is described. At first boot of the node, it creates a self-signed certificate with a validity of 14 days, and publishes a multiaddr containing the hash of that certificate. After 10 days, the node prepares the next certificate, setting the `NotBefore` date of that certificate to the expiration date (or shortly before that) of the first certificate, and an expiration of 14 days after that. The node continues using the old certificate until its expiry date, but it already advertises a multiaddr containing both certificate hashes. This way, clients will be able to connect to the node, even if they cache the multiaddr for multiple days.
+
+## Addressing
+
+Webtransport multiaddresses are composed of a QUIC multiaddr, followed by `/webtransport` and a list of multihashes of the certificates that the server uses.
+Examples:
+* `/ip4/1.2.3.4/udp/443/quic/webtransport/`
+* `/ip6/fe80::1ff:fe23:4567:890a/udp/1234/quic/webtransport/<hash1><hash2><hash3>`
```
willscott, line 35: might be useful to describe what the string serialization of hashes here are. would it be the base58 direct encoding of the multihash, or the CID encoding with a multibase prefix?

webtransport/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+If it is able to present a CA-signed certificate the list of certificate hashes SHOULD be empty. When using self-signed certificates, the server MUST include the hash(es) in the multiaddr.
+
+## WebTransport HTTP endpoint
+
+WebTransport needs a HTTPS URL to establish a WebTransport session, e.g. `https://example.com/webtransport`. As multiaddrs don't allow the encoding of URLs, this spec standardizes the endpoint. The HTTP endpoint of a libp2p WebTransport servers MUST be located at `/.well-known/libp2p-webtransport`.
```
willscott, line 41: multiaddrs can allow a conversion from URLs, we're using it as a signalling method in [legs](https://github.com/filecoin-project/go-legs/blob/main/httpsync/multiaddr/convert.go#L90)

webtransport/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+Unfortunately, the self-signed certificate doesn't allow the nodes to authenticate each others' peer IDs. It is therefore necessary to run an additional libp2p handshake on a newly established WebTransport connection.
+Once a WebTransport session is established, the clients opens a new stream and initiates a libp2p handshake (selecting the security protocol by running multistream, then performing the selected handshake).
+
+Note: Once we include the security protocol in the multiaddr (see https://github.com/libp2p/specs/pull/353), we will be able to shave off two (!!) round-trips here: Not running multistream saves one round trip. Furthermore, we'll be able to run the WebTransport handshake (i.e. the CONNECT request) in parallel with the cryptographic handshake, for example by transmitting the first handshake message as a URL parameter. The specifics of this is left to a future iteration of this spec.
+
+## Securing Streams
+
+All streams other than the stream used for the security handshake are protected using Salsa20. Two (symmetric) keys are derived from the master secrect established during the handshake, one for sending and for receiving on the stream.
```
willscott, line 56: It's unfortunate that there's no concept of the client identity at the transport level, so that we don't need an additional wrapping of encryption.

We should see if that's an API that can be proposed at the spec level.

webtransport/README.md
```diff
@@ -0,0 +1,60 @@
...
+WebTransport needs a HTTPS URL to establish a WebTransport session, e.g. `https://example.com/webtransport`. As multiaddrs don't allow the encoding of URLs, this spec standardizes the endpoint. The HTTP endpoint of a libp2p WebTransport servers MUST be located at `/.well-known/libp2p-webtransport`.
+
+To allow future evolution of the way we run the libp2p handshake over WebTransport, we use a URL parameter. The handshake described in this document MUST be signaled by setting the `type` URL parameter to `multistream`.
+
+Example: The WebTransport URL of a WebTransport server advertising `/ip4/1.2.3.4/udp/1443/quic/webtransport/` would be `https://1.2.3.4:1443/.well-known/libp2p-webtransport?type=multistream`.
+
+## Security Handshake
+
+Unfortunately, the self-signed certificate doesn't allow the nodes to authenticate each others' peer IDs. It is therefore necessary to run an additional libp2p handshake on a newly established WebTransport connection.
+Once a WebTransport session is established, the clients opens a new stream and initiates a libp2p handshake (selecting the security protocol by running multistream, then performing the selected handshake).
```
willscott, line 50: There should be something in the security handshake to specify version, so that a successor to Salsa20 can be introduced if needed.

webtransport/README.md
marten-seemann: Probably multibase, then we don't have to specify the encoding.

Looks like the encoding I suggested above doesn't work though. We can't have an optional hash argument.
I'm also not sure how to include the path (as you suggest in https://github.com/libp2p/specs/pull/404/files#r842506567) in the address, as `/webtransport` can't have 2 values: `/ip4/1.2.3.4/udp/443/quic/webtransport/<hash>/<encoded url>`.

Maybe the best solution is to introduce 2 new multiaddr components:
`/ip4/1.2.3.4/udp/443/quic/webtransport/certhashes/<hash1><hash2>/url/<encoded url>`, or even `/ip4/1.2.3.4/udp/443/quic/webtransport/certhash/<hash1>/certhash/<hash2>/url/<encoded url>`. wdyt?

webtransport/README.md
marten-seemann: The `?type` URL parameter gives us the option to switch to something else later. Unless the plan is to negotiate the algorithm, but that would make it more difficult to shave off the 2 RTTs with Protocol Select later.

webtransport/README.md
willscott: * I don't have a strong feeling about concatenation of hashes versus having each one in a separate `certhash` subcomponent.  Concatenation is probably less work for existing parsers that don't always think about repeated components.
* having a separate `/url/<encoded url>` seems reasonable as an optional over-ride to well-known.

marten-seemann: @mriise We definitely need to support 2 certs, but I don't see a reason to impose any restrictions that go further than the general length limit we have for multiaddrs.

The reason is the following: The way that `serverCertificateHashes` works is NOT by the browser telling the server "I'd like to see a cert with the following hash". Instead, the browser has a (list of) certificate hashes that it is willing to accept.

The node would generate one certificate for the first 2 weeks, another certificate for the 2 weeks after that, and so on. If you expect that peers will save your address for a prolonged period of time, you could advertise more than 2 hashes, but I expect that too be very rare.

mriise: > We definitely need to support 2 certs, but I don't see a reason to impose any restrictions that go further than the general length limit we have for multiaddrs.

This makes sense as a default definitely. Should there be checks to make sure they aren't the same hash to prevent people from only excepting one anyways?  

I guess a logical maximum for certs would be 14, 1 per day.

melekes: > We definitely need to support 2 certs

sorry for barging in, but I still don't understand why at least two certs are needed and not just one. could you explain or drop a link to spec or something (e.g. https://w3c.github.io/webtransport/#dom-webtransportoptions-servercertificatehashes)

webtransport/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+WebTransport needs a HTTPS URL to establish a WebTransport session, e.g. `https://example.com/webtransport`. As multiaddrs don't allow the encoding of URLs, this spec standardizes the endpoint. The HTTP endpoint of a libp2p WebTransport servers MUST be located at `/.well-known/libp2p-webtransport`.
+
+To allow future evolution of the way we run the libp2p handshake over WebTransport, we use a URL parameter. The handshake described in this document MUST be signaled by setting the `type` URL parameter to `multistream`.
+
+Example: The WebTransport URL of a WebTransport server advertising `/ip4/1.2.3.4/udp/1443/quic/webtransport/` would be `https://1.2.3.4:1443/.well-known/libp2p-webtransport?type=multistream`.
+
+## Security Handshake
+
+Unfortunately, the self-signed certificate doesn't allow the nodes to authenticate each others' peer IDs. It is therefore necessary to run an additional libp2p handshake on a newly established WebTransport connection.
```
melekes, line 49: is it possible to include a signature of the peer ID into the self-signed certificate? or it's just hacky and it's easier (more error prone) to negotiate noise on top?

webtransport/README.md
bertrandfalguiere: See line 28 of this document.

webtransport/README.md
marten-seemann: That's actually what we do for libp2p TLS.
As far as I can tell, there's no browser API to get access to the certificate obtained during the WebTransport handshake though.

webtransport/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+If it is able to present a CA-signed certificate the list of certificate hashes SHOULD be empty. When using self-signed certificates, the server MUST include the hash(es) in the multiaddr.
+
+## WebTransport HTTP endpoint
+
+WebTransport needs a HTTPS URL to establish a WebTransport session, e.g. `https://example.com/webtransport`. As multiaddrs don't allow the encoding of URLs, this spec standardizes the endpoint. The HTTP endpoint of a libp2p WebTransport servers MUST be located at `/.well-known/libp2p-webtransport`.
+
+To allow future evolution of the way we run the libp2p handshake over WebTransport, we use a URL parameter. The handshake described in this document MUST be signaled by setting the `type` URL parameter to `multistream`.
+
+Example: The WebTransport URL of a WebTransport server advertising `/ip4/1.2.3.4/udp/1443/quic/webtransport/` would be `https://1.2.3.4:1443/.well-known/libp2p-webtransport?type=multistream`.
```
Stebalien, line 45: nit: `?protocol=/multistream/1.0.0`

webtransport/README.md
Stebalien: We don't need that and we don't need double encryption. We can sufficiently identify the connection with the server certificate. Basically:

1. The client _knows_ it's talking to a server with a key that matches the certificate hash.
2. The client then sends a challenge of `(cert_hash, client_pid, server_pid, client_salt)`
3. The server responds with `server_salt, SIG(HASH(cert_hash || client_pid || server_pid || client_salt || server_salt))` to authenticate their end.
4. The client verifies this, then signs the same thing (maybe it signs the signature too?).

- In step 3, the server ties their peer ID to the connection.
- In step 4, the client ties _their_ peer ID to the connection.

marten-seemann: > 1. The client knows it's talking to a server with a key that matches the certificate hash.

That is correct iff it received the certificate hash in a signed peer record. Otherwise, the MITM could have generated a multiaddr containing its own cert hash.
Is that something we can require? Does js-libp2p support signed peer records?

> 2. The client then sends a challenge of (cert_hash, client_pid, server_pid, client_salt)

Note: Unfortunately, the client doesn't know which `cert_hash` was used by WebTransport. It only knows that _one_ of the hashes was considered valid, so it would have to send all of the cert hashes to the server.

> 3. The server responds with server_salt, SIG(HASH(cert_hash || client_pid || server_pid || client_salt || server_salt)) to authenticate their end.

It is crucial that the server validates that it owns _all_ cert hashes, otherwise a MITM attack is trivial. Furthermore (unless the certificate hash was signed in step 1), the client has to trust that the server actually performs this verification. Note that this kind of trust is not necessary when using TLS or Noise.

Here's a diagram of that handshake:
image

marten-seemann: @Kubuxu Would you mind taking a look at this proposal (the handshake that @Stebalien suggested in particular)?

Stebalien: > That is correct iff it received the certificate hash in a signed peer record. Otherwise, the MITM could have generated a multiaddr containing its own cert hash.

It doesn't need to be in a signed peer record, the server signs (and, as you noted, validates) this in step 2.

> Note: Unfortunately, the client doesn't know which cert_hash was used by WebTransport. It only knows that one of the hashes was considered valid, so it would have to send all of the cert hashes to the server.

Makes sense.

> It is crucial that the server validates that it owns all cert hashes, otherwise a MITM attack is trivial. Furthermore (unless the certificate hash was signed in step 1), the client has to trust that the server actually performs this verification. Note that this kind of trust is not necessary when using TLS or Noise.

Yes, the server needs to check it. But it's definitely not the clients problem if the server doesn't check. The server could just as easily leak their key.

marten-seemann: > It doesn't need to be in a signed peer record, the server signs (and, as you noted, validates) this in step 2.

Fair enough. As long as the client trusts the server to perform this check, we're good. What makes me a little bit nervous is that the client on its own can't verify that the connection isn't MITM'ed, which is different from the security guarantees you get from a regular TLS handshake. We probably can live with it (and as you say, the server could leak the key anyway), but it is a somewhat weaker security guarantee.

webtransport/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+Unfortunately, the self-signed certificate doesn't allow the nodes to authenticate each others' peer IDs. It is therefore necessary to run an additional libp2p handshake on a newly established WebTransport connection.
+Once a WebTransport session is established, the clients opens a new stream and initiates a libp2p handshake (selecting the security protocol by running multistream, then performing the selected handshake).
+
+Note: Once we include the security protocol in the multiaddr (see https://github.com/libp2p/specs/pull/353), we will be able to shave off two (!!) round-trips here: Not running multistream saves one round trip. Furthermore, we'll be able to run the WebTransport handshake (i.e. the CONNECT request) in parallel with the cryptographic handshake, for example by transmitting the first handshake message as a URL parameter. The specifics of this is left to a future iteration of this spec.
+
+## Securing Streams
+
+All streams other than the stream used for the security handshake are protected using Salsa20. Two (symmetric) keys are derived from the master secrect established during the handshake, one for sending and for receiving on the stream.
```
MarcoPolo, line 56: What's the reason to use Salsa20 instead of simply using TLS/Noise on top of WebTransport?

marten-seemann: TLS and Noise negotiate a key that's then used on a single byte stream. WebTransport is stream-multiplexed at the transport layer, so we need some way to encrypt every stream (while being careful not to reuse the same key).

However, with @Stebalien's proposal (https://github.com/libp2p/specs/pull/404#discussion_r846782316) this double encryption won't be necessary. I'll try to write it up in a more formal way later.

MarcoPolo: I guess my question is: What's the reason to not use Webtransport as if it were a single unencrypted byte stream and use TLS/Noise + Yamux on top of it?

Is it that that adds an extra layer of duplicated work and thus inefficient? Or is there something else that I'm missing?

Stebalien's proposal makes sense to me.

webtransport/README.md
```diff
@@ -0,0 +1,60 @@
...
+2. They use a self-signed certificate. According to the [w3c WebTransport certification](https://www.w3.org/TR/webtransport/), the validity of the certificate MUST be at most 14 days. Nodes then include the hash of one (or more) certificates in their multiaddr (see [#addressing]).
+
+When using self-signed certificates, nodes need to take care to regularly renew their certificate. In the following, the RECOMMENDED logic for rolling certificates is described. At first boot of the node, it creates a self-signed certificate with a validity of 14 days, and publishes a multiaddr containing the hash of that certificate. After 10 days, the node prepares the next certificate, setting the `NotBefore` date of that certificate to the expiration date (or shortly before that) of the first certificate, and an expiration of 14 days after that. The node continues using the old certificate until its expiry date, but it already advertises a multiaddr containing both certificate hashes. This way, clients will be able to connect to the node, even if they cache the multiaddr for multiple days.
```
Kubuxu, line 28: > The node continues using the old certificate until its expiry date

I would reword: s/using/accepts/

webtransport/README.md
Kubuxu: I think for announcement only one cert is needed. The node needs to be capable of accepting both of them.
So few days before cert expiry, the node start announcing cert2, but it still accepts connections coming with cert1.

In essence during that period the node has two valid multiaddresses, which I think is fine.

Kubuxu: Ahh, the server can only support one of them, and client has to specify both. Feel free to ignore what I said.

webtransport/README.md
Kubuxu: Let's set goals for this protocol, it needs to bind: the current channel and both PeerIDs.
TLS provides facilities for this by means of `ExportKeyingMaterial`. Result from it is unique per established channel, so even if someone were to try to MitM the session, `ExportKeyingMaterial` result will be different on either side.

I think, but I need to consider it more, we can just do:
Both sides do: `ekm = ExportKeyingMaterial("libp2p-auth")`, `ekm` is now shared secret on both sides, if they are using the same channel.
Then the Client sends the signature `Signature(ekm, ClientPrivKey)` and its pubkey. Server does the same, sends `Signature(ekm, ServerPrivKey)` and its pubkey.

Then to verify both sides verify that pubkeys match PeerIDs and they verify if the signature received is of `ekm`. The `ekm` is never transmitted over wire.

---

Questions is whether something like `ExportKeyingMaterial` is available on the browser end.

Kubuxu: PubKeys can also be exchanged out of band, IIRC libp2p has some other facilities for it.

Kubuxu: `ExportKeyingMaterial` is unavailable on the browser side.

Kubuxu: I think @Stebalien 's proposal can work. It is just there are a lot of details to work out that I don't feel comfortable deciding on (in essence we are rolling our own secure channel protocol, just without encryption).

An alternative to that proposal could be to reuse existing implementation of Noise secure channel to establish a secure session within WebTransport, for the purpose of co-authentication and authentication of the cert_hash. 
The proposal would be to start a libp2p-noise channel over WebCrypot and add `cert_hash` into the `Data` field of [HandshakePayload](https://github.com/libp2p/go-libp2p-noise/blob/95f0a223708e1bc01567dfbfe9ba7bab2723c781/pb/payload.pb.go#L25). An input of a predicate to assert onto the Data field would need to be added into libp2p-noise.

I think this is the lowest friction and risk solution.

marten-seemann:
I updated the spec, incorporating the ideas of @Stebalien and @Kubuxu for securing the connection and avoiding double-encryption:
1. I introduced a `/certhash` multiaddr component, which can be repeated to transmit multiple certificate hashes.
1. We now use Noise to perform the libp2p handshake, and transmit the list of certificate hashes the client was willing to accept as (unencrypted, but integrity-protected) payload of the first handshake message.
2. The server verifies that it possesses certificates matching ALL of the certificate hashes.
3. This also means that we won't be able to accept CA-signed certificates any more, as the Noise handshake described above is needed to tie the outer (WebTransport, TLS) connection to the Noise handshake. This simplifies the spec.

Kubuxu:
If we really wanted we could accept CA-signed certs but as self-signed certs are the primary use case, we might as well use just self-signed.

webtransport/README.md
```diff
@@ -0,0 +1,62 @@
...
+Certificate hashes are encoded using the following protobuf message:
+```proto
+syntax = "proto2";
+
+message WebTransport {
+  repeated bytes cert_hashes = 1;
+}
+```
+
+On receipt of the `e` message, the server MUST verify the list of certificate hashes. If the list is empty, it MUST fail the handshake. For every certificate in the list, it checks if it possesses a certificate with the corresponding hash. If so, it continues with the handshake. However, if there is even a single certificate hash in the list that it cannot associate with a certificate, it MUST abort the handshake.
```
Kubuxu, line 60: Thinking about it more, there is a small issue with it. 
It would be trivial for me to make any node impossible to connect to via WebTransport by, as 3rd party, announcing additional multiaddr with some random certhash.

I don't have good solution for this, especially if WebTransport browser APIs don't give us any feedback on which certhash was used.

marten-seemann: Not sure if I understand the attack you're suggesting. You're saying that an attacker can add a `/certhash/<attacker controlled hash>` to a WebTransport multiaddr, right? How is it different from an attacker modifying a TCP or QUIC address, changing the port number? That would prevent the victim from establishing a TCP / QUIC connection to that node as well.

I guess the answer to that is: the only multiaddr you can actually trust is an address transferred in a signed peer record. For all other addresses, the best you can do is try to connect and see what happens.

Kubuxu: I guess it is no different than specifying a bad port on the same IP address for a given peer id.

In face of this, we have to ensure that multiaddrs won't be merged on the client side.

SgtPooki:
> I introduced a /certhash multiaddr component, which can be repeated to transmit multiple certificate hashes.

Can we convert the '/certhash' pairs into an IPLD structure and do something like '/certhashs/<ipld_cid>' or something similar?

SgtPooki:
> Is there something I'm missing?

The CID shouldn't require lookup if it's an identity CID, other than that, probably not? ¯\_(ツ)_/¯

webtransport/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Security Handshake
+
+Unfortunately, the self-signed certificate doesn't allow the nodes to authenticate each others' peer IDs. It is therefore necessary to run an additional libp2p handshake on a newly established WebTransport connection.
+The first stream that the client opens on a new WebTransport session is used to perform a libp2p handshake using Noise (https://github.com/libp2p/specs/tree/master/noise). The client SHOULD start the handshake right after sending the CONNECT request, without waiting for the server's response.
+
+In order to verify that the end-to-end encryption of the connection, the peers need to establish that no MITM intercepted the connection. To do so, the client MUST include the certificate hash that was used to establish the connection as payload of the first Noise message (the `e` message). This payload is not encrypted, but the Noise handshake provides integrity protection.
+If the client was willing to accept multiple certificate hashes, but cannot determine with certificate was actually used to establish the connection (this will commonly be the case for browser clients), it MUST include a list of all certificate hashes.
```
mxinden, line 49:
```suggestion
If the client was willing to accept multiple certificate hashes, but cannot determine which certificate was actually used to establish the connection (this will commonly be the case for browser clients), it MUST include a list of all certificate hashes.
```

webtransport/README.md
```diff
@@ -0,0 +1,62 @@
...
+See the [lifecycle document][lifecycle-spec] for context about maturity level
+and spec status.
+
+## Introduction
+
+WebTransport is a way for browsers to establish a stream-multiplexed and bidirectional connection to servers using QUIC.
+
+The WebTransport protocol is currently under development at the IETF. Chrome has implemented and shipped support for [draft-02](https://datatracker.ietf.org/doc/draft-ietf-webtrans-http3/).
+
+The most exciting feature for libp2p (other than the improved performance and hole punching success rates that QUIC gives us) is that the W3C added a browser API allowing browsers to establish connections to nodes with self-signed certificates, provided they know the hash of the certificate in advance: `[serverCertificateHashes](https://www.w3.org/TR/webtransport/#dom-webtransportoptions-servercertificatehashes)`. This API is already [implemented in Chrome](https://chromestatus.com/feature/5690646332440576). Firefox is working on a WebTransport implementation and is [likely to implement](https://github.com/mozilla/standards-positions/issues/167#issuecomment-1015951396) `serverCertificateHashes` as well.
```
MarcoPolo, line 20:
```suggestion
The most exciting feature for libp2p (other than the improved performance and hole punching success rates that QUIC gives us) is that the W3C added a browser API allowing browsers to establish connections to nodes with self-signed certificates, provided they know the hash of the certificate in advance: [serverCertificateHashes](https://www.w3.org/TR/webtransport/#dom-webtransportoptions-servercertificatehashes). This API is already [implemented in Chrome](https://chromestatus.com/feature/5690646332440576). Firefox is working on a WebTransport implementation and is [likely to implement](https://github.com/mozilla/standards-positions/issues/167#issuecomment-1015951396) `serverCertificateHashes` as well.
```

Removes code ticks from the url (apologies if this was on purpose)

lidel: fwiw they will work when put inside:
```suggestion
The most exciting feature for libp2p (other than the improved performance and hole punching success rates that QUIC gives us) is that the W3C added a browser API allowing browsers to establish connections to nodes with self-signed certificates, provided they know the hash of the certificate in advance: [`serverCertificateHashes`](https://www.w3.org/TR/webtransport/#dom-webtransportoptions-servercertificatehashes). This API is already [implemented in Chrome](https://chromestatus.com/feature/5690646332440576). Firefox is working on a WebTransport implementation and is [likely to implement](https://github.com/mozilla/standards-positions/issues/167#issuecomment-1015951396) `serverCertificateHashes` as well.
```

webtransport/README.md
```diff
@@ -0,0 +1,62 @@
...
+* `/ip4/1.2.3.4/udp/443/quic/webtransport/certhash/<hash1>`
+* `/ip6/fe80::1ff:fe23:4567:890a/udp/1234/quic/webtransport/certhash/<hash1>/certhash/<hash2>/certhash/<hash3>`
+
+## WebTransport HTTP endpoint
+
+WebTransport needs a HTTPS URL to establish a WebTransport session, e.g. `https://example.com/webtransport`. As multiaddrs don't allow the encoding of URLs, this spec standardizes the endpoint. The HTTP endpoint of a libp2p WebTransport servers MUST be located at `/.well-known/libp2p-webtransport`.
+
+To allow future evolution of the way we run the libp2p handshake over WebTransport, we use a URL parameter. The handshake described in this document MUST be signaled by setting the `type` URL parameter to `noise`.
+
+Example: The WebTransport URL of a WebTransport server advertising `/ip4/1.2.3.4/udp/1443/quic/webtransport/` would be `https://1.2.3.4:1443/.well-known/libp2p-webtransport?type=noise`.
```
lidel, line 41: Is  `https://` a hard requirement, or would connection to any [secure context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (namely the no TLS `http://localhost`) work?

marten-seemann: According to the draft, it needs to be `https`: https://ietf-wg-webtrans.github.io/draft-ietf-webtrans-http3/draft-ietf-webtrans-http3.html#section-3.3

lidel: Yes, that is my hope.  Worst case, if we hit any issues here, we will work with Igalia to clarify  browser specs and behavior around this.

webtransport/README.md
```diff
@@ -0,0 +1,62 @@
...
+syntax = "proto2";
+
+message WebTransport {
+  repeated bytes cert_hashes = 1;
+}
+```
+
+On receipt of the `e` message, the server MUST verify the list of certificate hashes. If the list is empty, it MUST fail the handshake. For every certificate in the list, it checks if it possesses a certificate with the corresponding hash. If so, it continues with the handshake. However, if there is even a single certificate hash in the list that it cannot associate with a certificate, it MUST abort the handshake.
+
+For the client, the libp2p connection is fully established once it has sent the last Noise handshake message. For the server, receipt (and successful verification) of that message completes the handshake.
```
mxinden, line 62: > For the client, the libp2p connection is fully established once it has sent the last Noise handshake message.

Are the following assumptions correct?

- The Noise XX handshake has 3 messages (client -> server, server -> client, client -> server).
- The 3rd message contains the server's certificate hash send from the client to the server?
- The client depends on the server terminating the connection in case the certificate hash send by the client does not match one of its advertised certificate hashes, thus implying a MITM?

If the above is correct, would this handshake not be prone to a race condition, potentially making it insecure?

Say there is a MITM on a WebTransport connection between a client and a server.

1. The client sends the 3rd message of the Noise handshake containing the MITM certificate hash, not the correct server certificate hash.
2. The client assumes the connection to be secure, opens a second WebTransport channel and sends user-data on that channel.
3. The server notices the certificate hash mismatch and aborts the handshake.
4. The client notices that the server aborted the handshake.

Given that (2) happens before (4) the client sends user-data on a channel that is MITM.

marten-seemann: > The 3rd message contains the server's certificate hash send from the client to the server.

This assumption is not correct. The certificate hash is sent on the first message. This message is neither encrypted nor authenticated (at that point), but it is part of the rolling hash, which means it's authenticated retroactively: the handshake will fail if an attacker modifies it.

At least that's my understanding of the Noise handshake, unfortunately the spec is not really clear about this. Here's the Go code: https://github.com/flynn/noise/blob/d803f5c4b0f4c9bb4d8234603b9cf6c194f6b135/state.go#L443-L446

Kubuxu:
Interesting idea of using Noise Prologue instead of handshake payload data: https://github.com/libp2p/specs/pull/412/files#r931163931 if we don't need to actually send this data.

marten-seemann:
Interesting! Any suggestion what exactly to put in the prologue? Unfortunately, when using multiple certhashes (as we will do), the browser won't know which of the certificates was accepted.

Menduist:
To use this technique, you need to have a value known by the server & the client, where each of them has an impact on the value, before the Noise handshake begins.

So in webrtc case, concatenating both certificates does the trick, in webtransport case, I don't know if there is something similar applicable

Kubuxu:
> Interesting! Any suggestion what exactly to put in the prologue? Unfortunately, when using multiple certhashes (as we will do), the browser won't know which of the certificates was accepted.

You are right, I forgot about this caveat. In this situation, I don't think there is a nice way to use the Noise Prologue.

marten-seemann:
If https://github.com/w3c/webtransport/issues/411 is implemented, we could use that key exporter to generate a unique value tied to the TLS session, which could be used as the Noise prologue.

webtransport/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+Example: The WebTransport URL of a WebTransport server advertising `/ip4/1.2.3.4/udp/1443/quic/webtransport/` would be `https://1.2.3.4:1443/.well-known/libp2p-webtransport?type=noise`.
+
+## Security Handshake
+
+Unfortunately, the self-signed certificate doesn't allow the nodes to authenticate each others' peer IDs. It is therefore necessary to run an additional libp2p handshake on a newly established WebTransport connection.
+The first stream that the client opens on a new WebTransport session is used to perform a libp2p handshake using Noise (https://github.com/libp2p/specs/tree/master/noise). The client SHOULD start the handshake right after sending the CONNECT request, without waiting for the server's response.
+
+In order to verify that the end-to-end encryption of the connection, the peers need to establish that no MITM intercepted the connection. To do so, the client MUST include the certificate hash that was used to establish the connection as payload of the first Noise message (the `e` message). This payload is not encrypted, but the Noise handshake provides integrity protection.
```
marten-seemann, line 48: Alternatively, we could have the server send the hashes of all its certificates in the second handshake message. This might mean that we'd transfer one more certificate hash in some cases, but the overhead should be negligible.
The two advantages are:
1. This message would be encrypted.
2. It would shift the responsibility of verifying the certificate hashes onto the client.

@MarcoPolo @julian88110 @mxinden Thoughts?

MarcoPolo: This handshake is encrypted by the peer ids. The client should see the superset of the certhashes it used to start the connection sent by the expected peer id. I think this is the same as before, except of course the client gets to be sure the check was done. So I think this is a net win.

mxinden: > This message would be encrypted.

:+1: for choosing this new approach then.

> It would shift the responsibility of verifying the certificate hashes onto the client.

I can not think of a reason for either being beneficial over the other. I.e. for server or client verification to be better. Thus this does not effect my agreement with your proposal. Is there any reason?

> I think this is the same as before, except of course the client gets to be sure the check was done.

I read this as there being a reason. @MarcoPolo would you mind expanding if so?

MarcoPolo: > I read this as there being a reason. @MarcoPolo would you mind expanding if so?

Assume server keys aren't leaked (then all bets are off).

If the server is responsible for the verification then the client has to trust that the server (or the server's implementation) does the correct check. If the client does the verification, it doesn't have to trust that the server did this check it can do it itself. @marten-seemann mentioned this in his comment: https://github.com/libp2p/specs/pull/404#discussion_r846792263. The flip side of course, is that the server now trusts the client to do the check to avoid mitm, but this is how it works everywhere. In web2, servers can't force the browser to actually check the signature, but they do.

The point of the certificate hash is so that the client can validate that the server it's talking to is indeed the server it expects. The key thing here is that the _client_ cares about this. The client has no reason to believe that the server cares about this. In classic webtransport this is validated on the client side as well (this should be reason enough to do the same thing here), granted this needs to happen because clients don't know the server's publickey/peerid. 

Which brings up another thing you enable with client-side checking. You could say you want to dial any peer id that passes a given certhash check. So if you got a certhash over a trusted channel, you could use it to authenticate the connection without knowing the peer id (this is the classic webtransport case).

marten-seemann: Text change suggested in #455.

webtransport/README.md
```diff
@@ -0,0 +1,52 @@
+# libp2p WebTransport
+
+| Lifecycle Stage | Maturity                 | Status | Latest Revision |
+|-----------------|--------------------------|--------|-----------------|
+| 1A              | DRAFT                    | Active | r0, 2022-04-03  |
```
mxinden, line 5: Let's put one of the valid maturity levels here:

> Valid values are: Working Draft, Candidate Recommendation, Recommendation.

See https://github.com/libp2p/specs/blob/master/00-framework-02-document-header.md#status-table.

I would recommend:

```suggestion
| 1A              | Candidate Recommendation                    | Active | r0, 2022-04-03  |
```

webtransport/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+Authors: [@marten-seemann]
+
+Interest Group: TODO
```
mxinden, line 9:
```suggestion
Interest Group: [@mxinden]
```

Happy to be listed here.

webtransport/README.md
```diff
@@ -0,0 +1,52 @@
...
+See the [lifecycle document](../00-framework-01-spec-lifecycle.md) for context about maturity level
+and spec status.
+
+## Introduction
+
+WebTransport is a way for browsers to establish a stream-multiplexed and bidirectional connection to servers using QUIC.
+
+The WebTransport protocol is currently under development at the IETF. Chrome has implemented and shipped support for [draft-02](https://datatracker.ietf.org/doc/draft-ietf-webtrans-http3/).
+
+The most exciting feature for libp2p (other than the improved performance and hole punching success rates that QUIC gives us) is that the W3C added a browser API allowing browsers to establish connections to nodes with self-signed certificates, provided they know the hash of the certificate in advance: [`serverCertificateHashes`](https://www.w3.org/TR/webtransport/#dom-webtransportoptions-servercertificatehashes). This API is already [implemented in Chrome](https://chromestatus.com/feature/5690646332440576). Firefox is working on a WebTransport implementation and is [likely to implement](https://github.com/mozilla/standards-positions/issues/167#issuecomment-1015951396) `serverCertificateHashes` as well.
```
mxinden, line 20: > and hole punching success rates

Will we ever be able to leverage WebTransport for hole punching? Would the browsers allow us to reuse the same UDP port across connections?

webtransport/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+Example: The WebTransport URL of a WebTransport server advertising `/ip4/1.2.3.4/udp/1443/quic/webtransport/` would be `https://1.2.3.4:1443/.well-known/libp2p-webtransport?type=noise`.
+
+## Security Handshake
+
+Unfortunately, the self-signed certificate doesn't allow the nodes to authenticate each others' peer IDs. It is therefore necessary to run an additional libp2p handshake on a newly established WebTransport connection.
+The first stream that the client opens on a new WebTransport session is used to perform a libp2p handshake using Noise (https://github.com/libp2p/specs/tree/master/noise). The client SHOULD start the handshake right after sending the CONNECT request, without waiting for the server's response.
+
+In order to verify that the end-to-end encryption of the connection, the peers need to establish that no MITM intercepted the connection. To do so, the server MUST include the certificate hash of the currently used certificate as well as the certificate hashes of all future certificates it has already advertised to the network in the `webtransport_certhashes` Noise extension. The hash of recently used, but expired certificates SHOULD also be included.
```
mxinden, line 48:
```suggestion
In order to verify end-to-end encryption of the connection, the peers need to establish that no MITM intercepted the connection. To do so, the server MUST include the certificate hash of the currently used certificate as well as the certificate hashes of all future certificates it has already advertised to the network in the `webtransport_certhashes` Noise extension. The hash of recently used, but expired certificates SHOULD also be included.
```

mxinden, line 48: How about linking to the libp2p Noise specification here?

mxinden (commented):
Couple of comments. Mostly formalities.

webtransport/README.md
marten-seemann: I was just working on that, now that #453 was merged :)
Will push the change later.

webtransport/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+On receipt of the `webtransport_certhashes` extension, the client MUST verify that the certificate hash of the certificate that was used on the connection is contained in the server's list. If the client was willing to accept multiple certificate hashes, but cannot determine which certificate was actually used to establish the connection (this will commonly be the case for browser clients), it MUST verify that all certificate hashes are contained in the server's list. If verification fails, it MUST abort the handshake.
+
+For the client, the libp2p connection is fully established once it has sent the last Noise handshake message. For the server, receipt (and successful verification) of that message completes the handshake.
```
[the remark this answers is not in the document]
marten-seemann: I don't think so. The server still needs to process the last Noise handshake message to learn the client's libp2p identity.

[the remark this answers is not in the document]
mxinden: > and successful verification

I thought this refers to the verification of the certificate hashes, not the verification of the Noise handshake message of the client. I would suggest removing "and successful verification", though I am fine either way.

webtransport/README.md
marten-seemann: In principle it's possible to demultiplex QUIC connections by their Connection ID. In fact, that's what go-libp2p does. Currently, browsers use a new UDP socket per connection, but that might change in the future (maybe? hopefully?).

We should probably remove this sentence though, since it's not applicable today yet.

webtransport/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+Since most libp2p nodes don't possess a TLS certificate signed by a Certificate Authority, servers use a self-signed certificates. According to the [w3c WebTransport certification](https://www.w3.org/TR/webtransport/), the validity of the certificate MUST be at most 14 days, and must not use an RSA key. Nodes then include the hash of one (or more) certificates in their multiaddr (see [Addressing](#addressing)).
+
+Servers need to take care to regularly renew their certificate. In the following, the RECOMMENDED logic for rolling certificates is described. At first boot of the node, it creates a self-signed certificate with a validity of 14 days, and publishes a multiaddr containing the hash of that certificate. After 10 days, the node prepares the next certificate, setting the `NotBefore` date of that certificate to the expiration date (or shortly before that) of the first certificate, and an expiration of 14 days after that. The node continues using the old certificate until the expiry date, but it already advertises a multiaddr containing both certificate hashes. This way, clients will be able to connect to the node, even if they cache the multiaddr for multiple days.
```
elenaf9, line 26: > After 10 days, the node prepares the next certificate

Why only after 10 days, instead of generating and advertising the next certificate hash directly (= always advertise `.../<current-certificate-hash><next-certificate-hash>`)? Benefit would be that a new multiaddress would have to be announced less frequently and clients could cache addresses longer (not sure if they actually do that in practice).

marten-seemann: The motivation was that generating a new certificate costs some CPU, but then I realized this cost is negligible. I just forgot to update this document.

go-libp2p already implements exactly what you propose :)

webtransport/README.md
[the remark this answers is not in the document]
marten-seemann: Replaced by "processing".

webtransport/README.md
marten-seemann: I updated the text. PTAL.

webtransport/README.md
```diff
@@ -45,18 +45,8 @@ Example: The WebTransport URL of a WebTransport server advertising `/ip4/1.2.3.4
...
 
-In order to verify that the end-to-end encryption of the connection, the peers need to establish that no MITM intercepted the connection. To do so, the client MUST include the certificate hash that was used to establish the connection as payload of the first Noise message (the `e` message). This payload is not encrypted, but the Noise handshake provides integrity protection.
-If the client was willing to accept multiple certificate hashes, but cannot determine which certificate was actually used to establish the connection (this will commonly be the case for browser clients), it MUST include a list of all certificate hashes.
+In order to verify that the end-to-end encryption of the connection, the peers need to establish that no MITM intercepted the connection. To do so, the server MUST include the certificate hash of the currently used certificate as well as the certificate hashes of all future certificates it has already advertised to the network in the `webtransport_certhashes` Noise extension. The hash of recently used, but expired certificates SHOULD also be included.
```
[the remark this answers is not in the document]
marten-seemann: I added a link to the spec. I think it's preferable if readers go there to learn about the details of that mechanism.

webtransport/README.md
marten-seemann: Done, while also updating the date.

webtransport/README.md
marten-seemann: Done.

webtransport/README.md
marten-seemann: Fixed.

marten-seemann: I included a link to the Noise spec.

webtransport/README.md
```diff
@@ -0,0 +1,52 @@
...
+## Addressing
+
+Webtransport multiaddresses are composed of a QUIC multiaddr, followed by `/webtransport` and a list of multihashes of the certificates that the server uses.
+Examples:
+* `/ip4/1.2.3.4/udp/443/quic/webtransport/certhash/<hash1>`
+* `/ip6/fe80::1ff:fe23:4567:890a/udp/1234/quic/webtransport/certhash/<hash1>/certhash/<hash2>/certhash/<hash3>`
+
+## WebTransport HTTP endpoint
+
+WebTransport needs a HTTPS URL to establish a WebTransport session, e.g. `https://example.com/webtransport`. As multiaddrs don't allow the encoding of URLs, this spec standardizes the endpoint. The HTTP endpoint of a libp2p WebTransport servers MUST be located at `/.well-known/libp2p-webtransport`.
```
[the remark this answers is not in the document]
marten-seemann: Fixed.

webtransport/README.md
marten-seemann: I also added Marco, given that he's writing the JS implementation.

webtransport/README.md
marten-seemann: That would be super nice. I'd love to allow that, but I'd like to a generally agreed-upon format, and currently we don't have that.

webtransport/README.md
```diff
@@ -0,0 +1,60 @@
...
+See the [lifecycle document][lifecycle-spec] for context about maturity level
+and spec status.
+
+## Introduction
+
+WebTransport is a way for browsers to establish a stream-multiplexed connection to servers that allows bidirectional streaming.
+
+The WebTransport protocol is currently under development at the IETF. Chrome has implemented and shipped support for [draft-02](https://datatracker.ietf.org/doc/draft-ietf-webtrans-http3/).
+
+The most exciting feature for libp2p (other than the improved performance and hole punching success rates that QUIC gives us) is that the W3C added a browser API allowing browsers to establish connections to nodes with self-signed certificates, provided they know the hash of the certificate in advance: `[serverCertificateHashes](https://www.w3.org/TR/webtransport/#dom-webtransportoptions-servercertificatehashes)`. This API is already [implemented in Chrome](https://chromestatus.com/feature/5690646332440576). Firefox is working on a WebTransport implementation and is [likely to implement](https://github.com/mozilla/standards-positions/issues/167#issuecomment-1015951396) `serverCertificateHashes` as well.
```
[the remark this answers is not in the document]
marten-seemann: Thank you! I added the Bugzilla link.

webtransport/README.md
marten-seemann: Resolving, since we're using `noise` now.

marten-seemann:
Thank you everyone for the great discussions on this PR!

We've shipped (experimental) WebTransport support based on this specification in [go-libp2p v0.23](https://github.com/libp2p/go-libp2p/releases/tag/v0.23.0).

I've resolved all remaining nits in the specification. Would be grateful for another round of review and many approvals :)

MarcoPolo (approved):

webtransport/README.md
```diff
@@ -0,0 +1,53 @@
...
+Interest Group: [@MarcoPolo], [@mxinden]
+
+See the [lifecycle document](../00-framework-01-spec-lifecycle.md) for context about maturity level
+and spec status.
+
+## Introduction
+
+WebTransport is a way for browsers to establish a stream-multiplexed and bidirectional connection to servers using QUIC.
+
+The WebTransport protocol is currently under development at the IETF. Chrome has implemented and shipped support for [draft-02](https://datatracker.ietf.org/doc/draft-ietf-webtrans-http3/), and Firefox [is working](https://bugzilla.mozilla.org/show_bug.cgi?id=1709355) on WebTransport support.
```
elenaf9, line 18: If I understand it correctly, when we talk about **the** WebTransport  protocol in this doc, we actually mean the _WebTransport over HTTP-3_ protocol as described in [draft-ietf-webtrans-http3](https://datatracker.ietf.org/doc/draft-ietf-webtrans-http3/). 
WebTransport as specified in [draft-ietf-webtrans-overview](https://datatracker.ietf.org/doc/draft-ietf-webtrans-overview/) is a protocol framework and not a single protocol. I think we should add a note/ sentence on that. At least for me it was a bit confusing.

webtransport/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+Servers need to take care to regularly renew their certificate. In the following, the RECOMMENDED logic for rolling certificates is described. At first boot of the node, it creates one self-signed certificate with a validity of 14 days, starting immediately, and another certificate with the 14 day valididity period starting on the expiry date of the first certificate. The node advertises a multiaddr containing the certificate hashes of these two certificates.
+Once the first certificate has expired, the node prepares the next certificate, and updates the multiaddr it advertises.
+
+## Addressing
+
+Webtransport multiaddresses are composed of a QUIC multiaddr, followed by `/webtransport` and a list of multihashes of the certificates that the server uses.
+Examples:
+* `/ip4/1.2.3.4/udp/443/quic/webtransport/certhash/<hash1>`
+* `/ip6/fe80::1ff:fe23:4567:890a/udp/1234/quic/webtransport/certhash/<hash1>/certhash/<hash2>/certhash/<hash3>`
```
elenaf9, line 34: Based on this address I assume `certhash` is implying a fixed hash function? I think we should add info on this to the spec.

webtransport/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## WebTransport HTTP endpoint
+
+WebTransport needs a HTTPS URL to establish a WebTransport session, e.g. `https://example.com/webtransport`. As multiaddrs don't allow the encoding of URLs, this spec standardizes the endpoint. The HTTP endpoint of a libp2p WebTransport server MUST be located at `/.well-known/libp2p-webtransport`.
```
elenaf9, line 38: >  As multiaddresses don't allow the encoding of URLs

This sounds very absolute, which is not necessarily true. Multiaddresses just do not allow the encoding of URLs **yet**. If that ever changes and we forget to update this doc, it might be confusing. How about writing something like "At the point of writing multiaddresses don't allow the encoding of URLs, therefore this spec standardizes the endpoint".

webtransport/README.md
elenaf9, line 34: Why do we call it `certhash`, instead of just adding a more general `hash`? The latter could then also used in the future by other protocols that need to include a `hash` in the multiaddress for whatever reason.

webtransport/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+The most exciting feature for libp2p (other than the numberous performance benefits that QUIC gives us) is that the W3C added a browser API allowing browsers to establish connections to nodes with self-signed certificates, provided they know the hash of the certificate in advance: [`serverCertificateHashes`](https://www.w3.org/TR/webtransport/#dom-webtransportoptions-servercertificatehashes). This API is already [implemented in Chrome](https://chromestatus.com/feature/5690646332440576). Firefox is working on a WebTransport implementation and is [likely to implement](https://github.com/mozilla/standards-positions/issues/167#issuecomment-1015951396) `serverCertificateHashes` as well.
+
+## Certificates
+
+Since most libp2p nodes don't possess a TLS certificate signed by a Certificate Authority, servers use a self-signed certificates. According to the [w3c WebTransport certification](https://www.w3.org/TR/webtransport/), the validity of the certificate MUST be at most 14 days, and must not use an RSA key. Nodes then include the hash of one (or more) certificates in their multiaddr (see [Addressing](#addressing)).
```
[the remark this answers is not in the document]
MarcoPolo: MUST, it's part of the webtransport spec: https://www.w3.org/TR/webtransport/#web-transport-configuration

webtransport/README.md
MarcoPolo: There's some discussion here: https://github.com/multiformats/multiaddr/pull/130. But I think the important part is that we can be a bit more specific here to highlight that this is a certificate hash rather than a hash for some other (?) thing. We also don't have a use case for a generic hash, but do have this use case for the certificate hash.

webtransport/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+In order to verify end-to-end encryption of the connection, the peers need to establish that no MITM intercepted the connection. To do so, the server MUST include the certificate hash of the currently used certificate as well as the certificate hashes of all future certificates it has already advertised to the network in the `webtransport_certhashes` Noise extension (see Noise Extension section of the [Noise spec](/noise/README.md)). The hash of recently used, but expired certificates SHOULD also be included.
+
+On receipt of the `webtransport_certhashes` extension, the client MUST verify that the certificate hash of the certificate that was used on the connection is contained in the server's list. If the client was willing to accept multiple certificate hashes, but cannot determine which certificate was actually used to establish the connection (this will commonly be the case for browser clients), it MUST verify that all certificate hashes are contained in the server's list. If verification fails, it MUST abort the handshake.
```
julian88110, line 51: Is it necessary to specify how the peer should detect and retire expired certificate? or is it true that the webtransport connections are shortlived so that the expiration is usually not an issue?

webtransport/README.md
julian88110: Vote for more specific field name too. it consumes less encoding space and allows easier extension.

webtransport/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+To allow future evolution of the way we run the libp2p handshake over WebTransport, we use a URL parameter. The handshake described in this document MUST be signaled by setting the `type` URL parameter to `noise`.
+
+Example: The WebTransport URL of a WebTransport server advertising `/ip4/1.2.3.4/udp/1443/quic/webtransport/` would be `https://1.2.3.4:1443/.well-known/libp2p-webtransport?type=noise`.
+
+## Security Handshake
+
+Unfortunately, the self-signed certificate doesn't allow the nodes to authenticate each others' peer IDs. It is therefore necessary to run an additional libp2p handshake on a newly established WebTransport connection.
+The first stream that the client opens on a new WebTransport session is used to perform a libp2p handshake using Noise (https://github.com/libp2p/specs/tree/master/noise). The client SHOULD start the handshake right after sending the CONNECT request, without waiting for the server's response.
```
julian88110, line 47: This is just a curious question regarding handshake starting right after Connect. Is connect referring to establishing a socket level connection?

julian88110 (commented):
Good write-up. Some comments and questions in line.

webtransport/README.md
marten-seemann: This is WebTransport CONNECT request as described in https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3-03#section-3.2.

webtransport/README.md
marten-seemann: Excellent point. I applied your suggestion.

webtransport/README.md
marten-seemann: Yeah. Not the best naming from the IETF's side here...
Everybody thinks of WebTransport over HTTP/3 when WebTransport, but then there's the HTTP/2 fallback that not a lot of people care about to begin with...

I've added some text (and links) to clarify things. Let me know if that's less confusing now.

webtransport/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Certificates
+
+Since most libp2p nodes don't possess a TLS certificate signed by a Certificate Authority, servers use a self-signed certificates. According to the [w3c WebTransport certification](https://www.w3.org/TR/webtransport/), the validity of the certificate MUST be at most 14 days, and must not use an RSA key. Nodes then include the hash of one (or more) certificates in their multiaddr (see [Addressing](#addressing)).
+
+Servers need to take care to regularly renew their certificate. In the following, the RECOMMENDED logic for rolling certificates is described. At first boot of the node, it creates one self-signed certificate with a validity of 14 days, starting immediately, and another certificate with the 14 day valididity period starting on the expiry date of the first certificate. The node advertises a multiaddr containing the certificate hashes of these two certificates.
+Once the first certificate has expired, the node prepares the next certificate, and updates the multiaddr it advertises.
```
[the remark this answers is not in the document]
marten-seemann: Applied. Thank you for the text suggestion!

webtransport/README.md
marten-seemann: I think that being specific pays off here, by making us more future-proof.
Imagine we want to add another hash-like component to a multiaddr in the future. We'd now use `/hash` for the certificate hash and for that new thing. How would we know which of these is a certificate hash and which of them the new thing?

webtransport/README.md
marten-seemann: It's "MUST NOT", not "MUST not" ;)

This entire section needs a rewrite though. It _is_ possible to use a CA-signed certificate.

marten-seemann: I rewrote this section. Please take another look.

marten-seemann:
Thank you for this in-depth review, @elenaf9!

I rewrote the "Certificates" section to allow the use of CA-signed certificates.

webtransport/README.md
```diff
@@ -0,0 +1,63 @@
...
+Authors: [@marten-seemann]
+
+Interest Group: [@MarcoPolo], [@mxinden]
+
+See the [lifecycle document](../00-framework-01-spec-lifecycle.md) for context about maturity level
+and spec status.
+
+## Introduction
+
+[WebTransport](https://datatracker.ietf.org/doc/draft-ietf-webtrans-overview/) is a way for browsers to establish a stream-multiplexed and bidirectional connection to servers. The WebTransport protocol is currently under development at the IETF. The primary way to do by running on top of a HTTP/3 connection [WebTransport over HTTP/3](https://datatracker.ietf.org/doc/draft-ietf-webtrans-http3/). For situations where it is not possible to establish a HTTP/3 connection, there's a HTTP/2 fallback ([WebTransport using HTTP/2](https://datatracker.ietf.org/doc/draft-ietf-webtrans-http2/)).
```
elenaf9, line 16:
```suggestion
[WebTransport](https://datatracker.ietf.org/doc/draft-ietf-webtrans-overview/) is a way for browsers to establish a stream-multiplexed and bidirectional connection to servers. The WebTransport protocol is currently under development at the IETF. The primary way to do is by running on top of a HTTP/3 connection [WebTransport over HTTP/3](https://datatracker.ietf.org/doc/draft-ietf-webtrans-http3/). For situations where it is not possible to establish a HTTP/3 connection, there's a HTTP/2 fallback ([WebTransport using HTTP/2](https://datatracker.ietf.org/doc/draft-ietf-webtrans-http2/)).
```

webtransport/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+In this document, we mean WebTransport over HTTP/3 when using the term WebTransport.
+
+Chrome has implemented and shipped support for [draft-02](https://datatracker.ietf.org/doc/draft-ietf-webtrans-http3/02/), and Firefox [is working](https://bugzilla.mozilla.org/show_bug.cgi?id=1709355) on WebTransport support.
+
+The most exciting feature for libp2p (other than the numerous performance benefits that QUIC gives us) is that the W3C added a browser API allowing browsers to establish connections to nodes with self-signed certificates, provided they know the hash of the certificate in advance: [`serverCertificateHashes`](https://www.w3.org/TR/webtransport/#dom-webtransportoptions-servercertificatehashes). This API is already [implemented in Chrome](https://chromestatus.com/feature/5690646332440576). Firefox is working on a WebTransport implementation and is [likely to implement](https://github.com/mozilla/standards-positions/issues/167#issuecomment-1015951396) `serverCertificateHashes` as well.
```
elenaf9, line 22:
```suggestion
The most exciting feature for libp2p (other than the numerous performance benefits that QUIC gives us) is that the W3C added a browser API allowing browsers to establish connections to nodes with self-signed certificates, provided they know the hash of the certificate in advance: [`serverCertificateHashes`](https://www.w3.org/TR/webtransport/#dom-webtransportoptions-servercertificatehashes). This API is already [implemented in Chrome](https://chromestatus.com/feature/5690646332440576). Firefox is [likely to implement](https://github.com/mozilla/standards-positions/issues/167#issuecomment-1015951396) `serverCertificateHashes` as well.
```
The paragraph above already mentions that firefox is working on WebTransport. Is the repeat here on intentional?

webtransport/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Certificates
+
+According to the [w3c WebTransport specification](https://www.w3.org/TR/webtransport/), there are two ways for browser to validate the certificate used on a WebTransport connection.
```
elenaf9, line 26:
```suggestion
According to the [w3c WebTransport specification](https://www.w3.org/TR/webtransport/), there are two ways for a browser to validate the certificate used on a WebTransport connection. 
```

webtransport/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Certificates
+
+According to the [w3c WebTransport specification](https://www.w3.org/TR/webtransport/), there are two ways for browser to validate the certificate used on a WebTransport connection. 
+1. by verifying the chain of trust of the certificate. This means that the certificate has to be signed by a CA (Certificate Authority) that the browser trusts. This is how browsers verify certificates when establishing a regular HTTPS connection.
+2. by verifying that the cryptographic hash of the certificate matches a specific value, using the [`serverCertificateHashes`](https://www.w3.org/TR/webtransport/#dom-webtransportoptions-servercertificatehashes) option.
+
+libp2p nodes that possess a CA-signed TLS certificate MAY use that certificate on WebTransport connections.
```
elenaf9, line 30: Would they still announce the hash of that certificate as part of their multiaddress?

elenaf9 (approved):

webtransport/README.md
marten-seemann: Thanks, I rephrased the sentence.

webtransport/README.md
marten-seemann: Only the part saying that Firefox is also implementing `serverCertificateHashes`. I rephrased the sentence.

webtransport/README.md
marten-seemann: Fixed.

marten-seemann:
Thank you to everyone involved here! This was a big effort, and we now have a decent spec, implemented and released by go-libp2p, with another implementation in js-libp2p demonstrating interoperability.

Given that this PR has received 2 approvals from members of the libp2p team at PL, I'm going to merge it now.
This does NOT mean that this spec is perfect - both editorial as well as protocol changes are still possible via PRs to this repo.

Issue #328, opened 2021-05-24
Stream Migration Protocol


Stebalien:
Libp2p should support protocol agnostic stream migration. This will make upgrading to better transports "seamless".

## Requirement

1. Transport agnostic. Really, this means migrating at the stream level.
2. Minimal overhead. Overhead should be at most a small per-stream cost (no additional framing, etc.)
3. No interruption. Reading/writing should be continuous.
4. Transparent. Applications using migratable streams shouldn't notice anything.
5. Correct. There can't be any ambiguity (one side believing the migration happened, the other side disagreeing, etc.).

## Protocol

When opening a stream, if the target peer supports the stream migration protocol (discovered through identify) and the stream is "long lived" (opt in or opt out?):

1. First, open a stream (stream A).
2. Negotiate the stream migration protocol.
2. Send a message indicating that we're opening a new stream, along with a unique ID for the stream.
3. Negotiate the actual protocol.

Because we "know" that the peer supports this stream migration protocol, we can pipeline these steps so as not to take any additional round-trips.

To migrate a stream:

1. The initiator will open a new stream (stream B), on a new connection, to the receiver. This is the target stream for the migration.
2. On stream B, the initiator will negotiate the stream migration protocol.
3. On stream B, the initiator will send a message indicating that we're migrating a stream, along with the ID of the stream we're migrating
4. The receiver will acknowledge the migration (on stream B) and, from this point onward, treat any "EOF" (close) on stream A as a migration to stream B.
5. When the initiator receives the acknowledgement on stream B, it will close stream A for writing, and start writing on stream B.
6. When the receiver receives the EOF on stream A, it will send an EOF on stream B, switching over to stream A.
7. When the initiator receives an EOF on stream A, it will start reading on stream B.

At this point, the stream is fully migrated. 

**Resets**

If either stream is "reset" before both ends are closed, both streams must be reset and the stream as a whole should be considered "aborted" (reset).

**Half-Closed**

If stream A was half-closed (either for reading or writing), that state must be replicated on the new stream after the initial handshake. Importantly, there's an edge-case:

1. The receiver tries to stream A for writing.
2. The receiver receives a migration request on stream B, for stream A.
3. The receiver ACKs the migration request.
4. The initiator sees the ACK on stream B.
5. The initiator sees the EOF on stream A, and treats it as the migration EOF.

This is fine. The stream will be migrated and the EOF will be re-played on stream B, leaving stream B in the intended state.

## Analysis

[...]

PR #406, opened 2022-04-13
Add a Stream Migration spec

MarcoPolo:
Initial pass at creating a spec for https://github.com/libp2p/specs/issues/328. I'll start a PoC as well to start fleshing out some of the ideas and check my understanding.


connections/stream-migration.md
```diff
@@ -0,0 +1,217 @@
...
+note over Initiator, Responder: Assume at least 2 connections.
+
+Initiator <-> Responder: <b>Connection 1; stream A</b>
+Initiator <-> Responder: <b>Connection 2</b>
+
+== Stream Migration ==
+
+note over Initiator, Responder: Migrate <b>Stream A</b> to <b>Stream B</b>
+
+Initiator -> Responder: <b>Stream A:</b> Start stream migration, this stream is <b>id=A</b>
```
marten-seemann, line 55: This should probably be the other way around. The stream is named right after it is opened, and at some point later the migration is initiated.

connections/stream-migration.md
```diff
@@ -0,0 +1,217 @@
...
+Note: some of these steps may be pipelined.
+
+### Resets
+
+If either stream is "reset" before both ends are closed, both streams must be
+reset and the stream as a whole should be considered "aborted" (reset).
+
+### Half closed streams
+
+The final migrated stream should loook the same as the initial stream. If the
```
marten-seemann, line 112:
```suggestion
The final migrated stream should look the same as the initial stream. If the
```

connections/stream-migration.md
```diff
@@ -0,0 +1,217 @@
...
+
+<details>
+  <summary>Instructions to reproduce diagram</summary>
+``` plantuml
+@startuml stream-migration-half-closed
+skinparam sequenceMessageAlign center
+entity Initiator
+entity Responder
+
+note over Responder: <b>Stream A</b> is closed for writing
```
marten-seemann, line 130: This would probably be easier to understand this was the first arrow in diagram: <——EOF

connections/stream-migration.md
```diff
@@ -0,0 +1,217 @@
...
+
+note over Initiator, Responder: Stream A is now migrated to Stream B
+@enduml
+```
+To generate:
+```bash
+plantuml stream-migration.md -o stream-migration -tsvg
+```
+</details>
+
```
marten-seemann, line 173: What happens if the peer misbehaves, e.g. when B just doesn’t send an EOF on stream B? While A can “consider” the stream as closed, it still needs to be closed explicitly, otherwise the stream multiplexer can’t garbage-collect the stream after it has been EOFed from both sides.

connections/stream-migration.md
```diff
@@ -0,0 +1,217 @@
...
+- <https://github.com/libp2p/go-libp2p/issues/634>
+
+## Open Questions
+
+Some questions that will probably be resolved when a PoC is implemented.
+
+- How many streams are there total? 2 or 3? There is the old stream and new
+  stream, but is there a migration protocol specific stream? Or can the
+  migration be started from the old stream? If yes, how do we prevent
+  application data from looking like a migration without framing?
```
marten-seemann, line 212: I interpret your first diagram as saying that the migration is initiated on stream B by saying “migrate stream A to this stream”. Is this not what you meant?

marten-seemann (commented):
This is a good start!

A few thoughts:
* We need to define how the stream labels (A and B) are assigned (multistream).
* I’m wondering if each side should use their own namespace for streams, or if they should share a namespace (e.g. by say that client/server use even/odd stream IDs). Maybe then it would be possible to migrate streams that the peer initiated (is this even desireable?).

connections/stream-migration.md
MarcoPolo: I think if the Responder misbehaves and doesn't close B then the stream migration hasn't finished since the new stream isn't in the same state as the old stream. I'm not sure what else we can do besides consider it not spec-compliant.

connections/stream-migration.md
MarcoPolo: Yeah, that's my current thinking. but I think there's a difference in how stebalien presented the protocol in #328. Specifically step 3 in the first part. Maybe I'm missing something but I don't think you need that part (updated the diagram to reflect that). I think you only need the new stream that knows what the stream id is of the old stream.

MarcoPolo:
> I’m wondering if each side should use their own namespace for streams, or if they should share a namespace (e.g. by say that client/server use even/odd stream IDs). Maybe then it would be possible to migrate streams that the peer initiated (is this even desireable?).

Can we rely on the underlying stream to give us a stream ID we can use for identifying streams?

What do you mean by migrating streams that the peer initiated? If I'm migrating a stream from a connection a remote peer started to a connection I started should that matter? I'm just changing how the bytes are getting there but everything else should be the same? Maybe there's something I'm missing here.

marten-seemann:
> Can we rely on the underlying stream to give us a stream ID we can use for identifying streams?

Probably not. While stream IDs are unique for every, they're not across muxers. It's probably easier if we introduce a new ID here.

> What do you mean by migrating streams that the peer initiated? If I'm migrating a stream from a connection a remote peer started to a connection I started should that matter? I'm just changing how the bytes are getting there but everything else should be the same?

Maybe I'm just confused by the word "initiator" here. Do we mean the initiator of the connection or the initiator of the stream?

MarcoPolo:
I think this finally clicked for me yesterday. The stream migration protocol is a prefix on top of another protocol. That’s how you can send the `identify this stream as id=A` message. I’ll update the spec on Monday.

connections/stream-migration.md
```diff
@@ -0,0 +1,255 @@
...
+### Stream IDs
+
+In the above diagram stream IDs have the labels `A` and `B`. In practice this
+ID will be represented as an int defined by the initiator. Both sides can
+globally identify this stream by the namespacing the ID with the initiator's
+peer ID. e.g. `12D3Foo.1`.
+### Stream migration protocol id
+
+The stream migration protocol id should follow the format of
+`/streamMigration/1.0.0/<streamID>`. The stream should be an int.
```
vyzo, line 146: /libp2p/stream-migration/1.0.0

marten-seemann: `/libp2p/stream-migration/`, we can still version this later if necessary.

connections/stream-migration.md
```diff
@@ -43,18 +50,44 @@ skinparam sequenceMessageAlign center
...
 
-Initiator <-> Responder: <b>Connection 1; stream A</b>
-Initiator <-> Responder: <b>Connection 2</b>
+
+  note over Responder
+   The responder knows that this stream is called <b>stream A</b>
+  end note
+
+  ... <i>Continue negotiating another protocol</i> ...
+else Responder does <i>not</i> supports stream-migration
```
marten-seemann, line 76: Note: This should be _super rare_. We know from Identify which protocols the peer speaks, and we should only attempt the stream migration protocol if we know that the peer actually supports it. Otherwise we'll waste 1 RTT. We probably should a SHOULD somewhere in the text explaining this.

connections/stream-migration.md
```diff
@@ -100,6 +133,21 @@ plantuml stream-migration.md -o stream-migration -tsvg
 
 Note: some of these steps may be pipelined.
 
+
+### Stream IDs
+
+In the above diagram stream IDs have the labels `A` and `B`. In practice this
+ID will be represented as an int defined by the initiator. Both sides can
```
marten-seemann, line 140: uint64?

connections/stream-migration.md
```diff
@@ -100,6 +133,21 @@ plantuml stream-migration.md -o stream-migration -tsvg
...
+globally identify this stream by the namespacing the ID with the initiator's
```
marten-seemann, line 141: This would be an implementation detail, right? We wouldn't use this ID anywhere (on the wire).

connections/stream-migration.md
marten-seemann: The protocol name can't include the stream ID. The stream ID is the payload of this protocol.

connections/stream-migration.md
MarcoPolo: correct

connections/stream-migration.md
MarcoPolo: > The protocol name can't include the stream ID. The stream ID is the payload of this protocol.

Why not? This way you wouldn't need a separate payload message.

The initiator would send the stream-migration protocol ID+stream id, then can send the underlying protocol right away without having to wait for a response.

The responder would read the stream migration protocol and ack it (by echoing back as in multistream select. Note this doc doesn't say this part yet.), then continue negotiating the underlying protocol. If the other node for some reason doesn't support stream-migration (even if we thought it did), it would echo back na, and continue as before.

I

MarcoPolo: ah, it could be because we cache the protocols seen on the other side, correct?

marten-seemann: Yes, exactly. Even worse, we send an Identify Delta messages for every new protocol that we add.

Also, logically speaking, the stream ID is not part of the protocol ID, but it's a payload of the stream migration protocol. Counting the bytes on the wire, it makes no (or at least not more than a few byte) difference if it's in the protocol name or in the payload.

mxinden: Not sure this was mentioned before: For the sake of evolvability of the protocol in the future, can the stream ID be send embedded in a Protobuf? That way we can add new fields in the future.

MarcoPolo: Good call out. I did this in the poc, but I’ll update the spec to include this.

connections/stream-migration.md
```diff
@@ -0,0 +1,238 @@
...
+2. Check for simultaneous connect: If we have both inbound and outbound
+   connections, keep the ones initiated by the peer with the lowest peer ID. Open
+   Question: Some protocols behave differently depending on whether they are the
+   dialer or listener. Can we really consolidate these?
+3. Prefer the connection with the most streams.
+4. Break ties in the remaining connections by selecting the newest conn, to
+   match the swarm's behavior in best connection selection.
+
+Note that it's not required that all implementations (and all versions) follow
+the same heuristics since the initiator is driving the migration and specifies
```
mxinden, line 222: Below I am assuming that *Initiator* refers to the *connection* initiator, not the stream initiator. Please correct me in case I am misunderstanding this @MarcoPolo.

Say there are two nodes `A` and `B`. Connection `AB` is initiated by `A` to `B`. Conneciton `BA` is initiated by `B` to `A`.

Say that `A` and `B` follow different heuristics to pick the best connection. `A` chooses `AB` as the best connection, `B` chooses `BA` as the best connection.

If I understand the above correctly, this would result in `A` moving all the streams it created to `AB` and `B` moving all the streams it created to `BA`. Both connections would thus stay alive.

Potential solution: Instead of allowing both `A` and `B` to migrate streams, how about delegating the decision making to the peer with the lower peer ID, e.g. in this case `A`?

MarcoPolo: Yes!

I was thinking about this this weekend and came up with a similar solution. Glad to see you also came to the same conclusion. I'll update this spec to make this explicit.

marten-seemann: Few thoughts on this:
1. Currently each peer chooses its own IDs for streams, i.e. there are two distinct spaces of stream IDs. If we want to allow the receiver of a stream to migrate that stream, we need a single stream ID space. One way to realize this would be to mandate the client (roles as seen by the stream muxer) to use odd and the server to use even stream IDs.
2. I don't think this document should describe how peers would choose `AB` over `BA`. This document should only describe how to migrate one libp2p stream from one muxer stream to another. For all that this spec cares about, those streams might (or might not) live on the same underlying connection. We can _then_ use the stream migration protocol as a building block to converge onto one connection (and the peer ID comparison is quite a neat idea, I like it!), but that should probably be described in a different document.

MarcoPolo: 1. Yes. I was considering adding a boolean flag that would indicate which peer initially identified the stream it's referencing when migrating (was I the initiator of the from stream?). This is the same as using even and odd numbers, since that scheme effectively encodes this boolean in the least significant bit. I'm fine with either way. Maybe it's a little easier to think about even and odds, so I'll do that.

2. Agreed that describing how to sort connections is out of the scope of this document (I imagine that spec to iterate more and and possibly have more subtle details). But I do think this spec should define who is responsible for doing the stream migration. If we end up in the situation where we have two identical connections (`A` dialed `B` and `B` dialed `A` at roughly the same time) we should describe who is in charge of doing the stream migration. By defining which node starts the stream migration we simplify this protocol and also avoid having to handle cases where both sides start stream migration at the same time.

bertrandfalguiere: "Potential solution: Instead of allowing both A and B to migrate streams, how about delegating the decision making to the peer with the lower peer ID, e.g. in this case A?"

Wouldn't this create a biais toward lower peerIDs? Maybe we can hash the concatenation of the two peerIDs (the lower first). If the hash is even, use the lowest. Else use the highest. That way it is deterministic but no peerID is systematically favored over another.

MarcoPolo: Seems reasonable enough, I added that. Thanks!

mxinden: > Wouldn't this create a biais toward lower peerIDs? Maybe we can hash the concatenation of the two peerIDs (the lower first). If the hash is even, use the lowest. Else use the highest. That way it is deterministic but no peerID is systematically favored over another.

What would be the upside/downside for a node to be "systematically favored"? Also note that an attacker could still pre-generate its peer ID in order to have the lower or higher one.

An attacker could try to force streams of a remote onto a slower connection. Though I doubt this would (a) have a big impact and (b) be the only way to do so.

MarcoPolo: I don't think there's any security protection that this gives you. The one thing it does give you is a avoiding the case where a peer happens to be in the lower end of the key space and thus always has to handle this responsibility. With the hash approach (and cooperative nodes) peers have a 50/50 odds of handling this responsibility, so it should be more uniform.

MarcoPolo: Yeah on second thought I agree with Max. I actually don't see the benefits since it's still 50% odds either way.

marten-seemann: I'm still not convinced that we should specify anything here at all. Stream migration is a general feature, a building block.

The use case we have in mind _now_ is migrating all streams from one connection to another, but we might come up with other use cases in the future. I'd prefer to have stream migration just be a thing that any node, regardless of its peer ID, can use _in principle_.
For the _particular_ use case of converging on a single connection, comparing peer IDs seems reasonable.

MarcoPolo: Let me try to rephrase to see if I understand:

1. The protocol should only specify how a node would perform a stream migration.
2. It doesn't define who starts the migration or why.
3. Consolidating connections would be a layer on top of this that defines which nodes is in charge of migrating streams to empty and close connections.

We just have to make sure that what we design here doesn't block point 3.

If that seems accurate, then I agree we don't need this in here.

MarcoPolo: also @marten-seemann to highlight a recent change re:

> Currently each peer chooses its own IDs for streams, i.e. there are two distinct spaces of stream IDs. If we want to allow the receiver of a stream to migrate that stream, we need a single stream ID space. One way to realize this would be to mandate the client (roles as seen by the stream muxer) to use odd and the server to use even stream IDs.

I've specced something similar, except the lower peer id node uses even and the higher peer id node uses odds. This let's us avoid having to rely on the stream muxer to give us this role. And it also works across connections (it gets confusing if the stream muxer says we are the client on connection and the server on the other).

connections/stream-migration/streammigration.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto2";
+
+package streammigration.pb;
+
+message StreamMigration {
+  oneof type {
+    Label label = 1;
+    Migrate migrate = 2;
+  }
```
marten-seemann, line 9: Does this need an `AckMigrate` option?

connections/stream-migration.md
```diff
@@ -0,0 +1,217 @@
...
+If either stream is "reset" before both ends are closed, both streams must be
+reset and the stream as a whole should be considered "aborted" (reset).
+
+### Half closed streams
+
+The final migrated stream should look the same as the initial stream. If the
+initial stream `1` was half closed, then the final migrated stream `2` should
+also be half closed. Note this may involve an extra step by one of the nodes.
+If a node, had closed writes to its old stream before migration it should also
+close writes to the new stream after migration.
```
marten-seemann, line 205: We should probably call out what happens if a node sends new data if the stream was already closed in that direction: that is a connection error.

connections/stream-migration.md
```diff
@@ -0,0 +1,217 @@
...
+   happened, the other side disagreeing, etc.).
+
+## The Protocol
+
+The goal of the protocol is to move traffic from one stream to another
+seamlessly. The final state of the new stream should be the same as the initial
+state of the old stream.
+
+The protocol should only be used when the initiator knows the responder
+understands the stream-migration protocol. Otherwise we waste 1 round trip.
```
marten-seemann, line 46:
```suggestion
understands the stream-migration protocol. In that case the negotiation of the stream migration protocol can be pipelined with the negotiation of the application protocol, and therefore doesn't cost any additional round trips.
```

marten-seemann (approved):

connections/stream-migration/streammigration.proto
MarcoPolo: Ah yes, thank you

connections/stream-migration/streammigration.proto
```diff
@@ -0,0 +1,24 @@
+syntax = "proto2";
```
achingbrain, line 1: Proto3 is recommended for new protocols

```suggestion
syntax = "proto3";
```

connections/stream-migration/streammigration.proto
```diff
@@ -0,0 +1,24 @@
...
+  oneof type {
+    Label label = 1;
+    Migrate migrate = 2;
+    AckMigrate ack_migrate = 3;
+  }
+}
+
+message Migrate {
+  required uint64 id=1;
+  required uint64 from=2;
```
achingbrain, line 15: No `required` in proto3

```suggestion
  uint64 id=1;
  uint64 from=2;
```

PR #417, opened 2022-05-20
ROADMAP.md: Deprioritize "Full Observatory" below ProtocolSelect

mxinden:
_If I recall correctly, this is what we agreed on in Paris. I might be wrong._


marten-seemann (commented):
I don't think this is what we agreed on. We need the observability to prevent future bugs like our muxer prioritization. On the Go side, we'll most likely tackle observability first.

That said, I'm not a fan of using this list to express prioritization. Different implements can and will have different priorities, and that's fine. Also, we've always been running multiple work streams in parallel, and this will be the case even more so now that we have more people working on each implementation.

mxinden:
I am fine either way. Just opened this for the sake of completeness.

> Different implements can and will have different priorities, and that's fine. 

Agreed.

PR #419, opened 2022-05-23
Style/dark mode diagrams

6d7a:
A white background ensures that all of the diagrams' text is readable in dark mode as well.


mxinden (approved):
:+1: Thanks.

PR #412, opened 2022-05-11
webrtc/: Add libp2p WebRTC browser-to-server spec

mxinden:
This pull requests summarizes the status quo of the ongoing WebRTC in libp2p work. In other words, it represents the summary of all past discussions like https://github.com/libp2p/specs/issues/220.

While describing both the browser-to-server and browser-to-browser use-case, priority is on the former. Once the browser-to-server use-case is sufficiently specified, I will remove the browser-to-browser use-case, merge this pull request, and start iteration two, namely to sufficiently specify the browser-to-browser use-case.

### Remaining work items for browser-to-server:

- [x] Specification
  - [x] Server to start Noise handshake https://github.com/libp2p/specs/pull/412#discussion_r977698490
  - [x] Head of line blocking https://github.com/libp2p/specs/pull/412#discussion_r984720911
  - [x] `RTCDataChannel` ID negotiation https://github.com/libp2p/specs/pull/412#discussion_r984746570
  - [x] `RTCDataChannel` ID reuse https://github.com/libp2p/specs/pull/412#discussion_r984747802
  - [x] Fixing fingerprint hash algorithm to sha-256 https://github.com/libp2p/specs/pull/412#discussion_r968327480
  - [x] Document potential deprecation of WebRTC Star and WebRTC Direct once browser-to-browser support is specified
  - [x] Fingerprint sorting https://github.com/libp2p/specs/pull/412#discussion_r978398623
  - [x] Document label on `RTCConnection.createDataChannel`.  https://github.com/libp2p/specs/pull/412/files#r995877508
  - [x] Communicate message size limit in SDP https://github.com/libp2p/specs/pull/412#issuecomment-1287780905
- [ ] Implementations of this specification
  - [ ] JS https://github.com/little-bear-labs/js-libp2p-webrtc/
    - [ ] Refactoring to use `Upgrader` interface
      - [ ] https://github.com/libp2p/js-libp2p-interfaces/pull/290#issuecomment-1259305581
      - [ ] https://github.com/libp2p/js-libp2p/pull/1395#issuecomment-1259595179
      - [ ] https://github.com/little-bear-labs/js-libp2p-webrtc/pull/29
  - [ ] Go https://github.com/libp2p/go-libp2p/pull/1655
  - [ ] Rust https://github.com/libp2p/rust-libp2p/pull/2622
     - [ ] Message framing https://github.com/melekes/rust-libp2p/pull/10
[...]

PR #429, opened 2022-07-17
webtransport: use CBOR to encode cert hashes during Noise handshake

marten-seemann:
A typical protobuf sent here weighs ~75 bytes, whereas this encoding results in ~20 bytes more being sent.

Given that WebRTC (#412) will also use Noise's Early Data feature, it would be nice if the data structure sent in the early data was self-describing, to avoid confusing WebTransport's and WebRTC's early data in implementations that use a single instance of a Noise Transport (the Go implementation won't, but it's not completely crazy to imagine other implementations choosing this route). Having WebTransport use a CBOR struct with "webtransport-hashes" as a field name makes sure we don't accidentally define the same Protobuf (on the wire) and mix up things.

### DIFF
--- a/webtransport/README.md
+++ b/webtransport/README.md
@@ -48,12 +48,10 @@ The first stream that the client opens on a new WebTransport session is used to 
 In order to verify that the end-to-end encryption of the connection, the peers need to establish that no MITM intercepted the connection. To do so, the client MUST include the certificate hash that was used to establish the connection as payload of the first Noise message (the `e` message). This payload is not encrypted, but the Noise handshake provides integrity protection.
 If the client was willing to accept multiple certificate hashes, but cannot determine which certificate was actually used to establish the connection (this will commonly be the case for browser clients), it MUST include a list of all certificate hashes.
 
-Certificate hashes are encoded using the following protobuf message:
-```proto
-syntax = "proto2";
-
-message WebTransport {
-  repeated bytes cert_hashes = 1;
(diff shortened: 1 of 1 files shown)


mxinden:
I am not sure I understand the proposition. Am I paraphrasing this correctly?

> Protobuf is not self-describing. This particular CBOR is self-describing. Being self-describing makes sure the message is not accidentally used for a different purpose (e.g. WebRTC instead of WebTransport).

Reponse below applies to what I understand above.

> (the Go implementation won't, but it's not completely crazy to imagine other implementations choosing this route).

I don't think the Rust implementation would either.

> to avoid confusing WebTransport's and WebRTC's early data in implementations that use a single instance of a Noise Transport

I would not consider this a large issue. In other words, I think we have bigger foot guns.

> A typical protobuf sent here weighs ~75 bytes, whereas this encoding results in ~20 bytes more being sent.

As far as I know we don't use CBOR anywhere. I very much value the consistency, i.e. libp2p using *mostly* ProtoBuf.

Say we want to go the route of ensuring a message to not be mistaken as a different one, would a Protobuf *Dictionary* not do the job as well?

melekes:
> to avoid confusing WebTransport's and WebRTC's early data in implementations that use a single instance of a Noise Transport (the Go implementation won't, but it's not completely crazy to imagine other implementations choosing this route)

The problem seems a little far-fetched to me. Rust implementation also uses a separate instance of Noise. What would be the benefits of using the same instance between two transports?

Issue #411, opened 2022-05-10
AutoNAT: Clarify cases in which response status `E_DIAL_ERROR` should be returned


elenaf9:
Follow up on https://github.com/libp2p/rust-libp2p/pull/2618#discussion_r867529638.

Concerning the ResponseStatus E_DIAL_ERROR, the AutoNAT spec currently only states that:

> If all dials fail, the receiver sends a DialResponse message with the ResponseStatus E_DIAL_ERROR. 

However, the Go implementation also returns `DialError`s without even trying to the dial a remote peer. Concretely, this happens when [`skip_dial` ](https://github.com/libp2p/go-libp2p/blob/393e3518b397a1ae88fd825c85f892956868543f/p2p/host/autonat/dialpolicy.go#L20) returns true for the client's observed address, which is the case if it is e.g. a relayed or private address.
I don't think that this should count as an dial error since the server never actually tried to dial the client. Instead a `DialRefused` should be returned (which would also match the returned status_text  _"refusing to dial peer with blocked observed address"_). 

If folks disagree and are in favor of returning a `DialError`, this should be part of the spec.

marten-seemann:
I agree that it should be a `DialRefused`, the `DialError` has caused us a lot of headache in the past: It's not possible to give your peer a single address, and just probe that single address.

We can clarify this in the spec, however, we can't change the way that already deployed Go implementations behave. This means that probing a single address won't be possible until all nodes have upgraded (which, at the current speed of upgrades in the IPFS network means a couple of years). This was _one_ of the reasons for [Autonat v2 proposal](https://github.com/libp2p/go-libp2p-autonat/issues/10#issuecomment-906531640).

elenaf9:
I think if we agree that `DialError` should only be returned if an actual dial attempt was made then the current specs are fine. I just wanted to avoid that a `DialError` it is returned for a case that is not specified.
I'm happy to do a PR to go-libp2p to change it there (so that at least the upgrading nodes are aligned with this), unless you think this could cause problems in the IPFS network?

marten-seemann:
@elenaf9 Let's do it! And add a comment that we can't rely on this error code since all versions before v0.20.0 sent the wrong error code.

MarcoPolo:
> It's not possible to give your peer a single address, and just probe that single address.

I don't follow. Wouldn't you be able to set the dial message to have only one address?

I think this change could be backwards compatible since older nodes only check if the response was OK or not.

[...]

PR #430, opened 2022-07-19
Clarify use of E_DIAL_REFUSED in autonat

achingbrain:
Adds a paragraph detailing when `E_DIAL_REFUSED` should be sent as a `DialResponse` status as it's not mentioned in the current spec.

I think this is consistent with go-libp2p since https://github.com/libp2p/go-libp2p/pull/1527

### DIFF
--- a/autonat/README.md
+++ b/autonat/README.md
@@ -78,6 +78,12 @@ If all dials fail, the receiver sends a `DialResponse` message with the
 successfully, it sends a `DialResponse` with the `ResponseStatus` `OK`. It
 SHOULD include the address it successfully dialed in its response.
 
+If the receiver elects not to dial any addresses from the message - for example
+because they are invalid, because they resolve to the same host as the receiver,
+because they only include private addresses, because it does not support the
+relevant transports, or because it has no capacity, is sends a `DialResponse`
+message with the `ResponseStatus` `E_DIAL_REFUSED`.
+
 The initiator uses the responses obtained from multiple peers to determine its
 NAT status. If more than 3 peers report a successfully dialed address, the node
 SHOULD assume that it is not located behind a NAT and publicly accessible. On


autonat/README.md
```diff
@@ -78,6 +78,12 @@ If all dials fail, the receiver sends a `DialResponse` message with the
 successfully, it sends a `DialResponse` with the `ResponseStatus` `OK`. It
 SHOULD include the address it successfully dialed in its response.
 
+If the receiver elects not to dial any addresses from the message - for example
+because they are invalid, because they resolve to the same host as the receiver,
+because they only include private addresses, because it does not support the
+relevant transports, or because it has no capacity, is sends a `DialResponse`
```
mxinden, line 84:
```suggestion
relevant transports, or because it has no capacity, it sends a `DialResponse`
```

mxinden (approved):
Looks good to me. //CC @elenaf9 who has been looking into these in the past https://github.com/libp2p/specs/issues/411

mxinden:
Thanks :pray: @achingbrain!

marten-seemann (changes requested):
We should probably add that you can’t trust on your peer doing this correctly - the Go implementation has been sending the wrong error code for a long time (until Elena fixed it recently).

autonat/README.md
```diff
@@ -78,6 +78,12 @@ If all dials fail, the receiver sends a `DialResponse` message with the
 successfully, it sends a `DialResponse` with the `ResponseStatus` `OK`. It
 SHOULD include the address it successfully dialed in its response.
 
+If the receiver elects not to dial any addresses from the message - for example
+because they are invalid, because they resolve to the same host as the receiver,
+because they only include private addresses, because it does not support the
```
elenaf9, line 83:
```suggestion
because they are invalid, because it does not support the
```

> because they only include private addresses

In rust-libp2p we don't reject requests in this case. Directly after starting to listen, a peer won't know any of their external addresses and instead just their listening addresses (which are all private). The server still attempts to dial them, based on the address it observes this peer on, i.e. it replaces the IP of each address in the message with the observed IP. We do this anyway for DDoS protection (see https://github.com/status-im/nim-libp2p/pull/739#discussion_r917551966; @Tanguy suggested that we also add a section on for this, which imo would make sense).
Only if the observed address is private we may reject the peer (depending on whether this was set in our config or not).
> because they resolve to the same host as the receiver

Again, because we replace the IP with the observed one at least in rust-libp2p we don't really handle this case.

---
 Not sure how I think about having an exemplary list of rejection reasons. They seem rather go-lib2p2p specific right now. I'd propose to either properly specify in which cases a DialRequest shall be rejected, or keep it on a higher level to leave it up to the implementation (see suggestion).

autonat/README.md
elenaf9, line 84:
```suggestion
relevant transports, or because it has no capacity, it SHOULD send a `DialResponse`
```

> We should probably add that you can’t trust on your peer doing this correctly - the Go implementation has been sending the wrong error code for a long time (until Elena fixed it recently).

@marten-seemann What do you think about using the keyword _SHOULD_, in the sense that it is recommended but one can not trust that the other peer actually does it?

marten-seemann: Technically, this is a MUST. You _should_ be able to trust your peer's reporting here.

In practice you can't. But that's just because the Go implementation was just not compliant with the spec, which brought us to the mess we're in right now (one more reason for Autonat v2!). I suggest adding a note here that this was the case.

BigLep:
@achingbrain and @p-shahi : this has been open for a long time.  What are the next steps to get this landed?

PR #435, opened 2022-08-20
pubsub/gossipsub: Update rust-libp2p supporting v1.1

mxinden:
To the best of my knowledge, the Gossipsub implementation in rust-libp2p supports Gossipsub v1.1.

@AgeManning can you confirm? Am I missing something?

Follow up to https://github.com/libp2p/rust-libp2p/discussions/2825#discussioncomment-3430068.


AgeManning:
In my opinion, i think its fair to say we fully support 1.1. 

But technically, I need to finish this guy to fully support everything: https://github.com/libp2p/rust-libp2p/issues/2398

mxinden:
> In my opinion, i think its fair to say we fully support 1.1.

:+1: 

In case no one objects, I will merge here next week.

PR #439, opened 2022-08-31
Add SNI to TLS spec

MarcoPolo:
Adds SNI to the TLS negotiation spec. Increments the minor version.

We need SNI for wss (secure websockets) to work and we should document its usage.

### DIFF
--- a/tls/tls.md
+++ b/tls/tls.md
@@ -1,12 +1,14 @@
 # libp2p TLS Handshake
 
 | Lifecycle Stage | Maturity                 | Status | Latest Revision |
-|-----------------|--------------------------|--------|-----------------|
-| 2A              | Candidate Recommendation | Active | r0, 2019-03-23  |
+| --------------- | ------------------------ | ------ | --------------- |
+| 2A              | Candidate Recommendation | Active | r1, 2022-08-31  |
+
+Current Version: v1.1.0
 
 Authors: [@marten-seemann]
 
-Interest Group: [@Stebalien], [@jacobheun], [@raulk], [@Kubuxu], [@yusefnapora]
+Interest Group: [@Stebalien], [@jacobheun], [@raulk], [@Kubuxu], [@yusefnapora], [@marcopolo]
 
 [@marten-seemann]: https://github.com/marten-seemann
 [@Stebalien]: https://github.com/Stebalien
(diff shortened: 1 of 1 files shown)


marten-seemann (commented):
I'm not sure if this is the right document. This document describes how the libp2p identity is encoded into a TLS certificate. This doesn't require any SNI.

SNI is only needed for "regular" (CA-signed) certificates.

tls/tls.md
```diff
@@ -116,8 +121,15 @@ message PublicKey {
 
 **TODO: PublicKey.Data looks underspecified. Define precisely how to marshal the key.**
 
-## Future Extensibility
+## SNI
+Clients MAY use the Server Name Indication (SNI) in the `ClientHello` as defined in [RFC 6066 §3](https://www.rfc-editor.org/rfc/rfc6066#section-3).
```
mxinden, line 125: How about mentioning the requirement of using the protocol id (i.e. `/tls/1.1.0`) here?

tls/tls.md
```diff
@@ -116,8 +121,15 @@ message PublicKey {
...
+# Changelog
+## V1.1.0
 
-Future versions of this handshake protocol MAY use the Server Name Indication (SNI) in the `ClientHello` as defined in [RFC 6066, section 3](https://tools.ietf.org/html/rfc6066) to announce their support for other versions.
+Allow Clients to send an [SNI](#sni). Servers MAY use the SNI.
+### Backwards compatibility
 
-In order to keep this flexibility for future versions, clients that only support the version of the handshake defined in this document MUST NOT send any value in the Server Name Indication. Servers that support only this version MUST ignore this field if present.
+- Clients communicating with nodes only speaking `v1.0.0` MUST NOT send an SNI.
+- Servers communicating with nodes only speaking `v1.0.0` MUST ignore the SNI.
```
mxinden, line 135: This is confusing me a bit. Why would a `v1.0.0` client send an SNI?

I don't think this directly translates from the previous version:

> In order to keep this flexibility for future versions, clients that only support the version of the handshake defined in this document MUST NOT send any value in the Server Name Indication. Servers that support only this version MUST ignore this field if present.

Wouldn't the following be more accurate?

```
- Clients supporting `v1.0.0` only MUST NOT send an SNI.
- Servers supporting `v1.0.0` MUST ignore the SNI.
- ```

mxinden (commented):
> I'm not sure if this is the right document. This document describes how the libp2p identity is encoded into a TLS certificate. This doesn't require any SNI.

I don't have an opinion on this. @marten-seemann what would be a better place?

MarcoPolo:
I'm closing this since I agree with Marten. This isn't the spot for the SNI field. This spec is around how we tie a peer id to a tls connection with the libp2p extension. What I'm trying to do is outside of that. And is a way to specify an SNI for a tls connection in general (not specific to this libp2p extension).

I think to better explain this I need to better explain multiaddrs and how we interpret them. That's a different document. I'll close this and start the "how we interpret multiaddr" doc.

Thanks for the feedback all

PR #444, opened 2022-09-04
_archive/: Fix links to non archived resources #444

ChengYueJia:
same problem like #438


mxinden (approved):
Looks good to me. Thanks for the fixes @ChengYueJia!

PR #446, opened 2022-09-09
Muxer selection in security handshake

julian88110:
This spec describes a feature that carries out muxer selection in the security handshake process.


marten-seemann (commented):
This looks like a good start, thank you @julian88110!

Not sure if I was supposed to review it yet, so I just added some high-level comments.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,220 @@
...
+
+At the end of the handshake pattern, both the client and server have received
+the peer's early data. The Noise protocol does not perform the protocol
+selection as TLS does, rather it just delivers the early data to handshaking
+peers.
+
+The muxer selection logic runs out of the Noise handshake process, relying on
+the early data exchanged during the handshake. The early data is delivered in
+the form of a byte string. The supported muxers are passed in space separated
+string codes. An example early data string:
```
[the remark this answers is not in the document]
mxinden: I am very much in favor of using a protobuf here for the above mentioned reasons.

How about taking this one step further and use the _Protocol Select_ (https://github.com/libp2p/specs/pull/349) protobuf definition? To me this proposal actually reads as an extension to _Protocol Select_. Would people agree with that?

mxinden (commented):
Solid first specification proposal @julian88110! Thank you for the write-up!

connections/muxer-sel-in-sec-handshake.md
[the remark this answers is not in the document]
marten-seemann: I’m not convinced that would be a good idea.

1. This is specific to just the Noise handshake, there’s no reason to cause any coupling here
2. The Protocol Select spec might change before we actually deploy it, which would leave us with a zombie protobuf here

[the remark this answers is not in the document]
mxinden: > This is specific to just the Noise handshake

Related to this discussion, would it not make sense to use the same structure for both Noise and TLS? In case TLS only allows valid UTF-8 strings, we can use `multibase`.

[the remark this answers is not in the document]
mxinden: > The Protocol Select spec might change before we actually deploy it, which would leave us with a zombie protobuf here

In case it changes, we can do those changes in a backwards compatible way, given that we are using Protobuf, no?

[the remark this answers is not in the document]
mxinden: I am still failing to see the difference between the Protocol Select protocol and this proposal. The only difference to me is that this proposal exchanges the messages via the early data mechanism, whereas the Protocol Select protocol exchanges the messages on an _established_ channel. Am I missing something?

Both protocols
- Exchange a list of protocols
- Use the same prioritization algorithm
- Fail in the same scenarios (no protocols overlap)

How about adding an optional early-negotiation mechanism to Protocol Select, namely the exchange described in this pull request.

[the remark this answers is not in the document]
marten-seemann: > Related to this discussion, would it not make sense to use the same structure for both Noise and TLS? In case TLS only allows valid UTF-8 strings, we can use multibase.

That doesn't work with existing libraries, who do ALPN negotiation for you. They expect a certain format.

> In case it changes, we can do those changes in a backwards compatible way, given that we are using Protobuf, no?

I still don't see the advantage of using a more complicated than necessary Protobuf here. This negotiation step is very tightly bound to the handshake protocol, and it's reasonable to expect that different handshake protocols will use different encodings (as TLS and Noise already do).

> The only difference to me is that this proposal exchanges the messages via the early data mechanism, whereas the Protocol Select protocol exchanges the messages on an established channel. Am I missing something?

That's a huge difference. It saves 1-RTT *in every case*. In specific, early data in TLS 1.3 only works on resumed connections, whereas this proposal works on every connection.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,220 @@
...
+2. The selected security protocol performs handshaking and establishs a secure
+tunnel
+3. The multistream-selection protocol then will run again for muxer selection.
+4. Connection then is upgraded to a capable connection by the selected muxer.
+
+
+## Improved muxer selection
+
+The security protocol's ability of supporting higher level abstract protocol
+negotiation (for example, TLS's support of NextProtos, and Noise's support of
```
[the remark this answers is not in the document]
julian88110: yes, good point.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Cross version support
+
+The improved muxer selection approach MUST be interoperable with pervious
+libp2p versions which do not support this improved approach.
+
+### TLS case
+
+In the current version of libp2p, the "NextProtos" field is populated with a
+key "libp2p". By appending the key of "libp2p" to the end of the supported
```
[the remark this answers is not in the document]
julian88110: That's correct, I double checked the go implementation, Here is what it says: 
        // NextProtos is a list of supported application level protocols, in
	// order of preference. If both peers support ALPN, the selected
	// protocol will be one from this list, and the connection will fail
	// if there is no mutually supported protocol. If NextProtos is empty
	// or the peer doesn't support ALPN, the connection will succeed and
	// ConnectionState.NegotiatedProtocol will be empty.

The connection fails if there is not mutually supported ALPN, and can succeed if one of the ALPN fields is empty.  So we don't have to append "libp2p".

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,220 @@
...
+MUST fall back to the multistream-selection protocol to select the muxer.
+
+## Security
+
+The muxer list carried in TLS NextProtos field is part of the ClientHello
+message which is not encrypted. This feature will expose the supported muxers
+in plain text, but this is not a weakening of securiy posture. In the fuure
+when [ECH] is ready the muxer info can be protected too.
+
+The early data in Noise handshake is encrypted so that the muxer info carried
```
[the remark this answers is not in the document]
julian88110: For the Noise case, we plan to send early data in the second and third message in the xx pattern, (which I believe the libp2p implementation uses), those messages are encrypted so the early data can be encrypted too.

connections/muxer-sel-in-sec-handshake.md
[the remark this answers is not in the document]
julian88110: I like this section of discussion, it clarifies lot of things. 
I will add a protobuf definition for Noise.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,220 @@
...
+an empty result, and the multistream-selection protocol MUST be run to negotiate
+the muxer.
+
+(TBD: in some cases early abortion is desireable)
+
+
+### Muxer selection over Noise
+
+The libp2p Noise implementation allows the Noise handshake process to carry
+early data. [Noise-Early-Data] is carried in the second and third message of
```
marten-seemann, line 123: We need to specify which one exactly.

julian88110: sure I can add more specifics on handshaking/early data interaction.

connections/muxer-sel-in-sec-handshake.md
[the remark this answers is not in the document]
mxinden: > > The only difference to me is that this proposal exchanges the messages via the early data mechanism, whereas the Protocol Select protocol exchanges the messages on an established channel. Am I missing something?
> 
> That's a huge difference. It saves 1-RTT _in every case_. In specific, early data in TLS 1.3 only works on resumed connections, whereas this proposal works on every connection.

Sorry. I think I phrased this in a confusing way. I am definitely in favor of the 1RTT saving. My point was that I don't see the difference in the two that would justify different encodings. With the restriction on TLS' ALPN mechanism, this makes sense now.

> > Related to this discussion, would it not make sense to use the same structure for both Noise and TLS? In case TLS only allows valid UTF-8 strings, we can use multibase.
> 
> That doesn't work with existing libraries, who do ALPN negotiation for you. They expect a certain format.

Thanks for raising this. I was not aware of (/ didn't think of) this restriction. It is the case for Rust, I am guessing it is also the case for the other languages and their major TLS implementations.

Let's use the current proposal and not the Protocol Select Protobuf.

connections/muxer-sel-in-sec-handshake.md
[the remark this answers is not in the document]
MarcoPolo: We should probably document this. I don't think we have anything about this in this repo.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,248 @@
...
+These is no security weakening in this case either.
+
+
+## Protocol coupling
+
+This feature aggregates the multistream-selecion function and security
+handshake function. From function separation point of view, it introduces
+coupling between different functions. But the argument is that in the case of
+libp2p, the muxer and security are always needed at the same time, and it is a
+small price to pay to gain efficiency by ruducing one RTT.
```
MarcoPolo, line 231:
```suggestion
small price to pay to gain efficiency by reducing one RTT.
```

MarcoPolo (commented):
Looking good! I think we should have a Go implementation before merging this, since I think that will uncover some subtleties we probably want to mention.

Also I think the exact steps in the noise handshake are still a little fuzzy for me.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,248 @@
...
+   ServerHello message.  The application protocol negotiation can thus
+   be accomplished within the TLS handshake, without adding network
+   round-trips, and allows the server to associate a different
+   certificate with each application protocol, if desired.
+
+For the purpose of muxer selection, the types of muxers are coded as protocol
+names in the form of a list of strings, and inserted in the ALPN "NextProtos"
+field. An example list as following:
+
+    ["yamux/1.0.0", "mplux", "libp2p"]
```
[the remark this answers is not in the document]
marten-seemann: The complete ID for yamux is `/yamux/1.0.0`: https://github.com/libp2p/go-libp2p/blob/aa6d051e651fb1fc42c7f4d98023a3b6c01df6a0/defaults.go#L32-L36

Similarly, mplex is `/mplex/6.7.0`, see for example https://github.com/libp2p/go-libp2p/blob/aa6d051e651fb1fc42c7f4d98023a3b6c01df6a0/examples/ipfs-camp-2019/08-End/main.go#L44-L47

connections/muxer-sel-in-sec-handshake.md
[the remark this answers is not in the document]
marten-seemann: >  If both peers support ALPN, the selected
> protocol will be one from this list, and the connection will fail
> if there is no mutually supported protocol.

I'm surprised by this. I _thought_ that the Go standard library didn't enforce this. RFC 7301 doesn't require peers to  fail the handshake if ALPN negotiation fails, which is why we had to require that behavior in QUIC (see https://datatracker.ietf.org/doc/html/rfc9001#section-8.1).

As @MarcoPolo said, I agree that it would be valuable to be very precise about what we expect in various case (ALPN offered with muxers, APLN offered with only libp2p, no / empty ALPN offered), and to have tests for this behavior in our implementation.

connections/muxer-sel-in-sec-handshake.md
marten-seemann: There'd be an advantage to using the first and the second handshake message instead of the second and the third: After receiving the list of muxers from the client, the server would already know which muxer was chosen, and could already initialize that muxer and start opening streams.
This would allow the server to optimistically (at this point, the client is not yet authenticated, so we can't trust its peer ID) send some data.

For example, I could see us opening an identify stream to push our list of supported protocols at this point. We don't really care if this list is captured by a MITM, but it would allow the client to proceed faster after the handshake.

What do you think?

MarcoPolo: does the data actually get sent over the wire? I thought the write would be blocked until the noise handshake finishes.

connections/muxer-sel-in-sec-handshake.md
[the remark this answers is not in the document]
julian88110: Of course we will create test cases for this. It seems there are some differences between the RFC and the go lib implementation documentation. We need to flush out what is the actual behavior.

connections/muxer-sel-in-sec-handshake.md
julian88110: Given TLS already has muxer info in the plain text for the first TLS handshake message, and the fact muxer info is not sensitive data, I don't have objections to transmit the muxer info in the first and second Noise handsahke messages.

connections/muxer-sel-in-sec-handshake.md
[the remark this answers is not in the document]
julian88110: Good catch, thanks to Marco and Marten.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,248 @@
...
+The muxer selection logic runs out of the Noise handshake process, relying on
+the early data exchanged during the handshake. The early data is delivered in
+the form of a byte string in the second and third message of the XX handshake
+pattern. 
+ 
+The early data for this purpose is specified in the protobuf in the
+[Early-data-specification] section. The muxers are ordered by preference, with
+the most prefered muxer at the beginning.
+
+After the Noise handshake, the client and server run the muxer selection
```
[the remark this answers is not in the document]
julian88110: Great insight. Also added two examples to make the Noise handshake process more clear.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,248 @@
...
+and spec status.
+
+[lifecycle-spec]: https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md
+
+## Table of Contents
+
+- [Muxer selection in security handshake](#Muxer-selection-in-security-handshake)
+    - [Table of Contents](#table-of-contents)
+    - [Overview](#overview)
+    - [Design])(#design)
```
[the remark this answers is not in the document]
julian88110: Done, thanks!

connections/muxer-sel-in-sec-handshake.md
julian88110: done, thanks!

connections/muxer-sel-in-sec-handshake.md
MarcoPolo: > the server would already know which muxer was chosen, and could already initialize that muxer and start opening streams.
This would allow the server to optimistically (at this point, the client is not yet authenticated, so we can't trust its peer ID) send some data.

I don't think we could do this if we are encrypting the connection with the key derived from the noise handshake. The data that's sent after the handshake completes must be derived from all the DH operations. If the server optimistically starts sending data (e.g. initializing the muxer and opening streams) it will be encrypted with the wrong key after the noise handshake completes. The server needs to wait for the 3rd handshake message (in XX) for it to know what key to use to encrypt the data over the wire.

marten-seemann:
@julian88110 The Noise Extension Registry (#453) was merged. You can rebase this PR now, and add an extension entry for the muxer selection.

julian88110:
Thanks @marten-seemann  for the notice. Extension Registry added and pushed to the PR branch.

connections/muxer-sel-in-sec-handshake.md
julian88110: Per the discussion on #454, we decided to use first and second messages to send early data, so resolving this conversation thread.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,265 @@
...
+
+For the purpose of muxer selection, the types of muxers are coded as protocol
+names in the form of a list of strings, and inserted in the ALPN "NextProtos"
+field. An example list as following:
+
+    ["yamux/1.0.0", "/mplex/6.7.0", "libp2p"]
+
+The NextProtos list is ordered by preference, with the most prefered muxer at
+the beginning. The "libp2p" protocol code MUST always be the last item in the
+ALPN NextProtos list.
```
mxinden, line 109:
```suggestion
ALPN NextProtos list. See [#tls-case] for details on the special "libp2p" protocol code.
```

Would have helped me a bit on a first read.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,265 @@
...
+supported by the responder, that muxer is selected. If no mutually supported
+muxer is found, the muxer selection process MUST fall back to multistream
+-selection protocol.
+
+Example: Noise handshake between peers that have a mutually supported muxer.
+    Initiator supports: ["yamux/1.0.0", "/mplex/6.7.0"]
+    Responder supports: ["yamux/1.0.0"]
+
+    XX:
+    -> e ["yamux/1.0.0", "/mplex/6.7.0"]
```
mxinden, line 140: Ah, I though we settled on sending these in the second and third message? Sorry in case I am confusing the past discussions. Would appreciate any pointers.

This is also how I read the section below:

> The early data in Noise handshake is only sent afer the peers establish a
shared key, in the second and third handshake messages in the XX pattern. So
the early data is encrypted and the muxer info carried over is protected.
These is no security weakening in this case either.

mxinden (commented):
Thanks for the thorough write-up @julian88110!

connections/muxer-sel-in-sec-handshake.md
julian88110: Hi Max, thanks for the catch. Yes, you are right. I missed changing this part.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,265 @@
...
+   With ALPN, the client sends the list of supported application
+   protocols as part of the TLS ClientHello message.  The server chooses
+   a protocol and sends the selected protocol as part of the TLS
+   ServerHello message.  The application protocol negotiation can thus
+   be accomplished within the TLS handshake, without adding network
+   round-trips, and allows the server to associate a different
+   certificate with each application protocol, if desired.
+
+For the purpose of muxer selection, the types of muxers are coded as protocol
+names in the form of a list of strings, and inserted in the ALPN "NextProtos"
```
marten-seemann, line 102: “NextProtos” is the name of the field in the Go implementation (for historic reasons, because it was first an implementation of NPN, not ALPN). I suggest sticking with the RFC terminology here.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,265 @@
...
+field. An example list as following:
+
+    ["yamux/1.0.0", "/mplex/6.7.0", "libp2p"]
```
marten-seemann, line 105:
```suggestion
    ["/yamux/1.0.0", "/mplex/6.7.0", "libp2p"]
```

connections/muxer-sel-in-sec-handshake.md
julian88110: good point, changed that to "ALPN extesion"

julian88110: thanks for the catch, agreed. changed to "ALPN extension" which is aligned with the RFC

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,266 @@
...
+
+    ["/yamux/1.0.0", "/mplex/6.7.0", "libp2p"]
+
+The muxer list is ordered by preference, with the most prefered muxer at
+the beginning. The "libp2p" protocol code MUST always be the last item in the
+muxer list . See [#tls-case] for details on the special "libp2p" protocol code.
+
+The server chooses the supported protocol by going through its prefered
+protocol list and searchs if the protocol is supported by the client too. if no
+mutually supported protcol is found the TLS handshake will fail.
```
marten-seemann, line 113: Use normative language here.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,266 @@
...
+
+If the selected item from the muxer list is "libp2p" then the muxer selection
+process returns an empty result, and the multistream-selection protocol MUST be
+run to negotiate the muxer.
+
+
+### Muxer selection over Noise
+
+The libp2p Noise implementation allows the Noise handshake process to carry
```
marten-seemann, line 122: Specification, not implementation.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,266 @@
...
+early data. [Noise-Early-Data] is carried in the second and third message of
+the XX handshake pattern as illustrated in the following message sequence chart.
+The second message carries early data in the form of a list of muxers supported
+by the responder, ordered by preference. The initiator sends its supported
```
marten-seemann, line 126: Let's be more specific here. There are 4 messages in the server's 1st flight.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,266 @@
...
+process is fully done, the initiator and responder will both process the
+received eraly data and select the muxer to be used, they both iterate through
+the initiator's prefered muxer list in order, and if any muxer is also
+supported by the responder, that muxer is selected. If no mutually supported
+muxer is found, the muxer selection process MUST fall back to multistream
+-selection protocol.
+
+Example: Noise handshake between peers that have a mutually supported muxer.
+    Responder supports: ["yamux/1.0.0", "/mplex/6.7.0"]
+    Initiator supports: ["yamux/1.0.0"]
```
marten-seemann, line 137: Better: use an example that has [ mplex, yamux ], so that the preference becomes clear.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,266 @@
...
+    Responder supports: ["/mplex/6.7.0"]
+    Initiator supports: ["yamux/1.0.0"]
+
+    XX:
+    -> e
+    <- e, ee, s, es, ["/mplex/6.7.0"]
+    -> s, se, ["yamux/1.0.0"]
+    
+    After handshaking is done, early data processing will find no mutually
+    supported muxer, and falls back to multistream-selection protocol.
```
marten-seemann, line 158: We should fail the handshake in that case. There's no point in doing another multistream negotiation.

We can use the absence of the muxer entry in the protobuf for backwards compatibility.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,266 @@
...
+
+The muxer selection logic is run outside of the Noise handshake process. The
```
marten-seemann, line 160: Not sure what that means.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,266 @@
...
+format of he early data for this purpose is specified in the protobuf in the
+[Early-data-specification] section.
+
+#### Early data specification
```
marten-seemann, line 164: This seems to belong in the Noise doc. We shouldn't repeat the definition of the `NoiseExtensions` here, there should only be a single source of truth.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,266 @@
...
+  optional bytes identity_key = 1;
+  optional bytes identity_sig = 2;
+  optional NoiseExtensions extensions = 4;
+}
+```
+
+## Cross version support
+
+The improved muxer selection approach MUST be interoperable with pervious
+libp2p versions which do not support this improved approach.
```
marten-seemann, line 194: That seems like a weird place for a normative sentence. We should probably say in the introduction that this mechanism is designed to be backwards compatible.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,266 @@
...
+
+### TLS case
```
marten-seemann, line 196: Explain how backwards compatibility works in the TLS section of this document?

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,266 @@
...
+sets is always satisfied. When one peer runs the old version and the other peer
+runs the version that supports this feature, the negotiated protocol is
+"libp2p".
+
+In the case "libp2p" is the result of TLS ALPN, an empty result MUST be
+returned to the upgrade process to indicate that no muxer was selected. And the
+upgrade process MUST fall back to the multistream-selection protocol to
+to negotiate the muxer to be selected.
+
+### Noise case
```
marten-seemann, line 211: Explain how backwards compatibility works in the Noise section of this document?

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,266 @@
...
+in plain text, but this is not a weakening of securiy posture. In the fuure
+when [ECH] is ready the muxer info can be protected too.
+
+The early data in Noise handshake is only sent afer the peers establish a
+shared key, in the second and third handshake messages in the XX pattern. So
+the early data is encrypted and the muxer info carried over is protected.
+These is no security weakening in this case either.
+
+
+## Protocol coupling
```
marten-seemann, line 234: That also seems to belong in the introduction, and can probably be shortened to a single sentence.

marten-seemann (commented):
@julian88110 Can I ask you to do another editorial pass over this PR? There are quite a few typos in there, inconsistent tenses, half-finished sentences etc. Probably a good idea to use also use a spell checker.

In general, I think we should call it "muxer negotiation" instead of "muxer selection".

I added a few comments to the document.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,266 @@
...
+## Overview
+
+This document discribes an imporvement on the connection upgrade process. The
+goal of the improvement is to reduce the number of RTTs that takes to select the
+muxer of a connection. The solution relies on the ability of the security
+protocol's handshake process to negotiate higher level protocols, which enables
+the muxer selection to be carried out along with security protocol handshake.
+The proposed solution saves the RTT of multistream selection for muxers.
+
+For more context and the status of this work, please refer to [#426]
```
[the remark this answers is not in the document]
julian88110: done

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,266 @@
...
+            - [Muxer selection over Noise](#muxer-selection-over-noise)
+                - [Early data specification](#early-data-specification)
+    - [Cross version support](#cross-version-support)
+        - [TLS case](#tls-case)
+        - [Noise case](#noise-case)
+    - [Security](#security)
+    - [Protocol coupling](#protocol-coupling)
+    - [Alternative options considered](#alternative-options-considered)
+
+## Overview
```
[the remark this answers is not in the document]
julian88110: done

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,266 @@
...
+The current connection upgrade process is described in detail in [connections].
+As shown in this [sequence-chart], after a network connection is established,
+the following will happen to upgrade the conntection to a capable connection.
+
+1. The multistream-selection protocol is run over the connection to select the
+security protocol to be used.
+2. The selected security protocol performs handshaking and establishs a secure
+tunnel
+3. The multistream-selection protocol then will run again for muxer selection.
+4. Connection then is upgraded to a capable connection by the selected muxer.
```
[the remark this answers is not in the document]
julian88110: done

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,266 @@
...
+
+In order to achieve the above stated goal, each candidate muxer will be
+represented by a protocol name/code, and the candidate muxers are supplied to
+the security protocol's handshake process as a list of protocol names.
+
+If the client and server agree upon the common muxer to be used, then the
+result of the muxer selection is a muxer code represented by the selected
+protocol name. If no agreement is reached upon by the client and server
+then an empty muxer code is returned and the connection upgrade process
+MUST fall back to the multistream-selection protocol to negotiate the muxer.
```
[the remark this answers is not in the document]
julian88110: 1. That makes sense, got rid of the implementation details.
2. The behavior of falling back to multistream-selection instead of failing handshaking is a way of providing backward compatibility and also enhance security. For TLS the backword version carries non-empty ASPN, and for Noise early data is empty. And as the data is carried in variable size extension fields, I prefer to process the data only after the peers are authenticated to enhance security. Instead of introducing more edge cases and optimize for a transient phase, I prefer to simply fallback to multistream-selection in the current situation, if the negotiation is set to fail, it does not affect performace if it fails one RTT later. Open to hear other thoughts.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,266 @@
...
+
+
+### Muxer selection over TLS
```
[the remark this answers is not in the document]
julian88110: done

connections/muxer-sel-in-sec-handshake.md
julian88110: done

julian88110: done

connections/muxer-sel-in-sec-handshake.md
julian88110: done

connections/muxer-sel-in-sec-handshake.md
julian88110: Not sure I am following here. I am using the terminology found in the Noise spec https://github.com/libp2p/specs/tree/master/noise#xx, which refers to its messages by count. what is the 4th message here?

connections/muxer-sel-in-sec-handshake.md
julian88110: done

connections/muxer-sel-in-sec-handshake.md
julian88110: Please refer to the item 2 on previous comment for the reason of not prefering to fail.

connections/muxer-sel-in-sec-handshake.md
julian88110: done

connections/muxer-sel-in-sec-handshake.md
julian88110: removed the proto related section

connections/muxer-sel-in-sec-handshake.md
julian88110: relocated.

connections/muxer-sel-in-sec-handshake.md
julian88110: done

connections/muxer-sel-in-sec-handshake.md
julian88110: done

connections/muxer-sel-in-sec-handshake.md
julian88110: done

julian88110:
Thanks for putting up with my typos and imperfect English.  Hopefully I have corrected all of them.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,214 @@
...
+finished mutual authentication of the peers to enhance security.
+
+The format of the early data is specified in [Noise-handshake-payload]
+
+In the previous versions that do not support this feature, the Noise handshake
+messages carry an empty field of muxer list in the early data extension. When a
+version that supports this feature talks to an older version which does not
+support this feature, the multiplexer selection process on both peers will
+generate an empty list and they MUST fall back to the multistream-selection
+process to ensure backward compatibility.
```
marten-seemann, line 181: We already specified this before.
```suggestion
```

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,214 @@
...
+
+## Security
+
+The multiplexer list carried in the TLS ALPN extension field is part of the
+ClientHello message which is not encrypted. This feature will expose the
+supported multiplexers in plain text, but this is not a weakening of security
+posture. In the future when [ECH] is ready the multiplexer info can be protected
```
marten-seemann, line 188: It definitely is a weakening!
I don't think we should reference ECH here, as we don't have a deployment path for that protocol. Instead, point out that muxers are not really private information, since it's trivial to obtain that information by connecting to the node.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,214 @@
...
+The format of the early data is specified in [Noise-handshake-payload]
+
+In the previous versions that do not support this feature, the Noise handshake
+messages carry an empty field of muxer list in the early data extension. When a
+version that supports this feature talks to an older version which does not
+support this feature, the multiplexer selection process on both peers will
+generate an empty list and they MUST fall back to the multistream-selection
+process to ensure backward compatibility.
+
+## Security
```
marten-seemann, line 183: What you're discussing here is privacy, not security.

connections/muxer-sel-in-sec-handshake.md
```diff
@@ -0,0 +1,214 @@
...
+The multiplexer list carried in the TLS ALPN extension field is part of the
+ClientHello message which is not encrypted. This feature will expose the
+supported multiplexers in plain text, but this is not a weakening of security
+posture. In the future when [ECH] is ready the multiplexer info can be protected
+too.
+
+The early data in Noise handshake is only sent after the peers establish a
+shared key, in the second and third handshake messages in the XX pattern. So the
+early data is encrypted and the multiplexer info carried over is protected.
+There is no security weakening in this case either.
```
marten-seemann, line 194: Privacy, not security!

```suggestion
Early data in the Noise handshake is sent after the peers have established a
shared key, so an on-path observer won't be able to read the early data.
```

marten-seemann (changes requested):
Lots of comments. Most of them editorial, but there are some logic mistakes in here as well.
Furthermore, we should fail the handshake early, and not fall back to multistream negotiation.

marten-seemann:
I updated spec and condensed it down a bit. @mxinden and @MarcoPolo, can I ask you for a review?

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
+# Stream multiplexer negotiation in security handshake
```
p-shahi, line 1:
```suggestion
# Stream multiplexer negotiation in security handshake <!-- omit in toc -->
```

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
...
+[@marcopolo]: https://github.com/marcopolo
+[@mxinden]: https://github.com/mxinden
+[@julian88110]: https://github.com/julian88110
+
+See the [lifecycle document][lifecycle-spec] for context about maturity level
+and spec status.
+
+[lifecycle-spec]: https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md
+
+## Table of Contents
```
p-shahi, line 22:
```suggestion
## Table of Contents <!-- omit in toc -->
```

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
...
+- [multiplexer negotiation in security handshake](#multiplexer-negotiation-in-security-handshake)
+    - [Table of Contents](#table-of-contents)
+    - [Overview](#overview)
+    - [Design](#design)
+        - [Current connection upgrade process](#current-connection-upgrade-process)
+        - [Improved multiplexer negotiation](#improved-multiplexer-negotiation)
+            - [Multiplexer negotiation over TLS](#multiplexer-negotiation-over-tls)
+            - [Multiplexer negotiation over Noise](#multiplexer-negotiation-over-noise)
+    - [Security](#security)
+    - [Alternative options considered](#alternative-options-considered)
```
p-shahi, line 33:
```suggestion
- [Overview](#overview)
- [Design](#design)
  - [Multiplexer Negotiation over TLS](#multiplexer-negotiation-over-tls)
  - [Multiplexer Negotiation over Noise](#multiplexer-negotiation-over-noise)
- [Privacy](#privacy)
- [Alternative options considered](#alternative-options-considered)
```

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
...
+
+When using TLS, the [ALPN] extension is used to negotiate the multiplexer.
+
+The ALPN TLS extension allows the client to send a list of supported application
+protocols as part of the TLS ClientHello message.  The server chooses
+a protocol and sends the selected protocol as part of the TLS
+ServerHello message.
+
+For the purpose of multiplexer negotiation, the protocol IDs of the stream 
+multiplexers are sent, followed by "libp2p". The multiplexer list is ordered by
```
p-shahi, line 57:
```suggestion
multiplexers are sent, followed by the `"libp2p"` protocol code. The multiplexer list is ordered by
```

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
...
+the client's preference, with the most preferred multiplexer at the beginning.
+The server SHOULD respect the client's preference and pick the first protocol
+from the list that it supports.
+
+Example for a node supporting both yamux and mplex, with a preference for yamux:
+```json
+[ "/yamux/1.0.0", "/mplex/6.7.0", "libp2p" ]
+```
+
+The "libp2p" protocol code MUST always be the last item in the multiplexer list.
```
p-shahi, line 67:
```suggestion
The `"libp2p"` protocol code MUST always be the last item in the multiplexer list.
```

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
...
+According to [tls], nodes that don't implement the optimization described in
```
p-shahi, line 68:
```suggestion
According to [TLS], nodes that don't implement the optimization described in
```

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
...
+this document use "libp2p" for their ALPN. If "libp2p" is the result of the
```
p-shahi, line 69:
```suggestion
this document use `"libp2p"` for their ALPN. If `"libp2p"` is the result of the
```

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
...
+
+
+## Alternative options considered
+
+Instead of ALPN for multiplexer selection to reduce RTT, other options such as
+TLS extension and X.509 extension were considered. The pros and cons are explored
+and the discussion details can be found at [#454].
+
+
+
```
p-shahi, line 125:
```suggestion
[TLS]: https://github.com/libp2p/specs/blob/master/tls/tls.md
```

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
...
+ALPN process, nodes MUST use multistream negotiation of the stream multiplexer
+as described in [connections].
+
+### Multiplexer Negotiation over Noise
+
+The libp2p Noise Specification allows Noise handshake messages to carry
+early data. [Noise-Early-Data] is carried in the second and third message of
+the XX handshake pattern as illustrated in the following message sequence chart.
+The second message carries early data in the form of a list of multiplexers
+supported by the responder.The initiator sends its supported multiplexer list
```
p-shahi, line 79:
```suggestion
supported by the responder. The initiator sends its supported multiplexer list 
```

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
...
+The second message carries early data in the form of a list of multiplexers
+supported by the responder.The initiator sends its supported multiplexer list 
+in the third message of the handshake process, ordered by its preference. It
+MAY choose a single multiplexer from the responder's list and only send that
+value.
+
+The multiplexer to use is determined by picking the first item from the
+initiator's list that both parties support.
+
+Example: Noise handshake two clients that both support mplex and yamux. The
```
p-shahi, line 87:
```suggestion
Example: Noise handshake between two clients that both support mplex and yamux. The
```

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
...
+client prefers yamux, whereas the server prefers mplex. 
+
+```
+XX:
+-> e
+<- e, ee, s, es, [ "/mplex/6.7.0", "/yamux/1.0.0" ] 
+-> s, se, [ "/yamux/1.0.0", "/mplex/6.7.0" ] 
+```
+
+The result of this negotiation is "/mplex/6.7.0".
```
p-shahi, line 97:
```suggestion
The result of this negotiation is `"/mplex/6.7.0"`.
```

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
...
+If there is no overlap between the multiplexers support by client and server,
+the handshake MUST fail.
+
+The format of the early data is specified in [Noise-handshake-payload].
+
+
+## Privacy
+
+The list of multiplexers carried in the TLS ALPN extension field is part of the
+ClientHello message which is not encrypted. Using this optimiziation therefore
```
p-shahi, line 108:
```suggestion
unencrypted ClientHello message. Therefore, using this optimization
```

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
...
+exposes the list of supported multiplexers to an on-path observer. This leak can
+be considered insignificant, since a libp2p node reveals its list of supported
+multiplexers to any node that connects to it.
+
+The NoiseExtensions sent in the Noise handshake is sent after the peers have
```
p-shahi, line 113:
```suggestion
However, the NoiseExtensions in the Noise handshake are sent after the peers have 
```

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
...
+
+
+
+[#426]: https://github.com/libp2p/specs/issues/426
+[connections]: https://github.com/libp2p/specs/tree/master/connections
+[sequence-chart]: https://github.com/libp2p/specs/tree/master/connections#upgrading-connections
+[ALPN]: https://datatracker.ietf.org/doc/html/rfc7301
+[Noise-Early-Data]: https://github.com/libp2p/specs/tree/master/noise#the-libp2p-handshake-payload
+[ECH]: https://datatracker.ietf.org/doc/draft-ietf-tls-esni/
+[handshake-payload]: https://github.com/libp2p/specs/tree/master/noise#the-libp2p-handshake-payload
```
p-shahi, line 132:
```suggestion
```
not referenced anywhere?

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Alternative options considered
+
+Instead of ALPN for multiplexer selection to reduce RTT, other options such as
+TLS extension and X.509 extension were considered. The pros and cons are explored
+and the discussion details can be found at [#454].
+
+
+
+[#426]: https://github.com/libp2p/specs/issues/426
```
p-shahi, line 126:
```suggestion
```
not referenced anywhere?

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
...
+[connections]: https://github.com/libp2p/specs/tree/master/connections
+[sequence-chart]: https://github.com/libp2p/specs/tree/master/connections#upgrading-connections
```
p-shahi, line 128:
```suggestion
```
not referenced anywhere?

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
...
+[ALPN]: https://datatracker.ietf.org/doc/html/rfc7301
+[Noise-Early-Data]: https://github.com/libp2p/specs/tree/master/noise#the-libp2p-handshake-payload
```
p-shahi, line 130:
```suggestion
[Noise-Early-Data]: https://github.com/libp2p/specs/tree/master/noise#libp2p-data-in-handshake-messages
```
early data is described more in this section

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,135 @@
...
+[ECH]: https://datatracker.ietf.org/doc/draft-ietf-tls-esni/
+[handshake-payload]: https://github.com/libp2p/specs/tree/master/noise#the-libp2p-handshake-payload
+[#454]: https://github.com/libp2p/specs/issues/454
+[Noise-handshake-payload]: https://github.com/libp2p/specs/blob/b0818fa956f9940a7cdee18198e0daf1645d8276/noise/README.md#libp2p-data-in-handshake-messages
```
p-shahi, line 134:
```suggestion
[Noise-handshake-payload]: https://github.com/libp2p/specs/tree/master/noise#the-libp2p-handshake-payload
```
corrected link and referenced master

noise/README.md
```diff
@@ -220,6 +220,7 @@ syntax = "proto2";
 
 message NoiseExtensions {
     repeated bytes webtransport_certhashes = 1;
+    repeated string stream_muxers = 2;
```
p-shahi, line 223: increase the revision of this spec?

connections/README.md
```diff
@@ -234,6 +234,11 @@ Note: In the case where both peers initially act as initiators, e.g. during NAT
 hole punching, tie-breaking is done via the [multistream-select simultaneous
 open protocol extension][simopen].
 
+### Inlining Muxer Negotiation
```
p-shahi, line 237: increase the revision of this spec?

p-shahi (commented):
mostly nits

connections/README.md
marten-seemann: Done.

noise/README.md
marten-seemann: Done.

p-shahi (approved):

elenaf9 (commented):
Couple of questions. Apologies if they have already been discussed or if the answers are obvious. Parts of this are still new to me.

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,128 @@
...
+from the list that it supports.
+
+Example for a node supporting both yamux and mplex, with a preference for yamux:
+```json
+[ "/yamux/1.0.0", "/mplex/6.7.0", "libp2p" ]
+```
+
+The `"libp2p"` protocol code MUST always be the last item in the multiplexer list.
+According to [TLS], nodes that don't implement the optimization described in
+this document use `"libp2p"` for their ALPN. If `"libp2p"` is the result of the
```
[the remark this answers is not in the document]
marten-seemann: Yes, you could accept HTTP and libp2p at the same time. That's WebTransport does, by the way.
Negotiating HTTP though means that after completing the handshake, you're in HTTP land, not in libp2p land any more.

connections/inlined-muxer-negotiation.md
```diff
@@ -0,0 +1,128 @@
...
+ALPN process, nodes MUST use multistream negotiation of the stream multiplexer
+as described in [connections].
```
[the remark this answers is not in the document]
marten-seemann: This is to enable backwards-compatibility with legacy implementations. With these nodes, the result of the negotiation process will be "libp2p", so we need to do multistream.
Now we could say that we fail the handshake if there's anything other than "libp2p" in the list, but this would make it impossible to deploy a listener that speaks HTTP and libp2p without this optimization.

connections/inlined-muxer-negotiation.md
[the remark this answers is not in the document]
mxinden: > Unless I am missing something this is not true: the TLS spec does not specify what ALPN to use.

Good catch. Both Go and Rust do this. @marten-seemann can you add it to the TLS spec? Either in this pull request or a separate one?

mxinden (approved):
Apart from the comment above on documenting the default `libp2p` ALPN, this specification is good to merge from my end.

Thanks again for the document :pray: 

(Corresponding tracking issue in rust-libp2p: https://github.com/libp2p/rust-libp2p/issues/2994)

PR #451, opened 2022-09-14
New values for DHT Provider Record Republish and Expiration (22h/48h, RFM17)

yiannisbot:
This PR updates the description of the Provider Record settings and most importantly proposes new values for both the republish interval and the expiration interval. The new proposed values are:

- republish interval to be set to 24hrs, from its current 12hrs
- expiration interval to be set to 48hrs, from its current 

and they are based on the comprehensive study published here: https://github.com/protocol/network-measurements/blob/master/results/rfm17-provider-record-liveness.md

### DIFF
--- a/kad-dht/README.md
+++ b/kad-dht/README.md
@@ -279,9 +279,9 @@ heuristic of the value to make the decision.
 There are two things at play with regard to provider record (and therefore content)
 liveness and reachability:
 
-Content providers need to make sure that their content is reachable, despite peer churn;
-and nodes that store and serve provider records need to make sure that the Multihashes whose 
-records they store are still served by the content provider.
+Content needs to be reachable, despite peer churn;
+and nodes that store and serve provider records should not serve records for stale content,
+i.e., content that the original provider does not wish to make available anymore.
 
 The following two parameters help cover both of these cases.
 1. **Provider Record Republish Interval:** The content provider 
--- a/kad-dht/README.md
+++ b/kad-dht/README.md
@@ -2,7 +2,7 @@
 
 | Lifecycle Stage | Maturity       | Status | Latest Revision |
 |-----------------|----------------|--------|-----------------|
-| 3A              | Recommendation | Active | r1, 2021-10-30  |
+| 3A              | Recommendation | Active | r2, 2022-12-09  |
 
 Authors: [@raulk], [@jhiesey], [@mxinden]
 
@@ -284,21 +284,26 @@ and nodes that store and serve provider records should not serve records for sta
 i.e., content that the original provider does not wish to make available anymore.
 
 The following two parameters help cover both of these cases.
+
 1. **Provider Record Republish Interval:** The content provider 
 needs to make sure that the nodes chosen to store the provider record 
 are still online when clients ask for the record. In order to 
 guarantee this, while taking into account the peer churn, content providers
 republish the records they want to provide. Choosing the particular value for the
 Republish interval is network-specific and depends on several parameters, such as
-peer reliability and churn. For the IPFS network it is currently set to 22 hours.
+peer reliability and churn.
+
+   - For the IPFS network it is currently set to **22 hours**.
+
 2. **Provider Record Expiration Interval:** The network needs to provide
 content that content providers are still interested in providing. In other words,
 nodes should not keep records for content that content providers have stopped 
 providing (aka stale records). In order to guarantee this, provider records 
 should _expire_ after some interval, i.e., nodes should stop serving those records, 
 unless the content provider has republished the provider record. Again, the specific
-setting depends on the characteristics of the network. In the IPFS DHT the Expiration 
-Interval is set to 48hrs.
+setting depends on the characteristics of the network.
+
+   - In the IPFS DHT the Expiration Interval is set to **48 hours**.
 
 The values chosen for those parameters should be subject to continuous monitoring 
 and investigation. Ultimately, the values of those parameters should balance 
--- a/kad-dht/README.md
+++ b/kad-dht/README.md
@@ -75,12 +75,12 @@ nodes, unrestricted nodes should operate in _server mode_ and restricted nodes,
 e.g. those with intermittent availability, high latency, low bandwidth, low
 CPU/RAM/Storage, etc., should operate in _client mode_.
 
-As an example, running the libp2p Kademlia protocol on top of the Internet,
-publicly routable nodes, e.g. servers in a datacenter, might operate in _server
+As an example, running the libp2p Kademlia protocol on top of
+publicly routable nodes, e.g. servers in a datacenter, should operate in _server
 mode_ and non-publicly routable nodes, e.g. laptops behind a NAT and firewall,
-might operate in _client mode_. The concrete factors used to classify nodes into
+should operate in _client mode_. The concrete factors used to classify nodes into
 _clients_ and _servers_ depend on the characteristics of the network topology
-and the properties of the Kademlia DHT . Factors to take into account are e.g.
+and the properties of the Kademlia DHT. Factors to take into account are e.g.
 network size, replication factor and republishing period.
 
 Nodes, both those operating in _client_ and _server mode_, add another node to
@@ -228,7 +228,7 @@ Then we loop:
                becomes the new best peer (`Pb`).
 			2. If the new value loses, we add the current peer to `Po`.
 	2. If successful with or without a value, the response will contain the
-       closest nodes the peer knows to the key `Key`. Add them to the candidate
+       closest nodes the peer knows to the `Key`. Add them to the candidate
        list `Pn`, except for those that have already been queried.
 	3. If an error or timeout occurs, discard it.
 4. Go to 1.
@@ -256,7 +256,7 @@ type Validator interface {
 ```
 
 `Validate()` should be a pure function that reports the validity of a record. It
-may validate a cryptographic signature, or else. It is called on two occasions:
+may validate a cryptographic signature, or similar. It is called on two occasions:
 
 1. To validate values retrieved in a `GET_VALUE` query.
 2. To validate values received in a `PUT_VALUE` query before storing them in the
@@ -268,23 +268,55 @@ heuristic of the value to make the decision.
 
 ### Content provider advertisement and discovery
 
-Nodes must keep track of which nodes advertise that they provide a given key
-(CID). These provider advertisements should expire, by default, after 24 hours.
-These records are managed through the `ADD_PROVIDER` and `GET_PROVIDERS`
+There are two things at play with regard to provider record (and therefore content)
+liveness and reachability:
+
+Content providers need to make sure that their content is reachable, despite peer churn
+and nodes that store and serve provider records need to make sure that the CIDs whose 
+records they store are still served by the content provider.
+
+The following two parameters help cover both of these cases.
+1. **Provider Record Republish Interval (24hrs):** The content provider 
+needs to make sure that the nodes chosen to store the provider record 
+remain online when clients ask for the record. In order to 
+guarantee this, while taking into account the peer churn, content providers
+republish the records they want to provide every 24 hours.
+2. **Provider Record Expiration Interval (48hrs):** The network needs to provide
+content that content providers are still interested in providing. In other words,
+nodes should not keep records for content that content providers have stopped 
+providing (aka stale records). In order to guarantee this, provider records 
+_expire_ after 48 hours, i.e., nodes stop serving those records, 
+unless the content provider has republished the provider record.
+
+The values chosen for those parameters should be subject to continuous monitoring 
+and investigation. Ultimately, the values of those parameters should balance 
+the tradeoff between provider record liveness (due to node churn) and traffic overhead
+(to republish records).
+The latest parameters are based on the comprehensive study published
+in [provider-record-measurements].
+
+Provider records are managed through the `ADD_PROVIDER` and `GET_PROVIDERS`
 messages.
 
 #### Content provider advertisement
 
 When the local node wants to indicate that it provides the value for a given
-key, the DHT finds the closest peers to the key using the `FIND_NODE` RPC (see
+key, the DHT finds the (`k` = 20) closest peers to the key using the `FIND_NODE` RPC (see
 [peer routing section](#peer-routing)), and then sends an `ADD_PROVIDER` RPC with
-its own `PeerInfo` to each of these peers.
+its own `PeerInfo` to each of these peers. The study in [provider-record-measurements]
+proved that the replication factor of `k` = 20 is a good setting, although continuous
+monitoring and investigation.
 
 Each peer that receives the `ADD_PROVIDER` RPC should validate that the received
 `PeerInfo` matches the sender's `peerID`, and if it does, that peer should store
 the `PeerInfo` in its datastore. Implementations may choose to not store the
 addresses of the providing peer e.g. to reduce the amount of required storage or
-to prevent storing potentially outdated address information.
+to prevent storing potentially outdated address information. In the current implementation
+peers keep the network address (i.e., the `multiaddress`) of the providing peer for **the
+first 10 mins** after the provider record (re-)publication. The setting of 10 mins follows
+the DHT Routing Table refresh interval. After that, peers provide 
+the provider's `peerID` only, in order to avoid pointing to stale network addresses 
+(i.e., the case where the peer has moved to a new network address).
 
 #### Content provider discovery
 
@@ -470,3 +502,5 @@ multiaddrs are stored in the node's peerbook.
 [ping]: https://github.com/libp2p/specs/issues/183
 
 [go-libp2p-xor]: https://github.com/libp2p/go-libp2p-xor
+
+[provider-record-measurements]: https://github.com/protocol/network-measurements/blob/master/results/rfm17-provider-record-liveness.md
--- a/kad-dht/README.md
+++ b/kad-dht/README.md
@@ -75,13 +75,21 @@ nodes, unrestricted nodes should operate in _server mode_ and restricted nodes,
 e.g. those with intermittent availability, high latency, low bandwidth, low
 CPU/RAM/Storage, etc., should operate in _client mode_.
 
-As an example, running the libp2p Kademlia protocol on top of
-publicly routable nodes, e.g. servers in a datacenter, should operate in _server
+As an example, publicly routable nodes running the libp2p Kademlia protocol, 
+e.g. servers in a datacenter, should operate in _server
 mode_ and non-publicly routable nodes, e.g. laptops behind a NAT and firewall,
 should operate in _client mode_. The concrete factors used to classify nodes into
 _clients_ and _servers_ depend on the characteristics of the network topology
 and the properties of the Kademlia DHT. Factors to take into account are e.g.
 network size, replication factor and republishing period.
+
+For instance, setting the replication factor to a low value would require more
+reliable peers, whereas having higher replication factor could allow for less 
+reliable peers at the cost of more overhead. Ultimately, peers that act as 
+servers should help the network (i.e., provide positive utility in terms of 
+availability, reachability, bandwidth). Any factor that slows down network
+operations (e.g., a node not being reachable, or overloaded) for the majority
+of times it is being contacted should instead be operating as a client node.
 
 Nodes, both those operating in _client_ and _server mode_, add another node to
 their routing table if and only if that node operates in _server mode_. This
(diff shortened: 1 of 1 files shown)


kad-dht/README.md
```diff
@@ -75,12 +75,12 @@ nodes, unrestricted nodes should operate in _server mode_ and restricted nodes,
...
-publicly routable nodes, e.g. servers in a datacenter, might operate in _server
+As an example, running the libp2p Kademlia protocol on top of
+publicly routable nodes, e.g. servers in a datacenter, should operate in _server
 mode_ and non-publicly routable nodes, e.g. laptops behind a NAT and firewall,
-might operate in _client mode_. The concrete factors used to classify nodes into
+should operate in _client mode_. The concrete factors used to classify nodes into
 _clients_ and _servers_ depend on the characteristics of the network topology
-and the properties of the Kademlia DHT . Factors to take into account are e.g.
+and the properties of the Kademlia DHT. Factors to take into account are e.g.
 network size, replication factor and republishing period.
```
yiannisbot, line 84: AFAIK, this is not accurate, i.e., only connectivity is taken into account currently and furthermore, it's not clear how network size and replication factor (of what? the `k` parameter) can affect the decision between a node being a client and a server. Why not just include a description and pointer to AuthoNAT? If we want to include the extra items, it will be helpful to provide recommendations as to how these other factors can be useful.

kad-dht/README.md
```diff
@@ -268,23 +268,55 @@ heuristic of the value to make the decision.
...
 #### Content provider advertisement
 
 When the local node wants to indicate that it provides the value for a given
-key, the DHT finds the closest peers to the key using the `FIND_NODE` RPC (see
+key, the DHT finds the (`k` = 20) closest peers to the key using the `FIND_NODE` RPC (see
 [peer routing section](#peer-routing)), and then sends an `ADD_PROVIDER` RPC with
-its own `PeerInfo` to each of these peers.
+its own `PeerInfo` to each of these peers. The study in [provider-record-measurements]
+proved that the replication factor of `k` = 20 is a good setting, although continuous
+monitoring and investigation.
```
lidel, line 308:
```suggestion
monitoring and investigation may change this recommendation in the future.
```

lidel (commented):
Bumping Republish and Expiration sounds sensible (["reduce the overhead without interfering with the performance and reliability"](https://github.com/protocol/network-measurements/blob/master/results/rfm17-provider-record-liveness.md#5-conclusion)). 

@mxinden any concerns from non-IPFS side of things?

kad-dht/README.md
```diff
@@ -268,23 +268,55 @@ heuristic of the value to make the decision.
...
 
-Nodes must keep track of which nodes advertise that they provide a given key
-(CID). These provider advertisements should expire, by default, after 24 hours.
-These records are managed through the `ADD_PROVIDER` and `GET_PROVIDERS`
+There are two things at play with regard to provider record (and therefore content)
+liveness and reachability:
+
+Content providers need to make sure that their content is reachable, despite peer churn
+and nodes that store and serve provider records need to make sure that the CIDs whose 
+records they store are still served by the content provider.
```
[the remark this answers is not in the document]
Winterhuman: @lidel Not everywhere does, Bitswap apparently doesn't do this yet: https://github.com/ipfs/kubo/issues/4189#issuecomment-1201802260

kad-dht/README.md
```diff
@@ -75,12 +75,12 @@ nodes, unrestricted nodes should operate in _server mode_ and restricted nodes,
 e.g. those with intermittent availability, high latency, low bandwidth, low
 CPU/RAM/Storage, etc., should operate in _client mode_.
 
-As an example, running the libp2p Kademlia protocol on top of the Internet,
-publicly routable nodes, e.g. servers in a datacenter, might operate in _server
+As an example, running the libp2p Kademlia protocol on top of
+publicly routable nodes, e.g. servers in a datacenter, should operate in _server
```
mxinden, line 79: > running the libp2p Kademlia protocol on top of publicly routable nodes [...] should operate in _server mode_

Is this a valid sentence?

kad-dht/README.md
mxinden: I don't have a strong opinion here. Background: https://github.com/libp2p/specs/pull/392#discussion_r804946667

Maybe @aschmahmann does.

kad-dht/README.md
```diff
@@ -268,23 +268,55 @@ heuristic of the value to make the decision.
...
 ### Content provider advertisement and discovery
 
-Nodes must keep track of which nodes advertise that they provide a given key
-(CID). These provider advertisements should expire, by default, after 24 hours.
-These records are managed through the `ADD_PROVIDER` and `GET_PROVIDERS`
+There are two things at play with regard to provider record (and therefore content)
+liveness and reachability:
+
+Content providers need to make sure that their content is reachable, despite peer churn
+and nodes that store and serve provider records need to make sure that the Multihashes whose
```
mxinden, line 275:
```suggestion
and nodes that store and serve provider records need to make sure that the CID whose 
```

Isn't this a CID?

kad-dht/README.md
```diff
@@ -268,23 +268,55 @@ heuristic of the value to make the decision.
...
 `PeerInfo` matches the sender's `peerID`, and if it does, that peer should store
 the `PeerInfo` in its datastore. Implementations may choose to not store the
 addresses of the providing peer e.g. to reduce the amount of required storage or
-to prevent storing potentially outdated address information.
+to prevent storing potentially outdated address information. Implementations that choose
+to keep the network address (i.e., the `multiaddress`) of the providing peer should do it for **the
+first 10 mins** after the provider record (re-)publication. The setting of 10 mins follows
+the DHT Routing Table refresh interval. After that, peers provide 
+the provider's `peerID` only, in order to avoid pointing to stale network addresses 
+(i.e., the case where the peer has moved to a new network address).
```
mxinden, line 319: If I recall correctly, this is the status quo in the Golang implementation, correct? Is there any data backing up this decision? What I am surprised by is that addresses go stale so quickly. Is that really the case on IPFS today?

kad-dht/README.md
```diff
@@ -268,23 +268,55 @@ heuristic of the value to make the decision.
...
+and nodes that store and serve provider records need to make sure that the Multihashes whose 
+records they store are still served by the content provider.
+
+The following two parameters help cover both of these cases.
+1. **Provider Record Republish Interval (24hrs):** The content provider 
+needs to make sure that the nodes chosen to store the provider record 
+remain online when clients ask for the record. In order to 
+guarantee this, while taking into account the peer churn, content providers
+republish the records they want to provide every 24 hours.
+2. **Provider Record Expiration Interval (48hrs):** The network needs to provide
```
mxinden, line 284: Is there already a corresponding go-libp2p pull request to update these values?

mxinden (commented):
Wonderful to see these optimizations based on comprehensive studies!

Can you bump the revision of the specification at the top of the document?

kad-dht/README.md
```diff
@@ -268,23 +268,55 @@ heuristic of the value to make the decision.
...
 
-Nodes must keep track of which nodes advertise that they provide a given key
-(CID). These provider advertisements should expire, by default, after 24 hours.
-These records are managed through the `ADD_PROVIDER` and `GET_PROVIDERS`
+There are two things at play with regard to provider record (and therefore content)
+liveness and reachability:
+
+Content providers need to make sure that their content is reachable, despite peer churn
+and nodes that store and serve provider records need to make sure that the Multihashes whose 
+records they store are still served by the content provider.
```
mxinden, line 276: How do they do this? Is this happening today on the IPFS DHT?

kad-dht/README.md
aschmahmann: I'm pretty sure what @yiannisbot is effectively implying is that we almost certainly need different sections or even documents describing the IPFS Public DHTs implementation of this DHT with all it's parameters and associated recommendations.

Whether a node chooses to operate in client or server mode is a client decision, and whether servers should accept certain nodes into their routing tables is a network decision. When clients opt into being servers hopefully they do so with the intention of "helping" the network, but that could mean a number of things such as how much capacity (storage and bandwidth, which relates to network size, replication factor, and the type of load) the node has to give to the network, how available they are, and whether the network would currently benefit from unreliable servers (i.e. the replication factor is too high) or really needs reliable servers (i.e. replication factor is about right or too low). There's probably a lot that could go into this, I don't think we need to throw everything in here but giving some starting points for network developers to start from seems reasonable to me 🤷.

kad-dht/README.md
aschmahmann: the keys for provider records are multihashes (e.g. https://github.com/libp2p/go-libp2p-kad-dht/blob/dae5a9a5bd9c7cc8cfb5073c711bc308efad0ada/routing.go#L431)

TLDR on why:
- You are using provider records as a rendezvous point for all the parties who have advertised they have some data
- The same multihash can be in different CIDs (e.g. CIDv0 vs CIDv1 of a SHA-256 dag-pb object, or the same multihash but with different codecs such as dag-pb vs raw)
- This means you want to rendezvous on the minimal thing everyone agrees on, which is the multihash

kad-dht/README.md
```diff
@@ -268,23 +268,55 @@ heuristic of the value to make the decision.
...
+
+The following two parameters help cover both of these cases.
+1. **Provider Record Republish Interval (24hrs):** The content provider
```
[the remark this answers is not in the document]
yiannisbot: Good question. The results show that even with 36, or 40 things would still be ok, but I guess a shorter interval is a bit more conservative to make sure content is not disappearing. Then, the plan is that with continuous measurements we can get to a larger value. @cortze any more insights?

kad-dht/README.md
```diff
@@ -268,23 +268,55 @@ heuristic of the value to make the decision.
...
-(CID). These provider advertisements should expire, by default, after 24 hours.
-These records are managed through the `ADD_PROVIDER` and `GET_PROVIDERS`
+There are two things at play with regard to provider record (and therefore content)
+liveness and reachability:
+
+Content providers need to make sure that their content is reachable, despite peer churn
+and nodes that store and serve provider records need to make sure that the Multihashes whose 
+records they store are still served by the content provider.
+
+The following two parameters help cover both of these cases.
```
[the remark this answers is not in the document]
yiannisbot: I totally agree. I would love to have a list of the fixed/hardcoded parameters together with explanation that are easy to find. Not sure what's the best place for those to live in (here or in IPFS land), but would helpful to have somewhere :)

kad-dht/README.md
yiannisbot: Do you mean that it's not globally true, or that it doesn't make sense? 😁 

Do you have a better phrasing?

kad-dht/README.md
yiannisbot: Very useful TL;DR - perhaps even worth mentioning in the spec itself I would suggest? If not, we can resolve the comment.

kad-dht/README.md
yiannisbot: > this is the status quo in the Golang implementation, correct?
Yup.

> Is there any data backing up this decision?
Nope :) But plan to start some investigation asap. See: https://github.com/ipfs/kubo/issues/9264, https://github.com/protocol/prodeng/issues/22 and https://github.com/ipfs-shipyard/thunderdome/issues/91 as an optimisation.

> What I am surprised by is that addresses go stale so quickly. Is that really the case on IPFS today?

Given that IPFS DHT servers have public addresses, I doubt that they go stale so quickly. This might change when hole punching is widespread, but the optimisation proposed in the above issues won't hurt, I believe :)

kad-dht/README.md
yiannisbot: Not yet, AFAIK. The intention is to have the discussion in this spec repo and whatever is agreed here will be transferred to a PR there. That's my understanding at least.

kad-dht/README.md
yiannisbot: This is done (somewhat indirectly) through the expiration of the provider record: if the content provider does not republish the record (within the republish interval), then nodes do not serve those provider records after the expiration interval (assuming that the content provider is not interested having this content live anymore).

kad-dht/README.md
```diff
@@ -268,23 +268,55 @@ heuristic of the value to make the decision.
...
+1. **Provider Record Republish Interval (24hrs):** The content provider 
+needs to make sure that the nodes chosen to store the provider record 
+remain online when clients ask for the record. In order to 
+guarantee this, while taking into account the peer churn, content providers
+republish the records they want to provide every 24 hours.
```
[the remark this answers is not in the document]
yiannisbot: Right, so you suggest not having the _new_ republish interval the same as the _old_ expiration interval as this will become confusing and might have side effects as well? I guess a valid workaround is to have the republish interval set to something a bit smaller (say, 20hrs?) for the transition period? Any better approaches?

kad-dht/README.md
yiannisbot: Yup, sorry not making this clearer (here and elsewhere in this PR:). I agree we don't need to include everything here, but: i) it would be nice to have brief (a sentence or two) recommendations, or guides for developers, and ii) provide more details for the IPFS DHT.

kad-dht/README.md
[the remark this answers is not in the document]
cortze: > a shorter interval is a bit more conservative to make sure content is not disappearing

Exactly! In this context of suggesting upgrades to the live IPFS network, we wanted to ensure that our recommendations were inside the conservative margins we constructed from the study.

kad-dht/README.md
mxinden: > or that it doesn't make sense?

To me it does not make sense. How about:

> As an example, publicly routable nodes running the libp2p Kademlia protocol, e.g. servers in a datacenter, should operate in _server

kad-dht/README.md
mxinden: Given that we are not sure this optimization is a good idea, how about only documenting it in this specification once we know it is a good idea?

Otherwise new implementations like Iroh (//CC @dignifiedquire) would have to implement this despite not being an optimization in the first place.

kad-dht/README.md
mxinden: Sounds good. I suggest we only merge here once that pull request exists to make sure at least one implementation is up to date with the specification.

kad-dht/README.md
mxinden: I think the above phrasing implies this being an active process on the provider record storage node. What do you think of rephrasing this?

kad-dht/README.md
yiannisbot: I've added the following for now. I think a separate document to say how the IPFS DHT is configured/set up would be helpful, but that's for another PR (and perhaps for a different repository?).

"For instance, setting the replication factor to a low value would require more
reliable peers, whereas having higher replication factor could allow for less 
reliable peers at the cost of more overhead. Ultimately, peers that act as 
servers should help the network (i.e., provide positive utility in terms of 
availability, reachability, bandwidth). Any factor that slows down network
operations (e.g., a node not being reachable, or overloaded) for the majority
of times it is being contacted should instead be operating as a client node."

kad-dht/README.md
yiannisbot: I've added the bulletpoints at the end of that section as a clarification.

kad-dht/README.md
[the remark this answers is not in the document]
yiannisbot: I've rephrased the text for the Republish and Expiration interval to be more generic, but mentioned the specific values we choose for the IPFS DHT. I hope that's a better wording for this spec.

kad-dht/README.md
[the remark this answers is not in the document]
yiannisbot: Leaving it to that for now, more measurement will show if we need to reconsider.

kad-dht/README.md
yiannisbot: In fact, we've figured out that this has changed to 30mins: https://github.com/libp2p/go-libp2p/commit/c2821791bac7d638890accc98798bf4cbfe122e7 - I've rephrased the text accordingly to make it more generic and mention this value for the kubo implementation.

There's also this: https://github.com/ipfs/kubo/issues/9264 - any views more than welcome! :)

kad-dht/README.md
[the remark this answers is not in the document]
yiannisbot: I've set this to 22hrs. @mxinden @aschmahmann let me know if that works for you.

kad-dht/README.md
yiannisbot: I've rephrased to the following - let me know if it reads better:

"Content needs to be reachable, despite peer churn;
and nodes that store and serve provider records should not serve records for stale content,
i.e., content that the original provider does not wish to make available anymore."

kad-dht/README.md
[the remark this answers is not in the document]
aschmahmann: @yiannisbot while we're changing the times any reason not to do this too? Seems like it'd be reasonable since expiration is based on the same properties. It'd also make things easier to reason about.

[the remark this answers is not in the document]
yiannisbot: You mean in the PR? Yes, the intention is to change both the republish and the expiration interval. Otherwise it would make no sense (or well, it would be confusing). Or do you mean something else?

cc: @cortze who is working to submit the relevant PR.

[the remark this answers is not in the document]
cortze: I agree with changing them together. However, unless I am missing something in the code, they will have to be two separate PRs:
1. Update `kubo`'s default `Reprovider.Interval` [here](https://github.com/cortze/kubo/blob/0e0136100ed8aaea6c27c89c9b73471f542d8c1d/config/init.go#L78)
2. Update `go-libp2p-kad-dht`'s `ProvideValidity` [here](https://github.com/cortze/go-libp2p-kad-dht/blob/6a75b5910b5d6a60e17d6d69524d11783a57ad6b/providers/providers_manager.go#L29)

yiannisbot: Here it is: https://github.com/ipfs/kubo/pull/9326 :)

cortze:
Here is the second PR [libp2p/go-libp2p-kad-dht#793](https://github.com/libp2p/go-libp2p-kad-dht/pull/793) to increase the expiration time of the PRs from `24h` to `48h`.

kad-dht/README.md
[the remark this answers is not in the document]
aschmahmann: > You mean in the PR

No, I mean to change the expiration time for `PUT_VALUE` records in addition to `ADD_PROVIDER` records.

i.e. for the IPFS Public DHT the expiration time for `PUT_VALUE` (i.e. IPNS and the deprecated public key records) is 36hrs
https://github.com/libp2p/go-libp2p-kad-dht/blob/dae5a9a5bd9c7cc8cfb5073c711bc308efad0ada/internal/config/config.go#L117. It seems like this could be 48hrs as well rather than 36.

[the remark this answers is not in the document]
yiannisbot: @cortze should we change that too then? It makes sense. Do you want to create the PR and set it to 48hrs?

lidel (approved):
@mxinden @marten-seemann bumped revision in the header, mind giving this final review?

Kubo team wants to ship  and  with next Kubo 0.18 release, but want to make sure we have specs merged first

- [x] reprovide interval in kubo is merged: https://github.com/ipfs/kubo/pull/9326
- [x] expiration interval in  https://github.com/libp2p/go-libp2p-kad-dht/pull/793 and https://github.com/libp2p/go-libp2p-kad-dht/pull/794  is waiting for review/release 
  - @marten-seemann  lmk if it is ok for me to do it
- [x] this spec is in sync with the two above

fwiw I've updated revision as requested, and made it easier to find/eyeball both numbers:

>

Stebalien:
[Heil Hydra](https://github.com/libp2p/hydra-booster/), I guess. If the records are staying around that long, we should do this.

lidel:
There is also a general DX/UX improvement that comes with raising the ceiling of expiration.

Trying to keep IPNS website alive with spotty internet access is tricky. 
It is not hard to imagine a situation when the publisher of an IPNS record can't be online exactly every 24h.
Expiration set to 48h makes a better default ("publish once a day, exact hour does not matter").

--- 

In effort to include this in Kubo 0.18.0-rc1 before holidays, I've released [go-libp2p-kad-dht v0.20.0](https://github.com/libp2p/go-libp2p-kad-dht/releases/tag/v0.20.0) with 48h expiration. 
Kubo 0.18 will also have the default Republish Interval set to 22h (https://github.com/ipfs/kubo/issues/9389), being fully compliant with the spec from this PR.

@mxinden  
1.  should I open PR to adjust defaults in [rust-libp2p/protocols/kad/src/behaviour.rs](https://github.com/libp2p/rust-libp2p/blob/4fe518de86ecb0805421cca6e291db168c10eac1/protocols/kad/src/behaviour.rs#L182-L196), or are there other places?
1. ok to merge this PR?

mxinden (approved):
Thanks to everyone involved here.

mxinden:
> @mxinden
> 
>     1. should I open PR to adjust defaults in [rust-libp2p/protocols/kad/src/behaviour.rs](https://github.com/libp2p/rust-libp2p/blob/4fe518de86ecb0805421cca6e291db168c10eac1/protocols/kad/src/behaviour.rs#L182-L196), or are there other places?
> 
>     2. ok to merge this PR?

Tracked here https://github.com/libp2p/rust-libp2p/issues/3229. Contributions are always welcome. That said, not a requirement to move forward here. Thanks for stewarding this @lidel.

PR #452, opened 2022-09-14
noise: remove interface definition

marten-seemann:
The interface exposed by a libp2p Noise implementation is highly implementation-specific. Really, there’s no reason to even suggest something like `SecureInbound` and `SecureOutbound`, there are other ways to implement (and name) this that are equally conformant with this spec and interoperable with other implementations.

More importantly, the references made to early data in this section directly contradict what we’re planning to do with early data for WebTransport (#404), WebRTC (#412) and muxer negotiation (#446).

### DIFF
--- a/noise/README.md
+++ b/noise/README.md
@@ -52,12 +52,6 @@ and spec status.
 - [Noise Protocol Name](#noise-protocol-name)
 - [Wire Format](#wire-format)
 - [Encryption and I/O](#encryption-and-io)
-- [libp2p Interfaces and API](#libp2p-interfaces-and-api)
-  - [Initialization](#initialization)
-  - [Secure Transport Interface](#secure-transport-interface)
-    - [NoiseConnection](#noiseconnection)
-    - [SecureOutbound](#secureoutbound)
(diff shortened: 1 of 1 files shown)


mxinden (commented):
I am in favor of this change in particular and I am in favor of removal of programming API specifications in general. I don't think we should aim for libp2p implementations to be "compatible" on the API level, but only on the network level.

Thanks @marten-seemann.

I think it makes sense to bump the revision version, even though there is no change to the wire format. What do you think?

mxinden (approved):

marten-seemann:
> I think it makes sense to bump the revision version, even though there is no change to the wire format. What do you think?

I don't think so, this is just a (very large) editorial change. #453 will bump the version number anyway.

julian88110 (commented):
I agree the interface implementation details dose not belong to the spec. 

one general thought is that instead of defining the API, do we want to replace it with a brief general description of high level functions?

marten-seemann:
> do we want to replace it with a brief general description of high level functions?

We describe that this is a secure channel? Is there anything you'd like to add to that?

PR #453, opened 2022-09-14
noise: introduce an extension registry

### LINKED ISSUE #450: proposal: Noise extension registry
Problem: Noise doesn’t define a way to add general-purpose extensions to the handshake. These extensions could be used to negotiate optional features.

Currently, we have 3 (!) proposals in flight that would use Noise’s Early Data feature to either modify the Noise handshake or negotiate application-layer behavior:
1. WebTransport: #429, to tie the certificate hash to the (authenticated) peer IDs
2. WebRTC: #412, in a similar fashion, to tie the certificate fingerprint to the peer IDs
3. #446, to negotiate the stream multiplexer

By pure coincidence, these 3 use cases don’t overlap, i.e. you’d only want to use one OR the other, but not two or three at the same time. However, I can easily come up with use cases where you’d want to use one of those described above AND another extension.

As a point of reference, both TLS (via [RFC 6066](https://www.rfc-editor.org/rfc/rfc6066)) and QUIC (via transport parameters, see [Section 19.21 of RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000#section-19.21) provide such a general-purpose mechanism.

I propose to define a mechanism exactly along those lines: When defining a Noise extension, you’d pick a code point, and could then attach extension-defined data to that code point. It makes sense to use a Protobuf here, as our Noise handshake already uses Protobufs. We’re making use of the property of Protobufs that only fields that actually have a value are being serialized (I assume).

With the 3 use cases before, we’d start with something like this:
```proto
message NoiseExtension {
    optional bytes webrtc_fingerprint = 2;
    repeated string stream_muxers = 3; 
}
```

We could also add version field at the beginning of the protobuf, with a fixed version number, to make this even more future-proof.

Regarding document structure, this definition would probably live in a separate markdown file in the Noise directory. Specifications that want to use a Noise extension would need to update this file.

Regarding code points, we can reserve the short (1- and 2-byte varints) for approved specifications, and leave the space with longer varints open for experimentation on a first-come-first-serve basis.

@MarcoPolo @mxinden @julian88110 @p-shahi @BigLep, what do you think?

marten-seemann:
Fixes #450.

This PR removes the `data` field from the `NoiseHandshakePayload` protobuf, and adds an `extensions` field. `extensions` is `NoiseExtensions` protobuf, modeled after the extension registries that TLS and QUIC use.
By using a new ID (`data` was 3, `extensions` is 4), this change won’t break implementations that used early data (I’m not aware of any, but just in case).

I’ve added code points for WebTransport, WebRTC and Early Muxer Negotiation to this PR, mostly to demonstrate how this would look like. Given that none of the specs for these are merged yet, it might make sense to define an empty `NoiseExtensions` protobuf here, and have the respective PRs add their code point (we’ll have to deal with merge conflicts then). Let me know if I should make that change.

### DIFF
--- a/noise/README.md
+++ b/noise/README.md
@@ -219,8 +219,6 @@ syntax = "proto2";
 
 message NoiseExtensions {
     repeated bytes webtransport_certhashes = 1;
-    optional bytes webrtc_fingerprint = 2;
-    repeated string stream_muxers = 3; 
 }
 
 message NoiseHandshakePayload {
--- a/noise/README.md
+++ b/noise/README.md
@@ -5,7 +5,7 @@
 
 | Lifecycle Stage | Maturity       | Status | Latest Revision |
 |-----------------|----------------|--------|-----------------|
-| 3A              | Recommendation | Active | r3, 2022-09-20  |
+| 3A              | Recommendation | Active | r4, 2022-09-22  |
 
 Authors: [@yusefnapora]
 
@@ -200,18 +200,16 @@ The Noise Protocol Framework caters for sending early data alongside handshake
 messages. We leverage this construct to transmit:
 
 1. the libp2p identity key along with a signature, to authenticate each party to
-   the other.
-2. arbitrary data private to the libp2p stack. This facility is not exposed to
-   userland. Examples of usage include streamlining muxer selection.
+   the other, and
+2. extensions to the Noise handshake
 
-These payloads MUST be inserted into the first message of the handshake pattern
-**that guarantees secrecy**. In practice, this means that the initiator must not
-send a payload in their first message. Instead, the initiator will send its
-payload in message 3 (closing message), whereas the responder will send theirs
-in message 2 (their only message). It should be stressed, that the second
-message of the handshake pattern has forward secrecy, however the sender has not
-authenticated the responder, so this payload might be sent to any party,
-including an active attacker.
+Note that when sending extensions with the first message of the handshake
+pattern, the data is transmitted unencrypted. When sending the payload in
+message 3 (closing message), for the initiator, and in message 2, for the
+responder, the data will be send encrypted with forward secrecy.
+It should be stressed, that while the second message of the handshake pattern
+has forward secrecy, the sender has not yet authenticated the responder,
+so this payload might be sent to any party, including an active attacker.
 
 When decrypted, the payload contains a serialized [protobuf][protobuf]
 `NoiseHandshakePayload` message with the following schema:
@@ -219,10 +217,16 @@ When decrypted, the payload contains a serialized [protobuf][protobuf]
 ``` protobuf
 syntax = "proto2";
 
+message NoiseExtensions {
+    repeated bytes webtransport_certhashes = 1;
+    optional bytes webrtc_fingerprint = 2;
+    repeated string stream_muxers = 3; 
+}
+
 message NoiseHandshakePayload {
   optional bytes identity_key = 1;
   optional bytes identity_sig = 2;
-  optional bytes data         = 3;
+  optional NoiseExtensions extensions = 4;
 }
 ```
 
@@ -235,9 +239,8 @@ spec][peer-id-spec-signing-rules]. The data to be signed is the UTF-8 string
 `noise-libp2p-static-key:`, followed by the Noise static public key, encoded
 according to the rules defined in [section 5 of RFC 7748][rfc-7748-sec-5].
 
-The `data` field contains the "early data" provided to the Noise module when
-initiating the handshake, if any. The structure of this data is opaque to
-noise-libp2p and is defined in the connection establishment specs.
+The `extensions` field contains Noise extensions and is described in
+[Noise Extensions](#noise-extensions).
 
 Upon receiving the handshake payload, peers MUST decode the public key from the
 `identity_key` field into a usable form. The key MUST then be used to validate
@@ -282,6 +285,24 @@ Authentication section](#static-key-authentication) and may include other
 internal libp2p data.
 
 The XX handshake MUST be supported by noise-libp2p implementations.
+
+### Noise Extensions
+
+Since the Noise handshake pattern itself doesn't define any extensibility
+mechanism, this specification defines an extension registry, modeled after
+[RFC 6066](https://www.rfc-editor.org/rfc/rfc6066) (for TLS) and
+[RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000#section-19.21)
+(for QUIC).
+
+This spec currently defines 3 extension code points for the `NoiseExtensions`
+protobuf. Note that this document only defines the code point, and leaves
+it up to the protocol using that code point to define semantics associated
+with that code point.
+
+Code points above 1024 MAY be used for experimentation. Code points up to
+this value MUST be registered in this document before deployment.
+
+
 
 ## Cryptographic Primitives
 
--- a/noise/README.md
+++ b/noise/README.md
@@ -200,16 +200,17 @@ The Noise Protocol Framework caters for sending early data alongside handshake
 messages. We leverage this construct to transmit:
 
 1. the libp2p identity key along with a signature, to authenticate each party to
-   the other, and
-2. extensions to the Noise handshake
+   the other.
+2. extensions used by the libp2p stack.
 
-Note that when sending extensions with the first message of the handshake
-pattern, the data is transmitted unencrypted. When sending the payload in
-message 3 (closing message), for the initiator, and in message 2, for the
-responder, the data will be send encrypted with forward secrecy.
-It should be stressed, that while the second message of the handshake pattern
-has forward secrecy, the sender has not yet authenticated the responder,
-so this payload might be sent to any party, including an active attacker.
+The extensions are inserted into the first message of the handshake pattern
+**that guarantees secrecy**. Specifically, this means that the initiator MUST NOT
+send extensions in their first message. 
+The initiator sends its extensions in message 3 (closing message), and the 
+responder sends theirs in message 2 (their only message). It should be stressed,
+that while the second message of the handshake pattern has forward secrecy, 
+the sender has not authenticated the responder yet, so this payload might be
+sent to any party, including an active attacker.
 
 When decrypted, the payload contains a serialized [protobuf][protobuf]
 `NoiseHandshakePayload` message with the following schema:
--- a/noise/README.md
+++ b/noise/README.md
@@ -452,6 +452,14 @@ unsupported types like RSA.
 - Removed Noise Pipes and related handshake patterns
 - Removed padding within encrypted payloads
 
+### r3 - 2022-09-20
+
+- Change Protobuf definition to proto2 (due to the layout of the protobuf used, this is backwards-compatible change)
+
+### r4 - 2022-09-22
+
+- Add Noise extension registry
+
 
 [peer-id-spec]: ../peer-ids/peer-ids.md
 [peer-id-spec-signing-rules]: ../peer-ids/peer-ids.md#how-keys-are-encoded-and-messages-signed
--- a/noise/README.md
+++ b/noise/README.md
@@ -292,10 +292,9 @@ mechanism, this specification defines an extension registry, modeled after
 [RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000#section-19.21)
 (for QUIC).
 
-This spec currently defines 3 extension code points for the `NoiseExtensions`
-protobuf. Note that this document only defines the code point, and leaves
-it up to the protocol using that code point to define semantics associated
-with that code point.
+Note that this document only defines the `NoiseExtensions` code points, and
+leaves it up to the protocol using that code point to define semantics
+associated with these code point.
 
 Code points above 1024 MAY be used for experimentation. Code points up to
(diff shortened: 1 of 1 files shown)


noise/README.md
```diff
@@ -210,27 +210,32 @@ The Noise Protocol Framework caters for sending early data alongside handshake
...
+has forward secrecy, the sender has not yet authenticated the responder,
+so this payload might be sent to any party, including an active attacker.
 
 When decrypted, the payload contains a serialized [protobuf][protobuf]
 `NoiseHandshakePayload` message with the following schema:
 
 ``` protobuf
+message NoiseExtensions {
+    repeated bytes webtransport_certhashes = 1;
+    optional bytes webrtc_fingerprint = 2;
```
marten-seemann, line 230: @mxinden Is this `bytes` or should we use `string` here?

noise/README.md
```diff
@@ -210,27 +210,32 @@ The Noise Protocol Framework caters for sending early data alongside handshake
...
+}
+
+
 message NoiseHandshakePayload {
-  bytes identity_key = 1;
-  bytes identity_sig = 2;
-  bytes data         = 3;
+  bytes identity_key         = 1;
+  bytes identity_sig         = 2;
+  NoiseExtensions extensions = 4;
```
mxinden, line 238: Note that https://github.com/libp2p/specs/pull/412 would make use of the Noise [prologue](https://noiseprotocol.org/noise.html#prologue) feature. This is orthogonal to the early data mechanism. 

Thus far we were not planning to wrap the prologue data in a protobuf. That said, I am not opposed to it.

If I understand the Protobuf schema above correctly, this would require us to have a `identity_key` and `identity_sig` in the prologue encoding. Would we be setting that to an empty byte string? I have to give this more thought but as a first thought, this seems not very clean.

noise/README.md
mxinden: I think https://github.com/libp2p/specs/pull/412/ is not specific enough on this yet. Thanks for pointing that out. I suggest `string` for now which requires us to use multibase for the TLS certificate hashes. I still need to give this a bit more thought.

noise/README.md
```diff
@@ -291,6 +295,24 @@ internal libp2p data.
...
+[RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000#section-19.21)
+(for QUIC).
+
+This spec currently defines 3 extension code points for the `NoiseExtensions`
+protobuf. Note that this document only defines the code point, and leaves
+it up to the protocol using that code point to define semantics associated
+with that code point.
+
+Code points above 1024 MAY be used for experimentation. Code points up to
+this value MUST be registered in this document before deployment.
```
mxinden, line 312: As far as I understand we tried to use [multicodec](https://github.com/multiformats/multicodec/) for all magic numbers in the libp2p ecosystem thus far. If I understand this proposal correctly, it runs somewhat against this convention.

I am not opposed to that, I just think it is worth pointing out in the specification. E.g. a short paragraph of why we are not using multicodec for this.

What do you think @marten-seemann?

noise/README.md
marten-seemann: I don’t think we’d need this Protobuf at all when using Prologue, since prologues are not sent in the wire, they’re just hashed into the handshake transcript. 
I’ll remote the WebRTC extension entry here.

Orthogonal to this PR: Protobufs are great when sent over the wire. They’re not that great if you’re looking for a unique encoding or your data, which seems to be what we need for prologues. Just using a byte array / string seems to be the better solution for that use case.

noise/README.md
MarcoPolo: on the wire these are the same: https://developers.google.com/protocol-buffers/docs/encoding

noise/README.md
MarcoPolo: I don't think we should adopt this for the our prologue use. The reason is that the prologue is very specific to the protocol that's using it. Both sides must know the protocol and agree on what the prologue is before the handshake starts. Therefore it makes more sense to have each protocol (like webrtc) define how the prologue is set since it will only ever be set in the context of the protocol.

noise/README.md
```diff
@@ -210,27 +210,32 @@ The Noise Protocol Framework caters for sending early data alongside handshake
...
+   the other, and
+2. extensions to the Noise handshake
+
+Note that when sending extensions with the first message of the handshake
+pattern, the data is transmitted unencrypted. When sending the payload in
+message 3 (closing message), for the initiator, and in message 2, for the
+responder, the data will be send encrypted with forward secrecy.
+It should be stressed, that while the second message of the handshake pattern
+has forward secrecy, the sender has not yet authenticated the responder,
+so this payload might be sent to any party, including an active attacker.
```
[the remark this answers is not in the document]
marten-seemann: Prologue seems orthogonal to this PR. I agree that it would be helpful to document that using Prologues is an option. Do you want to create a separate PR for that?

noise/README.md
marten-seemann: Neither, it's a prologue. Resolving.

noise/README.md
marten-seemann: I removed WebRTC and stream muxer selection from the protobuf. WebRTC won't use early data, but a Noise prologue. For muxer selection, the respective PR (#446) will add that field.

noise/README.md
marten-seemann: Good point. I'm not super familiar with the design philosophy behind multicoded. @Stebalien and @lidel, could you help us out here?

noise/README.md
```diff
@@ -210,27 +210,30 @@ The Noise Protocol Framework caters for sending early data alongside handshake
...
-message of the handshake pattern has forward secrecy, however the sender has not
-authenticated the responder, so this payload might be sent to any party,
-including an active attacker.
+   the other, and
+2. extensions to the Noise handshake
+
+Note that when sending extensions with the first message of the handshake
+pattern, the data is transmitted unencrypted. When sending the payload in
+message 3 (closing message), for the initiator, and in message 2, for the
+responder, the data will be send encrypted with forward secrecy.
```
MarcoPolo, line 219:
```suggestion
responder, the data will be sent encrypted with forward secrecy.
```

noise/README.md
```diff
@@ -291,6 +293,24 @@ internal libp2p data.
...
 
+### Noise Extensions
+
+Since the Noise handshake pattern itself doesn't define any extensibility
+mechanism, this specification defines an extension registry, modeled after
+[RFC 6066](https://www.rfc-editor.org/rfc/rfc6066) (for TLS) and
+[RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000#section-19.21)
+(for QUIC).
+
+This spec currently defines 3 extension code points for the `NoiseExtensions`
```
MarcoPolo, line 304: I only see one code in this spec (`repeated bytes webtransport_certhashes = 1;`), right? Or am I missing something?

noise/README.md
```diff
@@ -210,27 +210,30 @@ The Noise Protocol Framework caters for sending early data alongside handshake
 messages. We leverage this construct to transmit:
 
 1. the libp2p identity key along with a signature, to authenticate each party to
-   the other.
-2. arbitrary data private to the libp2p stack. This facility is not exposed to
-   userland. Examples of usage include streamlining muxer selection.
-
-These payloads MUST be inserted into the first message of the handshake pattern
```
MarcoPolo, line 217: I don't think we want to remove this. If anything we should define a new payload type that is **only** sent on the first message. I think it's a footgun to have the `libp2p Handshake Payload` be sent either in plaintext or encrypted. 

I also think with the current usecases we don't need to define the type of data for the first handshake message. Everything we want to do can be done with the 2nd and 3rd message. So we can keep our original crypto guarantees.

MarcoPolo (changes requested):
Let's discuss the first message payload changes first.

noise/README.md
marten-seemann: I removed the other ones and forgot to update this sentence.

marten-seemann: We probably shouldn't count the extensions here anyway. This is just guaranteed to get out of sync in the future. I removed it in 8e4b017c6a8f4251a45acf604f0a36e6bc81df23.

noise/README.md
marten-seemann: I reverted this change.

marten-seemann: I rewrote this paragraph, to clarify that extensions are sent in message 2 and 3.

noise/README.md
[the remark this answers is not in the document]
mxinden: Sounds good. Added to my TODO list.

noise/README.md
```diff
@@ -211,26 +211,30 @@ messages. We leverage this construct to transmit:
...
+responder sends theirs in message 2 (their only message). It should be stressed,
+that while the second message of the handshake pattern has forward secrecy, 
+the sender has not authenticated the responder yet, so this payload might be
+sent to any party, including an active attacker.
 
 When decrypted, the payload contains a serialized [protobuf][protobuf]
 `NoiseHandshakePayload` message with the following schema:
 
 ``` protobuf
+message NoiseExtensions {
```
julian88110, line 229: For security reasons such as DOS, do we want to specify a size limit or item limit on the extensions?

noise/README.md
```diff
@@ -211,26 +211,30 @@ messages. We leverage this construct to transmit:
...
+2. extensions used by the libp2p stack.
+
+The extensions are inserted into the first message of the handshake pattern
+**that guarantees secrecy**. Specifically, this means that the initiator MUST NOT
+send extensions in their first message. 
+The initiator sends its extensions in message 3 (closing message), and the 
+responder sends theirs in message 2 (their only message). It should be stressed,
+that while the second message of the handshake pattern has forward secrecy, 
+the sender has not authenticated the responder yet, so this payload might be
+sent to any party, including an active attacker.
```
julian88110, line 223: As a security precaution is it helpful to specify that the "early data" should not be processed before the handshake is fully finished?  thus gives us more confidence on the data, what are your thoughts on this?

noise/README.md
marten-seemann: We should, good catch.

The amount of data we want to transmit might be on the larger side (all of our protocols), but then this is the second flight, so in practice we're limited by the host's congestion controller: If we try to send too much data, it won't fit into the congestion window and then we have to wait 1 RTT for acknowledgements to arrive, which would nullify the benefit of using an extension in the first place.
I'd assume that sending more than 1 kB would get us into problems here. We should probably point out this problem in the spec. The size limit for DoS protection can be higher than this, I guess 4 or 8kb would be fine?

noise/README.md
marten-seemann: In WebTransport, we process it immediately, and fail the handshake if the certificate hash doesn't match.
Maybe we should point out that consumers of early data need to be careful, and that delaying the processing / acting upon the early data can be one strategy to deal with this.

julian88110: Great, document that is a good point.

julian88110 (approved):

noise/README.md
```diff
@@ -243,9 +247,8 @@ spec][peer-id-spec-signing-rules]. The data to be signed is the UTF-8 string
 `noise-libp2p-static-key:`, followed by the Noise static public key, encoded
 according to the rules defined in [section 5 of RFC 7748][rfc-7748-sec-5].
 
-The `data` field contains the "early data" provided to the Noise module when
-initiating the handshake, if any. The structure of this data is opaque to
-noise-libp2p and is defined in the connection establishment specs.
+The `extensions` field contains Noise extensions and is described in
+[noise-extensions].
```
mxinden, line 251: I don't think this is a valid GitHub Markdown link. At least it does not resolve in a hyperlink when I view the markdown file directly.

mxinden (commented):
Looks good to me once updated to https://github.com/libp2p/specs/pull/456 and multicodec comment is addressed.

Thanks @marten-seemann for driving this.

marten-seemann:
I rebased this PR on top of master, bumped the version number and added a Changelog.

This should be good to merge now.

noise/README.md
marten-seemann: Fixed.

noise/README.md
```diff
@@ -201,28 +201,31 @@ messages. We leverage this construct to transmit:
...
 syntax = "proto2";
 
+message NoiseExtensions {
+    repeated bytes webtransport_certhashes = 1;
+}
+
 message NoiseHandshakePayload {
   optional bytes identity_key = 1;
   optional bytes identity_sig = 2;
-  optional bytes data         = 3;
```
mxinden, line 225: For the record, rust-libp2p does not make use of `data`. @marten-seemann do any of the other implementations?

mxinden (approved):
Just one question. Otherwise this looks good to me.

noise/README.md
marten-seemann: go-libp2p doesn't, and js-libp2p doesn't either (as far as I know).

mxinden: What about nim-libp2p cpp-libp2p jvm-libp2p @marten-seemann?

marten-seemann: I’m not familiar with their code bases. This PR was open for more than a week, that should’ve been plenty of time to chime in. Also note that this is backwards-compatible change since we didn’t reuse field 3 in the protobuf.

PR #455, opened 2022-09-16
webtransport: move certhash verification to the client

marten-seemann:
**This PR is targeting #404, not master**.

This implements the proposal discussed in https://github.com/libp2p/specs/pull/404/files#r968836367.


mxinden (approved):
Looks good to me.

> but cannot determine which certificate was actually used to establish the connection (this will commonly be the case for browser clients)

Oh, that is unfortunate.

MarcoPolo (approved):

PR #461, opened 2022-09-28
roadmap: rewrite the section about efficient handshakes

marten-seemann:
Over the last couple of weeks, we've broken the early muxer negotiation into its own project (#446), and found an arguably nicer solution than proposed in the Protocol Select spec (#349). The change to move the security address into the multiaddr is already split out into a separate PR: #354.

With these two changes having been separated from Protocol Select, we save 2 RTTs during a TCP handshake, reducing the total latency to the bare minimum consumed by the components used (TCP: 1 RTT, security handshake: 1 RTT).

Implementing the rest of the Protocol Select spec now appears less pressing. The main additional benefits would be:
1. Slightly shorter protocol names, since the protocol string would be wrapped in a protobuf instead of being prefixed with `/multistream/1.0.0`. This saving of ~10 bytes on a new stream is likely negligible in almost all situations.
2. The ability to propose multiple protocols on a stream at the same time, instead of sequentially. In theory, this saves roundtrips, but in practice, one can just open multiple streams, one for each of the protocols (and then even send application data right away, if desired).
3. The Protocol Select protobuf is as a basis for future extensions to the protocol, e.g. an extension that abbreviates protocol IDs with an integer, saving another ~10 bytes per stream.

(1) and (2) probably don't justify a high priority for Protocol Select. (3) is a valid point, but it would be preferable if the project was driving by a concrete use case that in itself would justify a high priority, rather than by the wish to create an extensible platform.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -322,21 +322,19 @@ be achieved with the [WebTransport](#✈️-webtransport) protocol.
 **High priority for: IPFS**
 
 **What?** Establishing a connection and performing the initial handshake
-should be as cheap and fast as possible. Supporting things like
-*selective* stream opening, *preemption*, *speculative* negotiation,
-*upfront* negotiation, protocol table *pinning*, etc. may enable us to
-achieve lower latencies when establishing connections, and even the
-0-RTT holy grail in some cases. These features are being discussed in
-the *Protocol Select* protocol design.
+should be as cheap and fast as possible. On TCP, The current libp2p
+handshakes spends one round-trip negotiating the security protocol, and
+another roundtrip negotiating the stream multiplexer.
+By using advanced features of the handshake protocol, we might even be
+able to reach the holy grail of a 0-RTT handshake in some cases.
 
-**Why?** Multistream 1.0 is chatty and naïve. Streams are essential to
-libp2p, and negotiating them is currently inefficient in a number of
-scenarios. Also, bootstrapping a multiplexed connection is currently
-guesswork (we test protocols one by one, incurring in significant
-ping-pong).
+**Why?** Applications rely on quick connection establishment. libp2p
+shouldn't make them wait for any longer than absolutely necessary.
 
 **Links:**
 
+- [security protocol in multiaddr](https://github.com/libp2p/specs/pull/353)
+- [muxer selection in security handshake](https://github.com/libp2p/specs/pull/446)
 - [Protocol Select specification](https://github.com/libp2p/specs/pull/349)
 
 ### 🛣️ Peer Routing Records
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -322,9 +322,9 @@ be achieved with the [WebTransport](#✈️-webtransport) protocol.
 **High priority for: IPFS**
 
 **What?** Establishing a connection and performing the initial handshake
-should be as cheap and fast as possible. On TCP, The current libp2p
-handshakes spends one round-trip negotiating the security protocol, and
-another roundtrip negotiating the stream multiplexer.
+should be as cheap and fast as possible. On TCP, the current libp2p
+handshake spends one round-trip negotiating the security protocol, and
+another round-trip negotiating the stream multiplexer.
 By using advanced features of the handshake protocol, we might even be
 able to reach the holy grail of a 0-RTT handshake in some cases.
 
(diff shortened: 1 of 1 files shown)


MarcoPolo:
With the caveat that I'm not as familiar with protocol select as others. I agree with these points.

On point 3, couldn't users already define an integer that represents their protocol? Is there anything that says it has to be a human readable string?

Even if not, we could imagine a small change to multistream-select that allows them to pass multicodecs instead of protocol ids with a similar extensions protobuf. This could be `/multistream/1.1.0` that gracefully falls back to the old method if it gets a `/multistream/1.0.0` response (we'd have to support semver matching in multistream code).

marten-seemann:
> On point 3, couldn't users already define an integer that represents their protocol? Is there anything that says it has to be a human readable string?

They could, but without a place to coordinate the small values, this will probably run into interoperability issues rather quickly.

I’m not convinced this is a problem worth solving in the first place. Even if you have a crazily chatty protocol that opens a stream in every single packet you send, the overhead of multistream will still just be 20 bytes (compared to a Protobuf with a protocol name compression extension). For reference, 20 bytes is the overhead of IPv6 vs IPv4, or 1.5% of the packet payload.

ROADMAP.md
```diff
@@ -322,21 +322,19 @@ be achieved with the [WebTransport](#✈️-webtransport) protocol.
...
-achieve lower latencies when establishing connections, and even the
-0-RTT holy grail in some cases. These features are being discussed in
-the *Protocol Select* protocol design.
-
-**Why?** Multistream 1.0 is chatty and naïve. Streams are essential to
-libp2p, and negotiating them is currently inefficient in a number of
-scenarios. Also, bootstrapping a multiplexed connection is currently
-guesswork (we test protocols one by one, incurring in significant
-ping-pong).
+should be as cheap and fast as possible. On TCP, The current libp2p
```
mxinden, line 325:
```suggestion
should be as cheap and fast as possible. On TCP, the current libp2p
```

mxinden (approved):
> Implementing the rest of the Protocol Select spec now appears less pressing.

Agreed that it became less pressing. Agreed with (1), (2) and (3).

The rephrasing of the section looks good to me.

I hope that we will get to Protocol Select one day.

ROADMAP.md
```diff
@@ -322,21 +322,19 @@ be achieved with the [WebTransport](#✈️-webtransport) protocol.
...
+handshakes spends one round-trip negotiating the security protocol, and
+another roundtrip negotiating the stream multiplexer.
+By using advanced features of the handshake protocol, we might even be
+able to reach the holy grail of a 0-RTT handshake in some cases.
+
+**Why?** Applications rely on quick connection establishment. libp2p
```
julian88110, line 331:
```suggestion
**Why?** Applications rely on quick connection establishment. Libp2p
```

julian88110 (approved):
Agreed, with the recent changes that are coming, this is not a priority anymore and actual work should only be committed when there are actual needs rather than speculated needs

ROADMAP.md
```diff
@@ -322,21 +322,19 @@ be achieved with the [WebTransport](#✈️-webtransport) protocol.
...
+shouldn't make them wait for any longer than absolutely necessary.
 
 **Links:**
 
+- [security protocol in multiaddr](https://github.com/libp2p/specs/pull/353)
+- [muxer selection in security handshake](https://github.com/libp2p/specs/pull/446)
```
p-shahi, line 337:
```suggestion
- [Security protocol in multiaddr](https://github.com/libp2p/specs/pull/353)
- [Muxer selection in security handshake](https://github.com/libp2p/specs/pull/446)
```
nit

ROADMAP.md
```diff
@@ -322,21 +322,19 @@ be achieved with the [WebTransport](#✈️-webtransport) protocol.
...
-0-RTT holy grail in some cases. These features are being discussed in
-the *Protocol Select* protocol design.
-
-**Why?** Multistream 1.0 is chatty and naïve. Streams are essential to
-libp2p, and negotiating them is currently inefficient in a number of
-scenarios. Also, bootstrapping a multiplexed connection is currently
-guesswork (we test protocols one by one, incurring in significant
-ping-pong).
+should be as cheap and fast as possible. On TCP, The current libp2p
+handshakes spends one round-trip negotiating the security protocol, and
```
p-shahi, line 326:
```suggestion
handshake spends one round-trip negotiating the security protocol, and
```

ROADMAP.md
```diff
@@ -322,21 +322,19 @@ be achieved with the [WebTransport](#✈️-webtransport) protocol.
...
+another roundtrip negotiating the stream multiplexer.
```
p-shahi, line 327:
```suggestion
another round-trip negotiating the stream multiplexer.
```

p-shahi (commented):
minor grammatical changes & a nit

ROADMAP.md
marten-seemann: libp2p should not be capitalized, as far as I know.

p-shahi (approved):

PR #462, opened 2022-09-29
roadmap: remove browser use cases

marten-seemann:
This is hopefully less controversial than the PR title suggests.

Reasons for removing this:
1. We have separate roadmap items for WebTransport and WebRTC, as well as a separate WASM item.
2. At this point, it's fair to say that raw TCP / UDP sockets in the browser is not going anywhere on the browser side. These efforts are effectively abandoned now.
3. Reducing the bundle size for js-libp2p seems like a worthwhile goal, and as far as I know js-libp2p has made significant progress towards that goal in recent months. This seems like a fine item for a js-libp2p specific roadmap.


MarcoPolo (approved):

mxinden (approved):
Not controversial in my eyes. Thanks for the change.

achingbrain (approved):

PR #463, opened 2022-09-29
roadmap: rewrite the telemetry section

marten-seemann:
Opt-in telemetry will only be of limited use, since there will be a very heavy selection bias (who turns on telemetry), as 99% of users never change the default settings.

Recent advances in cryptography, especially in multiparty computation, might allow us to develop a telemetry protocol that has privacy properties strong enough to justify turning on telemetry by default.

Providing a telemetry framework could be a major selling point of libp2p as library: Applications could use the same framework to gather application-level measurement results that come with the same strong privacy guarantees.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -31,7 +31,7 @@ third-party ownership of data.
         - [🧱 Composable routing](#🧱-composable-routing)
         - [💣 Attack resistance, threat models and security](#💣-attack-resistance-threat-models-and-security)
         - [📈 Proving we are scalable and interoperable](#📈-proving-we-are-scalable-and-interoperable)
-        - [🌡️ Opt-in telemetry protocol](#🌡️-opt-in-telemetry-protocol)
+        - [🌡️ Telemetry protocol](#🌡️-telemetry-protocol)
         - [📩 Message-oriented transports](#📩-message-oriented-transports)
         - [☎️ Reducing the dial fail rate](#️-reducing-the-dial-fail-rate)
         - [🔀 Peer exchange protocol](#🔀-peer-exchange-protocol)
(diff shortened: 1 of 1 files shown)


MarcoPolo:
Very cool. I agree this seems like a worthy endeavor!

ROADMAP.md
```diff
@@ -511,26 +511,29 @@ advances made by Mozilla's
 Not fully supporting the browser platform will hurt most of the projects betting
 on libp2p.
 
-### 🌡️ Opt-in telemetry protocol
+### 🌡️ Telemetry protocol
 
-**What?** By developing a *decentralized* telemetry protocol that nodes can opt
-into, we could build a world-wide real-time visualisation of our global libp2p
+**What?** By developing a *privacy-preserving* telemetry protocol,
```
julian88110, line 516: I agree telemetry is important. 
Should we split the telemetry protocol effort and metrics support effort into two separate items?  or change the title to be both the protocol and metrics?

marten-seemann: > Should we split the telemetry protocol effort and metrics support effort into two separate items?

Aren't they already? The metrics effort is described in the "Full Observability" section.

julian88110: I see, sorry my miss. Thanks for clarifying!  The "collect stats" statement made me misunderstood this item.

julian88110 (approved):

mxinden (approved):
:+1: looks good to me.

//CC @yiannisbot given that this is relevant for Probe Lab and given that you have been looking into this in the past.

MarcoPolo (approved):

marten-seemann:
Rebased PR to resolve merge conflict.

PR #469, opened 2022-10-14
connections: Document stream behaviour

thomaseizinger:
An initial draft for documenting the behaviour of libp2p streams. Feedback welcome!

## Open questions

- The mplex spec differs in regards to how reading from closed streams is handled. Which one is the better behaviour? What mplex specifies or what I wrote here? I took inspiration from our current implementations in `rust-libp2p`.
- Do we need specify more behaviour?
- The "Definitions" section specifies "They can also be "half closed", meaning that a stream can be closed for writing data but still open to receiving data and vice versa." I am not sure about the "vice-versa" part, i.e. actively closing the read-half but still writing to the stream. Neither mplex nor yamux support this. The Rust implementation doesn't have an interface for this. I am not sure about others. We built it for WebRTC with inspiration from QUIC but personally don't know when I'd use it.

### DIFF
--- a/connections/README.md
+++ b/connections/README.md
@@ -76,7 +76,7 @@ verifiable by the other peer.
 libp2p connection. They must support backpressure, which prevents receivers from
 being flooded by data from eager senders. They can also be "half closed",
 meaning that a stream can be closed for writing data but still open to receiving
-data and vice versa.
+data and vice versa. See [Stream Behaviour](#stream-behaviour) for details.
 
 Support for multiple streams ensures that a single connection between peers can
 support a wide variety of interactions, each with their own protocol. This is
--- a/connections/README.md
+++ b/connections/README.md
@@ -29,6 +29,10 @@ and spec status.
         - [multistream-select](#multistream-select)
     - [Upgrading Connections](#upgrading-connections)
     - [Opening New Streams Over a Connection](#opening-new-streams-over-a-connection)
+    - [Stream Behaviour](#stream-behaviour)
+      - [States](#states)
+      - [Resets](#resets)
+      - [State transition diagram](#state-transition-diagram)
     - [Practical Considerations](#practical-considerations)
         - [Interoperability](#interoperability)
         - [State Management](#state-management)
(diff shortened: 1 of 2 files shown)


connections/README.md
```diff
@@ -263,6 +267,49 @@ indicating whether the handler supports the protocol. This allows more flexible
...
+The default state after opening a new stream is "Open for reading and writing".
+Either party can at any point close the reading or writing side.
+
+The "Closed" state is the product of both halfs (read and write) of the stream being closed.
+This can happen because one party closes both halfs but also as a combination
+of both parties closing their write-half (which automatically closes other's read-half).
+
+Writing to a write-closed stream SHOULD fail with an error clearly signalling the closed state.
+
+Reading from a read-closed stream SHOULD succeed by reading 0 bytes.
```
mxinden, line 292: Intuitively this is to implementation specific. E.g. some may express this through a special event, some might express this through `0 bytes`. I don't think we should dictate the way of how an implementation expresses this.

connections/README.md
```diff
@@ -263,6 +267,49 @@ indicating whether the handler supports the protocol. This allows more flexible
...
+This allows for a robust "read to end"-behaviour.
+Layers on top of the stream MAY fail with an unexpected EOF error in case more data was expected.
```
mxinden, line 294: I don't think this needs to be in the specification.

connections/README.md
```diff
@@ -263,6 +267,49 @@ indicating whether the handler supports the protocol. This allows more flexible
...
+### Resets
+
+A stream reset is an abrupt termination of the stream.
+Implementations SHOULD reset a stream to signal a non-graceful close.
+Typically, this is during a resource cleanup like garbage-collection for streams
+which have not been closed properly (see [States](#states) above).
+
+Upon receiving a stream reset, any buffered data in the stream MAY be discarded.
+
+Reading and writing to a reset stream SHOULD fail with an error clearly signalling the reset.
```
mxinden, line 305: Cross referencing a past discussion for the curious reader here, namely that QUIC only closes the write side.

https://github.com/libp2p/specs/pull/412#discussion_r983113010

I don't have a strong opinion here. Maybe @marten-seemann does. Ideally we would be consistent across the different multiplexers. At least we should document the difference here.

mxinden (commented):
Thanks for documenting this! Very much appreciated.

Minus the comments, I think this is good to go. Bonus would be to document the behavior of QUIC, WebRTC, mplex and yamux here as well.

@marten-seemann want to give this a review as well?

marten-seemann (changes requested):
As I've explained on the WebRTC PR, I don't think we should define our stream state machine to match the behavior of legacy stream muxers like yamux or mplex. I'd like our stream state machine to take advantage of what QUIC has to offer.
By virtue of that, make it possible to map protocols like HTTP onto libp2p. The current proposal would prohibit this.

connections/README.md
thomaseizinger:
```suggestion
Reading from a read-closed stream SHOULD succeed.
```

Would that be better?

connections/README.md
thomaseizinger: Would you just omit it? I think I'd find it helpful because it explains, _why_ we want read operations on closed streams to be successful. Without this, people may rightfully ask, why reading from a read-closed stream shouldn't fail?

I am happy to reword it, but I think it is useful to have in the spec in one form or another.

thomaseizinger:
> As I've explained on the WebRTC PR, I don't think we should define our stream state machine to match the behavior of legacy stream muxers like yamux or mplex. I'd like our stream state machine to take advantage of what QUIC has to offer. By virtue of that, make it possible to map protocols like HTTP onto libp2p. The current proposal would prohibit this.

That does make sense.

Would you still keep it symmetric though? More specifically, can you reset a read-only stream? From [this conversation](https://github.com/libp2p/specs/pull/412#discussion_r991805871) it appears that QUIC cannot do that. I am curious to understand why. Does "closing" a read-only stream always communicate the same intent?

mxinden:
@thomaseizinger @marten-seemann what are the next steps here? What do you think of documenting the QUIC stream behavior as the reference plus how the other stream muxers deviate from that? While not ideal, it would document the status quo and where we are heading.

marten-seemann:
> What do you think of documenting the QUIC stream behavior as the reference plus how the other stream muxers deviate from that? While not ideal, it would document the status quo and where we are heading.

Sounds good to me.

thomaseizinger:
> @thomaseizinger @marten-seemann what are the next steps here? What do you think of documenting the QUIC stream behavior as the reference plus how the other stream muxers deviate from that? While not ideal, it would document the status quo and where we are heading.

I would still like to understand why resetting a read-stream is something that is unsupported. It seems like an obvious step from the two-stream mental model that you can reset both to signal an abrupt termination.

PR #474, opened 2022-11-05
Practical Example Section

btoms20:
This PR adds a Practical Example section to the Spec in an attempt to help developers (and end users) understand how the default IPFS Kad DHT works and how Keys on the DHT are derived for various namespaces.

#### What?
A brief example of how to format an RPCMessage in order to query the default IPFS Kad DHT for the PublicKey Record of the Bootstrap node with PeerID `QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ`.

#### Why?
No matter how well a Spec is written I often find myself searching for an example with real data. Real data, in this case, is hard due to the ephemeral nature of the DHT, but assuming the Bootstrap node used in this example persists, the querying of it for its PublicKey Record should continue to work.

#### More?
Hopefully we can add more examples to this section going forward. Maybe a similar example searching the `ipns` namespace and/or searching the KadDHT for Content Providers. 

#### Future
If the community thinks these Practical Example sections are worthwhile / helpful, we could add relevant examples to the rest of the Specs.

#### Notes 
- I placed the example in a `details` tag that defaults to closed as to not clutter the page. At the end of the day this is a Specs documentation and it should remain clear and concise, but if an implementer gets stuck, they could always dig deeper into the examples. Hopefully this doesn't break any formatting rules.
- Feel free to edit, change, delete this PR
- Feedback of all kinds is welcome

Thanks

### DIFF
--- a/kad-dht/README.md
+++ b/kad-dht/README.md
@@ -455,6 +455,53 @@ multiaddrs are stored in the node's peerbook.
 
 ---
 
+## Practical Examples
+<details>
+  <summary>Searching the <b>IPFS</b> Kad DHT for a Peer's Public Key</summary><blockquote>
+  
+  &nbsp;
+  ##### Lets find the Public Key for the IPFS Bootstrap node with the PeerID "QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ"
+  1. Derive the key to search the DHT for
+     1. We're searching for a public key record therefore we're going to use the `/pk/` namespace
+     2. And we want the public key associated with the PeerID `QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ`
+     3. Putting these together we get `/pk/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ`
+     4. But we need to turn this into it's raw byte representation
+     5. For the namespace and forward slashes we just convert the utf8 string into bytes
+     6. For the PeerID we need to remember that this is a [base58btc encoding](peer-ids.md#decoding) 
+      ---
+
+      | String | / | pk | / |  Qm | aCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ |
+      |--------|---|----|---|-----|----------------------------------------------|
+      | Part | slash | namespace | slash | Multihash prefix | Multihash digest | 
+      | Encoding | utf8 | utf8 | utf8 | base58btc | base58btc | 
+      | Bytes (UInt8) | 47 | 112, 107 | 47 | 18, 32 | 176, 74, 87, 212, 14, 202, 19, 136, 9, 241, 57, 167, 107, 18, 4, 67, 51, 195, 116, 3, 145, 201, 191, 28, 233, 216, 226, 26, 121, 33, 11, 253 | 
+
+      ---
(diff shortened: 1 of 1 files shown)


kad-dht/README.md
```diff
@@ -455,6 +455,53 @@ multiaddrs are stored in the node's peerbook.
 
 ---
 
+## Practical Examples
+<details>
+  <summary>Searching the <b>IPFS</b> Kad DHT for a Peer's Public Key</summary><blockquote>
```
mxinden, line 460: I think my message on https://discuss.libp2p.io/ was misleading. Sorry for that. `/pk` is an IPFS concept and not a llibp2p concept. Thus I don't think it should belong in libp2p/specs but instead in ipfs/specs.

Now unfortunately IPFS does not (yet) have a Kademlia specification, i.e. a document that describes how IPFS uses the libp2p Kademlia protocol. This is tracked in https://github.com/ipfs/specs/issues/345. //CC @lidel

I suggest moving the `/pk` related example to ipfs/specs.

kad-dht/README.md
```diff
@@ -455,6 +455,53 @@ multiaddrs are stored in the node's peerbook.
...
+  
+  &nbsp;
+  ##### Lets find the Public Key for the IPFS Bootstrap node with the PeerID "QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ"
+  1. Derive the key to search the DHT for
+     1. We're searching for a public key record therefore we're going to use the `/pk/` namespace
+     2. And we want the public key associated with the PeerID `QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ`
+     3. Putting these together we get `/pk/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ`
+     4. But we need to turn this into it's raw byte representation
+     5. For the namespace and forward slashes we just convert the utf8 string into bytes
+     6. For the PeerID we need to remember that this is a [base58btc encoding](peer-ids.md#decoding)
```
mxinden, line 470: I think the example of en-/ and decoding of a peer ID would be better placed in https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#decoding. We have some examples there, but none of them go all the way from private and public key to peer ID back to public key (in the case of ed25519). Would you be interested in extending the documentation there?

mxinden (commented):
Sorry for the late response here.

In my eyes this is very much needed documentation. That said, I don't think libp2p/specs is the right place for the `/pk` example (see below). This is my bad. I should have communicated better out of band.

I think the peer ID related example should live in https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#decoding.

@lidel do you think this is helpful for https://github.com/ipfs/specs/issues/345?

btoms20:
The part I wanted to focus on the most was actually key formation. When I first began interacting with the dht I naively converted the entire string '/pk/Qm...' directly into bytes. It took me a few hours to figure out you convert the namespace into bytes and append the original PeerID bytes to it. Pretty silly mistake but thought it might be helpful to jot down for others. 

I agree that this probably belongs somewhere in the ipfs docs.

kad-dht/README.md
lidel: :+1: to moving IPFS-specific things to ipfs/specs (fine to open a PR draft with copy&paste and then mention in https://github.com/ipfs/specs/issues/345 so we can take it from there).

PR #478, opened 2022-11-17
Introduce `perf` spec

MarcoPolo:
A simple protocol to run client-driven benchmarks inspired by https://datatracker.ietf.org/doc/html/draft-banks-quic-performance.

### DIFF
--- a/perf/perf.md
+++ b/perf/perf.md
@@ -106,7 +106,7 @@ To run this benchmark:
 3. once a connection is established, the client closes it and establishes
    another one.
 
-handshakes per second are calculated by taking the total number of connections
+Handshakes per second are calculated by taking the total number of connections
 successfully established and divide it by the time period of the test.
 
 # Security Considerations
--- a/perf/perf.md
+++ b/perf/perf.md
@@ -45,7 +45,8 @@ Client:
 1. Open a libp2p stream to the server.
 2. Tell the server how many bytes we want the server to send us as a single
    big-endian uint64 number. Zero is a valid number, so is the max uint64 value.
-3. Optional: Write some amount of data to the stream.
+3. Write some amount of data to the stream.
+     Zero is a valid amount.
 4. Close the write side of our stream.
 5. Read from the read side of the stream. This
    should be the same number of bytes as we told the server in step 2.
--- a/perf/perf.md
+++ b/perf/perf.md
@@ -1,0 +1,121 @@
+# Perf <!-- omit in toc -->
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+| --------------- | ------------- | ------ | --------------- |
+| 1A              | Working Draft | Active | r0, 2022-11-16  |
+
+Authors: [@marcopolo]
+
+Interest Group: [@marcopolo], [@mxinden], [@marten-seemann]
+
+[@marcopolo]: https://github.com/mxinden
+[@mxinden]: https://github.com/mxinden
+[@marten-seemann]: https://github.com/marten-seemann
+
+# Table of Contents <!-- omit in toc -->
+- [Context](#context)
+- [Protocol](#protocol)
+- [Benchmarks](#benchmarks)
+  - [Single connection throughput](#single-connection-throughput)
+  - [Handshakes per second](#handshakes-per-second)
+- [Security Considerations](#security-considerations)
+- [Prior Art](#prior-art)
+
+# Context
+
+The `perf` protocol represents a standard benchmarking protocol that we can use
+to talk about performance within and across libp2p implementations. This lets us
+analyze peformance, guide improvements, and protect against regressions.
+
+# Protocol
+
+The `/perf/1.0.0` protocol (from here on referred to as simply `perf`) is a
+client driven set of benchmarks. To not reinvent the wheel, this perf protocol
+is almost identical to Nick Bank's [_QUIC Performance_
+Internet-Draft](https://datatracker.ietf.org/doc/html/draft-banks-quic-performance#section-2.3)
+but adapted to libp2p.
+
+
+The protocol is as a follows:
+
+Client:
+
+1. Open a libp2p stream to the server.
+2. Tell the server how many bytes we want the server to send us as a single
+   big-endian uint64 number. Zero is a valid number, so is the max uint64 value.
+3. Write some amount of data to the stream.
+4. Close the write side of our stream.
+5. Read from the read side of the stream. This
+   should be the same number of bytes as we told the server in step 2.
+
+Server, on handling a new `perf` stream:
+1. Read the big-endian uint64 number. This is how many bytes we'll send back.
+2. Read from the stream until we get an `EOF` (client's write side was closed).
+3. Send the number of bytes defined in step 1 back to the client.
+4. Close the stream.
+
+# Benchmarks
+
+The above protocol is flexible enough to run the following benchmarks and more.
+The exact specifics of the benchmark (e.g. how much data to download or for how
+long) are left up to the benchmark implementation. Consider these rough
+guidelines for how to run one of these benchmarks.
+
+Other benchmarks can be run with the same protocol. The following benchmarks
+have immediate usefulness, but other benchmarks can be added as we find them
+useful. Consult the [_QUIC Performance_
+Internet-Draft](https://datatracker.ietf.org/doc/html/draft-banks-quic-performance#section-2.3)
+for some other benchmarks (called _scenarios_ in the document).
+
+## Single connection throughput
+
+For an upload test, the client sets the the server response size 0 bytes, writes
+some amount of data and closes the stream.
+
+For a download test, the client sets the server response size to N bytes, and
+closes the write side of the data.
+
+The measurements are gathered and reported by the client by measuring how many
+bytes were transferred by the total time it took from stream open to stream
+close.
+
+A timer based variant is also possible where we see how much data a client can
+upload or download within a specific time. For upload it's the same as before
+and the client closes the stream after the timer ends. For download, the client
+should request a response size of max uint64, then close the stream after the
+timer ends.
+
+## Handshakes per second
+
+This benchmark measures connection setup efficiency. A transport that takes many
+RTTs will perform worse here than one that takes fewer.
+
+To run this benchmark:
+1. Set up N clients
+2. Each client opens K connections/s to a single server
+3. once a connection is established, the client closes it and establishes
+   another one.
+
+handshakes per second are calculated by taking the total number of connections
+successfully established and divide it by the time period of the test.
+
+# Security Considerations
+
+Since this protocol lets clients ask servers to do significant work, it
+SHOULD NOT be enabled by default in any implementation. Users are advised not to
+enable this on publicly reachable nodes.
+
+Authentacting by Peer ID could mitigate the security concern by only allowing
+trusted clients to use the protocol. Support for this is left to the implementation.
+
+# Prior Art
+
+As mentioned above, this document is inspired by Nick Bank's: [QUIC Performance Internet-Draft](https://datatracker.ietf.org/doc/html/draft-banks-quic-performance)
+
+[iperf](https://iperf.fr)
+
+[@mxinden's libp2p perf](https://github.com/mxinden/libp2p-perf)
+
+[@marten-seemann's libp2p perf test](https://github.com/marten-seemann/libp2p-perf-test/)
+
+[@vyzo's libp2p perf test](https://github.com/vyzo/libp2p-perf-test/)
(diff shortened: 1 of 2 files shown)


mxinden (approved):
Proposal looks good to me. Thanks for the write-up!

Is there any value in merging this early? If not, I would suggest holding off until we have a first implementation.

perf/perf.md
```diff
@@ -0,0 +1,121 @@
...
+but adapted to libp2p.
+
+
+The protocol is as a follows:
+
+Client:
+
+1. Open a libp2p stream to the server.
+2. Tell the server how many bytes we want the server to send us as a single
+   big-endian uint64 number. Zero is a valid number, so is the max uint64 value.
```
Menduist, line 45: Why not varint to stay consistent?

MarcoPolo: I don’t think there’s much benefit. u64 is simple enough and avoids adding the concept of varints.

perf/perf.md
```diff
@@ -0,0 +1,121 @@
...
+3. Write some amount of data to the stream.
```
achingbrain, line 46: Perhaps specify some way of requsting a chunk size as well?  For example "send me 100MB in 100MB chunks" (i.e. one write to the stream), or 100MB in 500KB chunks (i.e. 200 writes to the stream).

achingbrain: This is useful for benchmarking muxers, among other things.

perf/perf.md
```diff
@@ -0,0 +1,121 @@
...
+Internet-Draft](https://datatracker.ietf.org/doc/html/draft-banks-quic-performance#section-2.3)
+but adapted to libp2p.
+
+
+The protocol is as a follows:
+
+Client:
+
+1. Open a libp2p stream to the server.
+2. Tell the server how many bytes we want the server to send us as a single
```
achingbrain, line 44: Is it worth specifying the type of bytes too?  E.g. do they have to be random?  Can they be all 0s?  Something else?

perf/perf.md
```diff
@@ -0,0 +1,121 @@
...
+successfully established and divide it by the time period of the test.
+
+# Security Considerations
+
+Since this protocol lets clients ask servers to do significant work, it
+SHOULD NOT be enabled by default in any implementation. Users are advised not to
+enable this on publicly reachable nodes.
+
+Authentacting by Peer ID could mitigate the security concern by only allowing
+trusted clients to use the protocol. Support for this is left to the implementation.
```
bajtos, line 109: Should we also recommend the implementors to implement a configurable limit on the maximum size that clients can request?

For example, in web2 space, when handling file uploads from the browser, web servers and frameworks allow developers to specify the maximum allowed size of an uploaded file. Often, this setting comes with a reasonable default.

Maybe this spec can go into more details regarding the allowed sizes. For example, we can say that by default, perf protocol implementations should set this limit to 1GB (we can pick a different value).

The downside of my proposal is additional complexity. When the client requests larger data than the server is willing to return, then we need some way how to signal this error from the server back to the client.

perf/perf.md
MarcoPolo: Good idea. I think they should be a predetermined arbitrary set of bytes. Zero is fine, but so is random bytes (as long as it's not a different set of random bytes, because then you may be limited by how fast you can generate random bytes)

perf/perf.md
MarcoPolo: Thanks for taking a look!

I don't think we should set a limit here. The client is a fully trusted actor in this setup, so the server trusts it will only ask for reasonable values.

perf/perf.md
```diff
@@ -0,0 +1,121 @@
...
+Other benchmarks can be run with the same protocol. The following benchmarks
+have immediate usefulness, but other benchmarks can be added as we find them
+useful. Consult the [_QUIC Performance_
+Internet-Draft](https://datatracker.ietf.org/doc/html/draft-banks-quic-performance#section-2.3)
+for some other benchmarks (called _scenarios_ in the document).
+
+## Single connection throughput
+
+For an upload test, the client sets the the server response size 0 bytes, writes
+some amount of data and closes the stream.
```
elenaf9, line 73: What do you think about also adding a benchmark for opening and sending data on _n_ streams in parallel? Could be interesting for HOL-blocking when running benchmarks on TCP.

perf/perf.md
```diff
@@ -0,0 +1,121 @@
...
+# Context
+
+The `perf` protocol represents a standard benchmarking protocol that we can use
+to talk about performance within and across libp2p implementations. This lets us
+analyze peformance, guide improvements, and protect against regressions.
+
+# Protocol
+
+The `/perf/1.0.0` protocol (from here on referred to as simply `perf`) is a
+client driven set of benchmarks. To not reinvent the wheel, this perf protocol
```
elenaf9, line 33: Should it be possible to use this protocol alongside with other libp2p protocols on a client? I would assume it does make a difference in the benchmarks whether we only run the `perf` protocol on a connection, or also multiplex it with other protocols like identify or gossipsub.

perf/perf.md
nazar-pc: I suspect compression can be used in some cases, so pseudo-random bytes from seed should be preferred.

perf/perf.md
```diff
@@ -0,0 +1,121 @@
...
+   big-endian uint64 number. Zero is a valid number, so is the max uint64 value.
+3. Write some amount of data to the stream.
+4. Close the write side of our stream.
+5. Read from the read side of the stream. This
+   should be the same number of bytes as we told the server in step 2.
+
+Server, on handling a new `perf` stream:
+1. Read the big-endian uint64 number. This is how many bytes we'll send back.
+2. Read from the stream until we get an `EOF` (client's write side was closed).
+3. Send the number of bytes defined in step 1 back to the client.
```
nazar-pc, line 54: Would there be a benefit of sending things back at the same time at least as an option rather than waiting til the end?

mxinden:
Wrote a draft implementation of this specification on top of rust-libp2p. See https://github.com/libp2p/rust-libp2p/pull/3508.

> **Draft only for now. Don't (yet) expect this to produce reliable performance metrics.**
> 
> Server implementation is deployed at `/dns4/libp2p-perf.max-inden.de/tcp/4001`.
> 
> Easiest way to test from your machine:
> 
> ```
> docker run -ti --rm --entrypoint perf-client mxinden/libp2p-perf --server-address /dns4/libp2p-perf.max-inden.de/tcp/4001
> 
> Finished run: Sent 10 MiB in 9.497557834 s with 8.423217989114063 MiBit/s and received 10 MiB in 6.494584597 s with 12.317954875351667 MiBit/s
> ```
> 
> For server machine metrics, see https://libp2p-perf.max-inden.de/d/rYdddlPWk/node-exporter-full

perf/perf.md
marten-seemann, line 46: Why? Sending data isn't very useful without a response from the server.

Consider making this a new upload handler, where the server explicitly acknowledges that it received all the sent data.

mxinden: Would you like to see an explicit acknowledgement for correctness, i.e. to make sure the data really made it from A to B?

I don't think we need this correctness guarantee. In my eyes this is a performance testing protocol only.

If the client wants a response from the server, it can request a single byte by sending a `1` early on.

marten-seemann: I can achieve amazing performance if I just disable my congestion controller and dump all packets into the network. Sure, most of them will be lost due to congestion, but since the sending and receiving part are completely decoupled in this protocol, I wouldn't even notice.

I don't understand what the point of the client sending data after the initial integer in this protocol is.

MarcoPolo: The networking stack should guarantee that the bytes made it to the other end. We aren't checking for correctness here, we assume the networking stack works. If we read an EOF on the server and no other error while reading from the stream we assume we got all the bytes intended for us.

> I can achieve amazing performance if I just disable my congestion controller and dump all packets into the network

Your networking stack should retry the lost packets and the server doesn't progress until it sees the write side is closed. So if you had reliable delivery, you wouldn't get amazing performance this way.

That said, we could add some protection here where the server sends back a uint64 of bytes received as the first chunk of data when it starts sending the requested response data. The client can then check this number.

perf/perf.md
MarcoPolo: Would opening two streams in parallel be enough to satisfy this use case?

perf/perf.md
MarcoPolo: I think that's a great test. I'm leaning towards not including it here right now to keep this document small.

perf/perf.md
MarcoPolo: I could see it being useful in debugging the pathological case. But for a well behaved stack, once you fill a write buffer your network stack should be sending full packets back to back, so not sure this is useful as a runtime parameter vs manually editing the client/server code for debugging.

marten-seemann: I still don’t see why the client would send anything after the integer. The server doesn’t act upon this data, it’s just added complexity of draining the stream.

MarcoPolo: This is how you benchmark the client's upload speed.

marten-seemann: That doesn’t work, as the server doesn’t send any response once it receives the EOF. I’d be in favor of cleanly separating uploads and downloads into two separate handlers.

mxinden:
> Is there any value in merging this early? If not, I would suggest holding off until we have a first implementation.

https://github.com/libp2p/rust-libp2p/pull/3508 just merged in case you were holding off merging here till there is at least one implementation @MarcoPolo.

perf/perf.md
shushen: Maybe full-duplex throughput within a single stream is interesting as well? This could be an optional flag in the client's request.

The referenced [draft](https://datatracker.ietf.org/doc/html/draft-banks-quic-performance#section-2.3) does ask the question whether it should be possible to send the traffic from the server before FIN:
> *Note* - Should the server wait for FIN before replying?

perf/perf.md
```diff
@@ -0,0 +1,124 @@
...
+
+Other benchmarks can be run with the same protocol. The following benchmarks
+have immediate usefulness, but other benchmarks can be added as we find them
+useful. Consult the [_QUIC Performance_
+Internet-Draft](https://datatracker.ietf.org/doc/html/draft-banks-quic-performance#section-2.3)
+for some other benchmarks (called _scenarios_ in the document).
+
+## Single connection throughput
+
+For an upload test, the client sets the the server response size 0 bytes, writes
```
thomaseizinger, line 75:
```suggestion
For an upload test, the client sets the the server response size to 0 bytes, writes
```

perf/perf.md
```diff
@@ -0,0 +1,124 @@
...
+# Table of Contents <!-- omit in toc -->
+- [Context](#context)
+- [Protocol](#protocol)
+- [Benchmarks](#benchmarks)
+  - [Single connection throughput](#single-connection-throughput)
+  - [Handshakes per second](#handshakes-per-second)
+- [Security Considerations](#security-considerations)
+- [Prior Art](#prior-art)
+
+# Context
```
thomaseizinger, line 24:
```suggestion
## Context
```
 Nit: Multiple L1 headings in this document.

thomaseizinger (commented):
Couple of stylistic notes. Take however you wish :)

perf/perf.md
MarcoPolo: I think we can get most of the value with using two streams. The full-duplex is interesting, but I'd hold off from doing that until we have a compelling use case since it adds a bit of complexity to support that option.

perf/perf.md
MarcoPolo: Why is that a problem? This is a top level section. If we made this `##` then it would be a child of ToC, which is wrong.

perf/perf.md
MarcoPolo: Compression should not be used because you may be slowed down by your compressor.

perf/perf.md
```diff
@@ -0,0 +1,124 @@
...
+2. Tell the server how many bytes we want the server to send us as a single
+   big-endian uint64 number. Zero is a valid number, so is the max uint64 value.
+3. Optional: Write some amount of data to the stream.
+4. Close the write side of our stream.
+5. Read from the read side of the stream. This
+   should be the same number of bytes as we told the server in step 2.
+
+Server, on handling a new `perf` stream:
+1. Read the big-endian uint64 number. This is how many bytes we'll send back in step 3.
+2. Read from the stream until we get an `EOF` (client's write side was closed).
```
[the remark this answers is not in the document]
MarcoPolo: I'd opt for not including this since I'd argue that we don't need to define everything implementations SHOULD NOT do unless there's a high risk it's likely they would do so otherwise and it would be bad if they did. I don't think it's likely folks would buffer this unnecessarily and if they did it would either be harmless or show up in the perf.

perf/perf.md
MarcoPolo: Ah I see the argument now. This could all be a child of the `# Perf` part. I think that's a good argument, but the counter argument is that then you would always lose one nesting level in this style since you'll always have that title heading.

Unless we have some specific templating system we are trying to make happy, I don't think it's necessary to only have a single `#` level. This whole doc is about `perf`, so it should be okay to have multiple `#` levels.

Issue #475, opened 2022-11-09
webrtc/: Add libp2p WebRTC browser-to-browser spec [tracking issue]


p-shahi:
Now that the specification for browser to server is merged as a Working Draft: https://github.com/libp2p/specs/pull/412, we can begin the second iteration of the browser to browser aspect. See: https://github.com/libp2p/specs/pull/412/commits/a46919c557aa4501afb285ffbec8261a7a7e595b for the beginnings.

mxinden:
@p-shahi can we close here or do you want to track the concrete implementations e.g. https://github.com/libp2p/go-libp2p/issues/2009 here?

PR #482, opened 2022-11-18
chore: proposal to update WebRTC Browser to Server to Candidate Recommendation

p-shahi:
### Motivation
WebRTC Transport (browser to server) has landed in rust-libp2p here https://github.com/libp2p/rust-libp2p/commit/a7148648858fe10e9ba4c2793c7e12392b49c0ab so there now exists one reference implementation.
In accordance with the [spec lifecycle](https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md), we can update this spec from a Working Draft to a Candidate Recommendation.

### Open Questions
* How will introducing [browser to browser](https://github.com/libp2p/specs/issues/475) affect the maturity level of this spec? Previously it was specified inside this document. Perhaps it should be specified in a separate document so it can have it's own maturity level/better separation of concerns.
* Any other criteria to consider before promoting from 1A to 2A?

### DIFF
--- a/webrtc/README.md
+++ b/webrtc/README.md
@@ -363,10 +363,11 @@ accept streams before completion of the handshake.
 
 ## Previous, ongoing and related work
 
-- Work in progress implementations of this specification:
-  - https://github.com/little-bear-labs/js-libp2p-webrtc/
-  - https://github.com/libp2p/go-libp2p/pull/1655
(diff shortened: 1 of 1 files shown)


marten-seemann (approved):
Agreed that browser to browser should live in its own document.

Issue #457, opened 2022-09-20
revisit revisions


marten-seemann:
> If I am not mistaken, a non-breaking change qualifies as a revision bump
> > Revision numbers start with lowercase r followed by an integer, which gets bumped whenever the spec is modified by merging a new PR.
>
> https://github.com/libp2p/specs/blob/master/00-framework-02-document-header.md

Indeed, that's what the document says. We should change that. It doesn't make sense to bump the version for every change, e.g. in the extreme case fixing a typo.

_Originally posted by @marten-seemann in https://github.com/libp2p/specs/issues/456#issuecomment-1252906980_

MarcoPolo:
Would semver here make sense? I think it's useful to differentiate between backward incompatible changes and small changes.

marten-seemann:
I'd like to distinguish a few cases:
1. trivial changes: easy rewording, typo fixes, etc. I don't think we need to change anything. This is only creating noise.
2. backwards-compatible changes: e.g. adding a field to a Protobuf with optional functionality. In that case, a revision bump seems to make sense.
3. backwards-incompatible changes: This effectively means standardizing a new protocol, and should live in a new document. There will still be nodes out there in the network speaking the old protocol version, and we shouldn't require people to go back in the Git history to understand how that version of the protocol works.

PR #484, opened 2022-11-23
chore: update lifecycle and maturity

p-shahi:
* Corrected a few lifecycle stages: addressing, autonat, hole-punching (1A -> 3A), WebTransport (1A -> 2A)
* DCUtR to Recommendation (from Working Draft)
* circuit-v2 to Recommendation (from Draft)
* TLS to Recommendation (from Candidate Recommendation)

### DIFF
--- a/addressing/README.md
+++ b/addressing/README.md
@@ -1,9 +1,9 @@
-# Addressing in libp2p
+# Addressing in libp2p  <!-- omit in toc -->
 > How network addresses are encoded and used in libp2p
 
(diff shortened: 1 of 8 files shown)


mxinden (commented):
:pray: thanks for the clean-ups Prithvi!

Just as a note, I think at some point we should touch the specification process specification. While it is great to have one, I think it overspecifies a bit (e.g. # of days between stages). See also https://github.com/libp2p/specs/issues/457.

mxinden (approved):

Issue #340, opened 2021-06-29
Formatting of specification documents


thomaseizinger:
I've noticed that most (all?) spec documents are formatted with hard line breaks based on line width.

I would like to suggest an alternative way of formatting specs, in particular using [semantic line breaks](https://sembr.org/).

The main advantage I would point out is that natural language tends to be changed on a per-sentence basis, i.e. add a complete sentence, rewrite an existing one, split one into two, merge two into one, etc.

Git on the other hand works on a line-by-line basis. If newlines occur only after each sentence or at meaningful places within a sentence, what shows up in the diff is much closer to the semantic changes that were made.

Note that semantic line breaks are allowed mid-sentence. Consequently, readability of the specs in plain-text shouldn't suffer too much from such a change, esp. if we strive for short sentences at the same time.

If people want to stick to the current formatting that is also fine by me but then I would suggest a script or tool to be added to the repository that produces the desired formatting. Manually formatting things is tedious.

mxinden:
Off the top of my head, given that it increases the barrier (for newcomers) to contribute,  I would deem it not worth the effort. That said, this is not a strong opinion.

As a middle-ground, I would very much appreciate if all specs would be wrapped at 80 (or X) characters, though not sure whether others see that as worth a debate by itself.

thomaseizinger:
Okay, I'll see to just automate it on my end then :)

To be perfectly honest, personally I've found it to be a bigger barrier to follow guidelines I have to guess (text wrapping at X characters) because there is no tooling to help / enforce it.

Might just be my OCD kicking in though :P

PR #490, opened 2022-12-05
chore: typo in TTL upper-bound + additional writing style updates

DannyS03:
## Context

- I believe:
  ```
  There may be a rendezvous point-specific upper bound on TTL, with a **minimum** such value of 72hrs
  ```

  should be:
  ```
  There may be a rendezvous point-specific upper bound on TTL, with a **maximum** such value of 72hrs
  ```

- General writing style tweaks.

### DIFF
--- a/rendezvous/README.md
+++ b/rendezvous/README.md
@@ -16,7 +16,7 @@ Interest Group: [@daviddias], [@whyrusleeping], [@Stebalien], [@jacobheun], [@yu
 [@yusefnapora]: https://github.com/yusefnapora
 [@vasco-santos]: https://github.com/vasco-santos
 
-See the [lifecycle document][lifecycle-spec] for context about maturity level
+See the [lifecycle document][lifecycle-spec] for context about the maturity level
 and spec status.
 
 [lifecycle-spec]: https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md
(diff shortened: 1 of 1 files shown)


rendezvous/README.md
```diff
@@ -16,7 +16,7 @@ Interest Group: [@daviddias], [@whyrusleeping], [@Stebalien], [@jacobheun], [@yu
 [@yusefnapora]: https://github.com/yusefnapora
 [@vasco-santos]: https://github.com/vasco-santos
 
-See the [lifecycle document][lifecycle-spec] for context about maturity level
+See the [lifecycle document][lifecycle-spec] for context about the maturity level
```
mxinden, line 19: Note that this is the same across all the specifications. In case you would like to see it changed, can you please propose it in a new pull request updating all the specifications?

mxinden (commented):
@thomaseizinger as you are most familiar with this specification, would you mind taking a look?

thomaseizinger (approved):
LGTM!

It is a pet peeve of mine but when reviewing this, I really wished [we would put each sentence on a new line](https://github.com/libp2p/specs/issues/340). Makes diffs a lot easier to read!

rendezvous/README.md
DannyS03: Yes, planning to go through some of the other specs as an opportunity to update writing style.

mxinden: Sorry potentially coming across as nit-picky here. Let's do this change in an atomic fashion, i.e. either all at once or not at all.

PR #491, opened 2022-12-06
identify:  length limit for unicode fields

lidel:
This PR formalizes ~normalization and~ length limit per string field ( `agentVersion`, `protocolVersion` and `protocols` array).

The goal is to reduce surprises and unify behavior across implementations. 

- Kubo PR: https://github.com/ipfs/kubo/pull/9465

### DIFF
--- a/identify/README.md
+++ b/identify/README.md
@@ -96,7 +96,7 @@ protocol families / networks.
 
 Example value: `/my-network/0.1.0`.
 
-Implementations SHOULD limit the string to 128 runes (Unicode code points) when
+Implementations SHOULD limit the string to 128 Unicode code points when
 displaying or processing this field. When displaying in terminals, logs, or user
 interfaces, implementations SHOULD sanitize the string as described in the
 [Unicode Sanitization](#unicode-sanitization) section below.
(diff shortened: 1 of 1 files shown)


marten-seemann (commented):
Can we make this UTF-8? I'd really like to not introduce any surprises here.

marten-seemann:
Nice UTF-8 art! :)

I don't really see reason not to. Limiting ourselves to ASCII is so 1990s style.

Winterhuman:
Perhaps this could be phrased as: "Implementations should discard non-ASCII characters and trim the string to 64 characters, but may choose to allow UTF-8 characters if potential for UTF-8 art/mimicry is acceptable"

I definitely wouldn't want UTF-8 support to be outright gone, using UTF-8 in protocol names (maybe containing CIDs with UTF-8 encodings) could have a lot of potential use-cases. For the agent and version strings though, that's perfectly understandable

Winterhuman:
In fact, maybe better idea, what about:

"Implementations should trim the string to 64 characters. Implementations MAY allow UTF-8 characters in the string, however, these strings should be visible to users as both UTF-8 and ASCII punycode (per IETF RFC 3492) to protect against UTF-8 art/mimicry."

marten-seemann:
I'd say let's fully embrace UTF-8. This is 2022, and we finally have a standard encoding that's universally supported.

Building on @Winterhuman's proposal:
> "Strings are UTF-8 encode. Implementations MAY trim the string to 64 characters. When made visible to users, implementations MAY output both UTF-8 and ASCII punycode (per IETF RFC 3492) to protect against UTF-8 mimicry."

MarcoPolo (commented):
I'd just call them Unicode code points rather than also call them runes.

PR #502, opened 2023-01-02
identify: add a sequence number to the Identify protobuf

### LINKED ISSUE #501: identify: multiple Push messages are racy
When a peer send multiple Identify push messages in succession, the outcome of this will be racy, since there's no ordering guarantee between the two Identify Push streams.

This can easily be solved by adding a sequence number to the Identify message.

marten-seemann:
Fixes #501.

### DIFF
--- a/identify/README.md
+++ b/identify/README.md
@@ -5,11 +5,11 @@
 
 | Lifecycle Stage | Maturity Level | Status | Latest Revision |
 |-----------------|----------------|--------|-----------------|
-| 3A              | Recommendation | Active | r1, 2021-08-09  |
+| 3A              | Recommendation | Active | r2, 2023-01-03  |
 
 Authors: [@vyzo]
 
-Interest Group: [@yusefnapora], [@tomaka], [@richardschneider], [@Stebalien], [@bigs]
+Interest Group: [@yusefnapora], [@tomaka], [@richardschneider], [@Stebalien], [@bigs], [@marten-seemann]
 
 [@vyzo]: https://github.com/vyzo
 [@yusefnapora]: https://github.com/yusefnapora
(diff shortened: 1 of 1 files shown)


identify/README.md
```diff
@@ -83,6 +86,13 @@ message Identify {
 }
 ```
 
+### sequence
+
+A sequence number for this identify message used to deal with reordered Identify
+messages. Senders SHOULD increment the sequence number for every Identify message
+they send. After processing an Identify with a certain sequence number, receivers
+MUST ignore all Identify messages with smaller sequence numbers.
```
mxinden, line 94: I am assuming that the monotonic sequence number is global, i.e. not per peer nor per connection, correct?

What happens on restarts or crashes? I am assuming that nodes are not expected to persist their sequence numbers, correct?

A timestamp instead of a simple counter might be an option. Though what happens if the node drifts in time? One could require that the timestamp is monotonically increasing, i.e. that time adjustments are only ever allowed to happen in forward direction.

Menduist: We can reuse this https://github.com/libp2p/specs/blame/master/RFC/0003-routing-records.md#L115-L119

marten-seemann: Yeah, I was planning to use a (monotonic) clock here, with some extra protections to make sure that we don't send the same sequence number twice. This will break if we send more than one Identify message per nanosecond, but I think we can live with that :)

mxinden: > We can reuse this https://github.com/libp2p/specs/blame/master/RFC/0003-routing-records.md#L115-L119

Good idea @Menduist. I suggest adopting that paragraph verbatim here.

MarcoPolo (changes requested):
Once the wording change is in, request a review and I'll approve

identify/README.md
marten-seemann: Added in d9de900.

marten-seemann:
Sorry for the delay, I just added the paragraph as suggested.

mxinden (approved):
Thanks for the addition.

MarcoPolo (approved):

p-shahi:
This PR has the sufficient approvals, @marten-seemann can this be merged?

marten-seemann:
@p-shahi Maybe, that's for you guys to decide now...

Looking back at the text, I'm not sure I like the fact that this is cross-connection.

PR #505, opened 2023-01-19
update(specs): non-test-net addresses -> special-purpose addresses for docs

DannyS03:
## Context

- This change aims to update all the addresses across the specs to use the special-purposes addresses specified in [RFC 6890](https://datatracker.ietf.org/doc/rfc6890/).

  - 192.0.2.0/24 (TEST-NET-1)
  - 198.51.100.0/24 (TEST-NET-2)
  - 203.0.113.0/24 (TEST-NET-3)

- Some lint-based updates included.

### DIFF
--- a/_archive/3-requirements.md
+++ b/_archive/3-requirements.md
@@ -12,13 +12,13 @@ In order to reason about possible transports, `libp2p` uses [multiaddr](https://
 /ip6/fe80::8823:6dff:fee7:f172/tcp/4001/ipfs/QmYJyUMAcXEw1b5bFfbBbzYu5wyyjLMRHXGUkCXpag74Fu
 
 # IPFS over uTP over UDP over IPv4 (UDP-shimmed transport)
-/ip4/203.0.113.0/udp/4001/utp/ipfs/QmYJyUMAcXEw1b5bFfbBbzYu5wyyjLMRHXGUkCXpag74Fu
+/ip4/162.246.145.218/udp/4001/utp/ipfs/QmYJyUMAcXEw1b5bFfbBbzYu5wyyjLMRHXGUkCXpag74Fu
 
 # IPFS over IPv6 (unreliable)
 /ip6/fe80::8823:6dff:fee7:f172/ipfs/QmYJyUMAcXEw1b5bFfbBbzYu5wyyjLMRHXGUkCXpag74Fu
 
 # IPFS over TCP over IPv4 over TCP over IPv4 (proxy)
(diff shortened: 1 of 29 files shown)


_archive/3-requirements.md
```diff
@@ -12,13 +12,13 @@ In order to reason about possible transports, `libp2p` uses [multiaddr](https://
 /ip6/fe80::8823:6dff:fee7:f172/tcp/4001/ipfs/QmYJyUMAcXEw1b5bFfbBbzYu5wyyjLMRHXGUkCXpag74Fu
 
 # IPFS over uTP over UDP over IPv4 (UDP-shimmed transport)
-/ip4/162.246.145.218/udp/4001/utp/ipfs/QmYJyUMAcXEw1b5bFfbBbzYu5wyyjLMRHXGUkCXpag74Fu
+/ip4/203.0.113.0/udp/4001/utp/ipfs/QmYJyUMAcXEw1b5bFfbBbzYu5wyyjLMRHXGUkCXpag74Fu
```
mxinden, line 15: I don't think there is value in touching any of the archived files. They are here for historic purposes only.

rendezvous/README.md
```diff
@@ -2,7 +2,7 @@
 
 | Lifecycle Stage | Maturity      | Status | Latest Revision |
 |-----------------|---------------|--------|-----------------|
-| 1A              | Working Draft | Active | r3, 2021-07-12  |
+| 1A              | Working Draft | Active | r3, 2022-12-20  |
```
mxinden, line 5: None of the below changes semantics, right? If so, I don't think the revision bump is needed here.

rendezvous/README.md
```diff
@@ -184,8 +191,8 @@ The protocol, as described so far, is susceptible to spam attacks from
 adversarial actors who generate a large number of peer identities and
 register under a namespace of interest (e.g., the relay namespace).
 
-It is TBD how exactly the protocol will mitigate such attacks.
-See https://github.com/libp2p/specs/issues/341 for a discussion on this
+It is TBD how exactly such attacks will be mitigated.
+See <https://github.com/libp2p/specs/issues/341> for a discussion on this
```
mxinden, line 195: In my eyes the previous wording made it clear, that it is the protocol that _should_ mitigate against the attacks.

mxinden (commented):
Thanks for the clean-ups.

rendezvous/README.md
DannyS03: Sorry, this was accidental (lint tool).

_archive/3-requirements.md
DannyS03: Good point.

mxinden (approved):

PR #497, opened 2022-12-15
feat(webrtc): add WebRTC (prev. browser-to-browser) spec

mxinden:
The libp2p WebRTC (prev. browser-to-browser) specification.

- Overarching tracking issue https://github.com/libp2p/specs/issues/220
- Browser-to-browser tracking issue: https://github.com/libp2p/specs/issues/475
- [Browser-to-server specification](https://github.com/libp2p/specs/blob/master/webrtc/README.md) added through https://github.com/libp2p/specs/pull/412.

Open items:

- [x] Signaling protocol name https://github.com/libp2p/specs/pull/497#discussion_r1073585736
- [x] Mechanism to advertise support for browser-to-browser https://github.com/libp2p/specs/pull/497#discussion_r1083223169

PR #506, opened 2023-01-21
Recommend proto3

### LINKED ISSUE #465: Recommend proto3
There seems to have been a misunderstanding in the past around proto2 vs proto3. My attempt here is to clear up the confusion, recommend proto3 in general, and explain why proto3 should be preferred.

Our main confusion is about field presence. That is, if a field is omitted from the serialized wire format does the user of the decoded message know the difference between if the field was unset or set as the default value. This document has a lot of good information and is worth the read: https://github.com/protocolbuffers/protobuf/blob/main/docs/field_presence.md

## Origins of the confusion

Proto2 would always serialize an explicitly set field, even if it was set to the default. This meant that you could know on the decoding side whether the field was set or not. This is called _Explicit Presence_. For example, in the Rust protobuf compiler, it would wrap these in `Option<T>`: https://github.com/tokio-rs/prost#field-modifiers. 

The confusing thing is that the language guide for [proto2](https://developers.google.com/protocol-buffers/docs/proto#optional) states:
> A well-formed message may or may not contain an optional element. When a message is parsed, if it does not contain an optional element, accessing the corresponding field in the parsed object returns the default value for that field.

The subtlety here is that this doesn't say anything about "hasField" accessors. Which may be provided by the implementation to check if the field was set or not. This is essentially with prost is doing with `Option<T>` types.

Another confusing thing is that this language guide doesn't mention "presence" a single time. Which is what we're talking about here.

In proto3, if a field was set to its default value it would not be serialized. This meant that the decoding sided wouldn't know if the field was omitted because it was unset or because it was the default value. This is called _No Presence_.

## Field Presence Proto2 vs Proto3
To clarify field presence in proto2 vs proto3:

From https://github.com/protocolbuffers/protobuf/blob/main/docs/field_presence.md#presence-in-proto2-apis

### Proto2

Field type                                   | Explicit Presence
-------------------------------------------- | -----------------
Singular numeric (integer or floating point) | ✔️
Singular enum                                | ✔️
Singular string or bytes                     | ✔️
Singular message                             | ✔️
Repeated                                     |
Oneofs                                       | ✔️
Maps                                         |

### Proto3

Field type                                   | `optional` | Explicit Presence
-------------------------------------------- | ---------- | -----------------
Singular numeric (integer or floating point) | No         |
Singular enum                                | No         |
Singular string or bytes                     | No         |
Singular numeric (integer or floating point) | Yes        | ✔️
Singular enum                                | Yes        | ✔️
Singular string or bytes                     | Yes        | ✔️
Singular message                             | Yes        | ✔️
Singular message                             | No         | ✔️
Repeated                                     | N/A        |
Oneofs                                       | N/A        | ✔️
Maps                                         | N/A        |

## Advantages in Proto3 compared to Proto2

* No `required` modifier
  * This is generally considered an [anti-pattern](https://developers.google.com/protocol-buffers/docs/proto#specifying-rules) since all future versions of this message will need to contain this field. Generally users should prefer custom validation.
* Opt-in explicit presence
  * It's good to be able to get the space advantages of no-presence while still being able to opt-in to explicit presence. If we pass in an empty byte array most of the time this is semantically the same as passing no byte array, so it's nice to avoid paying the byte-cost for this.
  * But if we do want explicit presence we can opt in to it. This is useful in case we do semantically care about knowing if there was _nothing_ set.
* Simple feature set: "The reason for removing these features is to make API designs simpler, more stable, and more performant. " from https://cloud.google.com/apis/design/proto3
* Better ecosystem support. As libraries develop, it's likely they will support the latest protobuf spec rather than continue supporting proto2. This is already the case with [protons](https://github.com/ipfs/protons), the compiler that JS-IP uses (see this [bug](https://github.com/ipfs/protons/issues/34)).
* User-defined default value for fields is no longer available. This was somewhat tricky to get right.

## Next steps

1. Come to mutual understanding around proto2 vs proto3.
1. Come to consensus around recommending proto3.
1. Make the change to `README.md`

MarcoPolo:
Updates the README to recommend proto3. Makes all existing protobuf without an explicit syntax have `syntax="proto2"`.

See #465 for the discussion around this change.

Closes #465

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,14 @@ and are used throughout the system.
 These specs define wire protocols that are used by libp2p for connectivity,
 security, multiplexing, and other purposes.
 
-The protocols described below all use [protocol buffers](https://developers.google.com/protocol-buffers/docs/proto?hl=en) (aka protobuf) to define message schemas. Version `proto2` is used unless stated otherwise.
+The protocols described below all use [protocol
+buffers](https://developers.google.com/protocol-buffers/docs/proto?hl=en) (aka
+protobuf) to define message schemas.
+
+Existing protocols may use `proto2`, and continue to use them. `proto3` is
+recommended for new protocols. `proto3` is a simplification of `proto2` and
+removes some footguns. For context and a discussion around `proto3` vs `proto2`,
+see [#465](https://github.com/libp2p/specs/issues/465).
 
 - [ping][spec_ping] - Ping protocol
 - [autonat][spec_autonat] - NAT detection
--- a/fetch/README.md
+++ b/fetch/README.md
@@ -88,6 +88,8 @@ The libp2p protocol ID for this protocol is `/libp2p/fetch/0.0.1`
 The messages in the Fetch protocol use on of the following protobufs (proto3 syntax):
 
 ```
+syntax = "proto2";
+
 message FetchRequest {
 	string identifier = 1;
 }
--- a/identify/README.md
+++ b/identify/README.md
@@ -73,6 +73,7 @@ whose values have changed.
 ## The Identify Message
 
 ```protobuf
+syntax = "proto2";
 message Identify {
   optional string protocolVersion = 5;
   optional string agentVersion = 6;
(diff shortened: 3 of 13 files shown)


marten-seemann (approved):

mxinden (approved):
This looks good to me. Thanks for the follow-up across the whole repository.

mxinden:
I will update #497 once this is merged.

PR #511, opened 2023-01-31
Update circuitv2 to proto3

### LINKED ISSUE #509: circuitv2: upgrade definition to proto3
- [ ] spec: https://github.com/libp2p/specs/pull/511
- [ ] go-libp2p: https://github.com/libp2p/go-libp2p/pull/2029
- [ ] rust-libp2p: (link to pr)

The current protobuf definitions for circuit v2 use proto2. For the new js implementation, it would be preferable to use proto3, since js-libp2p uses proto3 almost everywhere and protons only supports proto3. This change would require adding a default field to the `Status` enum as it is not currently compatible with proto3.

Example updated protobuf:
```protobuf
syntax = "proto3";
message HopMessage {
  enum Type {
    RESERVE = 0;
    CONNECT = 1;
    STATUS = 2;
  }

  Type type = 1;

  optional Peer peer = 2;
  optional Reservation reservation = 3;
  optional Limit limit = 4;

  optional Status status = 5;
}

message StopMessage {
  enum Type {
    CONNECT = 0;
    STATUS = 1;
  }

  Type type = 1;

  optional Peer peer = 2;
  optional Limit limit = 3;

  optional Status status = 4;
}

message Peer {
  bytes id = 1;
  repeated bytes addrs = 2;
}

message Reservation {
  uint64 expire = 1; // Unix expiration time (UTC)
  repeated bytes addrs = 2;   // relay addrs for reserving peer
  optional bytes voucher = 3; // reservation voucher
}

message Limit {
  optional uint32 duration = 1; // seconds
  optional uint64 data = 2;     // bytes
}

enum Status {
  // zero value field required for proto3 compatibility
  UNSPECIFIED             = 0;
  OK                      = 100;
  RESERVATION_REFUSED     = 200;
  RESOURCE_LIMIT_EXCEEDED = 201;
  PERMISSION_DENIED       = 202;
  CONNECTION_FAILED       = 203;
  NO_RESERVATION          = 204;
  MALFORMED_MESSAGE       = 400;
  UNEXPECTED_MESSAGE      = 401;
}
```

ckousik:
Updates the circuit v2 spec to use proto3. This closes #509.

### DIFF
--- a/relay/circuit-v2.md
+++ b/relay/circuit-v2.md
@@ -4,7 +4,7 @@ This is the version 2 of the libp2p Circuit Relay protocol.
 
 | Lifecycle Stage | Maturity        | Status | Latest Revision |
 |-----------------|-----------------|--------|-----------------|
-| 3A              | Recommendation  | Active | r1, 2021-12-17  |
+| 3A              | Recommendation  | Active | r2, 2023-01-31  |
 
 Authors: [@vyzo]
 
--- a/relay/circuit-v2.md
+++ b/relay/circuit-v2.md
@@ -348,11 +348,11 @@ The envelope domain is `libp2p-relay-rsvp` and uses the multicodec code `0x0302`
 
 The payload of the envelope has the following form, in canonicalized protobuf format:
 ```protobuf
-syntax = "proto2";
+syntax = "proto3";
 message Voucher {
-  required bytes relay = 1;
-  required bytes peer = 2;
-  required uint64 expiration = 3;
+  bytes relay = 1;
+  bytes peer = 2;
+  uint64 expiration = 3;
 }
 ```
 - the `relay` field is the peer ID of the relay.
@@ -365,7 +365,7 @@ The wire representation is canonicalized, where elements of the message are writ
 ## Protobuf
 
 ```protobuf
-syntax = "proto2";
+syntax = "proto3";
 message HopMessage {
   enum Type {
     RESERVE = 0;
@@ -373,7 +373,7 @@ message HopMessage {
     STATUS = 2;
   }
 
-  required Type type = 1;
+  Type type = 1;
 
   optional Peer peer = 2;
   optional Reservation reservation = 3;
@@ -388,7 +388,7 @@ message StopMessage {
     STATUS = 1;
   }
 
-  required Type type = 1;
+  Type type = 1;
 
   optional Peer peer = 2;
   optional Limit limit = 3;
@@ -397,12 +397,12 @@ message StopMessage {
 }
 
 message Peer {
-  required bytes id = 1;
+  bytes id = 1;
   repeated bytes addrs = 2;
 }
 
 message Reservation {
-  required uint64 expire = 1; // Unix expiration time (UTC)
+  uint64 expire = 1; // Unix expiration time (UTC)
   repeated bytes addrs = 2;   // relay addrs for reserving peer
   optional bytes voucher = 3; // reservation voucher
 }
@@ -413,6 +413,8 @@ message Limit {
 }
 
 enum Status {
+  // zero value field required for proto3 compatibility
+  UNUSED                  = 0;
   OK                      = 100;
   RESERVATION_REFUSED     = 200;
   RESOURCE_LIMIT_EXCEEDED = 201;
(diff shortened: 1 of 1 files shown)


p-shahi:
Can you bump the revision (to r2) and date on line 7?

relay/circuit-v2.md
```diff
@@ -365,15 +365,15 @@ The wire representation is canonicalized, where elements of the message are writ
...
+syntax = "proto3";
 message HopMessage {
   enum Type {
     RESERVE = 0;
     CONNECT = 1;
     STATUS = 2;
   }
 
-  required Type type = 1;
+  Type type = 1;
```
achingbrain, line 376: Are we expecting this to be backwards compatible with existing implementations that are expecting `proto2`?

If so this will break them when the enum value is `0` since it will be omitted on the wire as it's now a `singular` field and the rules are different.

Singular: https://developers.google.com/protocol-buffers/docs/proto3#specifying_field_rules

> It will be serialized to the wire unless it is the default value. 

Interestingly Google's protobuf.js gets this wrong - https://github.com/protobufjs/protobuf.js/issues/1822

Parsers following the older `proto2` schema will error as by their definition the field is `required` which means it needs to be present on the wire.

The only way for that to happen with a `proto3` definition is to mark the field `optional` and, at the application level, ensure it has a value set.

MarcoPolo: @ckousik can you add some fuzzing tests to https://github.com/ckousik/proto2to3? The tests should have caught that.

ckousik: @MarcoPolo If we consider a case where we try to send `HopMessage { type: RESERVE }`, it should serialize to an empty buffer. I wrote a test for proto2 to decode the empty buffer and it passes. https://github.com/ckousik/proto2to3/blob/master/src/main.rs#L117

MarcoPolo: Ah. Prost does not error here, although it should. See https://github.com/tokio-rs/prost#field-modifiers

Go would error here.

MarcoPolo: Two options:
1. Change all formerly required fields to `optional` fields so that serializers **always** serialize the value. Try to find any other cases of incompatibility. And prove that we are backwards compatible.
2. Keep proto2 and have protons do the correct thing.

2 is the preferred option I think it lets us avoid the compatibility problem completely. But I recognize that it's not easy to do. And little value if protocols from now on are going to use proto3. How did this work before? Most everything so far has been proto2.

1 Makes sense if we can prove that we won't have any compatibility problems. As @achingbrain pointed out, this is tricky since implementations can do things that aren't technically what the protobuf spec requires (error on missing required field for example).

A suggestion to prove 1 would be to have both go and rust generate random valid messages (fuzzing) with proto3 and make sure the proto2 implementation can correctly parse those messages. If we have this I'm okay with option 1.

I know this is a bit annoying and hard. But so is backwards compatibility :)

MarcoPolo: I'll do 1. and post an update on #509

MarcoPolo: I've updated this PR.

MarcoPolo (approved):
This looks good. I'll ask @achingbrain for a review as well here since they've thought a lot about proto2 vs proto3 and they caught the first bug :).

relay/circuit-v2.md
thomaseizinger: > I've run the experiments in [MarcoPolo/proto3-and-2-compat-experiments](https://github.com/MarcoPolo/proto3-and-2-compat-experiments). Could you update the PR to use this proto3 file? [MarcoPolo/proto3-and-2-compat-experiments@`main`/rust/src/circuitv2proto3.proto](https://github.com/MarcoPolo/proto3-and-2-compat-experiments/blob/main/rust/src/circuitv2proto3.proto?rgh-link-date=2023-02-01T02%3A15%3A34Z)

We are in the middle of transitioning to `quick-protobuf`. Let me add that to your experiments before you merge here please.

thomaseizinger (commented):
I'd like to include `quick-protobuf` in the experiments as we are currently transitioning to that in `rust-libp2p`. I'll report back with some results.

MarcoPolo:
quick-protobuf experiment looks good as well.

p-shahi:
@achingbrain pinging for re-review

achingbrain (approved):
LGTM

PR #515, opened 2023-02-13
better motivation for libp2p+HTTP

marten-seemann:


http/README.md
```diff
@@ -19,6 +19,14 @@ At the same time, nodes that are already connected via a libp2p connection, will
 
 Any protocol that follows request-response semantics can easily be mapped onto HTTP (mapping protocols that don’t follow a request-response flow can be more challenging). Protocols are encouraged to follow best practices for building REST APIs. Once a mapping has been defined, a single implementation can be used to serve both traditional libp2p as well as libp2p-HTTP clients.
 
+Specifically, using libp2p+HTTP will allow:
+
+1. HTTP servers to offer same endpoints over HTTP and libp2p streams, allowing clients to reuse existing libp2p connections
```
MarcoPolo, line 24: Is this a main selling point? I don't care if I have 2 connections vs a single one. Although I guess the reduced latency could be good.

http/README.md
```diff
@@ -19,6 +19,14 @@ At the same time, nodes that are already connected via a libp2p connection, will
...
+1. Defining services / protocols that are run on both public nodes and on nodes behind NATs / firewalls, making use of libp2p's NAT traversal magic
```
MarcoPolo, line 25: And leverage libp2p's connectivity story.

http/README.md
```diff
@@ -19,6 +19,14 @@ At the same time, nodes that are already connected via a libp2p connection, will
...
+1. Use existing peer and content discovery mechanisms to advertise HTTP-enabled multiaddresses, which can then be accessed either via plain HTTP(S) or via HTTP on top of libp2p
+1. Use peer authentication (both client and server auth) for a subset of HTTP endpoints
+
+
```
MarcoPolo, line 29:
```suggestion
1. Support edge compute directly. Many edge compute environments build on top of HTTP since it's a stateless request/response protocol. This includes services such as Cloudflare workers, AWS Lambda, Netflify Edge functions, and many more.
```

http/README.md
```diff
@@ -19,6 +19,14 @@ At the same time, nodes that are already connected via a libp2p connection, will
 
 Any protocol that follows request-response semantics can easily be mapped onto HTTP (mapping protocols that don’t follow a request-response flow can be more challenging). Protocols are encouraged to follow best practices for building REST APIs. Once a mapping has been defined, a single implementation can be used to serve both traditional libp2p as well as libp2p-HTTP clients.
 
+Specifically, using libp2p+HTTP will allow:
+
+1. HTTP servers to offer same endpoints over HTTP and libp2p streams, allowing clients to reuse existing libp2p connections
+1. Defining services / protocols that are run on both public nodes and on nodes behind NATs / firewalls, making use of libp2p's NAT traversal magic
+1. Use existing peer and content discovery mechanisms to advertise HTTP-enabled multiaddresses, which can then be accessed either via plain HTTP(S) or via HTTP on top of libp2p
+1. Use peer authentication (both client and server auth) for a subset of HTTP endpoints
+
```
MarcoPolo, line 28:
```suggestion
1. Support existing HTTP protocols like the S3 protocol. This would allow peers to fetch content seamlessly from an S3-compatible provider (S3, backblaze's B2, Cloudflare's R2).
```

For example, I imagine us shipping the S3 protocol support as part of go-libp2p. Then the provider records could include something like `/dns4/someS3Provider/https/s3/`. A client will see that provider when doing a content route lookup, and then be able to get the data directly. This could be piped to some other function that will verify the content hash.

MarcoPolo: This and the below point are the things I'm most excited about. Having edge DHT nodes would be really cool

http/README.md
marten-seemann: Definitely not. I'll move it down the list.

marten-seemann: Actually, I'll just remove it.

PR #518, opened 2023-02-16
autonat: document that the observed address may also be dialed

marten-seemann:
I don't like that this is the case, but that's what the Go implementation has been doing since forever, and for better or worse, we're stuck with it. The purpose of this PR is to document this behavior.

cc @sukunrt

### DIFF
--- a/autonat/README.md
+++ b/autonat/README.md
@@ -3,7 +3,7 @@
 
 | Lifecycle Stage | Maturity       | Status | Latest Revision |
 |-----------------|----------------|--------|-----------------|
-| 3A              | Recommendation | Active | r0, 2021-08-26  |
+| 3A              | Recommendation | Active | r1, 2023-02-16  |
 
 
 Authors: [@marten-seemann]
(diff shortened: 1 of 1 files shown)


mxinden (commented):
Given that the remote peer replaces the requested addresses with the observed IP anyways, why do you consider the current go-libp2p behavior bad @marten-seemann?

> In order to prevent attacks like the one described in [RFC 3489, Section
12.1.1](https://www.rfc-editor.org/rfc/rfc3489#section-12.1.1) (see excerpt
below), implementations MUST NOT dial any multiaddress unless it is based on the
IP address the requesting node is observed as.

marten-seemann:
> Given that the remote peer replaces the requested addresses with the observed IP anyways, why do you consider the current go-libp2p behavior bad @marten-seemann?
> 
> > In order to prevent attacks like the one described in [RFC 3489, Section
> > 12.1.1](https://www.rfc-editor.org/rfc/rfc3489#section-12.1.1) (see excerpt
> > below), implementations MUST NOT dial any multiaddress unless it is based on the
> > IP address the requesting node is observed as.

It prevents us from using AutoNAT v1 to test the reachability of different addresses. The observed address might be a TCP address, while what I actually want to test if I'm reachable via QUIC.

mxinden:
Makes sense. Thank you @marten-seemann.

mxinden (approved):

Issue #479, opened 2022-11-17
Proposal: provide error codes when closing connections and resetting streams


marten-seemann:
It would be really helpful to know _why_ a peer closed a connection or reset a stream. Unfortunately, we currently don’t have access to that information.

Here’s a proposal how to convey that piece of information.

## Connection Termination

Current situation:
* QUIC: uses a CONNECTION_CLOSE frame, which carries a 62 bit error code and a human-readable message (a string limited by the MTU).
* WebTransport: uses a CLOSE_WEBTRANSPORT_SESSION capsule, which carries a 32 bit error code and a human-readable message (up to 8k)
* yamux: has a GOAWAY frame that abuses the length field (32 bit) to carry an error code. Currently the spec only defines 3 distinct error codes
* mplex: don’t care

It seems straightforward to use a 32 bit error code space for libp2p. If we decide that transmitting an error message is important, we might be able to find a backwards-compatible yamux hack, similar to the one described in the next section.

We could have different error codes for: connections that are closed because they were dial-raced with other connections, disallowed by a connection gater, closed due to resource limitations, closed to make room for more valuable connections, closed for different kinds of protocol violations, etc.

Caveat: With TCP linger set to 0, the TCP connection is reset instead of properly closed. This also means that the error code might not be transmitted reliably.

## Stream Termination

Current situation:
* QUIC: the RESET_STREAM frames contains a 62 bit error code field (there are no human-readable messages for stream resets)
* WebTransport: limits stream reset error codes to 8 bits
* yamux: doesn’t allow transmitting any error code
* mplex: still don’t care

It seems like we’re therefore limit to 256 error codes. We’d need to reserve a subset of these for libp2p itself (for example, we need to convey that multistream negotiation failed, or that we didn’t even start multistream negotiation because of resource limits, etc.). The rest of the error codes would be defined by the application.

### yamux hack

Depending on how current implementations handle this, we could either:
* if implementations ignore data sent on frames that have the RESET flag set: attach the error code to that frame
* if implementations ignore stream data received on a stream that was set (I think go does): send the error code in a stream frame

marten-seemann:
We've been using a few error codes without having them defined beforehand. This probably means that we need to consider the error codes used there as "burnt". @elenaf9, can you compile a list of application error codes that QUIC in rust-libp2p has been using so far?
Fortunately, this only applies to connection-level error codes, and there are lots of numbers in the uint62 space, so this is nothing more than a mild annoyance.

[...]

PR #519, opened 2023-02-17
Add yamux specification

thomaseizinger:
cc @marten-seemann @mxinden @wemeetagain 

The specification itself is a verbatim copy from https://github.com/hashicorp/yamux/blob/master/spec.md modulo formatting changes.

The only addition is the "Implementation considerations" section at the bottom where we document the ACK backlog of 256. To my knowledge, only `go-libp2p` implements this at the moment. `rust-yamux` still needs to implement this: https://github.com/libp2p/rust-yamux/issues/150. I am not sure about the status of this in `js-libp2p-yamux`?

## TODO

- [x] Include yamux protocol string in spec
- [x] add @ianopolous to interest group

### DIFF
--- a/yamux/README.md
+++ b/yamux/README.md
@@ -115,7 +115,7 @@ The corresponding ACK does not need to be received.
 This is particularly well suited for an RPC system where a client wants to open a stream and immediately fire a request without waiting for the RTT of the ACK.
 
 This does introduce the possibility of a connection being rejected after data has been sent already.
-This is a slight semantic difference from TCP, where the conection cannot be refused after it is opened.
+This is a slight semantic difference from TCP, where the connection cannot be refused after it is opened.
 Clients should be prepared to handle this by checking for an error that indicates a RST was received.
 
 #### Closing a stream
@@ -152,5 +152,5 @@ The Length should be set to one of the following to provide an error code:
 ### ACK backlog
 
 Yamux allows for a stream to be opened (and used) before it is acknowledged by the remote.
-We defined the ACK backlog as the number of streams that a peer has opened which have not yet been acknowledged.
+The ACK backlog is defined as the number of streams that a peer has opened which have not yet been acknowledged.
 Implementations SHOULD at most allow an ACK backlog of 256 streams.
--- a/yamux/README.md
+++ b/yamux/README.md
@@ -6,10 +6,11 @@
 
 Authors: [@thomaseizinger]
 
-Interest Group: [@marten-seemann]
+Interest Group: [@marten-seemann], [@wemeetagain]
 
 [@thomaseizinger]: https://github.com/thomaseizinger
 [@marten-seemann]: https://github.com/marten-seemann
+[@wemeetagain]: https://github.com/wemeetagain
 
 See the [lifecycle document][lifecycle-spec] for context about maturity level and spec status.
 
--- a/yamux/README.md
+++ b/yamux/README.md
@@ -6,11 +6,12 @@
 
 Authors: [@thomaseizinger]
 
-Interest Group: [@marten-seemann], [@wemeetagain]
+Interest Group: [@marten-seemann], [@wemeetagain], [@ianopolous]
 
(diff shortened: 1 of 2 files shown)


yamux/README.md
```diff
@@ -0,0 +1,156 @@
...
+* 0x0 Normal termination
+* 0x1 Protocol error
+* 0x2 Internal error
+
+## Implementation considerations
+
+### ACK backlog
+
+Yamux allows for a stream to be opened (and used) before it is acknowledged by the remote.
+We defined the ACK backlog as the number of streams that a peer has opened which have not yet been acknowledged.
```
marten-seemann, line 155:
```suggestion
The ACK backlog is defined as the number of streams that a peer has opened which have not yet been acknowledged.
```

yamux/README.md
```diff
@@ -0,0 +1,156 @@
...
+The SYN flag should be set to signal a new stream.
+
+The receiver must then reply with either a data or window update frame with the StreamID along with the ACK flag to accept the stream or with the RST flag to reject the stream.
+
+Because we are relying on the reliable stream underneath, a connection can begin sending data once the SYN flag is sent.
+The corresponding ACK does not need to be received.
+This is particularly well suited for an RPC system where a client wants to open a stream and immediately fire a request without waiting for the RTT of the ACK.
+
+This does introduce the possibility of a connection being rejected after data has been sent already.
+This is a slight semantic difference from TCP, where the conection cannot be refused after it is opened.
```
marten-seemann, line 118:
```suggestion
This is a slight semantic difference from TCP, where the connection cannot be refused after it is opened.
```

marten-seemann (approved):

marten-seemann:
> The only addition is the "Implementation considerations" section at the bottom where we document the ACK backlog of 256. To my knowledge, only go-libp2p implements this at the moment. rust-yamux still needs to implement this: https://github.com/libp2p/rust-yamux/issues/150. I am not sure about the status of this in js-libp2p-yamux?

That's fine. Doesn't block the spec, as it's a SHOULD anyway.

yamux/README.md
thomaseizinger: Fixing upstream typos here 😂

thomaseizinger:
Just remembered that this is missing the protocol identifier that we use. Will add.

thomaseizinger:
> > The only addition is the "Implementation considerations" section at the bottom where we document the ACK backlog of 256. To my knowledge, only go-libp2p implements this at the moment. rust-yamux still needs to implement this: https://github.com/libp2p/rust-yamux/issues/150. I am not sure about the status of this in js-libp2p-yamux?
> 
> That's fine. Doesn't block the spec, as it's a SHOULD anyway.

Agreed. I am only mentioning it for visibility.

thomaseizinger:
Any other people I should add to the interest group?

Menduist:
Does it really make sense to copy the yamux spec here? It's not like we copied the quic spec in this repo for instance, we just specified how it's used inside libp2p
Maybe just point to a specific hash of the upstream spec, and add libp2p specifics

thomaseizinger:
> Does it really make sense to copy the yamux spec here? It's not like we copied the quic spec in this repo for instance, we just specified how it's used inside libp2p
> Maybe just point to a specific hash of the upstream spec, and add libp2p specifics

I see your point. The QUIC spec is hosted by the IETF though which I'd consider more trustworthy and I am sure there are other copies of it around too.

The yamux spec is owned by an unrelated GitHub account so I figured copying is better. It is also not too long and copying it allows us to improve wording or structure in the future if we want to.

I don't feel super strongly though. Curious to hear more opinions.

marten-seemann:
https://github.com/libp2p/specs/issues/479 will require us to make some (backwards-compatible) changes to yamux to encode errors, so having a spec in this repo that we can modify makes sense.

mxinden (approved):
Thanks @thomaseizinger.

I am in favor of copying.

thomaseizinger:
This is ready to merge from my end.

PR #526, opened 2023-02-28
circuitv2: clarify what to do when data limit is reached

achingbrain:
This documents the rust-libp2p behaviour when more data than is allowed is transferred over a relayed connection.  js-libp2p will take the same approach.  go-libp2p may need to be updated.

### DIFF
--- a/relay/circuit-v2.md
+++ b/relay/circuit-v2.md
@@ -259,7 +259,7 @@ If more data than the limit specified in the `data` field is transferred
 over the connection, the server should reset the stream. If the reservation
 for the connection has expired the server may then apply any connection
 management policy to the connection as normal otherwise it should retain
-the connection, unless doing so would prevent it from mainting it's resource
+the connection, unless doing so would prevent it from maintaining it's resource
 quotas.
 
 ***Note: Implementations _should not_ accept reservations over already relayed connections.***
--- a/relay/circuit-v2.md
+++ b/relay/circuit-v2.md
@@ -4,7 +4,7 @@ This is the version 2 of the libp2p Circuit Relay protocol.
 
 | Lifecycle Stage | Maturity       | Status | Latest Revision |
 | --------------- | -------------- | ------ | --------------- |
-| 3A              | Recommendation | Active | r2, 2023-01-31  |
+| 3A              | Recommendation | Active | r3, 2023-02-28  |
 
 Authors: [@vyzo]
 
--- a/relay/circuit-v2.md
+++ b/relay/circuit-v2.md
@@ -256,7 +256,7 @@ If a relay server becomes overloaded however, it may still drop a
 connection with reservations in order to maintain its resource quotas.
   
 If more data than the limit specified in the `data` field is transferred
-over the connection, the server should reset the stream. If the reservation
+over the relayed connection, or the relayed connection has been open for longer than `duration`, the relay should reset the stream to the source and the stream to the destination. If the reservation
 for the connection has expired the server may then apply any connection
 management policy to the connection as normal otherwise it should retain
 the connection, unless doing so would prevent it from maintaining its resource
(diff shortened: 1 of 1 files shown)


relay/circuit-v2.md
```diff
@@ -254,6 +254,13 @@ connection for the duration of any reservations and tag it to prevent
 accidental termination according to its connection management policy.
 If a relay server becomes overloaded however, it may still drop a
 connection with reservations in order to maintain its resource quotas.
+  
+If more data than the limit specified in the `data` field is transferred
+over the connection, the server should reset the stream. If the reservation
+for the connection has expired the server may then apply any connection
+management policy to the connection as normal otherwise it should retain
+the connection, unless doing so would prevent it from mainting it's resource
```
thomaseizinger, line 262:
```suggestion
the connection, unless doing so would prevent it from maintining it's resource
```

achingbrain:
```suggestion
the connection, unless doing so would prevent it from maintaining it's resource
```

Spelling is hard 🤣

marten-seemann: And there's another mistake ;)

achingbrain:
```suggestion
the connection, unless doing so would prevent it from maintaining its resource
```

Arrgh!

relay/circuit-v2.md
```diff
@@ -254,6 +254,13 @@ connection for the duration of any reservations and tag it to prevent
 accidental termination according to its connection management policy.
 If a relay server becomes overloaded however, it may still drop a
 connection with reservations in order to maintain its resource quotas.
+  
+If more data than the limit specified in the `data` field is transferred
+over the connection, the server should reset the stream. If the reservation
```
mxinden, line 259:
```suggestion
If more data than the limit specified in the `data` field is transferred
over the relayed connection, or the relayed connection has been open for longer than `duration`, the relay should reset the stream to the source and the stream to the destination. If the reservation
```

mxinden (commented):
Thanks @achingbrain improving the specification!

I suggest:

- Including going over `duration`.
- Differentiating by direct connection and *relayed* connection.
- Specifying that both streams, to source and destination, should be reset.

Let me know what you think.

relay/circuit-v2.md
```diff
@@ -254,6 +254,13 @@ connection for the duration of any reservations and tag it to prevent
...
+for the connection has expired the server may then apply any connection
+management policy to the connection as normal otherwise it should retain
+the connection, unless doing so would prevent it from maintaining it's resource
```
achingbrain, line 262:
```suggestion
the connection, unless doing so would prevent it from maintaining its resource
```

relay/circuit-v2.md
achingbrain: Much better, thanks!

p-shahi:
Should we update the revision to r3?

mxinden (approved):

p-shahi:
@MarcoPolo @marten-seemann all good for the go-libp2p side?

relay/circuit-v2.md
```diff
@@ -254,6 +254,13 @@ connection for the duration of any reservations and tag it to prevent
 accidental termination according to its connection management policy.
 If a relay server becomes overloaded however, it may still drop a
 connection with reservations in order to maintain its resource quotas.
+  
+If more data than the limit specified in the `data` field is transferred
+over the relayed connection, or the relayed connection has been open for longer than `duration`, the relay should reset the stream to the source and the stream to the destination. If the reservation
+for the connection has expired the server may then apply any connection
```
marten-seemann, line 260: Who's the "server" here. Is it the relay? We probably shouldn't be using 2 different terms here.

The following sentence doesn't really make sense to me, as a relayed connection doesn't really show up as a normal connection in the connection manager.

marten-seemann (changes requested):

relay/circuit-v2.md
thomaseizinger: > as a relayed connection doesn't really show up as a normal connection in the connection manager.

That sounds a bit too implementation-specific to me? I do agree though that the sentence is superfluous. "Do what you'd normally do" isn't really worth "specifying" I'd say.

marten-seemann:
> No activity here in 2 weeks and this is approved. @p-shahi is this ready to merge?

No, my review hasn’t been addressed yet.

marten-seemann (approved):

PR #531, opened 2023-03-21
circuitv2: specify timestamp units for reservation expiration

achingbrain:
The reservation expiration date is specified in seconds so be explicit about that in the spec in the same way we are explicit about the duration limit units.

### DIFF
--- a/relay/circuit-v2.md
+++ b/relay/circuit-v2.md
@@ -224,7 +224,7 @@ Reservation {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


marten-seemann (approved):
This is correct: https://github.com/libp2p/go-libp2p/blob/9e208249b588727cef726a7c2560a9db722404a1/p2p/protocol/circuitv2/client/reservation.go#L88.

Issue #536, opened 2023-04-03
Proposal: allow AutoNAT to dial all IP addresses, without risking amplification attacks


marten-seemann:
Our AutoNAT spec currently says:
> In order to prevent attacks like the one described in [RFC 3489, Section 12.1.1](https://www.rfc-editor.org/rfc/rfc3489#section-12.1.1) (see excerpt below), implementations MUST NOT dial any multiaddress unless it is based on the IP address the requesting node is observed as. This restriction as well implies that implementations MUST NOT accept dial requests via relayed connections as one can not validate the IP address of the requesting node.
>
>> RFC 3489 12.1.1 Attack I: DDOS Against a Target
>>
>> In this case, the attacker provides a large number of clients with the same faked MAPPED-ADDRESS that points to the intended target. This will trick all the STUN clients into thinking that their addresses are equal to that of the target. The clients then hand out that address in order to receive traffic on it (for example, in SIP or H.323 messages). However, all of that traffic becomes focused at the intended target. The attack can provide substantial amplification, especially when used with clients that are using STUN to enable multimedia applications.

The intention is to prevent an amplification attack against a target, and was designed with the properties of the STUN protocol in mind.

This is not the only way to prevent amplification attacks though. We can also just make the attack expensive enough such that it becomes unattractive. For example, we could require the requester to send a non-trivial amount of data, on the order of 10-100 kB. This would make AutoNAT completely uninteresting for an amplification attack, since a libp2p handshake is a lot smaller than 10 kB, and thus no amplification can be achieved.

On the other hand, for the AutoNAT server, receiving 10-100 kB is cheap enough to process that we're not placing a too large burden on the server. To further lighten the load on the server, this could be designed as a "retry" mechanism (borrowing QUIC terminology here): Only if the IP address doesn't match would the server request the client to send this additional data.

Obviously, as this is a new design, this would need to be incorporated into AutoNAT v2 (#503).

@sukunrt Thoughts?

sukunrt:
What benefits do we get by allowing this? 

Some preliminary thoughts, 
I think this scheme is optimal in some sybil attack sense. For an attacker to get 10 peers to dial the target the attacker will need to spend 10x resources. 

[...]

Issue #503, opened 2023-01-12
Proposal: AutoNAT v2


marten-seemann:
This proposal has been around for the longest time, but it’s living inside this go-libp2p issue: https://github.com/libp2p/go-libp2p/issues/1480#issuecomment-1122844443.
Writing this up here now, so we have all the protocol improvements in one place.

## Protocol Description

Compared to AutoNAT v1, AutoNAT v2 should provide the following features:
1. It should be a protocol to test the reachability of particular addresses (not list of addresses). This allows nodes to use AutoNAT to test address candidates (derived from different sources, e.g. Identify, circuit addresses etc.) for their reachability. The idea is to enable building of an address pipeline, where address candidates can go from “suspected” to “confirmed” listen address (with occasional re-confirmations of already confirmed addresses).
2. It makes it more difficult to lie to us. Now this only works in one direction: we can have a node prove that a connection attempt succeeded. The other direction isn’t provable: a node can always claim that it tried to dial us, but that dialing failed. This can be achieved by asking the node to provide a certain identifier (random string / number), and then wait for an incoming connection (on the right _transport_, we won’t be able to distinguish between different addresses on the same transport) where this identifier has to be presented.

(2) definitely is a breaking a protocol change, and (1) in practice is as well, because old go-libp2p versions would return a DIAL_FAILED instead of a DIAL_REFUSED error if a transport was unsupported.

## Rollout Strategy

AutoNAT v2 is only useful if a sufficient number of nodes support the protocol. During a transition period, nodes will need to support both versions (at least on the client side). Once a large enough fraction of the network (20%? tbd) supports AutoNAT v2, we can start disabling support for v1.

elenaf9:
Thanks for moving the discussion here @marten-seemann.
Concerning (2): I know that in go-libp2p AutoNAT servers uses a background host for the dial back and thus the PeerId is unknown. Could you explain why it is not possible in go-libp2p to do the dial as a "normal" dial with the server's PeerId? 
That's how it's done in rust-libp2p. It would avoid the need of the additional nonce, since the client could simply use the PeerId to check if it received an inbound connection.

marten-seemann:
How does rust-libp2p handle dial-back requests for the same TCP addressed received over a TCP connection? You won't be able to establish two TCP connections for the same 4-tuple when using SO_REUSEPORT. Do you special-case this?

Menduist:
[...]

PR #538, opened 2023-04-11
add autonat v2 spec

sukunrt:
First draft for autonat v2. https://github.com/libp2p/specs/issues/503

This protocol allows for testing reachability on exactly one address. This helps determine reachability at an address level. This also simplifies the protocol a lot. 

I'll change the spec to reflect the discussion on dialing a different ip address from the  nodes observed ip address: https://github.com/libp2p/specs/issues/536

Discussion for nonce in message is here: https://github.com/libp2p/go-libp2p/issues/1480
and this comment in particular https://github.com/libp2p/go-libp2p/issues/1480#issuecomment-1122844443


autonat/autonat-v2.md
```diff
@@ -0,0 +1,137 @@
...
+can test individual addresses discovered by different sources like identify, 
+upnp mappings, circuit addresses etc for reachability.
+
+
+## AutoNAT V2 Protocol
+
+The AutoNAT V2 Protocol uses the Protocol ID `/libp2p/autonat/2.0.0`. The node
+wishing to determine reachability of a given address sends a `DialRequest` 
+message. The `DialRequest` message contains a single multiaddress and a nonce 
+of 8 random bytes. Upon receiving  this message, the peer dials this address
```
marten-seemann, line 40: If we specify that it's 8 bytes, should we just make a uint64?

autonat/autonat-v2.md
```diff
@@ -0,0 +1,137 @@
...
+upnp mappings, circuit addresses etc for reachability.
+
+
+## AutoNAT V2 Protocol
+
+The AutoNAT V2 Protocol uses the Protocol ID `/libp2p/autonat/2.0.0`. The node
+wishing to determine reachability of a given address sends a `DialRequest` 
+message. The `DialRequest` message contains a single multiaddress and a nonce 
+of 8 random bytes. Upon receiving  this message, the peer dials this address 
+and sends the nonce in the `Dial` message. The peer MUST NOT dial any address
```
marten-seemann, line 41: You need to specify how this nonce is sent. Maybe define a new protocol ID for that?

autonat/autonat-v2.md
```diff
@@ -0,0 +1,137 @@
...
+
+The AutoNAT V2 Protocol uses the Protocol ID `/libp2p/autonat/2.0.0`. The node
+wishing to determine reachability of a given address sends a `DialRequest` 
+message. The `DialRequest` message contains a single multiaddress and a nonce 
+of 8 random bytes. Upon receiving  this message, the peer dials this address 
+and sends the nonce in the `Dial` message. The peer MUST NOT dial any address 
+other than the address in the DialRequest. The initiator SHOULD check that the 
+nonce received in the `Dial` message is the same as the nonce the initiator 
+sent in the `DialRequest` message. If the nonce received in the `Dial` message 
+is different the initiator should discard this `DialResponse`
```
marten-seemann, line 45:
```suggestion
is different the initiator MUST discard this `DialResponse`
```

autonat/autonat-v2.md
```diff
@@ -0,0 +1,137 @@
...
+> RFC 3489 12.1.1 Attack I: DDOS Against a Target
+>
+> In this case, the attacker provides a large number of clients with the same
+> faked MAPPED-ADDRESS that points to the intended target. This will trick all
+> the STUN clients into thinking that their addresses are equal to that of the
+> target. The clients then hand out that address in order to receive traffic on
+> it (for example, in SIP or H.323 messages). However, all of that traffic
+> becomes focused at the intended target. The attack can provide substantial
+> amplification, especially when used with clients that are using STUN to enable
+> multimedia applications.
```
marten-seemann, line 68: Sounds like after the discussion in https://github.com/libp2p/specs/issues/536#issuecomment-1497860644, we might not need this after all? But let's hear what @mxinden has to say first.

autonat/autonat-v2.md
```diff
@@ -0,0 +1,137 @@
...
+other than the address in the DialRequest. The initiator SHOULD check that the 
+nonce received in the `Dial` message is the same as the nonce the initiator 
+sent in the `DialRequest` message. If the nonce received in the `Dial` message 
+is different the initiator should discard this `DialResponse`
+
+If the dial fails, the receiver sends a `DialResponse` message with the
+`ResponseStatus` `E_DIAL_ERROR`. If the dial completes successfully, it 
+sends a `DialResponse` with the `ResponseStatus` `OK`. Nodes MUST NOT send a 
+`E_DIAL_ERROR` status when they didn't attempt a dial. 
+
```
marten-seemann, line 51: We need some text that talks about refusal of dials.

There's two (main) reasons why a peer might refuse a dial request:
1. The peer might apply some kind of rate limiting / blacklisting.
2. Or it might not be able to parse the multiaddr, or not be able to dial the transport.

I'm wondering if it makes sense to split these two reason into separate response codes, as this allows the requester to adjust their response:
1. If we're rate limited / blacklisted, retrying with a different address has little potential for success.
2. If the peer just can't dial the respective multiaddr, we could just open a new stream and request dialing of another multiaddr.

We could also think of adding a list of supported transport multicodecs to that response message.

What do you think?

autonat/autonat-v2.md
```diff
@@ -0,0 +1,137 @@
...
+> In this case, the attacker provides a large number of clients with the same
+> faked MAPPED-ADDRESS that points to the intended target. This will trick all
+> the STUN clients into thinking that their addresses are equal to that of the
+> target. The clients then hand out that address in order to receive traffic on
+> it (for example, in SIP or H.323 messages). However, all of that traffic
+> becomes focused at the intended target. The attack can provide substantial
+> amplification, especially when used with clients that are using STUN to enable
+> multimedia applications.
+
+The initiator uses the responses obtained from multiple peers to determine
```
marten-seemann, line 70: Can we make this a separate section? While the above normatively describes the AutoNAT v2 protocol, this section is more of a recommendation how to process the result. It's totally valid to come up with different heuristics and thresholds.

autonat/autonat-v2.md
```diff
@@ -0,0 +1,137 @@
...
+2. Sending the RPC `DialRequest` message.
+3. Listening for the RPC `DialResponse` message.
+
+`Dial` Message is sent by the responder by:
+1. Dialing an address
+2. Opening a new stream.
+3. Sending `Dial` message.
+
+```proto
+syntax = "proto2";
```
marten-seemann, line 96: Can we use proto3?

marten-seemann (commented):
Very nice, this is a solid starting point for the spec!

What's the plan for resolving #536? Would you open a new PR that targets this PR here?

autonat/autonat-v2.md
sukunrt: oh yes. I missed this. will fix.

sukunrt:
>What's the plan for resolving https://github.com/libp2p/specs/issues/536? Would you open a new PR that targets this PR here?

Yes, I'll open a PR with the changes for #536.

autonat/autonat-v2.md
sukunrt: I'm fine either way, do you think uint64 would be easier to handle?

autonat/README.md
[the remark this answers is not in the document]
sukunrt: sure. I'll do that.

autonat/autonat-v2.md
marten-seemann: Probably, if we use 8 bytes. However, I'm wondering if we might have any use case for longer sequences, and should leave the number of bytes unspecified (or just recommend a minimum).

autonat/autonat-v2.md
sukunrt: will fix

autonat/autonat-v2.md
sukunrt: I think separating them is a nice idea. Providing the list of supported transports might make the protocol more complex than we need. I'll change this to include the list of transports, we can then see if it is looking too complex.

autonat/README.md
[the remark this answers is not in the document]
sukunrt: fixed.

autonat/autonat-v2.md
sukunrt: I've changed this to uint64 because that does seem easier to handle for implementations. I'm not sure how to decide here, I can't think of a need for longer sequences, but that will be more flexible. Keeping this open for a bit

autonat/autonat-v2.md
sukunrt: fixed.

autonat/autonat-v2.md
sukunrt: fixed.

autonat/autonat-v2.md
sukunrt: fixed. Does the current version look better?

autonat/autonat-v2.md
sukunrt: fixed

autonat/autonat-v2.md
marten-seemann: Yes. Curious to hear what the other reviewers think.

autonat/autonat-v2.md
sukunrt: have separated response codes for these two cases. 

continuing the discussion for priority ordered address list here https://github.com/libp2p/specs/issues/539

sukunrt: opened a PR for the rest of the changes. We can continue the discussion there.

sukunrt: https://github.com/libp2p/specs/pull/541

autonat/autonat-v2.md
```diff
@@ -0,0 +1,163 @@
...
+Upon receiving this message the peer attempts to dial the address and opens a 
+stream with  Protocol ID `/libp2p/autonat/2.0.0/dial-attempt` and sends a 
+`DialAttempt` message with the nonce received in the `DialRequest`. The peer 
+MUST NOT dial any address other than the address provided in the `DialRequest` 
+message.
+
+Upon completion of the dial attempt, the peer sends a `DialResponse` to the 
+initiator node on the `/libp2p/autonat/2.0.0/dial-request` stream.
+
+The initiator SHOULD check that the nonce received in the `Dial` message is the
```
sukunrt, line 55: @marten-seemann should we make this a required feature?

autonat/autonat-v2.md
```diff
@@ -0,0 +1,193 @@
...
+reused to get information about other addresses
+
+Compared to `autonat v1` there are two major differences
+1. `autonat v1` allowed testing reachability for the node. `autonat v2` allows
+testing reachability for an individual address
+2. `autonat v2` provides a mechanism for nodes to verify whether the peer
+actually successfully dialled an address.
+
+
+## AutoNAT V2 Protocol
```
thomaseizinger, line 48: This section would benefit from a sequence diagram that shows what is exchanged on which stream at what point in time.

I've also made a naming suggestion further down.

autonat/autonat-v2.md
```diff
@@ -0,0 +1,193 @@
...
+}
+
+message AddressDialRequest {
+  bytes addr = 1;
+  fixed64 nonce = 2;
+}
+
+message DialRequest {
+  repeated AddressDialRequest addressDialRequests = 1;
+}
```
thomaseizinger, line 170: A different naming idea:

```suggestion
message Candidates {
  bytes addr = 1;
  fixed64 nonce = 2;
}

message DialRequest {
  repeated Candidates candidates = 1;
}
```

I've found the repetition of the words "addresses" and "request" hard for creating a mental model of what is happening upon first read.

autonat/autonat-v2.md
```diff
@@ -0,0 +1,193 @@
...
+index received in `DialResponse`. If the nonce is different, the initiator MUST
+discard this response.
+
+
+### Requirements for ResponseStatus
+
+On receiving a `DialRequest` the peer selects the first address on the list it
+is capable of dialing. This address is referred to as _addr_. The
+`ResponseStatus` sent by the peer in the `DialResponse` message MUST be set
+according to the following requirements
```
thomaseizinger, line 82: Instead of implicitly dialing only the first one we support, have we considered responding with a status code for each one by sending one `DialResponse` per address (on the same stream)?

This would provide a lot more information to a node about what addresses are supported and we can for example send the once a node can't parse or dial to a different autonat server.

autonat/autonat-v2.md
```diff
@@ -0,0 +1,193 @@
+# AutonatV2: spec
```
thomaseizinger, line 1: Should we handle addresses from private ranges like `10.x.y.z` in a dedicated way?

Two nodes might find themselves in a private network as reachable but they will still be behind NAT  in regards to the Internet.

https://en.m.wikipedia.org/wiki/Private_network

autonat/autonat-v2.md
```diff
@@ -0,0 +1,193 @@
...
+  bytes addr = 1;
+  fixed64 nonce = 2;
+}
+
+message DialRequest {
+  repeated AddressDialRequest addressDialRequests = 1;
+}
+
+message DialAttempt {
+    fixed64 nonce = 1;
```
thomaseizinger, line 173: What is the reason for a fixed64 and not a u64?

autonat/autonat-v2.md
```diff
@@ -0,0 +1,193 @@
...
+      OK                        = 0;
+      E_DIAL_ERROR              = 100;
+      E_DIAL_REFUSED            = 101;
+      E_TRANSPORT_NOT_SUPPORTED = 102;
+      E_BAD_REQUEST             = 200;
+      E_INTERNAL_ERROR          = 300;
+    }
+
+    ResponseStatus status = 1;
+    string statusText = 2;
```
thomaseizinger, line 187: What is the benefit of the status text if we have an enum?

autonat/autonat-v2.md
```diff
@@ -0,0 +1,193 @@
...
+
+message DialResponse {
+  enum ResponseStatus {
+      OK                        = 0;
+      E_DIAL_ERROR              = 100;
+      E_DIAL_REFUSED            = 101;
+      E_TRANSPORT_NOT_SUPPORTED = 102;
+      E_BAD_REQUEST             = 200;
+      E_INTERNAL_ERROR          = 300;
+    }
```
thomaseizinger, line 184: How should nodes handle unknown status codes? We might add some in the future.

autonat/autonat-v2.md
```diff
@@ -0,0 +1,193 @@
...
+## RPC Messages
+
+All RPC messages sent over a stream are prefixed with the message length in
+bytes, encoded as an unsigned variable length integer as defined by the
+[multiformats unsigned-varint spec][uvarint-spec]. 
+
+All RPC messages are of type `Message`. A `DialRequest` message is sent as a
+`Message` with the `dialRequest` field set and the `type` field set to
+`DIAL_REQUEST`. Other message types `DialResponse` and `DialAttempt` are handled
+similarly.
```
thomaseizinger, line 144: Is there a benefit of merging this into a single message if we have essentially two stream protocols?

None of the data types are shared between the protocols. We would avoid some complexity at compile-time if we would define one message per stream protocol.

Otherwise, we have error cases like "successfully parsed a `DialResponse` on a `/libp2p/autonat/2.0.0/dial-attempt` stream".

autonat/autonat-v2.md
```diff
@@ -0,0 +1,193 @@
...
+resource limit reached or blacklisting.
+
+`E_TRANSPORT_NOT_SUPPORTED`: the peer didn't have the ability to dial any of the
+requested addresses.
+
+`E_BAD_REQUEST`: the peer didn't attempt a dial because it was unable to decode
+the message. This includes inability to decode the requested address to dial.
+
+`E_INTERNAL_ERROR`: error not classified within the above error codes occured on
+peer that prevented it from completing the request.
```
thomaseizinger, line 98: Do we want an error code for "I have no resources for making a connection"?

autonat/autonat-v2.md
```diff
@@ -0,0 +1,193 @@
...
+
+A node wishing to determine reachability of its adddresses sends a `DialRequest`
+message to a peer on a stream with protocol ID
+`/libp2p/autonat/2.0.0/dial-request`. 
+
+This `DialRequest` message has a list of `AddressDialRequest`s. Each item in
+this list contains an address and a fixed64 nonce. The list is ordered in
+descending order of priority for verfication.
+
+Upon receiving this message the peer attempts to dial the first address from the
```
thomaseizinger, line 58: Do I understand correctly that we only ever dial a single address from this list? I think this should be stressed explicitly.

autonat/autonat-v2.md
```diff
@@ -0,0 +1,193 @@
...
+`E_DIAL_ERROR`: the peer attempted to dial _addr_ and was unable to connect. 
+
+`E_DIAL_REFUSED`: the peer didn't attempt a dial because of rate limiting,
+resource limit reached or blacklisting.
+
+`E_TRANSPORT_NOT_SUPPORTED`: the peer didn't have the ability to dial any of the
+requested addresses.
+
+`E_BAD_REQUEST`: the peer didn't attempt a dial because it was unable to decode
+the message. This includes inability to decode the requested address to dial.
```
thomaseizinger, line 95: I don't understand this one. If I fail to decode it, I don't know if I would support it.

When do I use this status code? In other words, when do I "pick" an address as a dial candidate but then respond with an error that I can't dial it? Wouldn't I just go to the next address?

thomaseizinger (commented):
Exciting! Thanks for your work. Left some comments :)

autonat/autonat-v2.md
sukunrt: I like this suggestion! thanks!

autonat/autonat-v2.md
sukunrt: I want the nonces to be 8 bytes long. An 8 byte integer will take more than 8 bytes with u64 and exactly 8 with fixed64.

autonat/autonat-v2.md
sukunrt: I added this line, https://github.com/libp2p/specs/blob/62123df073eabbf0a8260be694c36eb59f43b302/autonat/autonat-v2.md?plain=1#L100
Does this work?

autonat/autonat-v2.md
sukunrt: I've separated the message types and renamed the streams too. 
```/libp2p/autonat/2.0.0/dial``` stream exchanges ```DialMessage```
```/libp2p/autonat/2.0.0/attempt``` stream exchanges ```AttemptMessage```

File: autonat/autonat-v2.md (hunk 17 above)
Comment (sukunrt) [reply autonat/autonat-v2.md]: this should be E_DIAL_REFUSED. 
Do you think we need a separate code for rate limiting and resource limit reached?

File: autonat/autonat-v2.md (hunk 18 above)
Comment (sukunrt) [reply autonat/autonat-v2.md]: Is this line not clear enough
https://github.com/libp2p/specs/blob/62123df073eabbf0a8260be694c36eb59f43b302/autonat/autonat-v2.md?plain=1#L61-L63
?

File: autonat/autonat-v2.md (hunk 19 above)
Comment (sukunrt) [reply autonat/autonat-v2.md]: You are right. This cannot happen. I've removed this line.

Comment (sukunrt) [reply autonat/autonat-v2.md]: The expected behaviour is to proceed to the next address as you've mentioned

File: autonat/autonat-v2.md (hunk 11 above)
Comment (sukunrt) [reply autonat/autonat-v2.md]: autonat v2 started with the idea of sending a single address in a message. This is fine because creating a stream is cheaper than dialing an address, so the initiator can just open a new stream for all addresses it wants verified. 

While running a node, we may discover addresses or guess at our addresses(quic-v1 => webtransport). Depending on our node state we may or may not want to verify these addresses. If we already have a reachable tcp address, we probably don't want to verify another tcp address. On the other hand some addresses we do want to verify like a webrtc whose reachability we don't know. Requests to nodes for dialing webrtc addresses will likely fail because the support in the network is low. In this case we can add the lower priority tcp address to the list as a "nice to have". If there was no higher priority webrtc address, we'd not have tried verifying this low priority tcp address. 

The expectation is that nodes are not sending every address they know in every request. 
This previous discussion is [here](https://github.com/libp2p/specs/issues/539) 

@marten-seemann, thoughts? 


@thomaseizinger I'd love to get your thoughts on this

**Comment by sukunrt** [comment]
thanks for your review @thomaseizinger. 
I'd like your opinion on these two issues

Proposal: use a list of addresses in priority order for autonat v2 dial requests #539 
Proposal: allow AutoNAT to dial all IP addresses, without risking amplification attacks #536

File: autonat/autonat-v2.md (hunk 15 above)
Comment (thomaseizinger) [reply autonat/autonat-v2.md:185]: Works for me!

File: autonat/autonat-v2.md (hunk 17 above)
Comment (thomaseizinger) [reply autonat/autonat-v2.md]: Didn't see that one in my initial read, sorry for the confusion!

File: autonat/autonat-v2.md (hunk 18 above)
Comment (thomaseizinger) [reply autonat/autonat-v2.md]: It is clear but at the end of the paragraph. Given the importance of this aspect in the spec, I think it should be very explicit as soon as possible (lookup "frontloading" as a concept).

File: autonat/autonat-v2.md [hunk 20]
```diff
@@ -0,0 +1,207 @@
...
+- query multiple peers to determine reachability
+
+The suggested heuristic for implementations is to consider an address reachable
+if more than 3 peers report a successful dial and to consider an address
+unreachable if more than 3 peers report unsuccessful dials. 
+
+Implementations are free to use different heuristics than this one
+
+
+## RPC Messages
```
mxinden, line 140: **Stream behavior**

Do we have consensus that multiple `DialRequest`s should not be pipelined on a single stream? If so, mind adding a paragraph along the lines of:

```
Each `DialRequest` is sent on a new stream. Multiple `DialRequest`s are never sent on the same stream (i.e. no pipelining).
```

Also is a node supposed to close the write side of a stream once it sent the `DialRequest`, `DialResponse` and `AttemptMessage`?

autonat/autonat-v2.md
```diff
@@ -0,0 +1,207 @@
...
+
+message DialRequest {
+  repeated Candidate candidates = 1;
+}
+
+message DialResponse {
+  enum ResponseStatus {
+      OK                        = 0;
+      E_DIAL_ERROR              = 100;
+      E_DIAL_REFUSED            = 101;
```
mxinden, line 181: What is the difference between error and refused? What are examples for each?

autonat/autonat-v2.md
```diff
@@ -0,0 +1,207 @@
...
+
+message DialMessage {
+  enum Type {
+    DIAL_REQUEST  = 0;
+    DIAL_RESPONSE = 1;
+  }
+
+  Type type          = 1;
+  DialRequest dialRequest   = 2;
+  DialResponse dialResponse = 3;
```
mxinden, line 165:
```suggestion
  optional DialRequest dialRequest   = 2;
  optional DialResponse dialResponse = 3;
```

The proto3 presence rules still confuse me. From what I understand we should be introducing explicit presence here, no?

See also https://github.com/libp2p/specs/issues/465 for details.

mxinden (commented):
Great work @sukunrt. Thank you!

autonat/autonat-v2.md
marten-seemann: No pipelining, please!

I think it's sufficient to define when a stream is closed though, we don't need to explicitly forbid pipelining.

autonat/autonat-v2.md
sukunrt: in the section Requirements for ResponseStatus there are these lines 
```
E_DIAL_ERROR`: the peer attempted to dial _addr_ and was unable to connect. 

`E_DIAL_REFUSED`: the peer didn't attempt a dial because of rate limiting,
resource limit reached or blacklisting.
```
Is this not clear enough? Should I add an example? 
https://github.com/libp2p/specs/blob/0771babb17eb9e33d5334c78cb658fdf7e687a7f/autonat/autonat-v2.md?plain=1#L90-L94

autonat/autonat-v2.md
sukunrt: I think it's not required for message types. Please correct me if I'm wrong. 
 
this shouldn't be a problem because the default value of a message type should be NULL. For ex in go this would translate to a pointer to DialRequest ```*DialRequest``` and pointer to DialResponse whose value will be nil if it is not set. The message also has a Type field which signals which field to check 

```msg.Type == DIAL_REQUEST``` => ```msg.dialRequest``` should not be nil 

I'm fine with changing this if it simplifies implementation in other languages.

File: autonat/autonat-v2.md (hunk 21 above)
Comment (mxinden) [reply autonat/autonat-v2.md:181]: Please ignore my comment. I didn't read attentively.

File: autonat/autonat-v2.md (hunk 19 above)
Comment (sukunrt) [reply autonat/autonat-v2.md]: resolving this. feel free to open again in case you think something more should be done here.

File: autonat/autonat-v2.md (hunk 12 above)
Comment (sukunrt) [reply autonat/autonat-v2.md:1]: I've added a suggestion for implementations to only verify private address reachability with nodes on the same subnet. 

Do you think the spec should have information regarding what to do with this information? 

Consider the case that the node is reachable on a 192.168.1.100 address within its subnet. Now this node should advertise this address to other nodes(via identify) in case eventually this address reaches a node on the same subnet(via DHT or some other mechanism). Is my understanding correct here?

File: autonat/autonat-v2.md (hunk 18 above)
Comment (sukunrt) [reply autonat/autonat-v2.md]: I've moved the must not line near the top of the paragraph and added a clause(https://github.com/libp2p/specs/blob/autonat-v2/autonat/autonat-v2.md?plain=1#L30)
>As the server dials exactly one address from the list, autonat v2 allows nodes to determine reachability for individual addresses. 

in introduction. This should make it clearer.

File: autonat/autonat-v2.md (hunk 20 above)
Comment (sukunrt) [reply autonat/autonat-v2.md:148]: I've added lines. 

https://github.com/libp2p/specs/blob/autonat-v2/autonat/autonat-v2.md?plain=1#L57
>Each `DialRequest` is sent on a new stream.


https://github.com/libp2p/specs/blob/autonat-v2/autonat/autonat-v2.md?plain=1#L67-L68
>The peer MUST close this
stream after sending the `DialAttempt` message.


https://github.com/libp2p/specs/blob/autonat-v2/autonat/autonat-v2.md?plain=1#L81-L82
>The peer MUST close the stream after sending the response. The initiator MUST
close the stream after receiving the response.


Does this work?

File: autonat/autonat-v2.md (hunk 16 above)
Comment (sukunrt) [reply autonat/autonat-v2.md]: resolving this. You can open it again in case you think anything more is needed.

File: autonat/autonat-v2.md (hunk 11 above)
Comment (sukunrt) [reply autonat/autonat-v2.md]: Nice idea. As I understand browser nodes will have a lot of such cases where they understand the address but cannot dial it. 

I don't see any reason why not to do this. It only provides more information. 

@marten-seemann @mxinden any thoughts?

File: autonat/autonat-v2.md (hunk 12 above)
Comment (thomaseizinger) [reply autonat/autonat-v2.md:1]: My thinking was that we discard any addresses that are by design not globally reachable / routeable. I am not sure it makes sense for nodes to use Auto**NAT** to determine reachability on the same subnet? What we care about is reachability from the Internet, is it not?

To discover nodes on the same subnet, nodes can for example use mDNS.

File: autonat/autonat-v2.md [hunk 23]
```diff
@@ -0,0 +1,214 @@
...
+
+![Autonat V2 Interaction](autonat-v2.svg)
+
+
+A node wishing to determine reachability of its adddresses sends a `DialRequest`
+message to a peer on a stream with protocol ID `/libp2p/autonat/2.0.0/dial`.
+Each `DialRequest` is sent on a new stream.
+
+This `DialRequest` message has a list of `Candidate`s. Each item in this list
+contains an address and a fixed64 nonce. The list is ordered in descending order
```
MarcoPolo, line 60: Would be nice to have a sentence on why you want a unique nonce per address. I think the reasoning is a bit subtle, so would help implementers understand the point.

autonat/autonat-v2.md
```diff
@@ -0,0 +1,214 @@
...
+```proto
+syntax = "proto3";
+
+message DialMessage {
+  enum Type {
+    DIAL_REQUEST  = 0;
+    DIAL_RESPONSE = 1;
+  }
+
+  Type type          = 1;
```
MarcoPolo, line 171: We can use [oneof](https://protobuf.dev/programming-guides/proto3/#oneof) 

> If you have a message with many fields and where at most one field will be set at the same time, you can enforce this behavior and save memory by using the oneof feature.

MarcoPolo (commented):
On a brief skim, this looks good! I'm curious if we'll want to relax the "implementations MUST NOT dial any multiaddress unless it is based on the IP address the requesting node is observed as". Would it be useful to do this, and we can mitigate the amplification attack some other way?

It seems like there's a healthy discussion already going on, so I'll step back here and let other folks stay involved. If there's anything I can help with, please don't hesitate to ping.

sukunrt:
Thanks for your review @MarcoPolo 
>It seems like there's a healthy discussion already going on, so I'll step back here and let other folks stay involved. If there's anything I can help with, please don't hesitate to ping.


The suggested strategy is discussed here: https://github.com/libp2p/specs/issues/536
Please check if we've made any errors there or overlooked something.  

Here's the PR for those changes: https://github.com/libp2p/specs/pull/542
You can review it there, or here after I merge those changes.

autonat/autonat-v2.md
sukunrt: Great point! 
Now I'm not convinced of this requirement. I think a single nonce per request along with verifying that the transport on the local multiaddress where the nonce was received provides us with the same information. 

Consider the case where we are listening on a quic address and a tcp address. 
We ask the peer to dial [(pub_quic_addr, nonce1), (pub_tcp_addr, nonce2)]. It is a malicious peer which decides to dial the tcp address and tell us that it dialed quic address by sending nonce1. So locally our observation will be. 

(local_tcp_addr, pub_quic_addr, nonce1): Even without the address specific nonce we could have figured out that this observation is invalid. The connection we received was a tcp connection and not a quic one. 

Now consider a successful attempt to fool us. We are listening on a quic address and we have two candidates for a public quic address. So we ask the peer to dial (pub_quic_addr1, nonce1), (pub_quic_addr2, nonce2). The malicious peer dials pub_quic_addr2 but sends us nonce1 and informs us that it dialed pub_quic_addr1 with nonce1. 
Here our observation will be (local_quic_addr, pub_quic_addr1, nonce1). We cannot prevent this attack with multiple nonces. 

So I think single nonce plus recording local multiaddress provides us the the same benefits. This will also simplify the request objects.

sukunrt: cc @marten-seemann

autonat/autonat-v2.md
sukunrt: ah thanks! Using this!

autonat/autonat-v2.md
MarcoPolo: I originally thought the reason to have multiple nonces is so that, assuming honest peers, you could actually figure out which local interface/ip maps to which public IP address. Take your example `(pub_quic_addr1, nonce1), (pub_quic_addr2, nonce2)`. If the peer is honest, I'll know which local interface/ip maps to which public ip by checking the nonce.

However, rereading the spec:
> Upon completion of the dial attempt, the peer sends a `DialResponse` message to
the initiator node on the `/libp2p/autonat/2.0.0/dial` stream with the *index(0
based)* of the candidate that it attempted to dial and the appropriate
`ResponseStatus`

I realize that the honest peer would tell us which public addr it dialed and we'd know which local interface it was because we have the nonce. But the nonce here can just be for the request, not for each addr.

So I agree, I think we only need the single nonce.

autonat/autonat-v2.md
```diff
@@ -0,0 +1,222 @@
...
+
+message AttemptMessage {
+  oneof msg {
+    DialAttempt dialAttempt = 1;
+  }
+}
+
+message DialAttempt {
+    fixed64 nonce = 1;
+}
```
thomaseizinger, line 218: Does this structure provide us with much benefit? I am guessing the idea is evolvability. Now that we have a separate stream protocol for the attempt, we could also evolve things there if necessary (although at the cost of a roundtrip likely).

What do people think?

sukunrt: Can you elaborate on what you mean by 
>Now that we have a separate stream protocol for the attempt, we could also evolve things there if necessary

sukunrt: do you mean we can just send the nonce in the AttemptMessage?

thomaseizinger: > Can you elaborate on what you mean by 
> >Now that we have a separate stream protocol for the attempt, we could also evolve things there if necessary

This message is only used on the `/libp2p/autonat/2.0.0/dial-attempt` protocol.

Thus, if we wanted to change the message, we could also just change to: `/libp2p/autonat/3.0.0/dial-attempt` instead of extending the `oneof` with a new message type.

Currently, the `oneof` introduces an additional indirection and error path in the code.

thomaseizinger: > do you mean we can just send the nonce in the AttemptMessage? 

Yes.

b-zee:
I have a general remark that might be relevant to this spec.

About a month or so ago tested AutoNAT with QUIC (rust-libp2p). A node behind NAT could be dialed back successfully, falsely reporting the status as public. My working theory is that the dial-back happens to the same ip-port of a previously established connection. According to @mxinden on Matrix, in the Go implementation, a different port is supposedly used for the dial-back.

This v2 spec might solve that by allowing to probe for individual addresses but I am not totally sure. Perhaps my concern isn't relevant to this spec and is specific issue with the rust lib, feel free to ignore in that case.

marten-seemann:
> About a month or so ago tested AutoNAT with QUIC (rust-libp2p). A node behind NAT could be dialed back successfully, falsely reporting the status as public. My working theory is that the dial-back happens to the same ip-port of a previously established connection. According to @mxinden on Matrix, in the Go implementation, a different port is supposedly used for the dial-back.

Go uses a separate host for exactly that reason. If Rust dials back on the same 4-tuple, the connection attempt is guaranteed to succeed, rendering AutoNAT useless (actually, worse than useless, since it's reporting an incorrect result). This issue would still exist with AutoNAT v2.

b-zee:
Thanks for that insight, @marten-seemann!

Just to make sure I understand, the crux of the problem is then that the source port used for the `DialRequest` is the same as the address that is dialed back with `DialAttempt`? Is this a problem that is specific to UDP/QUIC (in the Rust impl), with TCP using a different source socket to establish an outbound connection (unless port reuse is enabled)?

marten-seemann:
As far as I know, rust-libp2p uses reuseport for TCP. In that case, the kernel will not allow you to bind another connection to the same 4-tuple, so I'd expect that rust-libp2p always returns a DIAL_FAILED error, which would also not be correct. I haven't tried it out though.

mxinden:
What @marten-seemann described above for rust-libp2p is correct. Let's move the rust-libp2p specific discussion to https://github.com/libp2p/rust-libp2p/issues/3900. Thanks @marten-seemann and @b-zee.

autonat/autonat-v2.md
```diff
@@ -0,0 +1,222 @@
...
+
+message DialResponse {
+  enum ResponseStatus {
+      OK                        = 0;
+      E_DIAL_ERROR              = 100;
+      E_DIAL_REFUSED            = 101;
+      E_TRANSPORT_NOT_SUPPORTED = 102;
+      E_BAD_REQUEST             = 200;
+      E_INTERNAL_ERROR          = 300;
```
marten-seemann, line 203: What error code would you send when you successfully established a TCP connection, but didn't support the security protocol or the muxer. Would that be a TRANSPORT_NOT_SUPPORTED, or should we mint a new code for that?

sukunrt: I think we need to split E_TRANSPORT_NOT_SUPPORTED into two error codes, E_TRANSPORT_UNKNOWN and E_TRANSPORT_NOT_SUPPORTED. Transport Unknown is for cases when you don't understand the address given to you. Transport not supported is for when you know the address but cannot dial it. After that this case will be Transport Not Supported

sukunrt: https://github.com/libp2p/specs/pull/538#discussion_r1176535618

autonat/autonat-v2.md
sukunrt: another great suggestion 🙇 

I have removed the AttemptMessage wrapper. 

I have also made the protocols `autonat/2/dial` and `autonat/2/attempt`. I don't see any benefit from using semver like protocol numbering here.

autonat/autonat-v2.md
sukunrt: I have incorporated this suggestion.

autonat/autonat-v2.md
sukunrt: I have introduced the error code `E_CONN_UPGRADE_FAILED` here.

autonat/autonat-v2.md
```diff
@@ -0,0 +1,255 @@
...
+the initiator node on the `/libp2p/autonat/2/dial` stream. This response
+contains a list of `DialStatus`es with a status for each address in the list up
+to and including the address that the peer attempted to dial. The `DialStatus`
+for an address is set according to [Requirements for
+DialStatus](#requirements-for-dialstatus). The response also contains an
+appropriate `ResponseStatus` set according to [Requirements For
+ResponseStatus](#requirements-for-responsestatus).
+
+The initiator MUST check that the nonce received in the `DialAttempt` is the
+same as the nonce the it sent in the `DialRequest`. If the nonce is different,
```
mxinden, line 97:
```suggestion
same as the nonce it sent in the `DialRequest`. If the nonce is different,
```

autonat/autonat-v2.md
```diff
@@ -0,0 +1,255 @@
...
+### Amplification Attack Prevention
+
+![Interaction](autonat-v2-amplification-attack-prevention.svg)
+
+
+When a client asks a server to dial an address that is not the client's observed
+IP address, the server asks the client to send him some non trivial amount of
+bytes as a cost to dial a different IP address. To make amplification attacks
+unattractive, the number of bytes is decided such that it's sufficiently larger
+than a new connection handshake cost.
```
mxinden, line 167: @sukunrt @marten-seemann do you have a recommendation on the number of bytes to request in mind?

autonat/autonat-v2.md
mxinden: Yes, thank you @sukunrt!

autonat/autonat-v2.md
```diff
@@ -0,0 +1,255 @@
...
+    }
+}
+
+message DialRequest {
+    repeated bytes addrs = 1;
+    fixed64 nonce = 2;
+}
+
+message DialDataRequest {
+    uint32 addrIdx = 1;
```
mxinden, line 222: Instead of an index, why not send the address itself? In my eyes just a small complexity reduction. Feel free to ignore. Based on intuition, the additional bytes (index < address) doesn't have a performance impact.

mxinden (approved):
This is good to merge from my end. Thank you for the work. Let's wait for reviews / approvals from others.

autonat/autonat-v2.md
```diff
@@ -0,0 +1,255 @@
...
+or reject the cost of dial. If the client rejects the cost, the client resets
+the stream and the `DialRequest` is considered aborted. If the client accepts
+the cost, the client starts transferring `numBytes` bytes to the server. The
+server on receiving `numBytes` bytes proceeds to dial the candidate address. 
+
+## Implementation Suggestions
+
+For any given address, implementations SHOULD do the following
+- periodically recheck reachability status
+- query multiple peers to determine reachability
```
MarcoPolo, line 185:
```suggestion
- Clients SHOULD NOT reuse their listening port when making a `DialRequest`. This avoids testing for reachability on an existing 4-tuple connection path. If both the client and server are reusing ports for dialing and listening, you'll end up using an existing hole-punched path rather than testing for reachability. 
```

Or something along those lines. The server could also control this by not reusing their listening port, but it's better for the client to do this since they can make sure they are doing the correct thing. And it avoids having the server allocate a new port.

MarcoPolo (commented):
One note to implementations

autonat/autonat-v2.md
thomaseizinger: The server being the node that wants its address to be probed for reachability right?

How would the server allocating a new port even work? If the server is behind a router that only forwards a particular port, then the AutoNAT probe would fail despite the original address actually being reachable.

autonat/autonat-v2.md
sukunrt: I think your suggestion is right. I'll resolve this while taking up the implementation. This will help us avoid sharing the slice between the thread that makes the dial request and the thread that receives the response. Similar effect can be had by sharing the idx to the requesting thread, but this seems simpler to me. I'll resolve this while implementing.

autonat/autonat-v2-amplification-attack-prevention.plantuml
```diff
@@ -0,0 +1,17 @@
...
+
+skinparam sequenceMessageAlign center
+skinparam defaultFontName monospaced
+
+
+== Amplification Attack Prevention ==
+
+Cli -> Srv: [dial] DialRequest:{nonce: 0xabcd, addrs: (addr1, addr2, addr3)}
+Srv -> Cli: [dial] DialDataRequest:{addrIdx: 1, numBytes: 120k}
+Cli -> Srv: [dial] {120k bytes}
```
thomaseizinger, line 13: I think we should have a message for this too, otherwise implementing the decoder for these messages is going to be quite annoying.

autonat/autonat-v2.md
marten-seemann: We're talking about an amplification attack, so number of bytes sent is the currency we should be thinking in.

If you want to introduce a PoW, you'd have to think about the exchange rate between CPU cycles and bandwidth, which is not at all obvious what a safe value would be.

autonat/autonat-v2-amplification-attack-prevention.plantuml
sukunrt: So the protocol is if the client accepts the DialDataRequest, it starts sending that amount of data. If the client doesn't accept the DialDataRequest, it resets the stream. You do not need to deserialise anything after you send DialDataRequest here. Just read that many bytes. 

Wrapping the entire data in a protobuf will make the protobuf size very large.

Do you think adding a `DataRequestAccepted` message before sending that amount of data as raw bytes will make things cleaner?

autonat/autonat-v2.md
sukunrt: The real cost here is getting the Server to do a handshake with a target. see @mxinden's [comment](https://github.com/libp2p/specs/issues/536#issuecomment-1505313751) for details. 

How should we measure the compute cost of a TLS or noise handshake?

autonat/autonat-v2-amplification-attack-prevention.plantuml
thomaseizinger: The reason I am asking for this is because at least in `rust-libp2p`, we typically implement "codecs" which are modules that turn byte streams into typed streams (and sinks) of messages. Reading bare half-way through does not play that well with strong type systems.

What do you mean it will make the protobuf large? Can't you just have a dynamically sized byte buffer in the protobuf? What is the difference?

thomaseizinger: > The reason I am asking for this is because at least in `rust-libp2p`, we typically implement "codecs" which are modules that turn byte streams into typed streams (and sinks) of messages. Reading bare half-way through does not play that well with strong type systems.

I guess we could unwrap the stream again and read a specified number of bytes.

sukunrt: >What do you mean it will make the protobuf large

You'll need to bring the whole serialised protobuf into memory before decoding. This will take memory of the order of 100kb. Not prohibitive, but I'd like to avoid this if we can.

If I understand correctly, you would prefer it to be a protobuf so you can have a typed stream which says first object received from the stream is of type A and the second object received is of type B, right?

sukunrt: cc @marten-seemann, @MarcoPolo since we had a small discussion on this [before](https://github.com/libp2p/specs/pull/542#discussion_r1171005513).

marten-seemann: > > > What do you mean it will make the protobuf large
> > 
> > 
> > You'll need to bring the whole serialised protobuf into memory before decoding. This will take memory of the order of 100kb. Not prohibitive, but I'd like to avoid this if we can.
> 
> Hmm, not sure how much of a hack this is but technically, the receiving end could declare the protobuf _without_ the payload field and thus avoid allocating it?

That might work, but you're now hardcoding assumptions about the wire format of the Protobuf, outside of your Protobuf library. I'd like to avoid that.



> Correct. Rust has tagged enums, meaning we can have rich types describing all fields of each message. But I don't really want to argue for it from an impl PoV. It is not that we are limited by Rust here. Instead, I think Rust's strong type systems makes it obvious that this isn't a consistent design

Not sure I agree with this framing, but in the end, this boils down to a matter of taste.

How about repeatedly sending a smaller message (limited by our usual Protobuf size), until we've reached the required number of bytes?

sukunrt: > Correct. Rust has tagged enums, meaning we can have rich types describing all fields of each message. But I don't really want to argue for it from an impl PoV. It is not that we are limited by Rust here. Instead, I think Rust's strong type systems makes it obvious that this isn't a consistent design:

So in the current proposal because the number of bytes that you need to read in the second message is not fixed and is equal to the number you asked for in the `DialDataRequest` this prevents you from writing a generic reader, right? 
If that's the case we can prefix the bytes sent with a unsigned-varint just like we prefix other protobuf messages sent. This will also be consistent with the other messages exchanged everywhere which I'd argue is the source of inconsistency in the design as opposed to bytes not being sent as a TLV object.

autonat/autonat-v2.md
sukunrt: No. The server is the node handling the dial request. The client is the initiator of the dial request. Apologies for the confusion here, I have reworded to consistently use client and server everywhere. 

>If the server is behind a router that only forwards a particular port, then the AutoNAT probe would fail despite the original address actually being reachable.

Assuming by server you mean the client that initiates the request for checking reachability.
 
This case will work fine. Say the node has setup port forwarding to forward router port 8080 to its listen port 5000. The node will not use this port to dial out, it will dial out of one of the higher ports, say 45000. The server which will handle the dial request will see the request to probe reachability for Client-IP:8080. It'll dial Client-IP:8080 and this will be forwarded correctly if it's reachable.

sukunrt: @MarcoPolo I added some text regarding this. Is this fine?

autonat/autonat-v2.md
sukunrt: done.

autonat/autonat-v2.md
mxinden: The above reasoning makes a lot of sense. Thanks @sukunrt and @marten-seemann.

I would expect future readers to ask the same question as @thomaseizinger did above. Can you include your reasoning in the document? I have found an `FAQ` section at the end of a specification useful, though I don't feel strongly about that structuring.

See e.g. https://github.com/libp2p/specs/blob/master/webrtc/webrtc.md#faq.

> do you have a recommendation on the number of bytes to request in mind?

I still think this is helpful for implementers.

sukunrt: >Can you include your reasoning in the document? 
Will do.

>do you have a recommendation on the number of bytes to request in mind?

I'll add this to the doc after calculating/measuring.
Off the top of my head, I think most handshakes finish within 10k bytes exchanged so something like 30KB-100KB should be enough.

sukunrt: added the recommendation.

sukunrt: >To make amplification attacks
unattractive, servers SHOULD ask for 30k to 100k bytes. Since most handshakes
cost less than 10k bytes in bandwidth, 30kB is sufficient to make attacks
unattractive.

autonat/autonat-v2-amplification-attack-prevention.plantuml
sukunrt: After doing a POC implementation, I think Marten's suggestion [here](https://github.com/libp2p/specs/pull/538#discussion_r1252319334) is the most suited to implementation. The problem from a go perspective is buffering in the protobuf msg reader on the server side. This can happen if someone does a pipelined implementation on the client side which sends data as soon as it has sent the request. see documentation for [NewDelimitedReader](https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/circuitv2/util/io.go#L28) for another case where this happens. While this situation is unlikely, it is simpler to just use protobufs for transferring data.

thomaseizinger: Yeah that makes sense. Would that message be a regular protobuf with a static array?

autonat/autonat-v2.md
```diff
@@ -0,0 +1,276 @@
...
+
+`E_REQUEST_REFUSED`: the server didn't attempt to serve the request because of
+rate limiting, resource limit reached or blacklisting.
+
+`E_INTERNAL_ERROR`: error not classified within the above error codes occured on
+server that prevented it from completing the request.
+
+Implementations MUST discard responses with status codes they do not understand. 
+
+### Amplification Attack Prevention
```
thomaseizinger, line 160: @mxinden mention that this design is takes inspiration from a different protocol but I forgot which one. Might be worth mentioning!

mxinden: E.g. related to the QUIC anti-amplification mechanism based on data size:

> Prior to validating the client address, servers MUST NOT send more than three times as many bytes as the number of bytes they have received. This limits the magnitude of any amplification attack that can be mounted using spoofed source addresses.

https://www.rfc-editor.org/rfc/rfc9000.html#section-8.1

I am not aware of an exact origin of the AutoNATv2 mechanism.

autonat/autonat-v2-amplification-attack-prevention.plantuml
sukunrt: I'm using bytes. 
```

message DialDataResponse {
    bytes data = 1;
}

```

autonat/autonat-v2.md
sukunrt: My original resolution didn't work. See [here](https://github.com/libp2p/specs/issues/562#issuecomment-1660451363) for details. Have changed this to a requirement on the server side. 

> Not sure how it works in other libp2ps, but in nim we connect to random nodes, not knowing what their capabilities are, and then we might ask them to dial-back if we find that they support autonat. We currently don't know at time of dialing if we are going to use them as AutoNat server.

sukunrt:
An implementation is available here: https://github.com/libp2p/go-libp2p/pull/2469/files
I have to clean up the PR a bit but the complete spec is implemented.

autonat/autonat-v2.plantuml
```diff
@@ -0,0 +1,20 @@
...
+participant Srv
+
+skinparam sequenceMessageAlign center
+skinparam defaultFontName monospaced
+
+
+== Dial Request Success==
+
+Cli -> Srv: [dial] DialRequest:{nonce: 0xabcd, addrs: (addr1, addr2, addr3)}
+Srv -> Cli: [attempt]addr2 DialAttempt:{nonce: 0xabcd}
```
marten-seemann, line 12: Not really sure how to achieve this, but it would be nice if it could be made clear that this is happening on a new connection (and a different 5-tuple), whereas the rest of the exchange happens on the same stream.

autonat/autonat-v2.md
sukunrt: should I mention QUIC?

autonat/autonat-v2.md
```diff
@@ -0,0 +1,287 @@
...
+`DialAttempt` message.
+
+Upon completion of the dial attempt, the server sends a `DialResponse` message
+to the client node on the `/libp2p/autonat/2/dial` stream. The response contains
+a list of `DialStatus`es with a status for each address in the list up to and
+including the address that the server attempted to dial. The `DialStatus` for an
+address is set according to [Requirements for
+DialStatus](#requirements-for-dialstatus). The response also contains an
+appropriate `ResponseStatus` set according to [Requirements For
+ResponseStatus](#requirements-for-responsestatus).
```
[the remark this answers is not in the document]
sukunrt: I think a single status is fine too. 
I wanted to specify the reason for refusing an address because that'd have allowed the client to infer more accurately that the particular address that the server refused to dial is a new transport not widely supported in the network. However multiple refusals to probe that address would support the same conclusion on a client.

autonat/autonat-v2.md
```diff
@@ -0,0 +1,287 @@
...
+### Requirements for ResponseStatus
+
+The `ResponseStatus` sent by the server in the `DialResponse` message MUST be
+set according to the following requirements
+
+`OK`: the server completed the request successfully. A request is considered
+completed successfully when the server either completes a dial(successfully or
+unsuccessfully) or rejects all addresses in the request as undialable. 
+
+`E_BAD_REQUEST`: the server was unable to decode the request.
```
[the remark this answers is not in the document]
sukunrt: Should we be sending errors on invalid requests? We do this in circuitv2. So if server receives a message on `dial-request` stream of type `DialResponse` or of type `DialDataResponse`, server sends `E_BAD_REQUEST`.

autonat/autonat-v2.md
```diff
@@ -0,0 +1,287 @@
...
+
+A client node wishing to determine reachability of its adddresses sends a
+`DialRequest` message to a server on a stream with protocol ID
+`/libp2p/autonat/2/dial`. Each `DialRequest` is sent on a new stream.
+
+This `DialRequest` message has a list of addresses and a fixed64 `nonce`. The
+list is ordered in descending order of priority for verification. AutoNAT V2 is
+only for testing reachability on Public Internet. Client MUST NOT send any
+private address as defined in [RFC
+1918](https://datatracker.ietf.org/doc/html/rfc1918#section-3) in the list.
```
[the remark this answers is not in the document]
sukunrt: I've changed it to `SHOULD NOT`

autonat/autonat-v2.md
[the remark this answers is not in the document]
sukunrt: I've changed it to sending a single status. cc @thomaseizinger

autonat/autonat-v2.md
```diff
@@ -0,0 +1,287 @@
...
+`/libp2p/autonat/2/dial`. Each `DialRequest` is sent on a new stream.
+
+This `DialRequest` message has a list of addresses and a fixed64 `nonce`. The
+list is ordered in descending order of priority for verification. AutoNAT V2 is
+only for testing reachability on Public Internet. Client MUST NOT send any
+private address as defined in [RFC
+1918](https://datatracker.ietf.org/doc/html/rfc1918#section-3) in the list. 
+
+Upon receiving this request, the server selects the first address from the list
+of addresses that it is capable of dialing. The server MUST NOT dial any address
```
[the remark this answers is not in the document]
sukunrt: I agree. this was unclear phrasing.

autonat/autonat-v2.md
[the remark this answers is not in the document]
marten-seemann: Why not just reset the stream (with an autonat-specific error code, once we have those)?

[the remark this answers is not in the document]
sukunrt: With error codes that would be fine. Without them it might be confusing.

autonat/autonat-v2.md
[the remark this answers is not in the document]
thomaseizinger: Works for me. The client probably doesn't need to know why a certain address was skipped. With a high probability, all previous addresses are not supported / understood.

Note that almost the same profiling can be done using the server's identify response and the advertised listen addresses.

autonat/autonat-v2.md
```diff
@@ -0,0 +1,287 @@
...
+
+All RPC messages sent over a stream are prefixed with the message length in
+bytes, encoded as an unsigned variable length integer as defined by the
+[multiformats unsigned-varint spec][uvarint-spec]. 
+
+All RPC messages on stream `/libp2p/autonat/2/dial` are of type `Message`. A
+`DialRequest` message is sent as a `Message` with the `msg` field set to
+`DialRequest`. `DialResponse` and `DialDataRequest` are handled similarly. 
+
+On stream `/libp2p/autonat/2/attempt`, a `DialAttempt` message is sent directly
```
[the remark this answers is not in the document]
sukunrt: done.

autonat/autonat-v2.md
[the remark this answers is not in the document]
sukunrt: leaving it as it is (client resets the stream)

umgefahren:
I read the spec and liked it a lot. I just have a tiny questiob:
Why do we use bandwidth to make the attack costly? Since a server also has to pay for bandwidth. Why are we not using a different mechanism like proof-of-work? Although I get why this is difficult, since the cost of the calculation, say prime factorization, has to exceed the bandwidth cost for the dial of the server. But since we took that from QUIC, I‘m sure it’s fine. I just wanted to know.

umgefahren:
> > Why do we use bandwidth to make the attack costly? Since a server also has to pay for bandwidth. Why are we not using a different mechanism like proof-of-work? Although I get why this is difficult, since the cost of the calculation, say prime factorization, has to exceed the bandwidth cost for the dial of the server. But since we took that from QUIC, I‘m sure it’s fine. I just wanted to know.
> 
> I had the same question and @mxinden explained it to me as: You want to make the attacker pay in the same "currency" as what gets used in the attack. Opening a new connection to a different node imposes bandwidth on that node's downlink, even if - in the best case - they drop all incoming packets and thus don't incur any other costs.
> 
> If you make an attacker pay with PoW, you'd have to figure out what the conversion-rate between bandwidth and CPU power is to make the attack economically nonsensical. If on the other hand, we make them pay with the same bandwidth they roughly gets sent as part of establishing a new connection, the attacker doesn't gain anything could as well just spent the bandwidth on the target itself without us having to worry what the economic cost relationship between our mitigation attempt and the actual attack is.

Thanks for clearing that up. I especially get that it's hard to figure out this conversion rate, however we can't deny that this will put strain on the servers that provide AutoNATv2 services.

thomaseizinger:
> we can't deny that this will put strain on the servers that provide AutoNATv2 services.

Yeah, I am not sure if we really have a choice TBH but @marten-seemann, @sukunrt or @mxinden can probably elaborate more on this design. Given that it came up the 2nd time now, should we add a more detailed explanation of the scheme to the spec?

marten-seemann:
How is this putting a strain on servers? It’s the server that decides when to require this verification, and for every incoming request, the server has the option to reject it.

umgefahren:
> How is this putting a strain on servers? It’s the server that decides when to require this verification, and for every incoming request, the server has the option to reject it.

Yes that's true.

umgefahren:
I need another clarification: In line 67 it is mentioned that every `DialRequest` will be sent on a stream with protocol ID `/libp2p/autonat/2/dial`. However in Line 219 it is also said that a `DialRequest` is a `Message` send on `/libp2p/autonat/2/dial-request`. This is confusing to me. Which one is it now?

Or is only the `DialRequest` a `Message` on `/libp2p/autonat/2/dial` and all other `Message`, like `DialDataRequest`, `DialDataResponse` are happening on `/libp2p/autonat/2/dial-request`. If that's correct, why use `Message` on `/libp2p/autonat/2/dial`?

sukunrt:
Thanks [umgefahren](https://github.com/umgefahren), I've fixed this. It should be `/libp2p/autonat/2/dial-request`

sukunrt:
@umgefahren 

> Or is only the DialRequest a Message on /libp2p/autonat/2/dial and all other Message, like DialDataRequest, DialDataResponse are happening on /libp2p/autonat/2/dial-request. If that's correct, why use Message on /libp2p/autonat/2/dial

on the stream `/libp2p/autonat/2/dial-request` we exchange messages, DialRequest, DialDataRequest, DialDataResponse, DialResponse. Since we exchange multiple types of messages on the stream from both sides, we wrap them in a Message type protobuf so that we can determine which type of message we have received. So a client on receiving a Message from server can determine whether it is a DialDataRequest or is it a DialResponse.

sukunrt:
> Another question, just to make sure: There will always be several DialDataResponse, since the upper bound for the amount of data per message is capped at 4096 bytes and it is required to send between 30kB and 100kB. Did I understand that correctly? That means the amount of data received so far have to be tracked, per connection.

Yes your understanding is correct, there'll be multiple DialDataResponse till you get the requested amount of data. You will have to track the amount of data received so far per dial-request stream. A peer can open multiple streams in parallel. Of course an implementation MAY choose to not allow this, but I don't think it is necessary to make this a MUST. 

>And an additional suggestion: What is the upper bound for the size of Message, i.e. how many Multiaddress can be send by the client?

I will think about this. Currently this is only limited by the size of the Message which again is implementation dependent. I think most implementations have a limit lower than 8kB. I think it is fine to add a suggestion to limit the number of addresses inspected.

umgefahren:
Another quick question, I probably missed something: When the server successfully dials the client and provides the nonce. The client closes the stream either way. How does the server know if the provided nonce was correct?

umgefahren:
In the spec it says that all private IP address should be excluded, but it also says it's just for checking reachability on the public internet. That said, we should exclude:

For IPv4:
- addresses on "this network" (i.e. IPv4 address starting with a 0)
- private IP addresses
- shared IP addresses
- loopback IP addresses (could create interesting behavior though)
- link local IP addresses
- reserved for future protocols
- documenting
- benchmarking
- reserved
- broadcast

For IPv6:
- Unspecified
- Loopback
- unicast link local
- unique local
- documentation
- IPv4 mapped
- IPv4-IPv6 translation
- Discard only
- IETF Protocol assignments

In the rust-libp2p implementation there was a PR discussing those globally reachable IP address: libp2p/rust-libp2p#3814

Also this list is probably not complete and not formal, it's also a small nitpick.

sukunrt:
@umgefahren it'd be better if you can add the comments as a review, commenting on the specific section. 

> In the spec it says that all private IP address should be excluded, but it also says it's just for checking reachability on the public internet. That said, we should exclude:


It means non public. Happy to change the wording to `non public`. 


> Another quick question, I probably missed something: When the server successfully dials the client and provides the nonce. The client closes the stream either way. How does the server know if the provided nonce was correct?

A correct server will always provide the correct nonce, no? This issue should be easy enough to debug for implementors without signalling from the client. Is there any benefit to the server knowing that it provided an incorrect nonce?

umgefahren:
> @umgefahren it'd be better if you can add the comments as a review, commenting on the specific section.

I will do that the next time. I'm sorry.
 
> > In the spec it says that all private IP address should be excluded, but it also says it's just for checking reachability on the public internet. That said, we should exclude:
> 
> It means non public. Happy to change the wording to `non public`.

Thanks for clarification.

> > Another quick question, I probably missed something: When the server successfully dials the client and provides the nonce. The client closes the stream either way. How does the server know if the provided nonce was correct?
> 
> A correct server will always provide the correct nonce, no? This issue should be easy enough to debug for implementors without signalling from the client. Is there any benefit to the server knowing that it provided an incorrect nonce?

I think there is a benefit. There is a pathological example where the network configuration or a NAT forward traffic to the wrong libp2p node. Not the one that requested the dial back, but a different one. In that case the server would report reachability on that address, but it's actually not reaching the peer in question. I'm not an expert enough here to think of any case where that might occur apart from a bad config or a malicious actor.

sukunrt:
In this case, the client sees that the server is reporting OK-Reachable but it has not received the nonce, so it should reject the response.

thomaseizinger:
> > > Another quick question, I probably missed something: When the server successfully dials the client and provides the nonce. The client closes the stream either way. How does the server know if the provided nonce was correct?
> > 
> > 
> > A correct server will always provide the correct nonce, no? This issue should be easy enough to debug for implementors without signalling from the client. Is there any benefit to the server knowing that it provided an incorrect nonce?
> 
> I think there is a benefit. There is a pathological example where the network configuration or a NAT forward traffic to the wrong libp2p node.

If we dial the node with `/p2p`, the connection will never be fully established if we end up at a different node so you can't send the nonce over.

umgefahren:
So since that is not possible, the implementation doesn't needs to handle this case, right?

But thanks for the clarification and sorry for the dumb questions.

thomaseizinger:
> So since that is not possible, the implementation doesn't needs to handle this case, right?

Yep I think you are right! We can assume that this will never happen. Feel free to use `debug_assert` if you want to be sure!

> But thanks for the clarification and sorry for the dumb questions.

No worries at all! I think your questions are pretty spot on actually :)

autonat/autonat-v2.plantuml
```diff
@@ -0,0 +1,20 @@
...
+Srv -> Cli: [dial] DialResponse:{status: OK, dialStatuses:(E_TRANSPORT_NOT_SUPPORTED, OK)}
```
thomaseizinger, line 13: This mentions `E_TRANSPORT_NOT_SUPPORTED` but that is missing from the protobufs?

umgefahren:
While doing the rust-libp2p implementation, we discovered a race condition, which we are now circumventing by a 100ms delay. You can read the finally comment by @thomaseizinger here: https://github.com/umgefahren/rust-libp2p/pull/1#issuecomment-1872977004

It happens when the server successfully performs a dial back, thus sends the confirmation of the address back to the client. However the client hasn't progressed enough to be notified of that successful dial back when receiving the confirmation. In that case the client wrongly assumed an address was confirmed where no dial back occurred.

thomaseizinger:
> In that case the client wrongly assumed an address was confirmed where no dial back occurred.

Minor correction here: The behaviour is usually that the client discards the "successful" confirmation because it has not yet processed the dial-back so it thinks the server is sending it a confirmation without having actually done the dial.

I think the correct way to solve this would be to add an ACK message from the client back to the server for the dial-back where the client can say: "Yes I've processed your dial-back". The server can then proceed to respond on the other stream and thus guarantee that we don't have a race condition between the two streams.

sukunrt:
You can read the closing of the stream as the ACK. See: https://github.com/libp2p/go-libp2p/blob/sukun/autonat-v2-2/p2p/protocol/autonatv2/server.go#L251-L257

The spec also dictates closing the stream: https://github.com/libp2p/specs/blame/autonat-v2/autonat/autonat-v2.md#L87

Do you think an explicit ACK is better?

thomaseizinger:
> You can read the closing of the stream as the ACK. See: [libp2p/go-libp2p@`sukun`/autonat-v2-2/p2p/protocol/autonatv2/server.go#L251-L257](https://github.com/libp2p/go-libp2p/blob/sukun/autonat-v2-2/p2p/protocol/autonatv2/server.go?rgh-link-date=2024-01-29T04%3A55%3A11Z#L251-L257)
> 
> The spec also dictates closing the stream: [`autonat-v2`/autonat/autonat-v2.md#L87 (blame)](https://github.com/libp2p/specs/blame/autonat-v2/autonat/autonat-v2.md#L87)
> 
> Do you think an explicit ACK is better?

Yeah I think so. I associate closing a stream with "I have no more data to write". The client never writes data so why wouldn't it immediately close the stream? Also, reading a stream and waiting for that to fail because it has been closed it also somewhat odd :man_shrugging:

sukunrt:
>The client never writes data so why wouldn't it immediately close the stream? 

That's a fair point. I'll add an ACK.

thomaseizinger (approved):
Nice, thank you!

autonat/autonat-v2.md
```diff
@@ -0,0 +1,299 @@
...
+
+
+If an attacker asks a server to dial a victim node, the only benefit the
+attacker gets is forcing the server and the victim to do a cryptographic
+handshake which costs some bandwidth and compute. The attacker by itself can do
+a lot of handshakes with the victim without spending any compute by using the
+same key repeatedly. The only benefit of going via the server to do this attack
+is not spending bandwidth required for a handshake. So the prevention mechanism
+only focuses on bandwidth costs. There is a minor benefit of bypassing IP
+blocklists, but that's made unattractive by the fact that servers may ask 5x
```
Stebalien, line 196: I don't think we can simply shrug this off. This is called a _reflection_ attack and has been a huge issue for open DNS resolvers.

Fixing the amplification side does go a long way, but paying a 5x bandwidth cost for a bunch of free IP addresses seems like a pretty reasonable tradeoff from an attacker's standpoint (especially because said attacker isn't paying for the bandwidth, but likely needs to compromise one machine per IP address).

Stebalien: Also note: home NAT users likely don't need this feature. That is:

1. They likely only need 1 dialable address.
2. They likely don't care which one.
3. Their outbound and inbound IPs are likely identical.

Being willing to dial other addresses _does_ matter for, e.g., AWS and other special settings where there are separate ingress IP addresses. But, in that case, maybe the user should just configure their node correctly rather than relying on AutoNAT? AutoNAT specifically exists to enable _home_ users.

sukunrt: It simplifies client implementations as they don't need to worry about IPv4 peer vs IPv6 peer. Though the benefit isn't huge since most IPv4 servers won't have IPv6 connectivity so they any way cannot check the IPv6 address. 

>Fixing the amplification side does go a long way, but paying a 5x bandwidth cost for a bunch of free IP addresses seems like a pretty reasonable tradeoff from an attacker's standpoint (especially because said attacker isn't paying for the bandwidth, but likely needs to compromise one machine per IP address).

Can you elaborate here? why isn't the attacker paying for the bandwidth.

p-shahi:
@sukunrt given that AutoNatv2 is merged in two reference implementations https://github.com/libp2p/rust-libp2p/pull/5526 ([released in 0.13.0](https://github.com/libp2p/rust-libp2p/blob/master/protocols/autonat/CHANGELOG.md#0130)) and https://github.com/libp2p/go-libp2p/pull/2469 ([released in 0.36.1](https://github.com/libp2p/go-libp2p/releases/tag/v0.36.1))

Are there any outstanding comments that need to be addressed before this pull request can be merged?
Also, the maturity should be either a Recommendation (I believe there is demonstrated interop between Go and Rust impls?)

umgefahren:
> Also, the maturity should be either a Recommendation (I believe there is demonstrated interop between Go and Rust impls?)

@sukunrt and I did interop testing and successfully verified that they are working together.

sukunrt:
The implementation is not used in go-libp2p yet. We should merge this after we start inferring reachability in go-libp2p.

autonat/autonat-v2.md
sukunrt: - This features enables the following attack:
    - Contact a large number of autonat v2 servers.
    - Give them the target's address to connect to
    - When requested for data: provide the data slowly:
          * The stream timeout is 1 minute
          * In the period, with a 1Gbps connection, you can send 60Gb ~= 6GB
          * Maximum dial data requirement is 100kB
          * So, in theory, you can run this with 60_000 peers in parallel. 
          * The servers have a random wait of up to 3 seconds precisely for this scenario. So in theory we can have 20k connections a second for 3 seconds to the target. 
    - We can make a bunch of implementation improvements to reduce the harm here. The simplest ones being: Only wait 10 seconds for the dial data, and wait for 5 seconds before dialing. That would reduce the max new connection rate to 2k / second, which is very manageable. 
    - The ideal solution is to introduce rate limits for new connections. 
- There's another problem with this feature related to implementation difficulty:
    - The primary use for this feature is to allow both IPv4 and IPv6 addresses to be tested without worrying about whether we have a v4 or a v6 connection. So you can ask a v4 peer to test your v6 address. This requires correctly reporting an error in case the server has no v6 connectivity, which is majority of servers.  
    - I'm not sure if the rust implementation correctly handles this case. @umgefahren  please correct me if I'm wrong here. 
            * See discussion around this comment: https://github.com/umgefahren/rust-libp2p/pull/1#issuecomment-2168354437 
    - I'm also not sure if we can rely on other implementations to correctly handle this. They might just make a dial, fail, and report unreachable. 
- If we have to ensure that we check v6 addresses with a v6 peer, it might be better to just disable this feature.

@MarcoPolo @Stebalien @umgefahren thoughts?

umgefahren: The potential attack vector seems to be described correctly, however I'm not sure that rust-libp2p is affected. The whole process is allowed to take a maximum of 10 seconds:
https://github.com/libp2p/rust-libp2p/blob/8ceadaac5aec4b462463ef4082d6af577a3158b1/protocols/autonat/src/v2/server/handler/dial_request.rs#L66
 However, we don't wait any time before dealing back. This mitigation is a quick fix, I can prepare.

Regarding IPv4 and IPv6 I stand with @thomaseizinger's comment on that matter: https://github.com/umgefahren/rust-libp2p/pull/1#issuecomment-2169530100

So it _correctly_ handles that case, in that we don't generate false positives.

autonat/autonat-v2.md
```diff
@@ -0,0 +1,299 @@
...
+
+![Autonat V2 Interaction](autonat-v2.svg)
+
+A client node wishing to determine reachability of its addresses sends a
+`DialRequest` message to a server on a stream with protocol ID
+`/libp2p/autonat/2/dial-request`. Each `DialRequest` is sent on a new stream.
+
+This `DialRequest` message has a list of addresses and a fixed64 `nonce`. The
+list is ordered in descending order of priority for verification. AutoNAT V2 is
+only for testing reachability on Public Internet. Client SHOULD NOT send any
```
MarcoPolo, line 71: I think this should be "MUST NOT". and "The server MUST NOT dial any private address".

MarcoPolo: Can you know that you are indeed on the same private network?

sukunrt: I'm not completely sure. 

If you see local connection address in private IP range and remote connection address in private IP range, is that enough to conclude that you're in some private network?
Note you cannot rely on https://datatracker.ietf.org/doc/html/rfc1918 subnet masks as you can make a private network from a collection of smaller private networks.

MarcoPolo: I think if the local address and remote address are in the same private subnet, then it would be okay.

How about adding "The server SHOULD NOT dial any private address". This leaves the door open in the spec.

I'm not sure the usefulness of doing this though, but maybe others might have a use for it.

sukunrt: done.

PR #542, opened 2023-04-15
allow autonat v2 to dial all ips

### LINKED ISSUE #536: Proposal: allow AutoNAT to dial all IP addresses, without risking amplification attacks

sukunrt:
Closes #536

### DIFF
--- a/autonat/autonat-v2.md
+++ b/autonat/autonat-v2.md
@@ -44,6 +44,10 @@ Compared to `autonat v1` there are two major differences
 testing reachability for an individual address
 2. `autonat v2` provides a mechanism for nodes to verify whether the peer
 actually successfully dialled an address.
+3. `autonat v2` provides a mechanism for nodes to dial an ip address different
+from the requesting node's observed ip address without risking amplification
+attacks. `autonat v1` disallowed such dials to prevent amplification attacks.
+
 
 
 ## AutoNAT V2 Protocol
@@ -62,10 +66,19 @@ of priority for verfication.
 
 Upon receiving this message the peer selects the first candidate from the list
 of candidates that it is capable of dialing. The peer MUST NOT dial any
-candidate other than this selected candidate. It dials the selected candidate's
-address, opens a stream with Protocol ID `/libp2p/autonat/2.0.0/attempt` and
-sends a `DialAttempt` message with the candidate nonce. The peer MUST close this
-stream after sending the `DialAttempt` message.
+candidate other than this selected candidate. If this selected candidate address
+has an ip address different from the requesting node's observed ip address, peer
+initiates the Amplification attack prevention mechanism (see [Amplification
+Attack Prevention](#amplification-attack-prevention) ). On completion, the peer
+proceeds to the next step. If the selected address has the same ip address as
+the requesting node's observed ip address, peer directly proceeds to the next
+step skipping Amplification Attack prevention steps.
+
+
+The peer dials the selected candidate's address, opens a stream with Protocol ID
+`/libp2p/autonat/2.0.0/attempt` and sends a `DialAttempt` message with the
+candidate nonce. The peer MUST close this stream after sending the `DialAttempt`
+message.
 
 Upon completion of the dial attempt, the peer sends a `DialResponse` message to
 the initiator node on the `/libp2p/autonat/2.0.0/dial` stream with the index(0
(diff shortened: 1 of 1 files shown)


autonat/autonat-v2.md
```diff
@@ -95,25 +104,23 @@ the message. This includes inability to decode the requested address to dial.
...
-> In this case, the attacker provides a large number of clients with the same
-> faked MAPPED-ADDRESS that points to the intended target. This will trick all
-> the STUN clients into thinking that their addresses are equal to that of the
-> target. The clients then hand out that address in order to receive traffic on
-> it (for example, in SIP or H.323 messages). However, all of that traffic
-> becomes focused at the intended target. The attack can provide substantial
-> amplification, especially when used with clients that are using STUN to enable
-> multimedia applications.
+Upon receiving a `DialDataRequest` the client starts transferring `numBytes`
+number of bytes to the server.
```
sukunrt, line 121: Should I add a step where the client first sends a message indicating that it has accepted/rejected the cost? 
@marten-seemann

sukunrt: i added this step.

autonat/autonat-v2.md
```diff
@@ -36,29 +36,38 @@ Implementations can generate low priority address guesses and add them to
...
 
 Compared to `autonat v1` there are two major differences
-1. `autonat v1` allowed testing reachability for the node. `autonat v2` allows
-testing reachability for an individual address
+1. `autonat v2` allows testing reachability for an individual address. `autonat
+   v1` allowed testing reachability for the node. 
 2. `autonat v2` provides a mechanism for nodes to verify whether the peer
 actually successfully dialled an address.
+3. `autonat v2` provides a mechanism for nodes to dial an ip address different
+from the requesting nodes observed ip address without risking amplification
```
marten-seemann, line 44:
```suggestion
from the requesting node's observed ip address without risking amplification
```

autonat/autonat-v2.md
```diff
@@ -167,7 +180,15 @@ message DialRequest {
...
-    fixed64 nonce = 1;
+  fixed64 nonce = 1;
+}
+
+message DialDataRequest {
+  uint64 numBytes = 1;
+}
+
+message DialDataResponse {
+  bool accepted = 1;
```
marten-seemann, line 191: Why not put the bytes into this protobuf? Then we wouldn't need to break the framing to send these bytes.

sukunrt: That would require memory of the order of 100kb for the protobufs on both ends, right? 
The current version can be implemented using a smaller buffer.

I don't have good intuition here, do you think this is fine?

sukunrt: @marten-seemann friendly ping

MarcoPolo: The server is going to discard these bytes right away anyways. No point to allocate the space to put the bytes in the message. I agree with the initial design. (Sorry for the drive-by comment, I'll defer to whatever you all decide).

autonat/autonat-v2.md
```diff
@@ -95,25 +104,26 @@ the message. This includes inability to decode the requested address to dial.
...
 
-In order to prevent attacks like the one described in [RFC 3489, Section
-12.1.1](https://www.rfc-editor.org/rfc/rfc3489#section-12.1.1) (see excerpt
-below), implementations MUST NOT dial any multiaddress unless it is based on the
-IP address the requesting node is observed as. This restriction as well implies
-that implementations MUST NOT accept dial requests via relayed connections as
-one can not validate the IP address of the requesting node.
+Upon receiving a `DialDataRequest` message the client responds with
+`DialDataResponse` message with a boolean `accepted` indicating if it accepted
+the request. If `accepted` is `false`, the `DialRequest` is considered complete
```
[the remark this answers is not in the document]
sukunrt: I am accepting this. This looks simpler to me. Even if we add a different cost for dial in future, say some POW like mechanism, the client can just reject the charge by resetting the stream.

autonat/autonat-v2.md
```diff
@@ -107,26 +120,24 @@ peer that prevented it from completing the request.
...
-> In this case, the attacker provides a large number of clients with the same
-> faked MAPPED-ADDRESS that points to the intended target. This will trick all
-> the STUN clients into thinking that their addresses are equal to that of the
-> target. The clients then hand out that address in order to receive traffic on
-> it (for example, in SIP or H.323 messages). However, all of that traffic
-> becomes focused at the intended target. The attack can provide substantial
-> amplification, especially when used with clients that are using STUN to enable
-> multimedia applications.
+On receiving a `DialRequest`, the server selects the first address it is capable
+of dialing. If this selected address has a ip different from the clients
```
marten-seemann, line 132:
```suggestion
of dialing. If this selected address has a ip different from the client's
```

autonat/autonat-v2.md
```diff
@@ -107,26 +120,24 @@ peer that prevented it from completing the request.
...
+observed ip, the server sends a `DialDataRequest` message with `numBytes` set to
+a sufficiently large value on the `/libp2p/autonat/2.0.0/dial-request` stream
 
+Upon receiving a `DialDataRequest` message, the client decides whether to accept
+or reject the cost of dial. If the client rejects the cost, the client resets
+the stream and the `DialRequest` is considered complete. If the client accepts
```
marten-seemann, line 138: Complete? Not aborted?

marten-seemann (approved):
LGTM.

Nit: I think it's spelled IP and not ip.

autonat/autonat-v2.md
sukunrt: that's a much better term

sukunrt:
>Nit: I think it's spelled IP and not ip.

oops, thanks!

PR #545, opened 2023-04-20
chore: [typo] Update README.md

FlyingDuck:
"more than" maybe a typo


marten-seemann (commented):
1. This is not a typo, this fundamentally changes the logic described here.
2. This is not correct.

FlyingDuck:
> 

got it. I miss the key info in reading.

Issue #471, opened 2022-10-20
Proposal: Back-pressure on opening new streams on yamux


thomaseizinger:
[Yamux](https://github.com/hashicorp/yamux/blob/master/spec.md) does not offer any form of control in how many streams can be opened by a remote.

This has been an annoyance for a while and transports like QUIC are useful because they do support managing the number of open streams.

I'd like to propose what I believe is a backwards-compatible extension of the yamux spec that gives us _some_ amount of back-pressure for streams.

The yamux specification states for [opening of a new stream](https://github.com/hashicorp/yamux/blob/master/spec.md#opening-a-stream):

> To open a stream, an initial data or window update frame is sent with a new StreamID. The SYN flag should be set to signal a new stream.
> The receiver must then reply with either a data or window update frame with the StreamID along with the ACK flag to accept the stream or with the RST flag to reject the stream.

My proposal is as follows:

> Implementations SHOULD keep track of how many streams are pending, i.e. they have not yet received an `ACK` for this stream. When receiving `RST` for a pending stream, implementations SHOULD take note of the number of currently pending streams as a "pending stream budget". For the remainder of the connection, implementations SHOULD NOT exceed this budget.
> Implementations MUST NOT assume for this budget to be a constant but shrink it in case the remote peer keeps sending `RST` flags for pending streams.

I think this is backwards-compatible because existing implementations already need to handle receiving `RST` for a stream at any point. What we are doing here is adding meaning to receiving `RST` whilst we have several pending substreams. This allows us to implement a bounded buffer on the receiver side where any streams exceeding that buffer will send a `RST` and be dropped. This means we will have 1 failed stream per physical connection that allows us to learn the other party's buffer size.

cc @mxinden @marten-seemann @rkuhn

marten-seemann:
The number of streams is probably the thing that peers will impose a limit on, not the number of pending streams. So the "pending stream budget" seems like the wrong thing to measure.

A global stream limit is not that useful to begin with. What we need in libp2p is a limit _per protocol_. This would require a coordination mechanism running on top of a stream muxer.

thomaseizinger:
> The number of streams is probably the thing that peers will impose a limit on, not the number of pending streams. So the "pending stream budget" seems like the wrong thing to measure.

Thanks for the input!

At least from an implementation PoV, I wouldn't immediately know, what a useful starting point for the max number of streams is and without an extension to the messages in yamux, we can't communicate a limit directly I think.

[...]

PR #547, opened 2023-05-18
yamux: recommend buffering of inbound streams

thomaseizinger:
Resolves https://github.com/libp2p/specs/issues/471.

cc (Interest-group) @mxinden @marten-seemann @wemeetagain @ianopolous

### DIFF
--- a/yamux/README.md
+++ b/yamux/README.md
@@ -165,5 +165,5 @@ The ACK backlog is defined as the number of streams that a peer has opened which
 To support a basic form of backpressure, implementions:
 
 - SHOULD at most allow an ACK backlog of 256 streams.
-- SHOULD buffer at least 256 unacknowledged inbound streams.
+- MAY buffer unacknowledged inbound streams instead of resetting them when the application currently cannot handle any more streams.
 - SHOULD NOT implicitly acknowledge streams but wait for the application to send the first DATA frame.
--- a/yamux/README.md
+++ b/yamux/README.md
@@ -167,4 +167,4 @@ To support a basic form of backpressure, implementions:
 - SHOULD at most allow an ACK backlog of 256 streams.
 - MAY buffer unacknowledged inbound streams instead of resetting them when the application currently cannot handle any more streams.
   Such a buffer MUST be bounded in size to mitigate DoS attacks.
-- SHOULD NOT implicitly acknowledge streams but wait for the application to send the first DATA frame.
+- MAY delay acknowledging new streams until the application has received or is about to send the first DATA frame.
--- a/yamux/README.md
+++ b/yamux/README.md
@@ -157,3 +157,7 @@ The Length should be set to one of the following to provide an error code:
 Yamux allows for a stream to be opened (and used) before it is acknowledged by the remote.
 The ACK backlog is defined as the number of streams that a peer has opened which have not yet been acknowledged.
 Implementations SHOULD at most allow an ACK backlog of 256 streams.
+Implementations SHOULD buffer at least 256 unacknowledged inbound streams.
+Implementations SHOULD NOT implicitly acknowledge streams but wait for the application to send the first DATA frame.
+
+Whilst the specification does not forsee a backpressure mechanism for opening new streams, implementing the ACK backlog as described above provides a limited form of backpressure that is backwards-compatible.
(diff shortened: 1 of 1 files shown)


mxinden (approved):
Thank you @thomaseizinger.

As far as I can tell, this only documents the status quo, i.e. it describes what go-libp2p already does today. Please correct me in case I am missing something @marten-seemann.

With that in mind, any objections moving forward here?

yamux/README.md
```diff
@@ -152,8 +152,18 @@ The Length should be set to one of the following to provide an error code:
...
+This presents a problem:
+A peer must read from the socket and decode the frames to make progress on existing streams.
+But any frame could also open yet another stream.
+
 The ACK backlog is defined as the number of streams that a peer has opened which have not yet been acknowledged.
-Implementations SHOULD at most allow an ACK backlog of 256 streams.
+To support a basic form of backpressure, implementions:
+
+- SHOULD at most allow an ACK backlog of 256 streams.
+- SHOULD buffer at least 256 unacknowledged inbound streams.
```
marten-seemann, line 168: Why? Resetting is a totally valid thing to do if you don’t want to buffer 256.

marten-seemann (changes requested):
Still not sure I understand what problem we’re trying to solve here. Is this really worth our time?

yamux/README.md
thomaseizinger: The practical issue is that the first thing that happens on a stream is multistream-select. Resetting a stream will be treated by most existing implementations as a negotiation failure.

So whilst the yamux spec allows this, existing implementations and our usage of yamux in libp2p doesn't really cope very well with that.

If I understand correctly, the meaning of SHOULD is that it is completely optional but when supported, gives access to more features.

In this case, buffering 256 streams allows an implementation to trigger the 256-backlog backpressure mechanism also documented in this section.

If you reset a stream, the other peer will just continue to open new streams. Interpreting _that_ as a way of signaling backpressure was actually my original idea in #471 but we moved away from that in favor of following go-libp2p's implementation.

thomaseizinger: Friendly ping @marten-seemann: Does the above explanation make sense? I'd like to close off this particular work stream so I'd appreciate some concrete suggestions on what to adjust in this PR so we can merge it and accurately reflect the status quo.

marten-seemann: Yes, my objections still stand.

thomaseizinger: Can we merge it with we change it to `MAY`?

thomaseizinger, line 168:
```suggestion
- MAY buffer unacknowledged inbound streams instead of resetting them when the application currently cannot handle any more streams.
```

yamux/README.md
```diff
@@ -152,8 +152,18 @@ The Length should be set to one of the following to provide an error code:
...
+This presents a problem:
+A peer must read from the socket and decode the frames to make progress on existing streams.
+But any frame could also open yet another stream.
+
 The ACK backlog is defined as the number of streams that a peer has opened which have not yet been acknowledged.
-Implementations SHOULD at most allow an ACK backlog of 256 streams.
+To support a basic form of backpressure, implementions:
+
+- SHOULD at most allow an ACK backlog of 256 streams.
+- MAY buffer unacknowledged inbound streams instead of resetting them when the application currently cannot handle any more streams.
```
thomaseizinger, line 168:
```suggestion
- MAY buffer unacknowledged inbound streams instead of resetting them when the application currently cannot handle any more streams.
  Such a buffer MUST be bounded in size to mitigate DoS attacks.
```

yamux/README.md
```diff
@@ -152,8 +152,19 @@ The Length should be set to one of the following to provide an error code:
...
+  Such a buffer MUST be bounded in size to mitigate DoS attacks.
+- SHOULD NOT implicitly acknowledge streams but wait for the application to send the first DATA frame.
```
marten-seemann, line 170:
```suggestion
- MAY delay acknowledging new streams until the application has received or is about to send the first DATA frame.
```

marten-seemann (approved):
LGTM modulo the suggestion for the last bullet point.

thomaseizinger:
> LGTM modulo the suggestion for the last bullet point.

Thanks! I've applied your suggestion verbatim :)

PR #548, opened 2023-05-30
[GossipSub 1.2] DONTSEND control message

Nashatyrev:
This PR introduces a new GossipSub version `1.2`

## New messages 

The new `DONTSEND` control message is added which notifies mesh peers to suspend sending back the full publish message based on its `messageId`

Other options for the message name: 
- `CHOKE_MESSAGE`: inspired by the `CHOKE` message (for a topic) from #413
- `IDONTWANT`: as opposed to `IWANT` 

## Simulation results 

Various simulations demonstrated ~30% traffic reduction and ~3-5% message delivery latency reduction

- @Menduist results involving the full TCP stack(the feature is named `choke` there) : https://hackmd.io/X1DoBHtYTtuGqYg0qK4zJw#4---BBR
- @Nashatyrev results (short cutting TCP stack): https://docs.google.com/spreadsheets/d/15tbZHBbWZmsDIY_xm9swNe9A2pcQ-gp4yE42iGDCJB4/edit#gid=1815594575

## Related work

Relates to #413

### DIFF
--- a/pubsub/gossipsub/gossipsub-v1.2.md
+++ b/pubsub/gossipsub/gossipsub-v1.2.md
@@ -1,0 +1,102 @@
+# Gossipsub v1.2
+
+# Overview
+
+This document aims to provide a minimal extension to the [gossipsub
+v1.1](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md)
+protocol.
+
+The proposed extensions are backwards-compatible and aim to enhance the
+efficiency (minimize amplification/duplicates and decrease message latency) of
+the gossip mesh networks for larger messages. 
+
+In more specific terms, a new control message is introduced: `DONTSEND`. It's primarily 
+intended to notify mesh peers that the node already received a message and there is no 
+need to send its duplicate.
+
+# Specification
+
+## Protocol Id
+
+Nodes that support this Gossipsub extension should additionally advertise the
+version number `1.2.0`. Gossipsub nodes can advertise their own protocol-id
+prefix, by default this is `meshsub` giving the default protocol id:
+- `/meshsub/1.2.0`
+
+## Parameters
+
+This section lists the configuration parameters that needs to agreed on across clients to avoid 
+ peer penalizations
+
+| Parameter               | Description                                                      | Reasonable Default |
+|-------------------------|------------------------------------------------------------------|--------------|
+| `max_dontsend_messages` | The maximum number of `DONTSEND` messages per heartbeat per peer | ???  |
+
+
+## DONTSEND Message
+
+### Basic scenario
+
+When the peer receives the first message instance it immediately broadcasts 
+(not queue for later piggybacking) `DONTSEND` with the `messageId` to all its mesh peers. 
+This could be performed prior to the message validation to further increase the effectiveness of the approach.    
+
+On the other side a node maintains per-peer `dont_send_message_ids` set. Upon receiving `DONTSEND` from 
+a peer the `messageId` is added to the `dont_send_message_ids` set. 
+When later relaying the `messageId` message to the mesh the peers found in `dont_send_message_ids` could be skipped. 
+
+Old entries from `dont_send_message_ids` could be pruned during heartbeat processing. 
+The prune strategy is outside of the spec scope and can be decided by implementations.
+
+`DONTSEND` message is supposed to be _optional_ for both receiver and sender. I.e. the sender may or may not utilize 
+this message. The receiver in turn may ignore `DONTSEND`: sending a message after the corresponding `DONTSEND` 
+should not be penalized.    
+
+The `DONTSEND` may have negative effect on small messages as it may increase the overall traffic and CPU load.
+Thus it is better to utilize `DONTSEND` for messages of a larger size.
+The exact policy of `DONTSEND` appliance is outside of the spec scope. Every implementation may choose whatever 
+is more appropriate for it. Possible options are either choose a message size threshold and broadcast `DONTSEND`
+on per message basis when the size is exceeded or just use `DONTSEND` for all messages on selected topics.
+
+To prevent DoS the number of `DONTSEND` control messages is limited to `max_dontsend_messages` per heartbeat  
+
+### Relying on `IHAVE`s
+
+Another potential additional strategy could be as follows. If a node receives `IHAVE` (from one or more peers)
+before the message is appeared in the mesh the node may request the message with `IWANT` and notify all mesh 
+peers that it don't want that message from them. 
+
+### Sending `IHAVE` to mesh peers who choked that particular message
+
+Reasonable addition to the later scenario would be to _immediately_ send `IHAVE` instead of a full message
+to those mesh peers who reported `DONTSEND`. That would notify mesh peers that the node has this message 
+and they could request it from you in case their `IWANT` requests fail in the previous scenario 
+
+### Cancelling `IWANT`
+
+If a node requested a message via `IWANT` and then occasionally receives the message from other peer it may 
+try to cancel its `IWANT` requests with the corresponding `DONTSEND` message. It may work in cases when a
+peer delays/queues `IWANT` requests and the `IWANT` request would be removed from the queue if not processed yet
+
+## Protobuf Extension
+
+The protobuf messages are identical to those specified in the [gossipsub v1.0.0
+specification](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md)
+with the following  control message modifications:
+
+```protobuf
+message RPC {
+ // ... see definition in the gossipsub specification
+}
+
+message ControlMessage {
+    // messages from v1.0
+    repeated ControlDontSend dontSend = 5;
+}
+
+message ControlDontSend {
+    required bytes messageID = 1;
+}
+
+```
+


vyzo:
That seems very reasonable; I like `IDONTWANT` more for the name.

@AgeManning any thoughts?

vyzo:
Actually, this might make some sense to implement with `IHAVE`, assuming we keep track of peer IHAVEs and avoid sending those messages to the peer.

However, this will conflate with heartbeat generated IHAVEs.
So overall, i think it is better to implement with an `IDONTWANT` message.

We can also effect some positive scoring on this action (if it pertains to messages we have seen or see shortly after)

Another consideration: if we sending the message to the peer already, would that be useful?
It is kind  of implicit.

PS:
@Nashatyrev do you want to join our episub group in telegram?

pubsub/gossipsub/gossipsub-v1.2.md
```diff
@@ -0,0 +1,102 @@
...
+- `/meshsub/1.2.0`
+
+## Parameters
+
+This section lists the configuration parameters that needs to agreed on across clients to avoid 
+ peer penalizations
+
+| Parameter               | Description                                                      | Reasonable Default |
+|-------------------------|------------------------------------------------------------------|--------------|
+| `max_dontsend_messages` | The maximum number of `DONTSEND` messages per heartbeat per peer | ???  |
```
Menduist, line 33: Is that necessary? We can just descore peer if they don't send duplicate, or we don't eventually see the message id
GossipSub already has too many parameters

pubsub/gossipsub/gossipsub-v1.2.md
```diff
@@ -0,0 +1,102 @@
...
+is more appropriate for it. Possible options are either choose a message size threshold and broadcast `DONTSEND`
+on per message basis when the size is exceeded or just use `DONTSEND` for all messages on selected topics.
+
+To prevent DoS the number of `DONTSEND` control messages is limited to `max_dontsend_messages` per heartbeat  
+
+### Relying on `IHAVE`s
+
+Another potential additional strategy could be as follows. If a node receives `IHAVE` (from one or more peers)
+before the message is appeared in the mesh the node may request the message with `IWANT` and notify all mesh 
+peers that it don't want that message from them.
```
Menduist, line 67: Not sure about this one, it means that by sending a single IHAVE to a peer, I'll make him send 6-12 DONTHAVE of the same size, so it's an amplification factor

pubsub/gossipsub/gossipsub-v1.2.md
```diff
@@ -0,0 +1,102 @@
...
+need to send its duplicate.
+
+# Specification
+
+## Protocol Id
+
+Nodes that support this Gossipsub extension should additionally advertise the
+version number `1.2.0`. Gossipsub nodes can advertise their own protocol-id
+prefix, by default this is `meshsub` giving the default protocol id:
+- `/meshsub/1.2.0`
```
Menduist, line 24: Is this necessary, since this is backward compatible? Could just send the the DONTSEND optimistically, and if the peer doesn't handle them, he will just skip them

vyzo:
Thinking a bit more about this, I am not entirely convinved we need any of this.

We can simply start keep tracking messages we have received from  a mesh peer recently; if they have send us a message, we should supress it. Otherwise we just send and the message and the IDONTWANT is implicit/redundant.

Menduist:
@vyzo the DONTSEND becomes useful when the messages get big enough that it actually takes time to transmit them

ie, a 0.5mb message over a 25mbps line will take 160ms to be sent to 8 peers, so sending a small DONTSEND before sending the full message gives us 160 ms more to avoid duplicates
Also, when more sophisticated relaying technique are used (for instance, sending to one peer after the other), this becomes even more useful

pubsub/gossipsub/gossipsub-v1.2.md
vyzo: I think we should bump minor when we add a new control msg.

But see my comments in the pr.

vyzo:
Ok, fair enough.
We should still do the supress thing however, for interacting with older peers.

Nashatyrev:
> Actually, this might make some sense to implement with `IHAVE`, assuming we keep track of peer IHAVEs and avoid sending those messages to the peer.
> 
> However, this will conflate with heartbeat generated IHAVEs. So overall, i think it is better to implement with an `IDONTWANT` message.

Yes, we discussed that option, and came to agreement that those messages have slightly different semantics which might make the difference under some conditions. For example a peer may send `IDONTWANT` prior to message validation which would be different from `IHAVE` as the message may appear to be invalid finally. 

>  I like IDONTWANT more for the name.

Yes I'm voting for `IDONTWANT` too 👍 

> Another consideration: if we sending the message to the peer already, would that be useful?

Yes, the basic strategy would be to broadcast corresponding `IDONTWANT` prior to broadcasting the message

pubsub/gossipsub/gossipsub-v1.2.md
Nashatyrev: > or we don't eventually see the message id

Sounds like a good idea 👍

pubsub/gossipsub/gossipsub-v1.2.md
Nashatyrev: You are right. That was just a potential usage option. But it indeed doesn't look great in that interpretation

pubsub/gossipsub/gossipsub-v1.2.md
Nashatyrev: I think @vyzo and @AgeManning had a similar conversation somewhere regarding protocol versioning. May be they have an opinion on that

vyzo: Bumping the version allows us to filter peers who wont understand the message, so we can save a bit of bandwidth.

pubsub/gossipsub/gossipsub-v1.2.md
mkalinin: > or we don't eventually see the message id

Considering `DONTSEND` is allowed to be sent before validation, a peer can be downscored if a message `DONTSEND` has been sent for appears to be invalid

Menduist: Do we have numbers on how much we can gain by allowing to send `DONTSEND`s before validation? (ie, how long the validation is in practice)

pubsub/gossipsub/gossipsub-v1.2.md
```diff
@@ -0,0 +1,102 @@
...
+ // ... see definition in the gossipsub specification
+}
+
+message ControlMessage {
+    // messages from v1.0
+    repeated ControlDontSend dontSend = 5;
+}
+
+message ControlDontSend {
+    required bytes messageID = 1;
```
Menduist, line 98: Just for discussion, what would be the pros/cons of also including the topic?
I think it's the first time that we reference messages only by their id on the wire, so needs some consideration

I guess the only con of include the topic is more bandwidth usage

pubsub/gossipsub/gossipsub-v1.2.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Basic scenario
+
+When the peer receives the first message instance it immediately broadcasts 
+(not queue for later piggybacking) `DONTSEND` with the `messageId` to all its mesh peers. 
+This could be performed prior to the message validation to further increase the effectiveness of the approach.    
+
+On the other side a node maintains per-peer `dont_send_message_ids` set. Upon receiving `DONTSEND` from 
+a peer the `messageId` is added to the `dont_send_message_ids` set. 
+When later relaying the `messageId` message to the mesh the peers found in `dont_send_message_ids` could be skipped.
```
mkalinin, line 46:
```suggestion
When later relaying the `messageId` message to the mesh the peers found in `dont_send_message_ids` should be skipped. 
```

IMO, SHOULD looks more appropriate in this context

pubsub/gossipsub/gossipsub-v1.2.md
vyzo: Note that validation is some networks can be slow, so there is real benefit by sending early.

vyzo:
@Menduist @mkalinin note the existence of the episub working group in telegram; our little Lair is open, reach out to me in Telegram if you want to join.

pubsub/gossipsub/gossipsub-v1.2.md
Nashatyrev: The `IDONTWANT` usage is totally optional and that behavior is not enforced in any means. So I'm not sure 'should' or 'could' makes any difference in that context. I'm ok with any variant though

mkalinin: This scenario describes potential `IDONTWANT` utilisation. SHOULD is an appropriate word considering the context, while `IDONTWANT` utilisation is optional (MAY be utilised) as you're writing it down the line

pubsub/gossipsub/gossipsub-v1.2.md
```diff
@@ -0,0 +1,102 @@
...
+On the other side a node maintains per-peer `dont_send_message_ids` set. Upon receiving `DONTSEND` from 
+a peer the `messageId` is added to the `dont_send_message_ids` set. 
+When later relaying the `messageId` message to the mesh the peers found in `dont_send_message_ids` could be skipped. 
+
+Old entries from `dont_send_message_ids` could be pruned during heartbeat processing. 
+The prune strategy is outside of the spec scope and can be decided by implementations.
+
+`DONTSEND` message is supposed to be _optional_ for both receiver and sender. I.e. the sender may or may not utilize 
+this message. The receiver in turn may ignore `DONTSEND`: sending a message after the corresponding `DONTSEND` 
+should not be penalized.
```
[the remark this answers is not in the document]
Menduist: https://github.com/libp2p/specs/pull/548#discussion_r1209989715

arnetheduck:
> Ok, fair enough. 

for context, here are two scenarios we're addressing with this upgrade, for large messages in particular:

* https://github.com/status-im/nim-libp2p/issues/850
* https://github.com/status-im/nim-libp2p/issues/859

ppopth:
@vyzo Can I also join the episub group? FYI, I am the network person of the EF research.

vyzo:
Of course! Reach out on telegram to add you.

vyzo:
Let's target this to #560

pubsub/gossipsub/gossipsub-v1.2.md
Nashatyrev: Removed those two usage scenarios

pubsub/gossipsub/gossipsub-v1.2.md
Nashatyrev: If no one is opposed or has any ideas on usage scenarios I would keep it without optional topic field to be more explicit

Nashatyrev:
Guys, what do you think about merging this proposal?

vyzo:
I think we are pretty much good here, but we should put some more structure to base pr first.

I'll try to get to that this weekend.

arnetheduck:
Ping :) This message is interesting for the upcoming ethereum hard fork.

AgeManning (approved):
This seems good to me and we are happy to implement

djrtwo:
Looks good to me!

pubsub/gossipsub/gossipsub-v1.2.md
```diff
@@ -4,16 +4,103 @@
...
+| `max_idontwant_messages` | The maximum number of `IDONTWANT` messages per heartbeat per peer | ???  |
+
+
+## IDONTWANT Message
+
+### Basic scenario
+
+When the peer receives the first message instance it immediately broadcasts 
+(not queue for later piggybacking) `IDONTWANT` with the `messageId` to all its mesh peers. 
+This could be performed prior to the message validation to further increase the effectiveness of the approach.
```
vyzo, line 58: Concerns about spam attacks triggering amplified IDONTWANT spam?

vyzo (commented):
Just a small comment, otherwise lgtm.

pubsub/gossipsub/gossipsub-v1.2.md
Nashatyrev: This doesn't look like a feasible attack vector to me: 
- `IDONTWANT` is primarily intended for larger messages, so the cumulative size of resulting `IDONTWANT` messages is expected to be significantly smaller than the original message
- If an attacker is sending invalid messages to initiate `IDONTWANT` spamming it would be pretty quickly banned due to negative scoring 
- And we have `max_idontwant_messages` limit as the last resort

pubsub/gossipsub/gossipsub-v1.2.md
```diff
@@ -4,16 +4,103 @@
...
+
+### Basic scenario
+
+When the peer receives the first message instance it immediately broadcasts 
+(not queue for later piggybacking) `IDONTWANT` with the `messageId` to all its mesh peers. 
+This could be performed prior to the message validation to further increase the effectiveness of the approach.    
+
+On the other side a node maintains per-peer `dont_send_message_ids` set. Upon receiving `IDONTWANT` from 
+a peer the `messageId` is added to the `dont_send_message_ids` set. 
+When later relaying the `messageId` message to the mesh the peers found in `dont_send_message_ids` SHOULD be skipped.
```
diegomrsantos, line 62: Why SHOULD instead of MUST?

Nashatyrev: In my mind since there is no way to enforce this behavior (after sending `IDONTWANT` a node may still receive the message from the correct peer) then there is basically no difference between `SHOULD` and `MUST`. 
I'm not sure about this though.

vyzo: yeah, if it cannot be enforced that it is a SHOULD.

Nashatyrev: That sounds reasonable as well. 
I'm totally fine with either option

Nashatyrev: @diegomrsantos your reasoning looks pretty fair. Thanks for comments 👍 
I've changed it to `MUST`.

pubsub/gossipsub/gossipsub-v1.2.md
```diff
@@ -4,16 +4,103 @@
...
+ // ... see definition in the gossipsub specification
+}
+
+message ControlMessage {
+    // messages from v1.0
+    repeated ControlIDontWant iDontWant = 5;
+}
+
+message ControlIDontWant {
+    required bytes messageID = 1;
```
ppopth, line 102: It seems that the official doc discourages using `required` label. See https://protobuf.dev/programming-guides/proto2/#field-labels

ppopth, line 102: Why not making it plural just like `IHAVE` and `IWANT`? I know that it's unnecessary because you can just put many `ControlIDontWant`s into the message, but isn't it better to make it consistent with `IHAVE` and `IWANT`?
```suggestion
    repeated bytes messageIDs = 1;
```

vyzo: Agreed, thats better.

arnetheduck: using `required` actively breaks forward compatibility and makes the protocol harder to upgrade - I'd strongly discourage anyone from using it - ie when you don't put required, you ensure that clients today are written in such a way that they gracefully can degrade their functionality iff the field ends up being not present. In an internet setting where you don't control when people upgrade their software, using `required` is quite problematic (ie it's reasonable for "private" software stacks where upgrades are controlled / mandatory)

the intent can be conveyed with a comment.

Nashatyrev: Yep, sounds reasonable 👍  Thanks for comment!

pubsub/gossipsub/gossipsub-v1.2.md
```diff
@@ -4,16 +4,103 @@
...
+- `/meshsub/1.2.0`
+
+## Parameters
+
+This section lists the configuration parameters that needs to agreed on across clients to avoid 
+ peer penalizations
+
+| Parameter                | Description                                                      | Reasonable Default |
+|--------------------------|------------------------------------------------------------------|--------------|
+| `max_idontwant_messages` | The maximum number of `IDONTWANT` messages per heartbeat per peer | ???  |
```
AgeManning, line 49: I guess this probably needs to be per-topic per heartbeat, rather than a total per heartbeat. 

It seems it could be tied in with the scoring for mesh message delivery rate. I.e the more messages we are expecting per topic, the more IDONTWANT messages we would expect to receive. 

One thought would be to add a behaviour penalty, similar to broken promises, if the number of IDONTWANT messages received from a peer exceeds the mesh message delivery rate. 

We intend to implement this fairly soon. Perhaps we can leave the scoring penalty here for a future PR if we dont want to specify it now.

Nashatyrev: Tracking `IDONTWANT` by topics would be perfect, however you don't know a message topic by its ID unless you already received this message.
`IDONTWANT` message is almost semantically equivalent to `IHAVE` message, so probably should have a similar anti-spam protection mechanism? 
Probably it could make sense to setup reasonable default: `maxIHaveLength * maxIHaveMessages` ?

pubsub/gossipsub/gossipsub-v1.2.md
```diff
@@ -4,16 +4,103 @@
...
+with the following  control message modifications:
+
+```protobuf
+message RPC {
+ // ... see definition in the gossipsub specification
+}
+
+message ControlMessage {
+    // messages from v1.0
+    repeated ControlIDontWant iDontWant = 5;
```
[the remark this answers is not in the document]
Nashatyrev: @jxs thanks!

pubsub/gossipsub/gossipsub-v1.2.md
ppopth: > If no one is opposed or has any ideas on usage scenarios I would keep it without optional topic field to be more explicit

The usage can be https://github.com/libp2p/specs/pull/548#discussion_r1534137418

pubsub/gossipsub/gossipsub-v1.2.md
Nashatyrev: > Is it a good reason to include topics to IDONTWANT and not care what IHAVE and IWANT are doing?

Including a topic would increase `IDONTWANT` traffic 2-3 times (messageID is 20 bytes, topic could be up to around 40 bytes). Not sure if it's worth just to enable more advanced spam protection...

Nashatyrev: Please also consider that `IHAVE` is sent on heartbeat with a batch of messages which may be grouped by topics and the topic overhead here could not be that significant. 
`IDONTWANT` on the other hand is intended to be flushed immediately and would most likely contain just a single messageID so the topic overhead would be more significant

AgeManning: We have built a version of this and are going to start testing. It doesn't discriminate based on topic and the implementation seems fine to me. 

Perhaps this first version is topic-independent and if we need to go back and handle topics in a subsequent PR we can do it (with the scoring)

AgeManning:
Hey all. 

We are very interested in testing this. What are our thoughts on merging this. We can do a future PR to handle scoring, or leave it up to implementors to handle their own dos prevention strategies. 

We want to start testing on live networks using the 1.2 protocol id. My concern is that if we do this, without this PR being merged and someone decides to change the protobuf, we will have then polluted the protocol-id space with an incompatible version. 

We will go with a `-beta` in the meantime, however it would be good to merge down to get some consensus.

pubsub/gossipsub/gossipsub-v1.2.md
Nashatyrev: Just realized that unlike `IHAVE` the `IDONTWANT` message IDs are never checked and thus could legally contain any random data. The `IDONTWANT` message IDs are supposed to be collected in a set which might be vulnerable to the hash map attack (we already fixed one in the jvm-libp2p impl). 
So it probably makes sense to: 
- Propose a more conservative limit. 
- Add a note to the spec to warn implementers about that possible issue and may be recommend to avoid using HashMap/HashSet implementations with weak hash function (like in Java)

ppopth: > The `IDONTWANT` message IDs are supposed to be collected in a set which might be vulnerable to the hash map attack

I think the implementers tend to use a hash set to remember the IDs that are seen in IDONTWANTs, but it's unreasonable to use the same hash set for `max_idontwant_messages`.

What would be reasonable is to just use a counter and increment it when seeing an IDONTWANT message without using a hash set at all.

arnetheduck:
fwiw, nimbus has implemented this as an extension to 1.1: https://github.com/vacp2p/nim-libp2p/blob/2b5319622c997ce1c80bc62c863e30f3349ee0d7/libp2p/protocols/pubsub/gossipsub/behavior.nim#L266 - it would indeed be nice to get this merged and properly give it a version number

pubsub/gossipsub/gossipsub-v1.2.md
Nashatyrev: > What would be reasonable is to just use a counter and increment it when seeing an IDONTWANT message without using a hash set at all.

@ppopth I'm not sure I'm following you. What are we supposed to count? It's expected that message IDs from a signle peer `IDONTWANT` messages would be unique.

arnetheduck: not sure it's been mentioned, but because the message id:s are not validated, it's possible to send (very) large ones - we're addressing this by hashing the message id which, together with the max-per-heartbeat gives a bound on mem usage overall.

ppopth: > not sure it's been mentioned, but because the message id:s are not validated, it's possible to send (very) large ones - we're addressing this by hashing (and salting) the message id which, together with the max-per-heartbeat gives a bound on mem usage overall.

This is a good point. I will do the same thing in `go-libp2p`.

arnetheduck: > not cleared at all

we use `mcache_len` for clearing it

ppopth: > we use `mcache_len` for clearing it

Yeah, what I meant is clearing it is the implementation detail and out of scope of the spec.

arnetheduck: I suspect it might be valuable to specify it - ie effectively, after mcache expires, the message should no longer exist and implementations should be able to rely on them being "resent" if they resurface after that time - this more faithfully keeps the protocol consistent in this aspect

Nashatyrev: >  I don't see why using a weak hash function for unwanted is a problem.

@ppopth I was meaning this [HashDoS attack](https://en.wikipedia.org/wiki/Collision_attack#Hash_flooding). 
The messageIDs in `IDONTWANT` are not validated: neither upon receive nor later (as for `IHAVE`). Thus an adversary may generate and send significant amount of messageIDs which yields the same hashCode in the context of `unwanted` hash set. 
Actually this attack vector could be addressed on the implementation side pretty easily. I just wanted to add a 'warning note' for implementers into the spec

Nashatyrev: > not sure it's been mentioned, but because the message id:s are not validated

@arnetheduck I believe as far as the `messageId` function is application specific, the validation of a `messageId` is left for an application responsibility. I believe Ethereum clients should all have such validation

AgeManning (approved):

vyzo (approved):
Soooo, lets merge it?
Any objections?

arnetheduck (approved):
lgtm, unless someone wants to edit in the various security recommendations

vyzo:
The go implementation in https://github.com/libp2p/go-libp2p-pubsub/pull/553 has merged.

PR #549, opened 2023-05-31
chore(roadmap): move webrtc to done

mxinden:
With https://github.com/libp2p/specs/pull/412 and
https://github.com/libp2p/specs/pull/497 merged, this roadmap item can be moved to "done".

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -648,7 +648,7 @@ connect to the rest of the libp2p network.
 
 ### WebRTC
 
-**Status**: In progress
+**Status**: Done
 
 **What?** WebRTC is a transport protocol supported by all major browsers. Those
 browsers allow the establishment of connections to remote endpoints that don't
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -23,7 +23,6 @@ third-party data ownership.
   - [⏱ Full Observability](#-full-observability)
   - [🧪 Automated compatibility testing](#-automated-compatibility-testing)
   - [Stream Migration Protocol](#stream-migration-protocol)
-  - [WebRTC](#webrtc)
   - [🤖 libp2p as a Wasm library](#-libp2p-as-a-wasm-library)
   - [🤝 Low latency, efficient connection handshake](#-low-latency-efficient-connection-handshake)
   - [🛣️ Peer Routing Records](#️-peer-routing-records)
@@ -38,6 +37,7 @@ third-party data ownership.
   - [🏹 RPC and other common node communication patterns](#-rpc-and-other-common-node-communication-patterns)
 - [Done](#done)
   - [🕸 Hole punching on TCP and QUIC](#-hole-punching-on-tcp-and-quic)
+  - [WebRTC](#webrtc)
 
 ## Core Tenets
 Before we dive into what libp2p should support and enable, let's outline the core tenets that underpin the project. As maintainers, we commit to ensuring libp2p is:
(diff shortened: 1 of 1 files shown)


ROADMAP.md
```diff
@@ -666,3 +645,25 @@ connect to the rest of the libp2p network.
 - [webtransport-go](https://github.com/marten-seemann/webtransport-go/)
 - [go-libp2p WebTransport](https://github.com/libp2p/go-libp2p/tree/master/p2p/transport/webtransport)
 - [js-libp2p WebTransport](https://github.com/libp2p/js-libp2p-webtransport/)
+
+### WebRTC
+
+**Status**: In progress
```
p-shahi, line 651:
```suggestion
**Status**: Done
```

p-shahi (approved):

ROADMAP.md
mxinden: Uups, thanks!

thomaseizinger (approved):

BigLep:
This feels a bit premature to me to close.  I can see the argument for doing so given the specs are done, but taking from the roadmap item text: "With WebRTC browsers will thus be able to connect to these previously unreachable nodes. In addition, being able to hole punch allows browsers to connect to nodes behind firewalls and NATs e.g. other browsers."  

The items I think we need here are:
1. Browser to go-libp2p nodes since go nodes are most plentiful public installation that we're aware of
2. Neither go nor rust has enabled browser to private server right?

I agree lots of great work has happened here including browser to browser, but my concerns are:
1. don't want to dilute the definition of done in general
2. looking at [the roadmap](https://github.com/libp2p/specs/blob/master/ROADMAP.md), there is nothing else about completing the connectivity matrix and I don't want to give the impression that we have full connectivity.

Maybe stepping back: what are we trying to signal or accomplish by marking this as done at this stage?

mxinden:
> Maybe stepping back: what are we trying to signal or accomplish by marking this as done at this stage?

By marking this as done I want to signal that establishing WebRTC connections in all combinations (i.e. all permutations of private/public to private/public) is now fully specified. In other words, any libp2p implementation can now implement these protocols without the need for specification, in other words without the need for driving consensus, across libp2p implementations.

> Browser to go-libp2p nodes since go nodes are most plentiful public installation that we're aware of

I would consider this a go-libp2p goal and thus an item for the go-libp2p roadmap.

> Neither go nor rust has enabled browser to private server right?

Correct. I think that should be tracked in the corresponding implementation roadmaps. E.g. see rust-libp2p's [`/webrtc` from the browser](https://github.com/libp2p/rust-libp2p/blob/master/ROADMAP.md#webrtc-in-the-browser-via-wasm) roadmap item.

Similar to marking WebRTC as done, [we also marked WebTransport as "Done"](https://github.com/libp2p/specs/pull/460). Given that rust-libp2p does not yet implement WebTransport, we have a [WebTransport roadmap item](https://github.com/libp2p/rust-libp2p/blob/master/ROADMAP.md#webtransport) in the rust-libp2p roadmap.

> looking at [the roadmap](https://github.com/libp2p/specs/blob/master/ROADMAP.md), there is nothing else about completing the connectivity matrix

From a specifications perspective, that is the case, no?

I guess ultimately this boils down to whether the roadmap in `libp2p/specs` should be the general libp2p roadmap or the libp2p specifications roadmap. I don't have a strong opinion on this. I defaulted to the latter in this pull request.

In case we go for the former, how do we decide when to mark an item as done? When the majority of libp2p implementations implement the protocol? How do we measure majority? E.g. does `py-libp2p` as one example of an unmaintained implementation even be considered?

PR #550, opened 2023-05-31
Document `/http`

MarcoPolo:
This is meant to clarify what libp2p is expecting of the `/http` component and avoid folks creating conflicting interpretations of the component. This purposely does not include anything libp2p+http (but hints at it) with the goal of clarifying expectations of the `/http` component early.

Please take a look:
@lidel 
@marten-seemann 
@willscott 
@masih 

This was prompted by https://github.com/ipni/go-libipni/issues/42 + historical issues referenced in https://github.com/ipni/go-libipni/issues/42#issuecomment-1570503551

### DIFF
--- a/http/transport-component.md
+++ b/http/transport-component.md
@@ -55,9 +55,9 @@ The following are examples of multiaddrs for HTTP transport capable nodes:
 * `/ip6/2001:0db8:85a3:0000:0000:8a2e:0370:7334/tcp/443/tls/http`
 * `/ip4/1.2.3.4/udp/50781/quic-v1/http`
 
-Note: When we use `/quic-v1/http` (or any other QUIC version)
-implementations MUST use the correct ALPN (e.g. `h3`) and not `libp2p` when
-using the HTTP transport.
+Note: When we use `/quic-v1/http` or `/tcp/443/tls/http` (or any other
+transport) implementations MUST use the correct HTTP ALPN (e.g. `h3` or `h2`
+respectively) and not `libp2p` when using the HTTP transport.
 
 ## HTTP Paths (and other HTTP Semantics)
 
--- a/http/transport-component.md
+++ b/http/transport-component.md
@@ -76,7 +76,7 @@ The HTTP Path exists in the user protocol level. In the HTTP semantics. You can
 use these semantics on any transport including, but not limited to, the HTTP
 transport.
 
-For example, say you want to fetch a file using the [IPFS trustless
+For example, say you want to fetch a file using the [IPFS trustless HTTP
 gateway](https://docs.ipfs.tech/reference/http/gateway/). It may be tempting to
 use the following URL to reference a file:
 
--- a/http/transport-component.md
+++ b/http/transport-component.md
@@ -72,7 +72,7 @@ information about a user protocol. However the `/http` component is not a user
 protocol, and it doesn't accept any parameters. It only signals that a node is
 capable of an HTTP transport.
 
-The HTTP Path exists in the user protocol level. In the HTTP semantics. You can
+The HTTP Path exists in the user protocol level. HTTP Semantics are transport-agnostic, and defined by [RFC 9110](https://httpwg.org/specs/rfc9110.html). You can
 use these semantics on any transport including, but not limited to, the HTTP
 transport.
 
--- a/http/transport-component.md
+++ b/http/transport-component.md
@@ -77,7 +77,7 @@ use these semantics on any transport including, but not limited to, the HTTP
 transport.
 
 For example, say you want to fetch a file using the [IPFS trustless HTTP
-gateway](https://docs.ipfs.tech/reference/http/gateway/). It may be tempting to
+gateway](https://specs.ipfs.tech/http-gateways/trustless-gateway/). It may be tempting to
 use the following URL to reference a file:
 
 ```
--- a/http/transport-component.md
+++ b/http/transport-component.md
@@ -52,7 +52,7 @@ The following are examples of multiaddrs for HTTP transport capable nodes:
 * `/dns/example.com/tls/http`
 * `/ip4/1.2.3.4/tcp/443/tls/http`
 * `/ip6/2001:0db8:85a3:0000:0000:8a2e:0370:7334/tcp/443/tls/http`
-* `/ip4/1.2.3.4/udp/50781/tls/http`
+* `/ip4/1.2.3.4/udp/50781/quic-v1/http`
 
 
 ## HTTP Paths
--- a/http/transport-component.md
+++ b/http/transport-component.md
@@ -53,8 +53,6 @@ left](https://github.com/multiformats/multiaddr#interpreting-multiaddrs)).
 The following are examples of multiaddrs for HTTP transport capable nodes:
 
 * `/dns/example.com/tls/http`
-* `/dns/example.com/https`
-* `/ip4/1.2.3.4/tcp/443/https`
 * `/ip4/1.2.3.4/tcp/443/tls/http`
 * `/ip6/2001:0db8:85a3:0000:0000:8a2e:0370:7334/tcp/443/tls/http`
 * `/ip4/1.2.3.4/udp/50781/tls/http`
--- a/http/transport-component.md
+++ b/http/transport-component.md
@@ -78,7 +78,7 @@ transport.
 
 For example, say you want to fetch a file using the [IPFS trustless HTTP
 gateway](https://specs.ipfs.tech/http-gateways/trustless-gateway/). It may be tempting to
-use the following URL to reference a file:
+use the following multiaddr to reference a file:
 
 ```
 /ip4/127.0.0.1/tcp/8080/http/httppath/ipfs%2fbafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi
--- a/http/transport-component.md
+++ b/http/transport-component.md
@@ -16,7 +16,8 @@ Interest Group: [@marcopolo], [@mxinden], [@marten-seemann]
 - [Context](#context)
 - [What is an HTTP transport](#what-is-an-http-transport)
 - [Multiaddr representation](#multiaddr-representation)
-- [HTTP Paths](#http-paths)
+- [HTTP Paths (and other HTTP Semantics)](#http-paths-and-other-http-semantics)
+  - [Recommendation on including HTTP semantics in multiaddrs](#recommendation-on-including-http-semantics-in-multiaddrs)
 
 
 ## Context
@@ -55,65 +56,40 @@ The following are examples of multiaddrs for HTTP transport capable nodes:
 * `/ip4/1.2.3.4/udp/50781/quic-v1/http`
 
 
-## HTTP Paths
+## HTTP Paths (and other HTTP Semantics)
 
 It may be tempting to add an HTTP path to end of the multiaddr to specify some
 information about a user protocol. However the `/http` component is not a user
 protocol, and it doesn't accept any parameters. It only signals that a node is
 capable of an HTTP transport.
 
-The HTTP Path exists in the user protocol level. HTTP Semantics are transport-agnostic, and defined by [RFC 9110](https://httpwg.org/specs/rfc9110.html). You can
-use these semantics on any transport including, but not limited to, the HTTP
-transports like [HTTP/1.1](https://www.rfc-editor.org/info/rfc7235), [HTTP/2](https://www.rfc-editor.org/info/rfc9113), or [HTTP/3](https://www.rfc-editor.org/info/rfc9114).
+The HTTP Path exists in the semantics level. HTTP Semantics are
+transport-agnostic, and defined by [RFC
+9110](https://httpwg.org/specs/rfc9110.html). You can use these semantics on any
+transport including, but not limited to, the HTTP transports like
+[HTTP/1.1](https://www.rfc-editor.org/info/rfc7235),
+[HTTP/2](https://www.rfc-editor.org/info/rfc9113), or
+[HTTP/3](https://www.rfc-editor.org/info/rfc9114).
 
-For example, say you want to signal that a node supports the [IPFS trustless
-HTTP gateway] protocol. It may be tempting to use the following multiaddr to
-signal that:
+### Recommendation on including HTTP semantics in multiaddrs
 
-```
-/ip4/127.0.0.1/tcp/8080/http/httppath/ipfs
-```
+In general, it's better to keep the multiaddrs as a way of addressing an
+endpoint and keep the semantics independent of any specific transport. This way
+you can use the same semantics among many specific transports.
 
-But `/http` is a transport and it doesn't accept any parameters. It would be the
-same as if we used the following multiaddr:
+However, sometimes it's helpful to share a single multiaddr that contains some
+extra application-level data (as opposed to transport data). The recommendation
+is to use a new [multicodec in the private
+range](https://github.com/multiformats/multicodec#private-use-area) for your
+application. Then apply whatever application parameters to the right of your new
+multicodec and transport information to the left. E.g.
+`<transport>/myapp/<parameters>`
+or `/ip4/127.0.0.1/tcp/8080/http/myapp/custom-prefix/foo%2fbar`. Your
+application has the flexibility to handle the parameters in any way it wants
+(e.g. set HTTP headers, an HTTP path prefix, cookies, etc).
 
-```
-/ip4/127.0.0.1/udp/1234/quic-v1/httppath/ipfs
-```
-
-What does `httppath/` mean here? `/quic-v1` is a transport and doesn't accept
-parameters. Who handles the input from `/httppath`?
-
-What we're really trying to do here is to highlight that the node that can be
-found at those Multiaddrs supports the [IPFS trustless HTTP gateway] protocol.
-That can and should be done some other way such as
-[identify](https://github.com/libp2p/specs/tree/master/identify) or soon the
-[`.well-known/libp2p`](https://github.com/libp2p/specs/pull/529) HTTP endpoint
-or some other custom application logic.
-
-If having this information on the multiaddr is desired and you are willing to
-make the tradeoff of potentially multiplying the number of multiaddrs you have
-by the number of protocols you want to signal, you could use a multicodec from
-the [private use
-area](https://github.com/multiformats/multicodec#private-use-area) and append
-this to your multiaddr. The result would be something like:
-
-```
-myProtocol = 0x300000
-
-/ip4/127.0.0.1/tcp/8080/http/myProtocol
-/ip4/127.0.0.1/udp/1234/quic-v1/myProtocol
-```
-
-Note that the problem appears when we want to add another protocol here:
-```
-myProtocol = 0x300000
-anotherProtocol = 0x300001
-
-/ip4/127.0.0.1/tcp/8080/http/myProtocol
-/ip4/127.0.0.1/udp/1234/quic-v1/myProtocol
-/ip4/127.0.0.1/tcp/8080/http/anotherProtocol
-/ip4/127.0.0.1/udp/1234/quic-v1/anotherProtocol
-```
-
-[IPFS trustless HTTP gateway]: (https://specs.ipfs.tech/http-gateways/trustless-gateway/)
+This is a bit cumbersome when you are trying to use multiple transports since
+you may end up with many multiaddrs with different transports but the same
+suffix. A potential solution here is to keep them separate. A list of multiaddrs
+for the transports being used, and another multiaddr for the application-level
+data. This is one suggestion, and many other strategies would work as well.
--- a/http/transport-component.md
+++ b/http/transport-component.md
@@ -31,8 +31,7 @@ avoid confusion and conflicting interpretations.
 ## What is an HTTP transport
 
 An HTTP transport is simply a node that can speak some standardized version of
-HTTP that is at least HTTP/1.1. Currently that means HTTP/1.1, HTTP/2, and
-HTTP/3. Intuitively if you can `curl` it with HTTP, then it speaks HTTP.
+HTTP. Intuitively if you can `curl` it with HTTP, then it speaks HTTP.
 
 Most environments will have a way to create an HTTP Client and Server, and the
 specific HTTP version used will be opaque. The client will negotiate (or [learn
--- a/http/transport-component.md
+++ b/http/transport-component.md
@@ -34,14 +34,11 @@ An HTTP transport is simply a node that can speak some standardized version of
 HTTP. Intuitively if you can `curl` it with HTTP, then it speaks HTTP.
 
 Most environments will have a way to create an HTTP Client and Server, and the
-specific HTTP version used will be opaque. The client will negotiate (or [learn
-about](https://www.rfc-editor.org/rfc/rfc9114.html#section-3.1.1)) the specific HTTP version to use when communicating
-with the HTTP server.
-
-For this opaque case, we use the `/http` component at the end of the multidadr.
-The end user agent decides on HTTP version to use, based on the multiaddr
-prefix, application, server negotiation, and specific use case. This follows
-what existing `http://` URL implementations do.
+specific HTTP version used will be opaque. We use the `/http` component at the
+end of the multidadr to signal that this server supports an HTTP transport. The
+end user agent decides on HTTP version to use, based on the multiaddr prefix,
+application, server negotiation, and specific use case. This follows what
+existing `http://` URL implementations do.
 
 ## Multiaddr representation
 
--- a/http/transport-component.md
+++ b/http/transport-component.md
@@ -55,6 +55,9 @@ The following are examples of multiaddrs for HTTP transport capable nodes:
 * `/ip6/2001:0db8:85a3:0000:0000:8a2e:0370:7334/tcp/443/tls/http`
 * `/ip4/1.2.3.4/udp/50781/quic-v1/http`
 
+Note: When we use `/quic-v1/http` (or any other QUIC version)
+implementations MUST use the correct ALPN (e.g. `h3`) and not `libp2p` when
+using the HTTP transport.
 
 ## HTTP Paths (and other HTTP Semantics)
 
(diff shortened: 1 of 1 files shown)


thomaseizinger (commented):
Currently, we expect to immediately run `multistream-select` after a connection is established. Connections are described by `multiaddr`.

How does this interact with the `/http` transport component?

I am guessing we don't run `multistream-select` on connections using the `/http` component?

thomaseizinger:
> Currently, we expect to immediately run `multistream-select` after a connection is established.

Actually, that is not true for QUIC.

Anyway, it might be worth mentioning this briefly.

MarcoPolo:
We know this is an HTTP transport. Multistream-select is not used here. The specifics I believe belong in the libp2p+http doc since it's how libp2p uses the HTTP transport.

http/transport-component.md
```diff
@@ -0,0 +1,121 @@
...
+protocol, and it doesn't accept any parameters. It only signals that a node is
+capable of an HTTP transport.
+
+The HTTP Path exists in the user protocol level. In the HTTP semantics. You can
+use these semantics on any transport including, but not limited to, the HTTP
+transport.
+
+For example, say you want to fetch a file using the [IPFS trustless
+gateway](https://docs.ipfs.tech/reference/http/gateway/). It may be tempting to
+use the following URL to reference a file:
```
BigLep, line 81: I think the message still comes across, but maybe a stronger example is for a multiaddr to a node that speaks the IPFS trustless HTTP gateway API?  It seems more likely that someone advertises "I speak HTTP trustless gateway" rather than "you can fetch a specific file from me".

Given the current landscape, I could imagine someone thinking that `/ip4/127.0.0.1/tcp/8080/http/httppath/ipfs` should map to an HTTP trustless gateway instance.

BigLep (commented):
Thanks for creating and getting this written down.

http/transport-component.md
```diff
@@ -0,0 +1,121 @@
...
+Most environments will have a way to create an HTTP Client and Server, and the
+specific HTTP version used will be opaque. The client will negotiate (or [learn
+about](https://www.rfc-editor.org/rfc/rfc9114.html#section-3.1.1)) the specific HTTP version to use when communicating
+with the HTTP server.
+
+For this opaque case, we use the `/http` component at the end of the multidadr.
+If a node wants to advertise specific HTTP versions it may advertise a
+`/http-1.1`, `/http-2`, or `/h3` instead. Although these components are not
+currently [registered](https://github.com/multiformats/multiaddr), this document
+claims their use.
```
mxinden, line 46: I don't think we should have two sources of truth. Either it is `multiformats/multiaddr` or `libp2p/specs`, but not both. Mind creating a pull request against `multiformats/multiaddr`?

http/transport-component.md
```diff
@@ -0,0 +1,121 @@
...
+## Multiaddr representation
+
+The multiaddr of a node with an HTTP transport ends with `/http` and is prefixed
+by information that would let an HTTP client know how to reach the server
+(remember that multiaddrs are [interpreted right to
+left](https://github.com/multiformats/multiaddr#interpreting-multiaddrs)). 
+
+The following are examples of multiaddrs for HTTP transport capable nodes:
+
+* `/dns/example.com/tls/http`
```
mxinden, line 58: `/tls/http` is deprecated in favor of `/https` according to https://github.com/multiformats/multiaddr/blob/master/protocols.csv. Is it necessary to list it here?

mxinden (commented):
With my limited knowledge of the libp2p+http effort, this looks good to me apart from the smaller comments below.

http/transport-component.md
```diff
@@ -0,0 +1,121 @@
...
+information about a user protocol. However the `/http` component is not a user
+protocol, and it doesn't accept any parameters. It only signals that a node is
+capable of an HTTP transport.
+
+The HTTP Path exists in the user protocol level. In the HTTP semantics. You can
+use these semantics on any transport including, but not limited to, the HTTP
+transport.
+
+For example, say you want to fetch a file using the [IPFS trustless
+gateway](https://docs.ipfs.tech/reference/http/gateway/). It may be tempting to
```
lidel, line 80: When we have specs, prefer them over any user docs:
```suggestion
gateway](https://specs.ipfs.tech/http-gateways/trustless-gateway/). It may be tempting to
```

http/transport-component.md
lidel, line 81:
```suggestion
use the following multiaddr to reference a file:
```

http/transport-component.md
lidel, line 46: We should not mention things that do not exist, nor encourage people to invent more codes  the rest of the ecosystem is expected to support.

I feel this only muddies the takeaway for the document, could we remove the notion of "you could explicitly state the HTTP version" from this doc? It is not possible with URLs, so maybe replace this with a note:

```suggestion
The end user agent decides on HTTP version to use, based on the multiaddr prefix, application, server negotiation, and specific use case. This follows what existing `http://` URL implementations do.
```

http/transport-component.md
```diff
@@ -0,0 +1,121 @@
...
+
+
+## HTTP Paths
+
+It may be tempting to add an HTTP path to end of the multiaddr to specify some
+information about a user protocol. However the `/http` component is not a user
+protocol, and it doesn't accept any parameters. It only signals that a node is
+capable of an HTTP transport.
+
+The HTTP Path exists in the user protocol level. In the HTTP semantics. You can
```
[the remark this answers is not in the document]
lidel:
```suggestion
The HTTP Path exists in the user protocol level. In the HTTP semantics defined by [RFC 9110](https://httpwg.org/specs/rfc9110.html). You can
```

http/transport-component.md
```diff
@@ -0,0 +1,121 @@
...
+use these semantics on any transport including, but not limited to, the HTTP
+transport.
```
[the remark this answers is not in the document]
lidel:
```suggestion
use these semantics on any transport including, but not limited to, the HTTP
transports like [HTTP/1.1](https://www.rfc-editor.org/info/rfc7235), [HTTP/2](https://www.rfc-editor.org/info/rfc9113), or [HTTP/3](https://www.rfc-editor.org/info/rfc9114).
```

http/transport-component.md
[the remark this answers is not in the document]
lidel:
```suggestion
The HTTP Path exists in the user protocol level. HTTP Semantics are transport-agnostic, and defined by [RFC 9110](https://httpwg.org/specs/rfc9110.html). You can
```

http/transport-component.md
```diff
@@ -0,0 +1,121 @@
...
+
+The following are examples of multiaddrs for HTTP transport capable nodes:
+
+* `/dns/example.com/tls/http`
+* `/dns/example.com/https`
+* `/ip4/1.2.3.4/tcp/443/https`
+* `/ip4/1.2.3.4/tcp/443/tls/http`
+* `/ip6/2001:0db8:85a3:0000:0000:8a2e:0370:7334/tcp/443/tls/http`
+* `/ip4/1.2.3.4/udp/50781/tls/http` // We can infer this is an HTTP/3 address
+* `/ip4/1.2.3.4/udp/50781/quic-v1/http` // Not necessary to specify that the
```
marten-seemann, line 64: Let's not infer things. There might other protocols (and new QUIC versions) that also run on top of UDP.

http/transport-component.md
```diff
@@ -0,0 +1,121 @@
...
+by information that would let an HTTP client know how to reach the server
+(remember that multiaddrs are [interpreted right to
+left](https://github.com/multiformats/multiaddr#interpreting-multiaddrs)). 
+
+The following are examples of multiaddrs for HTTP transport capable nodes:
+
+* `/dns/example.com/tls/http`
+* `/dns/example.com/https`
+* `/ip4/1.2.3.4/tcp/443/https`
+* `/ip4/1.2.3.4/tcp/443/tls/http`
```
marten-seemann, line 61: Can we define an unambiguous way to encode this? For reference, check what we decided for regarding `/tls/ws` and `/wss`.

http/transport-component.md
```diff
@@ -0,0 +1,121 @@
...
+
+## What is an HTTP transport
+
+An HTTP transport is simply a node that can speak some standardized version of
+HTTP that is at least HTTP/1.1. Currently that means HTTP/1.1, HTTP/2, and
+HTTP/3. Intuitively if you can `curl` it with HTTP, then it speaks HTTP.
+
+Most environments will have a way to create an HTTP Client and Server, and the
+specific HTTP version used will be opaque. The client will negotiate (or [learn
+about](https://www.rfc-editor.org/rfc/rfc9114.html#section-3.1.1)) the specific HTTP version to use when communicating
```
marten-seemann, line 39: This is probably not the best document to reference any more. Nowadays you'd look at the SVCB DNS record, and soon the AltSvcB HTTP header.

http/transport-component.md
```diff
@@ -0,0 +1,121 @@
...
+with the HTTP server.
+
+For this opaque case, we use the `/http` component at the end of the multidadr.
+If a node wants to advertise specific HTTP versions it may advertise a
+`/http-1.1`, `/http-2`, or `/h3` instead. Although these components are not
```
marten-seemann, line 44: Why would you do that? HTTP has its own version negotiation mechanism (ALPN for HTTP/1.1 and HTTP/2, more complicated (see above) for HTTP/3. I don't think we need to do anything here.

http/transport-component.md
```diff
@@ -0,0 +1,121 @@
...
+
+The multiaddr of a node with an HTTP transport ends with `/http` and is prefixed
+by information that would let an HTTP client know how to reach the server
+(remember that multiaddrs are [interpreted right to
+left](https://github.com/multiformats/multiaddr#interpreting-multiaddrs)). 
+
+The following are examples of multiaddrs for HTTP transport capable nodes:
+
+* `/dns/example.com/tls/http`
+* `/dns/example.com/https`
```
marten-seemann, line 59: Missing IP protocol and port.

http/transport-component.md
```diff
@@ -0,0 +1,121 @@
...
+doesn't make any assertions about how libp2p should interact with that
+transport. That will be defined in a future document.
+
+This exists to clarify the role of the `/http` component in Multiaddrs early to
+avoid confusion and conflicting interpretations.
+
+## What is an HTTP transport
+
+An HTTP transport is simply a node that can speak some standardized version of
+HTTP that is at least HTTP/1.1. Currently that means HTTP/1.1, HTTP/2, and
```
marten-seemann, line 34: Is it necessary to specify this? HTTP/1.0 is not really in use any more, but it would be nice if we didn't have to say anything here.

http/transport-component.md
MarcoPolo: will do after we agree on this.

http/transport-component.md
MarcoPolo: I went back and forth. I think `/https` is a bit more ergonomic than `/tls/http` (I at least often double check if tls comes before or after). But either way we must still support this since it's in the wild.

http/transport-component.md
MarcoPolo: Yeah, I can remove the list. We should still specify at least 1.1.

http/transport-component.md
MarcoPolo: They are equivalent. I think what we do in go-libp2p is transform `/wss` to `/tls/ws` for consistency.

http/transport-component.md
MarcoPolo: You're right. We shouldn't include the quic-v1 line. Just the udp and port part and the http client can figure out what HTTP transport to use (which right now only means h3).

http/transport-component.md
marten-seemann: Is there any precedent for this? Multiaddrs are all about being explicit.

http/transport-component.md
marten-seemann: Why?

MarcoPolo: Because HTTP/1.0 is still in use (It's the default for nginx: https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_http_version). We should be explicit that we only support at least 1.1

marten-seemann: That seems like an argument for not specifying it. Let the HTTP layer do the HTTP things.

http/transport-component.md
MarcoPolo: `/wss` is derectated just like `/https`. But that doesn't mean it isn't a valid example of a multiaddr you would see in the wild. See also this comment: https://github.com/libp2p/specs/pull/550#discussion_r1213381244.

MarcoPolo: I could add a note that this is deprecated.

http/transport-component.md
MarcoPolo: Do you want to support HTTP/1.0? I don't. If we don't set a minimum requirement we are effectively saying we support HTTP/1.0

http/transport-component.md
MarcoPolo: Yes, off the top of my head: https://github.com/multiformats/js-multiaddr-to-uri#usage converts `('/dnsaddr/protocol.ai/https') -> https://protocol.ai`.

This is in line with how to interpret multiaddrs. The stuff left of the `/http` component goes to the HTTP transport. As long as the HTTP transport knows how to handle `/dns/example.com` that's all that's necessary.

Another example is the `webrtc` transport. We don't specify SCTP anywhere in there.

http/transport-component.md
MarcoPolo: You can avoid paying an extra round trip for version negotiation. If your server prefers HTTP/2 and always upgrades clients to that, you can tell clients ahead of time to skip the round trip cost for the upgrade to HTTP/2 and just start with http/2.

http/transport-component.md
MarcoPolo: Could you link a better reference? I found https://www.ietf.org/archive/id/draft-ietf-dnsop-svcb-https-12.html but it's still a draft.

http/transport-component.md
MarcoPolo: What do you mean? This is only for `/http` transports. To deploy a new UDP based transport you would append it to the end of a udp multiaddr (e.g. `/ip4/1.2.3.4/udp/54321/utp)

http/transport-component.md
marten-seemann: We’re not. We’re just saying we don’t get involved in HTTP internals. How you configure your HTTP client and server is on you.

http/transport-component.md
MarcoPolo: I think it's useful to be able to signal that you support a specific HTTP version so that nodes don't have to waste a round trip in version negotiation. Supporting these is easy, since if you can't or won't act on this information you can interpret these as an opaque HTTP endpoint and fall back to HTTP version negotiation (e.g. in the browser).

If we agree on these I can register these codes in the table.

http/transport-component.md
marten-seemann: That’s because there’s no WebRTC without SCTP, as there also is no QUIC v1 without TLS.

http/transport-component.md
MarcoPolo: * We're defining a standard here that folks will reference when they want their nodes to interoperate with the network.
* We don't want them to use HTTP/1.0.

What parts do you disagree with?

http/transport-component.md
marten-seemann: There’s no round trip cost. ALPN negotiation happens during the TLS handshake.

If you made the argument that SVCB contains a protocol hint, that would make me reconsider. Although we could now end up in the awkward situation where DNS tells us one thing and the (/dns) multiaddr another.

Anyway, at this point I think we don’t really have a problem we’re solving, this seems like premature optimization.

http/transport-component.md
marten-seemann: We’re also not telling people which cipher suites they need to support when using TLS / QUIC. Somebody could configure some obscure cipher suites, which would make cipher suite negotiation fail during the handshake.

BigLep:
Related issue: https://github.com/multiformats/multiaddr/issues/63

http/transport-component.md
MarcoPolo: > There’s no round trip cost. ALPN negotiation happens during the TLS handshake.

Yes, you're correct. I was thinking of the Upgrade header which would only happen in `h2c` which would be the rare case (e.g. doing HTTP not on tls).

I'll remove this section.

http/transport-component.md
MarcoPolo: I'm going to remove this section per discussion in another thread. You don't pay for a round trip from HTTP/1.1 to HTTP/2, and HTTP/3 will be obvious when using a udp port.

http/transport-component.md
MarcoPolo: I don't feel strongly on this. I will remove the HTTP/1.1 requirement.

http/transport-component.md
MarcoPolo: Not necessary. Just helpful. we can remove that link.

http/transport-component.md
MarcoPolo: I removed `/https`. I think your original comment was a typo since `/https` is deprecated in favor of `/tls/http`.

http/transport-component.md
MarcoPolo: We also don't specify if the webrtc connection is over UDP vs TCP (since I believe you can run it on top of TCP as well). This is handled automatically by the webrtc transport. There's actually a lot we don't specify in the `/webrtc` transport. The only thing that is specified is the bootstrap connection.

But anyways, my earlier point stands. Anything that is valid for an http transport is valid for the components left of `/http`.

http/transport-component.md
```diff
@@ -0,0 +1,111 @@
...
+
+The following are examples of multiaddrs for HTTP transport capable nodes:
+
+* `/dns/example.com/tls/http`
+* `/ip4/1.2.3.4/tcp/443/tls/http`
+* `/ip6/2001:0db8:85a3:0000:0000:8a2e:0370:7334/tcp/443/tls/http`
+* `/ip4/1.2.3.4/udp/50781/tls/http`
+
+
+## HTTP Paths
```
MarcoPolo, line 58: I almost want to remove this section since it's kind of a tangent. It's useful in making the case why `/httppath` shouldn't be appended to `/http` and what you can do instead, but it also muddies the spec.

http/transport-component.md
MarcoPolo: Thanks, I've reworded this. Resolving this comment, but please reopen if needed.

http/transport-component.md
lidel: Hm.. maybe replace the tangent with a list of HTTP Semantics that one would be tempted to expose after `/http`, like I've hinted in https://github.com/ipni/go-libipni/issues/42#issuecomment-1570503551?

> After all, paths are not special, it is only one of many attributes that may impact HTTP endpoint's behavior.  If we allow HTTP paths, the unanswered question is: how one adds basic auth or bearer token headers? or a cookie? or an Accept header for content type version negotiation? [Do we add `/foo` for every HTTP feature? Where is the line in the sand?]
> 
> There are prior discussions in
> - https://github.com/multiformats/multiaddr/issues/63
> - https://github.com/multiformats/multiaddr/issues/87

masih: Should this section make recommendations on how paths should be handled? The HTTP rest APIs commonly support this. I think at the very least there should be recommendations on how should one handle HTTP API path prefixes when expressed as multiaddr.

http/transport-component.md
```diff
@@ -0,0 +1,119 @@
...
+by information that would let an HTTP client know how to reach the server
+(remember that multiaddrs are [interpreted right to
+left](https://github.com/multiformats/multiaddr#interpreting-multiaddrs)). 
+
+The following are examples of multiaddrs for HTTP transport capable nodes:
+
+* `/dns/example.com/tls/http`
+* `/ip4/1.2.3.4/tcp/443/tls/http`
+* `/ip6/2001:0db8:85a3:0000:0000:8a2e:0370:7334/tcp/443/tls/http`
+* `/ip4/1.2.3.4/udp/50781/tls/http`
```
marten-seemann, line 55:
```suggestion
* `/ip4/1.2.3.4/udp/50781/quic/http`
```

http/transport-component.md
```diff
@@ -0,0 +1,119 @@
...
+## Multiaddr representation
+
+The multiaddr of a node with an HTTP transport ends with `/http` and is prefixed
+by information that would let an HTTP client know how to reach the server
+(remember that multiaddrs are [interpreted right to
+left](https://github.com/multiformats/multiaddr#interpreting-multiaddrs)). 
+
+The following are examples of multiaddrs for HTTP transport capable nodes:
+
+* `/dns/example.com/tls/http`
```
marten-seemann, line 52:
```suggestion
* `/dns/example.com/tcp/443/tls/http`
```

MarcoPolo: Let's keep the discussion on https://github.com/libp2p/specs/pull/550#discussion_r1213520210

http/transport-component.md
marten-seemann: Actually, I agree. If you're doing DNS, you have SVCB, so that's fine.

BigLep:
@MarcoPolo : I know you want to have more sync time during https://pl-strflt.notion.site/IPFS-Implementers-Sync-2023-06-15-b8556243125442f2b0bf48c5db90d056?pvs=4 .  I assume you'll consolidate down the open questions and areas of disagreeement in advance.

http/transport-component.md
MarcoPolo: I believe @marten-seemann's argument is that we should specify this HTTP stack is sitting on quic-v1 (rather than some other future hypothetical UDP based transport).

The general counter argument is that everything left of the `/http` component is handled by the HTTP transport. If the HTTP transport needs more information than the [authority](https://datatracker.ietf.org/doc/html/rfc3986#section-3) and whether we're using tcp vs udp, then the extra information should be supplied. Right now, I don't think there's ambiguity here. If someday there is ambiguity we can add the extra information before the `/http` component. e.g. `/ip4/1.2.3.4/udp/54321/new-transport/http`.

http/transport-component.md
masih: Many thanks for adding this @MarcoPolo 

Using private ranges to specify path is certainly one way to do it. However, I am conscious of the increase in barrier for adoption. This is something that hopefully we can iterate on with the help of the lip2p community as HTTP over libp2p matures.

http/transport-component.md
```diff
@@ -0,0 +1,95 @@
...
```
thomaseizinger, line 53: If this says `/tls`, does that mean we expect both ends to conform to the libp2p TLS spec with its mandatory extension for signing the public key? Or is this vanilla TLS like a browser would perform with a server?

If this would not end in `/http`, it would definitely mean the former, correct?

But that would make it impossible for existing HTTP clients to connect as they don't understand our mandatory extension.

The same question applies to `/quic-v1`.

I think technically, we could say that - because multiaddresses are interpreted from right to left - the `/http` component defines that `/tls` and `/quic-v1` mean vanilla TLS without our extension. This sounds very confusing to me though.

thomaseizinger (approved):
One question regarding the semantics of TLS, other than that, this is a really good document and I am looking forward to this!

http/transport-component.md
MarcoPolo: > If this says /tls, does that mean we expect both ends to conform to the libp2p TLS spec with its mandatory extension for signing the public key? Or is this vanilla TLS like a browser would perform with a server?

The latter, vanilla TLS. Authentication will be optional in the HTTP transport.

> If this would not end in /http, it would definitely mean the former, correct?

I think so because we only support authenticated streams. And we only have defined the libp2p TLS extension as the auth mechanism.

The libp2p TLS extension is about authentication. If we require auth then we need it (unless we define an alternate way to do auth). In the HTTP transport, auth will be optional so we don't require the libp2p TLS extension.

> The same question applies to /quic-v1.

Yup, and hopefully this same answer makes sense.

thomaseizinger: I just realized that the TLS spec only applies to the `/tls/1.0.0` stream protocol and not to the multiaddr transport component.

However, for QUIC, we explicitly define that the TLS authentication is required if the `/quic-v1` component is used: https://github.com/libp2p/specs/tree/master/quic#peer-authentication

I think this confusion would not arise if we would define `/http1`, `/http2` and `/http3`.

Somewhere else you say that we only want to define that a node speaks the HTTP semantics but since we are talking about transports in the multiaddr, wouldn't it make sense to also talk about the various HTTP versions (which are transport related?).

thomaseizinger: > > However, for QUIC, we explicitly define that the TLS authentication is required if the /quic-v1 component is used: [`master`/quic#peer-authentication](https://github.com/libp2p/specs/tree/master/quic?rgh-link-date=2023-06-15T17%3A46%3A48Z#peer-authentication)
> 
> This spec is specific for libp2p streams on QUIC. `/quic-v1` in the context of `/http` can have its own semantics (e.g. optional peer auth).

At least to me, that is completely non-obvious to me from reading the specs.

Multiaddress components currently always mean the same thing, right? This spec would be the first one where we _change_ the meaning of an existing component.

MarcoPolo: Added in e71d7c5

MarcoPolo:
All outstanding concerns have been addressed. We brought this up in the ipfs implementers meeting and no objections were raised there either. I think we can merge this as soon as we get an approval.

thomaseizinger (approved):

BigLep:
What are the next steps for merging this?  Are we waiting to also include https://github.com/libp2p/specs/issues/554 or is that going to be a separate PR?

MarcoPolo:
#554 Will happen in a separate PR. I'd like @marten-seemann to give his approval here before merging

http/transport-component.md
```diff
@@ -0,0 +1,98 @@
...
+* `/ip4/1.2.3.4/tcp/443/tls/http`
+* `/ip6/2001:0db8:85a3:0000:0000:8a2e:0370:7334/tcp/443/tls/http`
+* `/ip4/1.2.3.4/udp/50781/quic-v1/http`
+
+Note: When we use `/quic-v1/http` (or any other QUIC version)
+implementations MUST use the correct ALPN (e.g. `h3`) and not `libp2p` when
+using the HTTP transport.
```
marten-seemann, line 60: That applies to any transport, doesn't it?

marten-seemann (approved):

http/transport-component.md
MarcoPolo: Correct, I'll mention that as well. Thanks

PR #556, opened 2023-07-07
HTTP: Reformat the spec from the Point of View of an implementer + Peer ID Auth

MarcoPolo:
Roughly, this PR:
* Removes sections that aren't directly relevant to an implementer.
* Proposes a Peer ID Authentication scheme that's similar to how we do authentication in WebTransport and WebRTC.

Besides the peer id authentication, I don't think there is anything new here compared to the `http` branch.

⚠️ Note: this targets the `http` branch. I would like to merge these changes into the `http` branch soon and continue discussion on that branch.

### DIFF
--- a/http/README.md
+++ b/http/README.md
@@ -6,7 +6,7 @@
 
 Authors: [@marten-seemann, @MarcoPolo]
 
-Interest Group: [@MarcoPolo, @marten-seemann]
+Interest Group: [todo]
 
 [@marten-seemann]: https://github.com/marten-seemann
 [@MarcoPolo]: https://github.com/MarcoPolo
@@ -19,7 +19,7 @@ This document defines how libp2p nodes can offer and use an HTTP transport along
 - HTTP only edge workers can run application protocols and respond to peers on the network.
 - `curl` from the command line can make requests to other libp2p nodes.
 
-As well as allowing application protocols to make use of HTTP middle boxes such as HTTP caching and layer 7 proxying and load balancing. This is all in addition to the existing features that libp2p provides such as:
+As well as allowing application protocols to make use of HTTP intermediaries such as HTTP caching and layer 7 proxying and load balancing. This is all in addition to the existing features that libp2p provides such as:
 
 - Connectivity – Work on top of WebRTC, WebTransport, QUIC, TCP, or an HTTP transport.
 - Hole punching – Work with peers behind NATs.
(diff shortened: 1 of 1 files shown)


http/README.md
```diff
@@ -1,172 +1,106 @@
...
-
-The libp2p protocol identifier is `/http1.1`. After negotiating this protocol using multistream-select, nodes treat the stream as a HTTP/1.1 stream for a single HTTP request (i.e. nodes MUST NOT use request pipelining).
+The resource contains a mapping of application protocols to their respective URL. For example, this configuration file would tell a client
 
-## Outlook: Interaction with Intermediaries
+1. That the Kademlia protocol is available at `/kademlia` and
+2. The [IPFS Path Gateway API](https://specs.ipfs.tech/http-gateways/path-gateway/) is mounted at `/`.
 
-One of the advantages of running HTTP is that there’s widely deployed caching infrastructure (CDNs). Content-addressed data is infinitely cacheable. Assuming a properly design data transfer protocol, retrieval for CIDs could be cached by the CDN and made available via a POP (geographically) close to the user, dramatically reducing retrieval latencies.
+It is valid expose a service at `/`. It is RECOMMENDED that the server resolve more specific URLs before less specific ones. e.g. a path of `/kademlia/foo` should be routed to the Kademlia protocol rather than the IPFS HTTP API.
```
marten-seemann, line 68:
```suggestion
It is valid to expose a service at `/`. It is RECOMMENDED that the server resolve more specific URLs before less specific ones. e.g. a path of `/kademlia/foo` should be routed to the Kademlia protocol rather than the IPFS HTTP API.
```

marten-seemann: > It is RECOMMENDED that the server resolve more specific URLs before less specific ones

Do we need to specify this? Is there an RFC that deals with path resolution that we could link instead?

http/README.md
```diff
@@ -1,172 +1,106 @@
...
 
-Services SHOULD specify the caching properties (if any), and set the appropriate cache headers (according to RFC 9111).
+## Peer ID Authentication
 
-CDNs can also be used to increase censorship resistance, since the CDN effectively hides the IP address of the origin server. With the upcoming introduction of ECHO (Encrypted ClientHello) in TLS, all that an on-path observer will be able to see is that a client is establishing a connection to a certain CDN, but not to which domain name.
+When using the HTTP Transport, peer id authentication is optional. You only pay for it if you need it. This is benefits use cases that don’t need peer authentication (e.g. fetching content addressed data) or authenticate some other way (not tied to libp2p peer ids).
```
marten-seemann, line 72:
```suggestion
When using the HTTP Transport, peer id authentication is optional. You only pay for it if you need it. This benefits use cases that don’t need peer authentication (e.g. fetching content addressed data) or authenticate some other way (not tied to libp2p peer ids).
```

http/README.md
```diff
@@ -1,172 +1,106 @@
...
 
-The level of delegation between the origin node and the CDN can be adjusted. In the simplest configuration, the origin node is the only node that holds the libp2p private key, thus requests to the `server-auth` protocol would be forwarded from the CDN to the origin server. In a more advanced configuration, it would be possible to move the private key to a worker on the edge of the CDN, and perform the signing operation there (thereby reducing the request latency for `server-auth` requests).
+Peer ID authentication in the HTTP Transport follows a similar to pattern to how libp2p adds Peer ID authentication in WebTransport and WebRTC. We run the standard libp2p Noise handshake, but using `IX` for client and server authentication or `NX` for just server authentication.
 
-## FAQ
+### Authentication flow
```
marten-seemann, line 76: This seems fine for now, but should we point out that this is just _one_ method you can do authentication? It would be nice to use https://datatracker.ietf.org/doc/draft-ietf-httpbis-unprompted-auth/ once it becomes available.

MarcoPolo: Agreed that this is only one way you can do _peer id_ authentication. I'll mention that, thanks.

MarcoPolo: The preceding section covers this already:

> This benefits use cases that don’t need peer authentication (e.g. fetching content addressed data) or *authenticate some other way (not tied to libp2p peer ids).*

Any suggestions to make this clearer?

MarcoPolo:
Merging this to consolidate discussion to a single PR.

http/README.md
marten-seemann: It's _one way_ to do **libp2p peer authentication**. We might come up with better ways later (e.g. the draft I linked above).

MarcoPolo: Something to that affect is on the other pr

PR #557, opened 2023-07-13
fix(pubsub): remove "Implementation note" about setting `message_id_fn`

cce:
This updates an old note in the message ID section of the pubsub spec to reflect that go-libp2p-pubsub currently supports topic-specific message ID configuration, using [`WithTopicMessageIdFn`](https://pkg.go.dev/github.com/libp2p/go-libp2p-pubsub#WithTopicMessageIdFn).

### DIFF
--- a/pubsub/README.md
+++ b/pubsub/README.md
@@ -198,12 +198,6 @@ on the [signature policy](#signature-policy) configured for the topic.**
 Whichever the choice, it is crucial that **all peers** participating in a topic
 implement identical message ID calculation logic, or the topic will malfunction.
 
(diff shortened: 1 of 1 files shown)


mxinden (commented):
To me the two options are equivalent:

- a top-level `message_id_fn(message)` that returns different `id`'s dependent on `message.topic`
- one `message_id_fn(message)` per topic, where each individual function ignores the `message.topic` field.

Unless I am missing something and there indeed is a difference, I suggest simply removing the paragraph.

mxinden (approved):

mxinden (approved):

PR #558, opened 2023-07-13
[GossipSub 1.2] PING/PONG control messages

Nashatyrev:
This PR introduces a new GossipSub version 1.2

Co-authored with @Menduist 

## New messages

The new `PING/PONG` control messages are added which would help gossip router to gather additional statistics about connection characteristics (e.g. latency and bandwidth) and optimize message sending (send staggering mostly) 

Additionally that could potentially be used to keep TCP connection 'warm' (avoid TCP slow start restart) by sending quite frequent `PING` requests (with a period of 1-2 RRTs)

### DIFF
--- a/pubsub/gossipsub/gossipsub-v1.2.md
+++ b/pubsub/gossipsub/gossipsub-v1.2.md
@@ -1,0 +1,102 @@
+# Gossipsub v1.2
+
+# Overview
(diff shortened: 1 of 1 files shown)


vyzo:
Let's target this to #560

PR #561, opened 2023-07-14
Simple request/response

MarcoPolo:
@mxinden / @thomaseizinger  I believe this is what rust-libp2p does, could you confirm please?

The motivation for this PR is to introduce a way of defining request response style application protocols without requiring it to be built on top of HTTP semantics directly. This is meant to be minimal and very agnostic as to what the application protocol does. For example, I'm not defining a way to include metadata or headers here. This is to keep it minimal, but also to preserve backwards compatibility.

I'm open to naming suggestions for this instead of "simple request/response".

This is an optional spec in the sense that no implementation is required to implement an API for this, but when I talk about a simple request/response that can make use of all libp2p transports this is what I mean. It would be helpful to agree on that part.

### DIFF
--- a/simple-request-response/README.md
+++ b/simple-request-response/README.md
@@ -1,0 +1,75 @@
+# A simple request/response abstraction
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+| --------------- | ------------- | ------ | --------------- |
+| 1A              | Working Draft | Active | r0, 2023-07-14  |
+
+Authors: [@MarcoPolo]
+
+todo: others welcome to join
+Interest Group: [@mxinden @marten-seemann]
+
+[@marten-seemann]: https://github.com/marten-seemann
+[@MarcoPolo]: https://github.com/MarcoPolo
+[@mxinden]: https://github.com/mxinden
+
+## Introduction
+
+Many application protocols map well to request/response semantics where a peer
+makes a request to another peer, and receives a response back. For example,
+requesting a file from a peer and getting the file back. In contrast to
+request/response, stream semantics allow two peers to continously send data
+back and forth with no specific ordering. Both styles have their uses. This
+document focuses on _one_ way of implementing request/response semantics and
+how to map it to stream based transports and an HTTP transport.
+
+## Goals of this document
+
+A goal of this document is to define _a_ simple way to write request/response
+style application protocols that can make use of all libp2p transports
+seamlessly (all stream transports and the HTTP transport). This is only one way
+of writing request/response style protocols, and users may choose to not use it.
+
+Another goal is to be backwards compatible with existing application protocols
+that follow these semantics to make upgrading them easier so that they may take
+advantage of the new HTTP transport without sacrificing backwards compatibility.
+
+The final goal is to define something simple such that it's easy to follow,
+implement, and use.
+
+## The abstraction
+
+This abstraction takes in a blob from the requesting peer (aka client) that
+represents the request and delivers the blob to the responding peer (aka
+server). The server responds with a blob that respresents the response. This
+abstraction is agnostic to what the blobs are or how they are encoded. That's up
+to the application protocol to decide.
+
+## How to run on top of a libp2p stream
+
+Each request and response should happen in a single stream. There MUST NOT be
+pipelining. After sending a request, the client SHOULD close its write side
+(signalling EOF to the peer). After handling the response, the client SHOULD
+close the stream. After sending the response the server SHOULD close the stream.
+
+## How to run on top of an HTTP transport
+
+The client's request is placed in the body of an HTTP POST request. The server
+places it's response in the body of the HTTP response. Headers are unused by the
+application protocol (but may be used by the libp2p implementation to provide
+authentication). The HTTP path used is defined by the server's
+`.well-known/libp2p` HTTP resource (see the [HTTP](../http/README.md) spec for
+more details).
+
+
+## Prior Art
+
+This spec is inspired by existing work to use request response protocols on top
+of libp2p streams including, but not limited to:
+- go-libp2p-kad-dht
+- Identify Push
+- AutoNATV2
+
+This is also inspired by rust-libp2p's [Request/Response
+crate](https://docs.rs/libp2p-request-response/0.25.0/libp2p_request_response/).
+


simple-request-response/README.md
```diff
@@ -0,0 +1,75 @@
...
+authentication). The HTTP path used is defined by the server's
+`.well-known/libp2p` HTTP resource (see the [HTTP](../http/README.md) spec for
+more details).
+
+
+## Prior Art
+
+This spec is inspired by existing work to use request response protocols on top
+of libp2p streams including, but not limited to:
+- go-libp2p-kad-dht
```
MarcoPolo, line 69: Not sure if rust-libp2p kad does this as well

simple-request-response/README.md
```diff
@@ -0,0 +1,75 @@
...
+This abstraction takes in a blob from the requesting peer (aka client) that
+represents the request and delivers the blob to the responding peer (aka
+server). The server responds with a blob that respresents the response. This
+abstraction is agnostic to what the blobs are or how they are encoded. That's up
+to the application protocol to decide.
+
+## How to run on top of a libp2p stream
+
+Each request and response should happen in a single stream. There MUST NOT be
+pipelining. After sending a request, the client SHOULD close its write side
```
mxinden, line 51: _Just for the record, given that I had to look this up myself, no changes needed:_

rust-libp2p's `libp2p-request-response` follows `SHOULD` here. I.e. it does require a custom request-response protocol (implementation of `libp2p-request-response::Codec`) to depend on the client closing the stream. Instead the custom request-response protocol may use e.g. length prefixing to determine the end of a message.

mxinden, line 51: _Again for the record, no changes needed:_

rust-libp2p's `libp2p-request-response` does not allow pipelining. The stream life-cycle is managed by `libp2p-request-response` and not by the custom request-response protocol (i.e. implementation of `libp2p-request-response::Codec`).

simple-request-response/README.md
```diff
@@ -0,0 +1,75 @@
...
+(signalling EOF to the peer). After handling the response, the client SHOULD
+close the stream. After sending the response the server SHOULD close the stream.
```
mxinden, line 53: > After handling the response, the client SHOULD close the stream.

I think in all of our transport protocols we have a notion on closing the write-side. I am not sure the same is true for closing the stream in general. Thus I suggest dropping this sentence.

mxinden, line 53:
```suggestion
(signalling EOF to the peer). After handling the response, the client SHOULD
close the stream. After sending the response the server SHOULD close its write side (signalling EOF to the peer).
```

To be consistent with the client side write-side closing above.

simple-request-response/README.md
mxinden: I don't understand the question. In case this is of some help:

- rust-libp2p's `libp2p-kad` does not reuse outgoing streams but allows a remote to reuse (pipeline) incoming streams. If I remember correctly this is needed for backwards compatibility with go-libp2p.
- We considered replacing `libp2p-kad`'s stream handling logic with rust-libp2p's `libp2p-request-response`. This is what we already do in `libp2p-autonat`. https://github.com/libp2p/rust-libp2p/blob/4c889fd6d5eface44aca38a8fa1664b7248c62ca/protocols/autonat/src/behaviour.rs#L164-L169

mxinden (approved):
This looks good to me. Thank you Marco for writing it up.

I have one suggestion in regards to stream closing, i.e. only requiring write-side closing not general stream closing.

simple-request-response/README.md
```diff
@@ -0,0 +1,75 @@
+# A simple request/response abstraction
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+| --------------- | ------------- | ------ | --------------- |
+| 1A              | Working Draft | Active | r0, 2023-07-14  |
+
+Authors: [@MarcoPolo]
+
+todo: others welcome to join
+Interest Group: [@mxinden @marten-seemann]
```
mxinden, line 10:
```suggestion
Interest Group: [@mxinden] [@marten-seemann]
```

thomaseizinger:
I'll review tomorrow.

simple-request-response/README.md
```diff
@@ -0,0 +1,75 @@
+# A simple request/response abstraction
```
thomaseizinger, line 1: I'd suggest to drop the "simple" keyword.

Request-response implies that only one message will be sent in each direction. Consequently, I think it is save to require closing the write-side after each message (see other comment below). Thus, I believe this is the most-basic form of request-response and I don't see there being another primitive.

To conclude: I don't think there will be another req-res primitive that isn't compatible with this one, meaning we don't need a qualifier but can just say that this is how you do any form of request-response on top of libp2p streams.

simple-request-response/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Introduction
+
+Many application protocols map well to request/response semantics where a peer
+makes a request to another peer, and receives a response back. For example,
+requesting a file from a peer and getting the file back. In contrast to
+request/response, stream semantics allow two peers to continously send data
+back and forth with no specific ordering. Both styles have their uses. This
+document focuses on _one_ way of implementing request/response semantics and
+how to map it to stream based transports and an HTTP transport.
```
thomaseizinger, line 24: Nit: The sentence from "This document" is already part of the next section IMO because it describes the goal.

simple-request-response/README.md
```diff
@@ -0,0 +1,75 @@
...
+A goal of this document is to define _a_ simple way to write request/response
+style application protocols that can make use of all libp2p transports
+seamlessly (all stream transports and the HTTP transport). This is only one way
+of writing request/response style protocols, and users may choose to not use it.
+
+Another goal is to be backwards compatible with existing application protocols
+that follow these semantics to make upgrading them easier so that they may take
+advantage of the new HTTP transport without sacrificing backwards compatibility.
+
+The final goal is to define something simple such that it's easy to follow,
```
thomaseizinger, line 37: I am struggling with defining something as "simple". IMO, it is non-descriptive. The S in SOAP also stands for "simple" and it certainly is not :)

simple-request-response/README.md
```diff
@@ -0,0 +1,75 @@
...
+back and forth with no specific ordering. Both styles have their uses. This
+document focuses on _one_ way of implementing request/response semantics and
+how to map it to stream based transports and an HTTP transport.
+
+## Goals of this document
+
+A goal of this document is to define _a_ simple way to write request/response
+style application protocols that can make use of all libp2p transports
+seamlessly (all stream transports and the HTTP transport). This is only one way
+of writing request/response style protocols, and users may choose to not use it.
```
thomaseizinger, line 31: Why combine these two in one document? There doesn't seem to be anything that is shared between the definition of doing request-response over a libp2p stream and doing request-response over HTTP.

simple-request-response/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## The abstraction
+
+This abstraction takes in a blob from the requesting peer (aka client) that
+represents the request and delivers the blob to the responding peer (aka
+server). The server responds with a blob that respresents the response. This
+abstraction is agnostic to what the blobs are or how they are encoded. That's up
+to the application protocol to decide.
+
+## How to run on top of a libp2p stream
```
thomaseizinger, line 48: My suggestion would to be to make this the main section of this document and define that this is how we do request-response on libp2p streams. I'd change the `SHOULD`s to `MUST` to make sure implementations can rely on that.

As mentioned above, I don't see how closing the write-side would be incompatible with any other way of doing request-response which is why I think it is safe to create this as a baseline abstraction.

thomaseizinger:
> This is an optional spec in the sense that no implementation is required to implement an API for this, but when I talk about a simple request/response that can make use of all libp2p transports this is what I mean. It would be helpful to agree on that part.

This is the first time I am hearing about such a thing as an explicitly optional spec. I thought that is the default?

simple-request-response/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+Each request and response should happen in a single stream. There MUST NOT be
+pipelining. After sending a request, the client SHOULD close its write side
+(signalling EOF to the peer). After handling the response, the client SHOULD
+close the stream. After sending the response the server SHOULD close the stream.
+
+## How to run on top of an HTTP transport
```
thomaseizinger, line 55: To me, this is a separate document. Other than sharing a title, nothing that is / has been defined in this document is related to how we do this on HTTP.

simple-request-response/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+This spec is inspired by existing work to use request response protocols on top
+of libp2p streams including, but not limited to:
+- go-libp2p-kad-dht
+- Identify Push
+- AutoNATV2
+
+This is also inspired by rust-libp2p's [Request/Response
+crate](https://docs.rs/libp2p-request-response/0.25.0/libp2p_request_response/).
+
```
thomaseizinger, line 75: I'd like an additional section here `## Implementations considerations` that describes that implementations need to be aware of DoS attacks.

Relying on the other side closing the stream is convenient but in the absence of a TLV encoding, we need to make sure to only read a maximum amount of data from the stream while waiting for EOF.

simple-request-response/README.md
thomaseizinger: We could add here that implementations are free to use _additional_ ways of encoding the boundaries of req and res, for example by means of a TLV encoding.

thomaseizinger (commented):
Thank you for writing this up. I have some ideas & questions, let me know what you think :)

simple-request-response/README.md
MarcoPolo: I see "request-response" being a more general idea. And this is a specific implementation of it. We could imagine alternative implementations that have associated metadata (like HTTP headers), or specific encodings (like an RPC system), or allow pipelining.

I want to be able to talk about "request-response" protocols without also talking about a specific implementation.

I'm not attached to the "simple" here. But I'd prefer to not use "request-response" directly.

simple-request-response/README.md
MarcoPolo: I'm saying that this abstraction can run on all transports, which will soon include the HTTP transport. Let's continue the discussion in the other comment, and I'll resolve this one.

simple-request-response/README.md
MarcoPolo: We need to define how to use this abstraction on all libp2p transports. That will soon include the HTTP transport, and thus we define that here. We could put it in a separate document "How to run simple request-response on top of an HTTP transport", but that doesn't give us anything besides adding a layer of indirection.

If we agree that "How to run on top of a libp2p stream" makes sense in this document, then "How to run on top of an HTTP transport" must also make sense in this document for the same reasons.

simple-request-response/README.md
MarcoPolo: Thanks, that answers my question.

simple-request-response/README.md
MarcoPolo: I think "MUST" would break backwards compatibility since existing clients may not close the write side after sending a request.

MarcoPolo:
go-libp2p-kad-dht is compatible with this, despite it doing some pipelining. It tries to pipeline, but if that fails it will use a new stream.

I think we should change the implementation to not pipeline. And I think this is still backwards compatible as existing clients will try a new stream if the server closes the stream.

simple-request-response/README.md
MarcoPolo: The server should also close its read side to signal to the client that the server will no longer accept any more writes (and thus prevent pipelining). This is necessary for backwards compatibility with go-libp2p-kad-dht.

simple-request-response/README.md
MarcoPolo: What are the DoS attacks you have in mind?

> We need to make sure to only read a maximum amount of data from the stream

Isn't this an application level protocol consideration? If I'm requesting a file, this can be quite big.

simple-request-response/README.md
MarcoPolo: Would you like to propose an alternative?

simple-request-response/README.md
thomaseizinger: > What are the DoS attacks you have in mind?

If I rely on the stream being closed before considering the message as received, a malicious peer can send me data until I run out of memory to allocate.

This is how we solve this in `rust-libp2p`: https://github.com/libp2p/rust-libp2p/blob/d3b4714730cafb850be46cbbb7038658d0d560bb/protocols/request-response/src/cbor.rs#L90-L94

> If I'm requesting a file, this can be quite big.

Yes, but there still needs to be an upper limit.

simple-request-response/README.md
thomaseizinger: > If we agree that "How to run on top of a libp2p stream" makes sense in this document, then "How to run on top of an HTTP transport" must also make sense in this document for the same reasons.

What exactly are we defining outside of these two sections?

Unless I am missing something, you are just describing what request-response is: A message in one direction with a message back, that is a pretty universally understood definition, no? Other things like disallowing of pipelining or closing of the write-side is already stream-specific.

Perhaps I am nitpicking here but I just don't see what we are actually defining that one can refer to unless you are talking about request-response on a particular transport (stream vs HTTP). Those two sections again have nothing in common so I don't quite see why they would be the same document.

Differently put, if I ask "What is request-response in libp2p if we are not talking about specific transports?", what would be your answer?

simple-request-response/README.md
thomaseizinger: I think I don't yet fully understand what you are trying to define with this spec (see other comments below). But if we move towards the "always close the stream after writing a message" definition then we could consider a name like: "EOF-indicated request-response abstraction".

simple-request-response/README.md
MarcoPolo: > What exactly are we defining outside of these two sections?

What's defined in "The abstraction" section. 

>  you are just describing what request-response is

I'm not. I'm describing a very limited request-response abstraction. Notably this doesn't include metadata about the request (headers) because it aims to be backwards compatible. For example, consider some alternative request-response abstractions:

- HTTP request/response semantics
- JSON RPC
- gRPC
- [Scuttlebutt](https://ssbc.github.io/scuttlebutt-protocol-guide/#rpc-protocol)
- Simple request-response, but we add metadata via a protobuf at the beginning of the request.

>  I just don't see what we are actually defining that one can refer to 

Consider this a "thin waist layer". On one side we have user protocols that are building on top of this request response abstraction. On the other side we have the transports. Each side agrees on this abstraction and how to use it, and thus can develop independently. Application protocols building on this abstraction don't have to think about specific transports, and transports don't have to think about specific application protocols.  

> Differently put, if I ask "What is request-response in libp2p if we are not talking about specific transports?", what would be your answer?

Hopefully the above paragraph explains this a bit better. This, like the stream abstraction, is an abstraction that lets application protocols not think about specific transports.

"What is request-response in libp2p?" It is this document, or HTTP semantics, or even some user specific request-response abstraction. It's any request-response abstraction that acts as a "thin waist" between application protocols and transports.

simple-request-response/README.md
MarcoPolo: I added a note to the doc. This should be defined by the application, so I made a section called "Considerations for applications".

simple-request-response/README.md
thomaseizinger: > > What exactly are we defining outside of these two sections?
> 
> What's defined in "The abstraction" section.

I see, so basically along the same lines as the [`Service`](https://crates.io/crates/tower-service) trait in the Rust ecosystem?

simple-request-response/README.md
```diff
@@ -0,0 +1,84 @@
...
+side (signalling EOF to the peer as well as signalling no future writes will be accepted.).
+
+## How to run on top of an HTTP transport
+
+The client's request is placed in the body of an HTTP POST request. The server
+places its response in the body of the HTTP response. Headers are unused by the
+application protocol (but may be used by the libp2p implementation to provide
+authentication). The HTTP path used is defined by the server's
+`.well-known/libp2p` HTTP resource (see the [HTTP](../http/README.md) spec for
+more details).
```
thomaseizinger, line 63: Nit: That isn't entirely true, right?

The HTTP path is defined by: The mapping in the `.well-known` resource AND whatever the application-protocol defines as an additional path for a certain protocol interaction.

For example, I might mount the kademlia protocol under `/kademlia` and the kademlia protocol might define that "GetRecord" is available as `GET /record/<base64-encoded-key>` which results in an HTTP request of:

```http
GET /kademlia/record/<base64-encoded-key>
```

Am I understanding this correctly?

simple-request-response/README.md
thomaseizinger: Based on [this](https://github.com/libp2p/specs/pull/561/files#r1265023182) discussion, I think a better name for this document would be "A request-response interface for protocols on any transport".

simple-request-response/README.md
```diff
@@ -0,0 +1,84 @@
...
+The final goal is to define something simple such that it's easy to follow,
+implement, and use.
+
+## The abstraction
+
+This abstraction takes in a blob from the requesting peer (aka client) that
+represents the request and delivers the blob to the responding peer (aka
+server). The server responds with a blob that respresents the response. This
+abstraction is agnostic to what the blobs are or how they are encoded. That's up
+to the application protocol to decide.
```
thomaseizinger, line 46: I would put this before the "goal" section such that we can extend the goal section to say that what we want is to define an interface that application protocols can rely on.

Plus, I'd perhaps reword this to:

```
## The interface

At its core, request-response means:
- taking _one_ blob from the requesting peer (aka the client)
- delivering this blob to the responding peer (aka the server)
- taking _one_ blob from the responding peer
- delivering this blob to the requesting peer

The defining characteristics are:

- Each party can only send one message
- The interaction is linear in time (the response can only be sent after the request has been received)

This is in contrast to for example streaming protocols where the responding peer may send any number of messages or event/notification-based protocols where a message may be received without a prior request.

What the blobs are and how they are encoded is subject to the application protocol and not defined in this interface.
```

I think the contrast to other communication patterns is important here:

- one message per peer vs many messages
- interaction is synchronous in nature, i.e. a client can "block" its control-flow on receiving a response

These are essentially the things we are defining that allow the creation of an interface (in code) that we can use as an abstraction for other protocols to build upon.

simple-request-response/README.md
thomaseizinger: Based on https://github.com/libp2p/specs/pull/561/files#r1267873831, I'd say the primary goal of this document is the definition of an _interface_ that application protocols can rely on and that must be fulfilled by transports. We can then go on about how the various transports fulfill it (stream & HTTP).

simple-request-response/README.md
thomaseizinger: > > plus it is easy to adapt existing implementations to additionally close the stream which makes them follow this spec.
> 
> Yes, it's easy to change new clients. But you can't change existing clients, thus breaking backwards compatibility.

Technically yes, but does it affect us? Is it important that our currently deployed protocols follow this spec? As far as I understand, we are mostly writing this to be able to deploy existing protocols over HTTP which requires code changes anyway. Can you outline a scenario where the incompatibility of existing clients with this spec hurts us?

> > each protocol has to define an additional mechanism like a TLV encoding to be able to reliably parse the messages
> 
> The application protocol themselves can say MUST. This doesn't prohibit that. If I make a new application protocol `/my-app/1` I can state in the spec that clients MUST close the stream when using the simple request-response abstraction.

Right but does that give us anything? If our goal is to write a generic piece of code, we cannot rely on what the individual protocols specify, right? In code terms, if a transports exposes an interface like:

```rust
async fn send_request(req: Res) -> Res;
```

How do you implement this if you cannot rely on receiving an EOF from the other party? If there is the possibility that a protocol does not say MUST, you effectively can't provide the above interface, otherwise it is not compatible with all possible protocols following this spec, right?

An implementation could be stricter and elevate the SHOULD to a MUST but that would limit which other protocols & implementations it is compatible with.

simple-request-response/README.md
thomaseizinger: Thanks, this is resolved for me (I can't mark it as such).

simple-request-response/README.md
MarcoPolo: For this interface The path under the root application protocol path is unused.
Everything is a POST request to `/<application-protocol-path-defined-by-servers-well-known>`.

simple-request-response/README.md
MarcoPolo: renamed

thomaseizinger (approved):
Great effort Marco!

This is now much clearer for me. You might want to rename the file still before merging.

request-response-interface/README.md
```diff
@@ -0,0 +1,107 @@
...
+Another goal is to be backwards compatible with existing application protocols
+that follow these semantics to make upgrading them easier so that they may take
+advantage of the new HTTP transport without sacrificing backwards compatibility.
+
+The final goal is to define something simple such that it's easy to follow,
+implement, and use.
+
+## How to map to a libp2p stream
+
+Each request and response should happen in a single stream. There SHOULD NOT be
```
marten-seemann, line 73:
```suggestion
Each request and response should happen in a single stream. There MUST NOT be
```

request-response-interface/README.md
```diff
@@ -0,0 +1,107 @@
...
+pipelining. After sending a request, the client SHOULD close its write side
+(signalling EOF to the peer). After handling the response, the client SHOULD
+close the stream. After sending the response the server SHOULD close the stream
```
marten-seemann, line 76: This sentence doesn't make sense, the client already closed the stream.

marten-seemann (changes requested):

request-response-interface/README.md
MarcoPolo: For backwards compatibility, this can't be a MUST.

I'd like to have existing request/response over streams (like identify), to already be compliant with this spec.

request-response-interface/README.md
MarcoPolo: Both sides can close the stream, right?

request-response-interface/README.md
MarcoPolo: Previous discussion: https://github.com/libp2p/specs/pull/561#discussion_r1265018181

simple-request-response/README.md
mxinden: Cross-posting https://github.com/libp2p/specs/pull/561#discussion_r1264295430 here. I am not sure all our transports even support the notion of closing ones read side. Thus I don't think we should require it in a specification.

MarcoPolo: It's currently recommended, not required. Am I missing something?

mxinden: :+1: fine as a recommendation. Sorry for not checking that earlier.

Note that rust-libp2p's stream interface doesn't even support closing of the read side. Users can only close the write side through `AsyncWrite::poll_close` trait method implementation. Obviously the specification should not adjust to a single application but rather the other way around.

PR #508, opened 2023-01-22
add HTTP spec

marten-seemann:
This PR adds a specification for libp2p+HTTP. 
It builds on countless discussions with various people interested in this topic, and incorporates the thinking outlined in #477 and #481.

Peer ID authentication over HTTP is deferred to a separate proposal: https://github.com/libp2p/specs/pull/564. That work is put on hold for now until we find a strong use case for it.

The work was started by Marten & Marco, and continued by @MarcoPolo. Recent edits are from @MarcoPolo .

PR #564, opened 2023-08-02
HTTP: Peer ID Authentication

MarcoPolo:
I've removed the peer id authentication scheme from the core spec of #508 for a couple reasons:

1. Auth is optional, the MVP may not include it.
2. This needs a bit more discussion since this is the first time we are doing a peer id auth scheme that relies on web PKI.
3. I've also changed the auth scheme to not use Noise because we didn't get as much code reuse as I was hoping for and I think it adds unnecessary complexity to this. It's now a more traditional challenge-response scheme similar to [SSH auth](https://datatracker.ietf.org/doc/html/rfc4252#section-7) and [WebAuthn](https://w3c.github.io/webauthn/) and the [first version](https://github.com/libp2p/specs/blob/8a57943a151d891dfd47e06abddd770987ebd033/http/README.md?plain=1#L126) of this auth.

On 2, this is the first time we are doing a peer id auth scheme that relies on web PKI, and thus has slightly different security properties than what we've done in the past. Instead of tying a peer id to an underlying encrypted channel we are tying it to a domain name. If the client can't trust the domain name (e.g. has enterprise root CAs installed) then their connection can be mitmd. In practice I don't think this is a serious concern because:
* If a 3rd party has installed root CAs on a client, they probably also have root on the client. Thus they don't have to mitm, since they own the client.
* If web PKI breaks, it will be very obvious since many other things will break. Libp2p users will still be able to use a different authentication scheme.

---

go-libp2p implementation: https://github.com/libp2p/go-libp2p/pull/2854


http/peer-id-auth.md
```diff
@@ -0,0 +1,101 @@
...
+This spec defines one way of authenticating Peer IDs over HTTP using a
+challenge-response scheme.
+
+## Mutual Client and Server Peer ID Authentication
+
+1. The server initiates the authentication by responding to a request that must
+   be authenticated with the response header `WWW-Authenticate: Libp2p-Challenge
+   challenge="<base64-encoded-challenge>, Libp2p-Challenge-Server-Only"`. The
+   challenge MUST be randomly generated from server for sole purpose of
+   authenticating the client. The server SHOULD store the challenge temporarily
```
marten-seemann, line 22: This breaks the stateless nature of HTTP. Subsequent requests might end up at a different backend server.

http/peer-id-auth.md
```diff
@@ -0,0 +1,101 @@
...
+## Introduction
+
+This spec defines one way of authenticating Peer IDs over HTTP using a
+challenge-response scheme.
+
+## Mutual Client and Server Peer ID Authentication
+
+1. The server initiates the authentication by responding to a request that must
+   be authenticated with the response header `WWW-Authenticate: Libp2p-Challenge
+   challenge="<base64-encoded-challenge>, Libp2p-Challenge-Server-Only"`. The
```
marten-seemann, line 20: Can we use structured fields here, or is this one of the legacy fields where retrofitting sf was too difficult?

http/peer-id-auth.md
```diff
@@ -0,0 +1,101 @@
...
+1. The client sends a request and sets the `Authorization`
+   [header](https://www.rfc-editor.org/rfc/rfc9110.html#section-11.6.2) header
+   to the following:
+   ```
+   Libp2p-Challenge peer-id="<encoded-peer-id-bytes>",client-challenge="<base64-encoded-client-challenge>",sig="<base64-signature-bytes>"
+   ```
+   * The peer-id is encoded per the [peer-ids spec](../peer-ids/peer-ids.md).
+   * The signature is over the concatenated result of:
+   ```
+     <varint-length> + "origin=" + server-name +
```
marten-seemann, line 33: How do you concat a varint and a string?

http/peer-id-auth.md
```diff
@@ -0,0 +1,101 @@
...
+   * The peer-id is encoded per the [peer-ids spec](../peer-ids/peer-ids.md).
+   * The signature is over the concatenated result of:
+   ```
+     <varint-length> + "origin=" + server-name + 
+     <varint-length> + "client-challenge=" + base64-encoded-client-chosen-client-challenge + 
+     <varint-length> + "challenge=" + base64-encoded-challenge
+   ```
+   * The client chosen client-challenge MUST be randomly generated.
+   * The client chosen client-challenge SHOULD be at least 32 bytes.
+   * The client MUST use the same server-name as what is used for the TLS
```
marten-seemann, line 39: Can this now be used on subsequent requests?

http/peer-id-auth.md
MarcoPolo: This is what webauthn does. I copied it from there.

An alternative is to use some encrypted value from the server, but this opens you up to replay attacks. They can be mitigated by including a validity time window. I went with the less risky approach that is widely deployed in webauthn.

http/peer-id-auth.md
MarcoPolo: Yes, it's one of those legacy fields unfortunately. I did a version with sfv, but it was a bit too awkward.

http/peer-id-auth.md
MarcoPolo: Maybe? I'm not sure. I could be better to have a separate scheme for that. Like the previous libp2p token.

http/peer-id-auth.md
```diff
@@ -0,0 +1,101 @@
+# Peer ID Authentication over HTTP
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+| --------------- | ------------- | ------ | --------------- |
+| 1A              | Working Draft | Active | r0, 2023-01-23  |
+
+Authors: [@MarcoPolo]
```
thomaseizinger, line 7: Doesn't render as link.

http/peer-id-auth.md
```diff
@@ -0,0 +1,101 @@
...
+Interest Group: Same as [HTTP](README.md)
+
+## Introduction
+
+This spec defines one way of authenticating Peer IDs over HTTP using a
+challenge-response scheme.
+
+## Mutual Client and Server Peer ID Authentication
+
+1. The server initiates the authentication by responding to a request that must
```
thomaseizinger, line 18: I am assuming this is as part of a 401 response?

http/peer-id-auth.md
thomaseizinger: > This breaks the stateless nature of HTTP. Subsequent requests might end up at a different backend server.

I don't think this is true. HTTP being stateless means that each message is entirely self-contained and doesn't depend on _who_ sends the message.

For example, FTP is not stateless because it contains commands like `LIST` which do different things depending on which directory the client is in. That is a stateful protocol because the server has to keep state _per client_ to actually fulfill the request.

Depending on server-side state for an application protocol does not break HTTP's stateless aspect.

If you load balance an HTTP interface, it is fairly well understood that you need to share application state between those to "hide" the load balancing from the client. For example, if a client first creates a resource via `POST` and then tries to `GET` it, your application needs to ensure that this works despite the load balancing.

http/peer-id-auth.md
```diff
@@ -0,0 +1,101 @@
...
+   challenge MUST be randomly generated from server for sole purpose of
+   authenticating the client. The server SHOULD store the challenge temporarily
+   until the authentication is done. The challenge SHOULD be at least 32 bytes.
+1. The client sends a request and sets the `Authorization`
+   [header](https://www.rfc-editor.org/rfc/rfc9110.html#section-11.6.2) header
+   to the following:
+   ```
+   Libp2p-Challenge peer-id="<encoded-peer-id-bytes>",client-challenge="<base64-encoded-client-challenge>",sig="<base64-signature-bytes>"
+   ```
+   * The peer-id is encoded per the [peer-ids spec](../peer-ids/peer-ids.md).
```
thomaseizinger, line 30:
```suggestion
   * The peer-id is encoded per the string encoding described in the [peer-ids spec](../peer-ids/peer-ids.md).
```

http/peer-id-auth.md
thomaseizinger: Why not use a delimiter that is outside of the base64 range and not a valid URL character? We could use `&` for example.

http/peer-id-auth.md
```diff
@@ -0,0 +1,101 @@
...
+   * The signature is over the concatenated result of:
+   ```
+     <varint-length> + "origin=" + server-name + 
+     <varint-length> + "client-challenge=" + base64-encoded-client-chosen-client-challenge + 
+     <varint-length> + "challenge=" + base64-encoded-challenge
+   ```
+   * The client chosen client-challenge MUST be randomly generated.
```
thomaseizinger, line 37: Nit: I think this needs a hyphen?

```suggestion
   * The client-chosen client-challenge MUST be randomly generated.
```

http/peer-id-auth.md
thomaseizinger, line 20: Why not just name them `-Client` and `-Server`? Aren't the fully composable?

Like, wouldn't it be possible to only authenticate the client for example?

```suggestion
   be authenticated with the response header `WWW-Authenticate: Libp2p-Challenge-Client
   challenge="<base64-encoded-challenge>, Libp2p-Challenge-Server"`. The
```

http/peer-id-auth.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Authentication Endpoint
+
+Because the client needs to make a request to authenticate the server, and the
+client may not want to make the real request before authenticating the server,
+the server MAY provide an authentication endpoint. This authentication endpoint
+is like any other application protocol, and it shows up in `.well-known/libp2p`,
+but it only does the authentication flow. It doesn’t send any other data besides
+what is defined in the above authentication flows. The protocol id for the
+authentication endpoint is `/http-peer-id-auth/1.0.0`.
```
thomaseizinger, line 89: Should we specify that the client SHOULD NOT send any payload with the HTTP request?

thomaseizinger (commented):
Thanks Marco!

A few thoughts. I think we could split this into two flows: `Libp2p-Challenge-Client` and `Libp2p-Challenge-Server` and specify that they can be combined as necessary or used standalone.

http/peer-id-auth.md
MarcoPolo: It depends on how the application is structured. There are cases where you won't read your writes (GET after a POST will fail temporarily until some backend is consistent).

I think auth is an exception, and it's okay to store some state temporarily to avoid replay attacks. This is the same thing WebAuthn does: 
> As a cryptographic protocol, Web Authentication is dependent upon randomized challenges to avoid replay attacks. Therefore, the values of both [PublicKeyCredentialCreationOptions](https://w3c.github.io/webauthn/#dictdef-publickeycredentialcreationoptions).[challenge](https://w3c.github.io/webauthn/#dom-publickeycredentialcreationoptions-challenge) and [PublicKeyCredentialRequestOptions](https://w3c.github.io/webauthn/#dictdef-publickeycredentialrequestoptions).[challenge](https://w3c.github.io/webauthn/#dom-publickeycredentialrequestoptions-challenge) MUST be randomly generated by [Relying Parties](https://w3c.github.io/webauthn/#relying-party) in an environment they trust (e.g., on the server-side), and the returned [challenge](https://w3c.github.io/webauthn/#dom-collectedclientdata-challenge) value in the client’s response MUST match what was generated. This SHOULD be done in a fashion that does not rely upon a client’s behavior, e.g., the [Relying Party](https://w3c.github.io/webauthn/#relying-party) SHOULD store the challenge temporarily until the operation is complete. Tolerating a mismatch will compromise the security of the protocol.

https://w3c.github.io/webauthn/#sctn-cryptographic-challenges

http/peer-id-auth.md
MarcoPolo: > How do you concat a varint and a string?

I can add the following section:

> The string should be utf8 encoded and the bytes should be concatenated to end end of the varint bytes.

I'll include an example as well.

MarcoPolo: > Why not use a delimiter that is outside of the base64 range and not a valid URL character? We could use `&` for example.

Why does it matter that it's not a valid URL character? This isn't being sent over the wire. Also do you suggest a delimiter instead of a length prefix? I have a slight preference for length prefix as it's harder to mess up. You don't have to worry about properly escaping the delimiter if the delimiter shows up in the field.

http/peer-id-auth.md
MarcoPolo: Let's use the bearer token [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750) for this. It's well defined and is pretty much what we want. Essentially, it's uses the request header of "Authorization: Bearer <token>".

I'll update the spec to include information on how the server can return a bearer token.

http/peer-id-auth.md
MarcoPolo: These don't exactly compose in the way I think you are thinking about. It's not standard to include two credentials in an `Authorization` header (which I think is required for the composition you want).

`Libp2p-Challenge` authenticates both the client and server. It's different than a "authenticate client only mode". We could introduce a new mode to only authenticate the client (`Libp2p-Challenge-Client`), but what's the use case? When would a server not have a key to authenticate with, but the client does?

I would suggest not doing it now and adding it later, if a good use case arises.

http/peer-id-auth.md
MarcoPolo: I've updated the spec to include a blurb about how the server may return a bearer token.

http/peer-id-auth.md
thomaseizinger: > > It's not standard to include two credentials in an `Authorization` header (which I think is required for the composition you want).
> 
> Unless I am misunderstanding the article, MDN says something different: [developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate#syntax](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate#syntax)
> 
> I think the following would be entirely legitimate (I renamed a few things to make them more concise):
> 
> ```
> WWW-Authenticate: Libp2p-Client challenge=<base64-encoded-challenge>
> WWW-Authenticate: Libp2p-Server challenge=<base64-encoded-challenge>
> ```

I just realized that this provides the client with two _options_ of how to authenticate instead of requiring them both. In that case, we probably need to go with the single scheme and two parameters:

```
> WWW-Authenticate: Libp2p server-challenge=<base64-encoded-challenge>, client-challenge=<base64-encoded-challenge>
```

Nit: I am unsure about the naming of the auth-scheme. It would be great if we can somehow reference _this_ particular spec with it. Perhaps it should be called `Libp2p-PeerId`? I find `Libp2p-Challenge` verbose and not very descriptive because `WWW-Authenticate` itself already specifies that its value is some kind of challenge.

MarcoPolo: > > It's not standard to include two credentials in an `Authorization` header (which I think is required for the composition you want).
> 
> Unless I am misunderstanding the article, MDN says something different: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate#syntax
> 
> I think the following would be entirely legitimate (I renamed a few things to make them more concise):
> 
> ```
> WWW-Authenticate: Libp2p-Client challenge=<base64-encoded-challenge>
> WWW-Authenticate: Libp2p-Server challenge=<base64-encoded-challenge>
> ```

The server can provide multiple authentication challenge kind (e.g. basic and bearer), but the client can only respond with a single scheme in `Authorization`. The [RFC section for `WWW-Authenticate`](https://www.rfc-editor.org/rfc/rfc9110#section-11.6.1) says, as you say, any number of challenges are allowed (`WWW-Authenticate = #challenge`). But for the [`Authorization`](https://www.rfc-editor.org/rfc/rfc9110#section-11.6.2), it's a single credential (`Authorization = credentials`).

But that's tangential. I think we both agree we should have a single auth scheme that has multiple parameters. So I'll change the spec to be that.

http/peer-id-auth.md
MarcoPolo: Thinking about this some more, applications should be able to decide where they want to fall on the replay protection spectrum. If they want no possibility of replay in exchange for some auth state, they could randomly generate the token. If they are okay with a small window of replay (maybe that's fine for the application), they could sign a timestamp and a nonce and use that to statelessly authenticate the client.

With that in mind, I'll change the wording to allow both, and applications/implementations can decide what they want to do.

MarcoPolo:
I've update the spec with a couple changes:

* A single auth scheme with parameters to control who is authenticated. This allows for mutual, client only, and server only auth.
* Change the wording to allow for using a server-encrypted value as the `challenge-client` value.

http/peer-id-auth.md
thomaseizinger: > You don't have to worry about properly escaping the delimiter if the delimiter shows up in the field.

A domain-name can only include valid URL characters, right? So we would be save with `&` as a delimiter because it cannot appear in domain-names and it is not in the base64 alphabet.

I am okay with sticking to varint prefixing aswell.

http/peer-id-auth.md
```diff
@@ -0,0 +1,131 @@
...
+Authors: [@MarcoPolo]
+
+[@MarcoPolo]: https://github.com/MarcoPolo
+
+Interest Group: Same as [HTTP](README.md)
+
+## Introduction
+
+This spec defines one way of authenticating Peer IDs over HTTP using a
+challenge-response scheme. The authentication scheme is called `Libp2p-PeerID``
```
thomaseizinger, line 16:
```suggestion
challenge-response scheme. The authentication scheme is called `Libp2p-PeerID`.
```

thomaseizinger, line 16:
```suggestion
This spec defines an authentication scheme in accordance with [RFC7235](https://datatracker.ietf.org/doc/html/rfc7235).
The authentication scheme is called `Libp2p-PeerID`.
```

http/peer-id-auth.md
```diff
@@ -0,0 +1,131 @@
...
+   * The signature is over the concatenated result of:
+   ```
+     <varint-length> + "origin=" + server-name + 
+     [<varint-length> + "challenge-server=" + base64-encoded-client-chosen-challenge-server + ]
+     <varint-length> + "challenge-client=" + base64-encoded-challenge
+   ```
+   * Strings are UTF-8 encoded.
+   * If the challenge server was omitted in the `Authorization` header it MUST
+     be omitted in the signature.
+   * If provided, the client-chosen `challenge-server`` MUST be randomly generated.
```
thomaseizinger, line 47:
```suggestion
   * If provided, the client-chosen `challenge-server` MUST be randomly generated.
```

Too many backticks I think.

http/peer-id-auth.md
```diff
@@ -0,0 +1,131 @@
...
+    ```
+    12 // (18 bytes)
+    6f726967696e3d6578616d706c652e636f6d // (origin=example.com)
+
+    3d // (61 bytes)
+    636c69656e742d6368616c6c656e67653d7171717171717171717171717171717171717171717171717171717171717171717171717171717171716f3d // (client-challenge=qq...o=)
+
+    36 // (54 bytes)
+    6368616c6c656e67653d414141414141414141414141414141414141414141414141414141414141414141414141414141414141413d // (challenge=AA...A=)
+    ```
```
thomaseizinger, line 74: This is great! I'd appreciate if we can have a more elaborate set of test vectors at the end of the document.

thomaseizinger (approved):
Great work! I've made a few more comments but this looks sound from my end :)

http/peer-id-auth.md
MarcoPolo: I'll make a note to include this as well when I have an implementation :)

http/peer-id-auth.md
MarcoPolo: My editor being "helpful"

http/peer-id-auth.md
MarcoPolo: https://datatracker.ietf.org/doc/html/rfc9110 is the latest standard :)

http/peer-id-auth.md
```diff
@@ -0,0 +1,137 @@
...
+     <varint-length> + "challenge-client=" + base64-encoded-challenge
+   ```
+   * Strings are UTF-8 encoded.
+   * If the challenge server was omitted in the `Authorization` header it MUST
+     be omitted in the signature.
+   * If provided, the client-chosen `challenge-server` MUST be randomly generated.
+   * The client-chosen `challenge-server` SHOULD be at least 32 bytes.
+   * The client MUST use the same server-name as what is used for the TLS
+     session.
+   * If the client _only_ wants to authenticate the server and the server does
```
sukunrt, line 52: Can we make this a separate section? 
Mutual Authentication
Client Only Authentication
Server Only Authentication

http/peer-id-auth.md
```diff
@@ -0,0 +1,137 @@
...
+    36 // (54 bytes)
+    6368616c6c656e67653d414141414141414141414141414141414141414141414141414141414141414141414141414141414141413d // (challenge=AA...A=)
+    ```
+
+    All together:
+    ```
+    Message To Sign in hex:
+    126f726967696e3d6578616d706c652e636f6d3d636c69656e742d6368616c6c656e67653d7171717171717171717171717171717171717171717171717171717171717171717171717171717171716f3d366368616c6c656e67653d414141414141414141414141414141414141414141414141414141414141414141414141414141414141413d
+    ```
+2. The server MUST verify the signature using the server name used in the TLS
```
sukunrt, line 82: How does the server get the client's public key corresponding to the peer id? I assume that's what the client is supposed to sign with to prove its peer id.

http/peer-id-auth.md
```diff
@@ -0,0 +1,137 @@
...
+is like any other application protocol, and it shows up in `.well-known/libp2p`,
+but it only does the authentication flow. The client and server SHOULD NOT send
+any data besides what is defined in the above authentication flows. The protocol
+id for the authentication endpoint is `/http-peer-id-auth/1.0.0`.
+
+
+## Considerations for Implementations
+
+* Implementations MUST only authenticate over a secured connection (i.e. TLS).
+* Implementations SHOULD limit the maximum length of any variable length field.
```
sukunrt, line 125: Should we provide a recommended min and max length limit to support?

http/peer-id-auth.md
MarcoPolo: Good point. If it can't be embedded in the peer id, the client/server should send it as a field in the header.

http/peer-id-auth.md
MarcoPolo: That makes sense. Maybe an overview section and then separate sections for specifics.

http/peer-id-auth.md
```diff
@@ -0,0 +1,194 @@
...
+Signatures sign some set of parameters. The parameters are sorted
+alphabetically, prepended with a varint length prefix, and concatenated together
+to form the data to be signed. The signing algorithm is defined by the key type
+used. Refer to the [Peer ID
+spec] for
+specifics on the signing algorithm. The set of parameters is prefixed with the auth scheme "libp2p-PeerID"
+
+As an example, if we wanted to sign the parameters `hostname="example.com",
+challenge-client=base64String` we would first structure the parameters as a byte
+slice containing:
```
sukunrt, line 44: Do we need to specify that base64 string is `UTF-8` encoded before signing? 

Should we sign the raw bytes after base64 decoding? Even with raw bytes we'd need to specify utf-8 for things like `hostname`. The present scheme is fine too. I'm curious why you chose(base64 encoded) over the other(raw after base64 decoding).

http/peer-id-auth.md
```diff
@@ -0,0 +1,194 @@
...
+   parameter back to the server. The server MAY use the opaque parameter to
+   encode state.
+3. The client makes another POST request to the authentication URI and sets the header:
+   ```
+   Authorization: libp2p-PeerID peer-id=<string-representation-of-client-peer-id>, opaque=<opaque-from-server>, challenge-server=<base64-encoded-challenge-server>, sig=<base64-signature-bytes>
+   ```
+
+   The `sig` param represents a signature over the parameters:
+    - `hostname`
+    - `challenge-client` in its base64 encoded form.
```
sukunrt, line 81: For RSA keys we need a header for sending the public key, right?

http/peer-id-auth.md
```diff
@@ -0,0 +1,194 @@
...
+reason this is not a multibase is to aid clients or servers who can not or
+prefer not to import a multibase dependency.
+
+## Mutual Client and Server Peer ID
+
+The following protocol allows both the client and server to authenticate each
+other's peer id by having them each sign a challenge issued by the other. The
+protocol operates as follows:
+
+1. The client makes a POST request to the authentication URI.
```
sukunrt, line 66: Does this need to be the authentication URI? Can this work for any resource with the server responding with a 401 and a challenge and client doing a retry of the request with the auth headers?

sukunrt (approved):

http/peer-id-auth.md
```diff
@@ -0,0 +1,194 @@
...
+   authenticated the client's peer id. The server MUST respond with status code
+   200 (OK) and set the headers:
+   ```
+   Authentication-Info: libp2p-PeerID peer-id=<server-peer-id-string>, sig=<base64-signature-bytes>
+   Authorization: libp2p-Bearer <base64-encoded-opaque-blob>
+   ```
+   The `sig` param represents a signature over the parameters:
+    - `hostname`
+    - `challenge-server` in its base64 encoded form.
+    - `client` the string representation of the client's Peer ID.
```
sukunrt, line 94: peerid spec has two string representations for the peer id base58BTC and cid-v1.

http/peer-id-auth.md
MarcoPolo: I'll specify that all strings are utf8-encoded. Varints of course are inserted into the byte array as is.

> I'm curious why you chose(base64 encoded)

It saves a decoding step. A client/server don't have to decode the challenge, they just paste the challenge into the byte array for the signature.

http/peer-id-auth.md
MarcoPolo: Yes, thanks for the callout. This is implemented, but I need to spec it here.

http/peer-id-auth.md
MarcoPolo: It could, but that's a bit trickier. I think the simpler flow is:

1. client does mutual auth with the authentication URI.
2. client makes a request to the authenticated endpoint with the bearer token.

That said, I don't want to forbid it. I'll try to change the language to allow it while also recommending this simpler strategy.

http/peer-id-auth.md
MarcoPolo: Either is fine. The string representation is whatever is defined by the peer id spec.

http/peer-id-auth.md
```diff
@@ -0,0 +1,194 @@
...
+Then sign the resulting byte slice. See the test vectors below for a
+examples.
+
+
+## Base64 Encoding
+
+The base64 encoding follows Base 64 Encoding with URL and Filename Safe Alphabet
+from [RFC 4648](https://datatracker.ietf.org/doc/html/rfc4648#section-5). The
+reason this is not a multibase is to aid clients or servers who can not or
+prefer not to import a multibase dependency.
```
MarcoPolo, line 58: I think it's still better to keep this simple. But let me be the first to point out that users of this auth scheme implicitly rely on multibase when they handle peer id decoding.

http/peer-id-auth.md
sukunrt: The problem is which representation should the client use to verify the signature? 

How about saying the server SHOULD use the string representation provided by the client in the request? This allows servers to use a different representation but only if it makes things better for them.

http/peer-id-auth.md
sukunrt: I think this is fine. Multibase is useful if you want to share the bytes with other people. We have no such need here.

http/peer-id-auth.md
```diff
@@ -0,0 +1,194 @@
...
+to form the data to be signed. The signing algorithm is defined by the key type
+used. Refer to the [Peer ID
+spec] for
+specifics on the signing algorithm. The set of parameters is prefixed with the auth scheme "libp2p-PeerID"
+
+As an example, if we wanted to sign the parameters `hostname="example.com",
+challenge-client=base64String` we would first structure the parameters as a byte
+slice containing:
+```
+libp2p-PeerID<varintprefix>challenge-client=<base64String><varintprefix>hostname="example.com"
```
sukunrt, line 46:
```suggestion
libp2p-PeerID<varintprefix>challenge-client=<base64String><varintprefix>hostname=example.com
```

Why do we want to quote this? `challenge-client=<base64string>` doesn't have the base64 string quoted

http/peer-id-auth.md
```diff
@@ -0,0 +1,194 @@
...
+    - `hostname`
+    - `challenge-client` in its base64 encoded form.
+4. The server MUST verify the signature using the server name used in the TLS
+   session. The server MUST return 401 Unauthorized if the server fails to
+   validate the signature. If the signature is valid, the server has
+   authenticated the client's peer id. The server MUST respond with status code
+   200 (OK) and set the headers:
+   ```
+   Authentication-Info: libp2p-PeerID peer-id=<server-peer-id-string>, sig=<base64-signature-bytes>
+   Authorization: libp2p-Bearer <base64-encoded-opaque-blob>
```
sukunrt, line 89: Is Authorization generally used with HTTP Requests? 
From: RFC 9110:  https://www.rfc-editor.org/rfc/rfc9110.html#section-11.6.2
>The "Authorization" header field allows a user agent to authenticate itself with an origin server -- usually, but not necessarily, after receiving a [401 (Unauthorized)](https://www.rfc-editor.org/rfc/rfc9110.html#status.401) response.

http/peer-id-auth.md
sukunrt: We can also just sign the public key.

http/peer-id-auth.md
MarcoPolo: It's following the ABNF of what goes in these headers. Specifically this is an `auth-param` which can be a token or a quoted string. `auth-param = token BWS "=" BWS ( token / quoted-string )`. However, reviewing this again, I don't think the padding `=` of base64url makes it count as a token. So it needs to also be a quoted string.

http/peer-id-auth.md
MarcoPolo: These are passed in the header. They can't be arbitrary unicode, but they can be ["quoted strings"](https://datatracker.ietf.org/doc/html/rfc9110#section-5.6.4) which I think is almost unicode with some exceptions.

http/peer-id-auth.md
MarcoPolo: I like that. Let's do that instead.

http/peer-id-auth.md
MarcoPolo: (assuming you mean with HTTP Responses) You're right. I'm not sure where I got the notion it's used by both requests and response.

I'll include this in the Authorization field.

http/peer-id-auth.md
```diff
@@ -0,0 +1,194 @@
...
+   ```
+   The opaque parameter is opaque to client. The client MUST return the opaque
+   parameter back to the server. The server MAY use the opaque parameter to
+   encode state.
+3. The client makes another POST request to the authentication URI and sets the header:
+   ```
+   Authorization: libp2p-PeerID peer-id=<string-representation-of-client-peer-id>, opaque=<opaque-from-server>, challenge-server=<base64-encoded-challenge-server>, sig=<base64-signature-bytes>
+   ```
+
+   The `sig` param represents a signature over the parameters:
```
Stebalien, line 79: It would be nice to include the certificate (or public key?) hash, if possible. That way, _if_ the client has some way of validating it they can not rely on the CA system and instead validate the certificate.

MarcoPolo: This could be easy to do, but I'm not sure it would be used. The optimal solution here is to authenticate with some TLS exported key material ([RFC 5705](https://datatracker.ietf.org/doc/html/rfc5705)). That would bind the peer id to the TLS session and can happen with no additional round trips. If a peer can access the certificate they likely can access the key material exporter.

The only problem with this approach is that the key material exporters aren't yet exposed by browsers. But I expect that to change soon-ish as https://datatracker.ietf.org/doc/draft-ietf-httpbis-unprompted-auth/ gets published and implemented by browsers.

My thought here is to focus on the browser use case, and, if it would be used and useful, create a new spec for the exported key material scenario.

Stebalien: > The optimal solution here is to authenticate with some TLS exported key material ([RFC 5705](https://datatracker.ietf.org/doc/html/rfc5705)). That would bind the peer id to the TLS session and can happen with no additional round trips. If a peer can access the certificate they likely can access the key material exporter.

I agree, but that's likely harder to get at.

> My thought here is to focus on the browser use case, and, if it would be used and useful, create a new spec for the exported key material scenario.

What about making the format extensible? I.e., allow the server to sign multiple things that can be extended later?

Stebalien: > I agree, but that's likely harder to get at.

Ah, I see. That's what that proposal is about. Nice!

http/peer-id-auth.md
MarcoPolo: Update to sign the public key bytes instead.

http/peer-id-auth.md
MarcoPolo: I've updated to quote all of the params.

http/peer-id-auth.md
MarcoPolo: Added a section.

http/peer-id-auth.md
MarcoPolo: > What about making the format extensible? I.e., allow the server to sign multiple things that can be extended later?

There is room in this spec to extend this later. A future version could for example define a new parameter that is passed by the client. Example: a SessionID parameter which is derived from keying material from the tls session just for this purpose, the server would check that it matches the expected SessionID it sees or fail the request.

Stebalien: This has to be covered by a signature. I.e., either:

1. It's covered by the client signature and the server refuses to authenticate if it doesn't match
2. It's covered by the server's signature and the client rejects the authentication if it doesn't match.

I want to make sure there's some sane upgrade path where we can add support for this later.

http/peer-id-auth.md
```diff
@@ -52,55 +52,56 @@ examples.
...
-   Authorization: libp2p-PeerID peer-id=<string-representation-of-client-peer-id>, opaque=<opaque-from-server>, challenge-server=<base64-encoded-challenge-server>, sig=<base64-signature-bytes>
+   Authorization: libp2p-PeerID peer-id="<string-representation-of-client-peer-id>", opaque="<opaque-from-server>", challenge-server="<challenge-string>"[,encoded-public-key="<base64-encoded-public-key-bytes>" ], sig="<base64-signature-bytes>"
    ```
 
+   The `encoded-public-key` param is optional and represents the client's public key. This is only needed when the client's public key is not included in the PeerID (e.g. not using the "identity" multihash).
+
    The `sig` param represents a signature over the parameters:
     - `hostname`
-    - `challenge-client` in its base64 encoded form.
+    - `challenge-client`
```
Stebalien, line 84: If possible, I'd sign _everything_ the client sent us (basically, each step of the handshake should witness the previous step).

http/peer-id-auth.md
MarcoPolo: Yes, sorry. The signature SHOULD include the SessionID as part of the parameters it signs.

http/peer-id-auth.md
MarcoPolo: This is what the client signs and the server has sent (challenge-client is the challenge for the client). The only thing here not signed that was sent by the server is `opaque` field. Why do you think that needs to be signed?

On the other side, the only thing the server doesn't sign is the returned `opaque` field as well. (and the string form of the peer id, but the server MUST verify that is equivalent/derived to/from the public key bytes we do sign)

http/peer-id-auth.md
MarcoPolo: Just to be clear with the example for a future extension here:

1. Each side derives a `TLSSessionID` using TLS' Keying Material exporters in a TBD way.
2. Each side passes that as a parameter to the `libp2p-PeerID` auth scheme. And includes it as parameter in the signature.
3. Each side MUST verify the provided `TLSSessionID` matches the expected value (as well as verifying the signature as before).
4. Either side MUST fail the authentication if there is a mismatch.

Both sides need to sign and compare it.

http/peer-id-auth.md
MarcoPolo: Would you prefer if the wording were changed to say something like "all parameters must be covered by the signature except for the `opaque` parameter, and the peer-id parameter (as it's covered by the public key bytes)"?

MarcoPolo:
I'm considering making one more change to this spec: Allow the client to send a challenge in the initial request. This would allow the client to authenticate the server on the first round trip, and to send application data after only one round trip. It is similar in spirit to how the noise XX handshake works where the responder (server in our case) is authenticated after the first round trip. 

I can't think of any compelling reason to not allow this, and many reasons to support it. The main reason being that in cases you know you need to authenticate, it removes a round trip.

Are there any compelling cases you can think of @Stebalien, @sukunrt, @aschmahmann?

---

Essentially supporting this flow:
```mermaid
sequenceDiagram
    C->>S: challenge-server
    S-->>C: challenge-client+server-sig [server authenticated]
    C->>S: client-sig + application data [client authenticated]
    S-->>C: resp
```

Alongside what is currently defined:
```mermaid
sequenceDiagram
    C->>S: initial request
    S-->>C: challenge-client
    C->>S: client-sig+challenge-server [client authenticated]
    S-->>C: server-sig [server authenticated]
    C->>S: application data
    S-->>C: resp
```

Stebalien:
I agree, the client should send its challenge in the first message.

http/peer-id-auth.md
```diff
@@ -0,0 +1,331 @@
...
+
+   The opaque parameter is opaque to client. The client MUST return the opaque
+   parameter back to the server. The server MAY use the opaque parameter to
+   encode state.
+
+3. The client makes another HTTP request to the same authenticated resource and
+   sets the header:
+
+   ```
+   Authorization: libp2p-PeerID public-key="<base64-encoded-public-key-bytes>", opaque="<opaque-from-server>", challenge-server="<challenge-string>", sig="<base64-signature-bytes>"
```
sukunrt, line 149: It's clear from context but do you want to rename `sig` to `sig-client` and `sig-server`?

sukunrt (approved):

http/peer-id-auth.md
```diff
@@ -0,0 +1,331 @@
...
+   The `bearer` token allows the client to make future Peer ID authenticated
+   requests. The value is opaque to the client, and the server may use it to
+   store authentication state such as:
+     - The client's Peer ID.
+     - The `hostname` parameter.
+     - The token creation date (to allow tokens to expire).
+
+5. The client MUST verify the signature. After verification the client has
+   authenticated the server's Peer ID. The client SHOULD send the `bearer`
+   token for Peer ID authenticated requests.
```
sukunrt, line 184: If it's mutual auth, both the client and server need to exchange bearer tokens, right?

MarcoPolo: the bearer token is to allow the server to statelessly authenticate the client. Clients are expected to store the state that they have authenticated the server.

sukunrt: How would this work for browsers? Should they re-authenticate servers after a TTL?

MarcoPolo: What are the issues specific to browsers?

On the Go client implementation, the client stores a map from hostname to
```
type tokenInfo struct {
	token      string
	insertedAt time.Time
	peerID     peer.ID
}
```

http/peer-id-auth.md
MarcoPolo: No preference. Happy to change, but I'd prefer calling the client-sig and server-sig.

sukunrt: I am fine with keeping this as sig.

http/peer-id-auth.md
sukunrt: I think this is fine. 

I thought it might cause an issue with the browser since they don't control the transport connection. But we don't do anything special in go either. 

Browsers can have an implementation defined `TokenTTL` similar to the go implementation.

sukunrt: This is fine. Even if the connection is broken and a new connection is established after the auth is run. It'll be as secure as the WebPKI.

PR #568, opened 2023-08-24
libp2phttp: Define the multiaddr URI

MarcoPolo:
This defines a multiaddr URI scheme. This would enable HTTP redirects to other hosts and resources. In Go, this would enable a single `http.RoundTripper` to work on any transport. In JS, we would be able to create a new [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request) with this URI and use it with a custom [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) to make HTTP requests to any transport.

Example:
```
multiaddr:/ip4/1.2.3.4/udp/54321/quic-v1/p2p/12D.../httppath/.well-known%2Flibp2p
```

If we approve this, we should likely register this scheme with IANA. PL has registered other schemes in the past, so I think it is doable (`ipfs` and `ipns`).

### DIFF
--- a/http/README.md
+++ b/http/README.md
@@ -126,6 +126,22 @@ Application protocols using HTTP semantics can run over any libp2p stream transp
 
 HTTP/1.1 is chosen as the minimum bar for interoperability, but other encodings of HTTP semantics are possible as well and may be specified in a future update.
 
+## Multiaddr URI scheme
+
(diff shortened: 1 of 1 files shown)


achingbrain (approved):
LGTM

Issue #570, opened 2023-09-01
Proposal: Remove Peer Exchange in GossipSub Prune Message


diegomrsantos:
**Introduction**  

Peer Exchange in GossipSub may introduce more issues than it aims to solve. I propose that we consider removing it.

**Current Benefits of Peer Exchange:**
- **Mesh Recovery**: Helps pruned peers find alternative nodes to reestablish a mesh.

**Concerns:**
- **Unverifiable Data**: Peer Exchange is the only feature that introduces data that can't be validated in the protocol. A malicious agent could generate fake identities and inject them, leaving no way to verify this data.
- **Message Size**: We've observed prune messages with sizes up to 9KB and 210 peerIDs, and this size is primarily made up of Peer Exchange data. If signed records were populated, sizes could get substantially larger, raising both efficiency and potential DoS concerns.
- **Privacy and Security Risks**: Publicizes all connected peers in a global topic, making them easier targets for attacks.
- **Suboptimal or Malicious Peer Suggestions**: Possible that a malicious actor could flood the network with bad peer suggestions, reducing network quality.
- **Limited Applicability for Peer Discovery**: Peer Exchange may aid in mesh recovery for pruned peers, but it does not replace the need for a dedicated peer discovery mechanism crucial for forming the initial mesh. As signed peer records containing the node's dialable addresses are currently optional, this feature is even less useful. Finally, waiting to be pruned to receive a list of peers is a slow, unpredictable, and unreliable strategy for bootstrapping a network from a single or small set of peers.

**Proposal**  
I propose that we reconsider the existence of Peer Exchange in GossipSub to reduce potential vulnerabilities and to keep the protocol streamlined and focused on its primary goals.

vyzo:
It is impossible to bootstrap from a star without PX unfortunately.

Also note that it clearly stated that this should be limited to trusted entities, like bootstrappers.

vyzo:
Also note that signed peer records are a MUST.

vyzo:
Another point I would like to make: PX really is optional and ignorable; it is not even enabled by default. So the risks you suggest are simply not there, unless your systems programmer shoots himself in the foot.

So to summarize my opinion on the matter: not only the risks suggested are not really there, but removing it is going to remove a very important capability from the protocol: the ability to bootstrap from a star.

So here it is:  **I strongly oppose this motion to remove PX**.

vyzo:
> > Also note that signed peer records are a MUST.
> 
[...]

PR #576, opened 2023-09-14
[Gossip 1.1] Limit the size of PRUNE Peer Exchange list

Nashatyrev:
At the moment the `ControlPrune.peers` list is not limited by any boundary. 

This PR suggests to add the `MaxExchangePeers` parameter for that. 
Instead of hardcoding a 'reasonable default' I'd suggest to parametrize it mostly to have the ability to turn off this feature (by setting this parameter to `0`). 

See also this issue: https://github.com/ethereum/consensus-specs/issues/3500

### DIFF
--- a/pubsub/gossipsub/gossipsub-v1.1.md
+++ b/pubsub/gossipsub/gossipsub-v1.1.md
@@ -126,7 +126,9 @@ In this case, the pruned peer will have to rely on the ambient peer discovery se
 
 #### Protobuf
 
-The `ControlPrune` message is extended with a `peers` field as follows.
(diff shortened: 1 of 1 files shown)


vyzo (commented):
Sounds good to me.

diegomrsantos:
If to disable the feature we need to set the parameter to `0`, making 16 the default value will make it enabled by default.

diegomrsantos:
@Nashatyrev how about making `0` the default value and effectively disabling it in the implementations? Library users who wish to enable the feature should change it to a number greater than `0`.

diegomrsantos:
It seems that was the intention as mentioned here https://github.com/libp2p/specs/issues/570#issuecomment-1703829575

PR #578, opened 2023-09-21
webrtc(browser-browser): specify init channel for connection establishment

sukunrt:
Connection establishment doesn't work if we don't create a channel before creating the SDP offer. 
Not exactly very helpful but some info on this SO question: https://stackoverflow.com/questions/38867763/why-i-have-to-open-data-channel-before-send-peer-connection-offer/38872920#38872920

js-libp2p already does this: https://github.com/libp2p/js-libp2p/blob/master/packages/transport-webrtc/src/private-to-private/handler.ts#L116-L119

go-libp2p: https://github.com/libp2p/go-libp2p/blob/webrtcprivate/transport/p2p/transport/webrtcprivate/transport.go#L242

I think this should be a part of the spec.

### DIFF
--- a/webrtc/webrtc.md
+++ b/webrtc/webrtc.md
@@ -27,6 +27,7 @@ On a historical note, this specification replaces the existing [libp2p WebRTC st
    Note that further steps depend on the relayed connection to be authenticated, i.e. that data sent on the relayed connection can be trusted.
 
 4. _A_ (outbound side of relayed connection) creates an `RTCPeerConnection` provided by a W3C compliant WebRTC implementation (e.g. a browser).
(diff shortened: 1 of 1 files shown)


thomaseizinger:
I think in Rust we handle this by including the relevant `m-line`?

https://github.com/libp2p/rust-libp2p/blob/484fc5693c0a7ad9ad2acb0a08445a7a4f5f7fd9/misc/webrtc-utils/src/sdp.rs#L84

sukunrt:
I think that's for webrtc-direct. Can something similar can be done for webrtc browser-browser too?

thomaseizinger:
> I think that's for webrtc-direct. Can something similar can be done for webrtc browser-browser too?

You are right, that is for `webrtc-direct`. We haven't implemented `webrtc` yet so I can't say.

achingbrain (approved):

mxinden (approved):
Thank you for improving the specs.

Issue #575, opened 2023-09-12
webrtc: figure out how to properly close datachannels


marten-seemann:
Closing WebRTC datachannels is not straightforward. When closing the datachannel, it is instead reset, leading to data sent on that datachannel (potentially) being discarded. See RFC 8831: https://datatracker.ietf.org/doc/html/rfc8831#name-closing-a-data-channel:
> Closing of a data channel MUST be signaled by resetting the corresponding outgoing streams [[RFC6525](https://datatracker.ietf.org/doc/html/rfc6525)]. This means that if one side decides to close the data channel, it resets the corresponding outgoing stream. When the peer sees that an incoming stream was reset, it also resets its corresponding outgoing stream. Once this is completed, the data channel is closed.

This has led to interop failures in go-libp2p, and delayed the release of WebRTC from the go-libp2p v0.31 release. See https://github.com/libp2p/go-libp2p/pull/2337#issuecomment-1696619731 for details.

The problem is that if we don't close a datachannel, the underlying WebRTC / SCTP stack will need to continue keeping track of that datachannel / stream, leading to a memory leak, since a WebRTC connection can use up to 64k streams per association. This is a known bug in rust-libp2p: https://github.com/libp2p/rust-libp2p/issues/4333.

---

### How can we solve this?

This is annoying, and yet another time where WebRTC's suboptimal stream state machine is coming to bite us (can't wait for a WebRTC running on top of QUIC!).

First, here's a workaround: Limit the total number of streams to a value lower than 64k, e.g. 500. This is roughly the number of _concurrent_ streams that other stream multiplexers allow, and it allows for some initial interaction between two peers. Important: Once that number is reached, close the underlying WebRTC connection (don't just return an error from the `NewStream` method). This might lead to some unacknowledged data to be lost, but peers going away in the middle of a transfer is something that application protocols need to deal with anyway.
Obviously, this is not ideal, but it's better than releasing code containing a significant memory leak.

Here's how we could actually solve this: Assuming there's no better way to reliably close a stream, we could add some additional signaling on the libp2p layer. We already use a Protobuf to mirror parts of the QUIC state machine, and this would just be one additional message. Specifically, we'd need to add an acknowledgement for the FIN. Once the FIN-ACK has been received, a peer could actually close the underlying datachannel. We'll need to carefully check how this affects the send and the receive side of the stream.

thomaseizinger:
[...]

PR #582, opened 2023-09-27
webrtc: add FIN_ACK to close datachannels without data loss

achingbrain:
Specify closing datachannels by mutual agreement to ensure all data has been received by the remote before closing.

Refs: #575

### DIFF
--- a/webrtc/README.md
+++ b/webrtc/README.md
@@ -174,8 +174,7 @@ to the user. Consequently we must add an additional layer of signaling to ensure
 reliable data delivery.
 
 When a node wishes to close a stream for writing, it SHOULD send a message with
-the `FIN` flag set, then proceed to wait for a `FIN_ACK` message from the remote
-node.
+the `FIN` flag set. A node SHOULD only consider its write-half closed once it received a `FIN_ACK`.
 
 If a `FIN` flag is received the node SHOULD respond with a `FIN_ACK`.
 
@@ -183,7 +182,7 @@ When a `FIN_ACK` and a `FIN` have been received, a node may close the datachanne
 
 The node MAY close the datachannel without receiving a `FIN_ACK`, for example in
 the case of a timeout, but there will be no guarantee that all previously sent
-messages have been received.
+messages have been received by the remote.
 
 If a node has previously sent a `STOP_SENDING` flag to the remote node, it MUST
 continue to act on any flags present in received messages in order to
@@ -196,16 +195,16 @@ finishes writing.
 
 ```mermaid
 sequenceDiagram
-    NodeA->>NodeB: DATA
-    NodeA->>NodeB: FIN
-    NodeB->>NodeA: FIN_ACK
-    NodeB->>NodeA: DATA
-    NodeB->>NodeA: FIN
-    NodeA->>NodeB: FIN_ACK
+    A->>B: DATA
+    A->>B: FIN
+    B->>A: FIN_ACK
+    B->>A: DATA
+    B->>A: FIN
+    A->>B: FIN_ACK
 ```
 
-After NodeA has received the `FIN` it is free to close the datachannel since it
-has previously received a `FIN_ACK`. If NodeB receives the `FIN_ACK` before this
+After _A_ has received the `FIN` it is free to close the datachannel since it
+has previously received a `FIN_ACK`. If _B_ receives the `FIN_ACK` before this
 it may close the channel since it previously received a `FIN`.
 
 This way the channel can be closed from either end without data loss.
--- a/webrtc/README.md
+++ b/webrtc/README.md
@@ -67,9 +67,9 @@ message Message {
     // The sender abruptly terminates the sending part of the stream. The
     // receiver MAY discard any data that it already received on that stream.
     RESET_STREAM = 2;
-    // The receiver has previously sent a message with a FIN flag. On receiving
-    // the FIN_ACK flag they may now close the stream as they know the remote has received
-    // all data that was sent before the FIN.
+    // Sending the FIN_ACK flag acknowledges the previous receipt of a message
+    // with the FIN flag set. Receiving a FIN_ACK flag gives the recipient
+    // confidence that the remote has received all sent messages.
     FIN_ACK = 3;
   }
 
@@ -168,7 +168,7 @@ of the `RTCDataChannel` `label` property.
 ## Closing an `RTCDataChannel`
 
 Some WebRTC implementations do not guarantee that any queued messages will be
-sent after a datachannel is closed.  Other implementatations maintain separate
+sent after a datachannel is closed.  Other implementations maintain separate
 outgoing message and transport queues, the status of which may not be visible
 to the user. Consequently we must add an additional layer of signaling to ensure
 reliable data delivery.
@@ -177,14 +177,9 @@ When a node wishes to close a stream for writing, it SHOULD send a message with
 the `FIN` flag set, then proceed to wait for a `FIN_ACK` message from the remote
 node.
 
-If a `FIN` flag is received and the node has finished writing data, it SHOULD
-respond with a `FIN_ACK` immediately.
+If a `FIN` flag is received the node SHOULD respond with a `FIN_ACK`.
 
-If a `FIN` flag is received but the node has not finished writing data, it SHOULD
-delay sending a `FIN_ACK` response until it has finished writing data, sent a
-`FIN` flag, and finally received a `FIN_ACK` in response.
-
-When a `FIN_ACK` has been sent and received, a node may close the datachannel.
+When a `FIN_ACK` and a `FIN` have been received, a node may close the datachannel.
 
 The node MAY close the datachannel without receiving a `FIN_ACK`, for example in
 the case of a timeout, but there will be no guarantee that all previously sent
@@ -203,11 +198,17 @@ finishes writing.
 sequenceDiagram
     NodeA->>NodeB: DATA
     NodeA->>NodeB: FIN
+    NodeB->>NodeA: FIN_ACK
     NodeB->>NodeA: DATA
     NodeB->>NodeA: FIN
     NodeA->>NodeB: FIN_ACK
-    NodeB->>NodeA: FIN_ACK
 ```
+
+After NodeA has received the `FIN` it is free to close the datachannel since it
+has previously received a `FIN_ACK`. If NodeB receives the `FIN_ACK` before this
+it may close the channel since it previously received a `FIN`.
+
+This way the channel can be closed from either end without data loss.
 
 ## Previous, ongoing and related work
 
--- a/webrtc/README.md
+++ b/webrtc/README.md
@@ -174,9 +174,11 @@ to the user. Consequently we must add an additional layer of signaling to ensure
 reliable data delivery.
 
 When a node wishes to close a stream for writing, it SHOULD send a message with
-the `FIN` flag set. A node SHOULD only consider its write-half closed once it received a `FIN_ACK`.
+the `FIN` flag set.
 
 If a `FIN` flag is received the node SHOULD respond with a `FIN_ACK`.
+
+A node SHOULD only consider its write-half closed once it received a `FIN_ACK`.
 
 When a `FIN_ACK` and a `FIN` have been received, a node may close the datachannel.
 
--- a/webrtc/README.md
+++ b/webrtc/README.md
@@ -170,17 +170,19 @@ of the `RTCDataChannel` `label` property.
 Some WebRTC implementations do not guarantee that any queued messages will be
 sent after a datachannel is closed.  Other implementations maintain separate
 outgoing message and transport queues, the status of which may not be visible
-to the user. Consequently we must add an additional layer of signaling to ensure
-reliable data delivery.
+to the user. Consequently we must add an additional layer of signaling to
+ensure reliable data delivery.
 
 When a node wishes to close a stream for writing, it MUST send a message with
 the `FIN` flag set.
 
 If a `FIN` flag is received the node SHOULD respond with a `FIN_ACK`.
 
-A node SHOULD only consider its write-half closed once it has received a `FIN_ACK`.
+A node SHOULD only consider its write-half closed once it has received a
+`FIN_ACK`.
 
-When a `FIN_ACK` and a `FIN` have been received, the node may close the datachannel.
+When a `FIN_ACK` and a `FIN` have been received, the node may close the
+datachannel.
 
 The node MAY close the datachannel without receiving a `FIN_ACK`, for example in
 the case of a timeout, but there will be no guarantee that all previously sent
@@ -192,8 +194,8 @@ successfully process an incoming `FIN_ACK`.
 
 ### Example of closing an `RTCDataChannel`
 
-NodeA closes for writing, NodeB delays allowing the channel to close until it also
-finishes writing.
+NodeA closes for writing, NodeB delays allowing the channel to close until it
+also finishes writing.
 
 ```mermaid
 sequenceDiagram
--- a/webrtc/README.md
+++ b/webrtc/README.md
@@ -67,6 +67,10 @@ message Message {
     // The sender abruptly terminates the sending part of the stream. The
     // receiver MAY discard any data that it already received on that stream.
     RESET_STREAM = 2;
+    // The receiver has previously sent a message with a FIN flag. On receiving
+    // this flag they may now close the stream as they know the remote has received
+    // all data that was sent.
+    FIN_ACK = 3;
   }
 
   optional Flag flag=1;
@@ -132,8 +136,8 @@ real-world experiments.
 
 `RTCDataChannel`s are negotiated in-band by the WebRTC user agent (e.g. Firefox,
 Pion, ...). In other words libp2p WebRTC implementations MUST NOT change the
-default value `negotiated: false` when creating a standard libp2p stream 
-of type `RTCDataChannel` via `RTCPeerConnection.createDataChannel`. 
+default value `negotiated: false` when creating a standard libp2p stream
+of type `RTCDataChannel` via `RTCPeerConnection.createDataChannel`.
 Setting `negotiated: true` is reserved only for creating Noise handshake channels
 under certain protocol conditions.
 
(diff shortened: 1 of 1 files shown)


webrtc/README.md
```diff
@@ -67,6 +67,10 @@ message Message {
     // The sender abruptly terminates the sending part of the stream. The
     // receiver MAY discard any data that it already received on that stream.
     RESET_STREAM = 2;
+    // The receiver has previously sent a message with a FIN flag. On receiving
+    // this flag they may now close the stream as they know the remote has received
```
mxinden, line 71:
```suggestion
    // the FIN_ACK flag they may now close the stream as they know the remote has received
```

Avoids the confusion with this referring to FIN.

webrtc/README.md
```diff
@@ -67,6 +67,10 @@ message Message {
...
+    // all data that was sent.
```
mxinden, line 72:
```suggestion
    // all data that was sent before the FIN.
```

No strong opinion.

webrtc/README.md
```diff
@@ -161,6 +165,50 @@ MUST pass an empty string. When receiving an `RTCDataChannel` via
...
 
+## Closing an `RTCDataChannel`
+
+Some WebRTC implementations do not guarantee that any queued messages will be
+sent after a datachannel is closed.  Other implementatations maintain separate
+outgoing message and transport queues, the status of which may not be visible
+to the user. Consequently we must add an additional layer of signaling to ensure
+reliable data delivery.
+
+When a node wishes to close a stream for writing, it should send a message with
```
mxinden, line 176:
```suggestion
When a node wishes to close a stream for writing, it SHOULD send a message with
```

webrtc/README.md
```diff
@@ -161,6 +165,50 @@ MUST pass an empty string. When receiving an `RTCDataChannel` via
...
+the `FIN` flag set, then proceed to wait for a `FIN_ACK` message from the remote
+node.
+
+If a `FIN` flag is received and the node has finished writing data, it should
```
mxinden, line 180:
```suggestion
If a `FIN` flag is received and the node has finished writing data, it SHOULD
```

webrtc/README.md
```diff
@@ -161,6 +165,50 @@ MUST pass an empty string. When receiving an `RTCDataChannel` via
...
+respond with a `FIN_ACK` immediately.
+
+If a `FIN` flag is recieved but the node has not finished writing data, it should
```
mxinden, line 183:
```suggestion
If a `FIN` flag is received but the node has not finished writing data, it SHOULD
```

webrtc/README.md
```diff
@@ -161,6 +165,50 @@ MUST pass an empty string. When receiving an `RTCDataChannel` via
...
+delay sending a `FIN_ACK` response until it has finished writing data, sent a
+`FIN` flag, and finally received a `FIN_ACK` in response.
+
+When a `FIN_ACK` has been sent and received, a node may close the datachannel.
+
+The node may close the datachannel without receiving a `FIN_ACK`, for example in
```
mxinden, line 189:
```suggestion
The node MAY close the datachannel without receiving a `FIN_ACK`, for example in
```

webrtc/README.md
```diff
@@ -161,6 +165,50 @@ MUST pass an empty string. When receiving an `RTCDataChannel` via
...
+the case of a timeout, but there will be no guarantee that all previously sent
+messages have been received.
+
+If a node has previously sent a `STOP_SENDING` flag to the remote node, it must
```
mxinden, line 193:
```suggestion
If a node has previously sent a `STOP_SENDING` flag to the remote node, it MUST
```

Note that this is a breaking change. I.e. existing implementations will not adhere to this requirement. Though one could argue that it is rather minor.

mxinden, line 193:
```suggestion
If a node has previously sent a `STOP_SENDING` flag to the remote node before a FIN, it must
```

Just to make sure we are on the same page, one does not expect to receive a FIN_ACK after a STOP_SENDING, correct?

achingbrain: The protobuf definition says about `STOP_SENDING`;

> The sender will no longer read messages on the stream. Incoming data is being discarded on receipt.

My reading of that was that the sender would not process any incoming messages at all after sending this flag but I was trying to call out that they should actually continue to process the flags from incoming messages (ignoring any stream data) in order to receive a `FIN_ACK` message.

Maybe it could be better worded?

achingbrain: Also I think you could send `FIN` (e.g. close write) then `STOP_SENDING` (e.g. close read) before receiving a `FIN_ACK` - and adding the `before a FIN` in the suggestion above excludes this scenario?

webrtc/README.md
```diff
@@ -67,6 +67,10 @@ message Message {
     // The sender abruptly terminates the sending part of the stream. The
     // receiver MAY discard any data that it already received on that stream.
     RESET_STREAM = 2;
+    // The receiver has previously sent a message with a FIN flag. On receiving
+    // the FIN_ACK flag they may now close the stream as they know the remote has received
```
thomaseizinger, line 71:
```suggestion
    // the FIN_ACK flag they may now close the write-side of the stream as they know the remote has received
```

We should still continue reading, right?

achingbrain: RTCDataChannels don't support half-closed streams, the whole thing is either open or closed, so saying "close the write-side" could be a bit confusing?

thomaseizinger: This is refering to our framing mechanism which supports half-closing, right?

achingbrain: Ah, ok - this is my mistake then, the idea here is this is the datachannel rather than the libp2p stream.  It should probably say that..

thomaseizinger: Unfortunately, the browser APIs don't give us access to the SCTP APIs which would support all of these things. Hence we needed to make our own framing. AFAIK, all we can do on a data channel is send data and close it entirely. No half-closing, no resetting.

webrtc/README.md
```diff
@@ -161,6 +165,50 @@ MUST pass an empty string. When receiving an `RTCDataChannel` via
...
+finishes writing.
+
+```mermaid
+sequenceDiagram
+    NodeA->>NodeB: DATA
+    NodeA->>NodeB: FIN
+    NodeB->>NodeA: DATA
+    NodeB->>NodeA: FIN
+    NodeA->>NodeB: FIN_ACK
+    NodeB->>NodeA: FIN_ACK
```
sukunrt, line 209: How does this work when NodeA is a go-libp2p node and NodeB is a chromium node. 

A -> B: FIN
B -> A: Data
B -> A: FIN
A -> B: FIN_ACK
B -> A: FIN_ACK
Now as B is a chromium node the final FIN_ACK will be dropped since B is a chromium node and we don't know if the final message will be sent reliably. 
Should A just exit after a fixed period of time and exit if no FIN_ACK sort of like a timewait state?

sukunrt: should we mention handling this using onClose event handler on the datachannel and to ensure checking bufferedAmount is 0 before closing?

sukunrt: @achingbrain you can resolve this, if my understanding here is correct. 
specifically this:
>Since B closes the datachannel, A will be notified of the datachannel close with the onClose event handler.

webrtc/README.md
mxinden: How about `before or after a FIN`? What I want to avoid is that a reader misunderstands that a `STOP_SENDING` is also acknowledged by a `FIN_ACK`.

webrtc/README.md
achingbrain: Indeed, yes.

Anyway what I meant was it should probably say:

```suggestion
    // the FIN_ACK flag they may now close the datachannel as they know the remote has received
```

webrtc/README.md
achingbrain: Yes, this is my understanding too.

```
1. A -> B: FIN         // A tells B it wants to close the datachannel
2. B -> A: Data        // B still sends data
3. B -> A: FIN         // B is done sending data, wants to close the datachannel
4. A -> B: FIN_ACK     // A agrees, yes, close the datachannel
5. B -> A: FIN_ACK     // B agrees, yes close the datachannel (or has closed it already?)
```

The side to close the channel is the first one to have received a `FIN_ACK` and to have sent one too (B after step 4. above).

The other side would only close the datachannel if it had sent the `FIN_ACK` to the remote (step 4 above) and the remote has not closed the channel after a timeout.

I guess then that the sending of the final `FIN_ACK` is redundant?  Can you think of a case where that would not be true?

achingbrain: Sounds good.

webrtc/README.md
achingbrain: Maybe we could explicitly state that with something like:

> The FIN_ACK flag acknowledges the previous receipt of a message with the FIN flag set.
On receiving the FIN_ACK flag the recipient may now close the datachannel as they know the remote has received all data that was sent before the FIN.

It might be getting a bit wordy, what do you think to reducing it to just:

> Sending the FIN_ACK flag acknowledges the previous receipt of a message with the FIN flag set.

...and limiting any other exposition to the "Closing an `RTCDataChannel`" section?

webrtc/README.md
thomaseizinger: I don't think this is true, right? The remote is just acknowledging that _we_ closed our side of the stream. Unless the remote has also sent us a `FIN`, we need to keep reading.

webrtc/README.md
thomaseizinger: > > I guess then that the sending of the final FIN_ACK is redundant? Can you think of a case where that would not be true?
> 
> I can't. Any scheme which relies on the final FIN_ACK being correctly delivered can't work because chromium might drop that final message. The present scheme looks fine to me. I am also fine with keeping the final FIN_ACK as it is since it doesn't cost us much.

This is a [two-generals problem](https://en.wikipedia.org/wiki/Two_Generals%27_Problem), right? It doesn't matter how many ACKs you send, you can never be sure of the final one :)

Closing the channel can be arbitrarily interleaved, meaning we cannot depend on one party being the "first" one to close the channel. Both sides may perceive themselves as being the first one.

I think the best we can do is:

- Send `FIN` to start closing the write-half of the stream
- Keep reading from the stream
- At any point, if a node sent & received a `FIN`, close the underlying data channel
- If you receive a `FIN_ACK`, keep reading

As @sukunrt said above, what may happen is that a browser-node within Chromium receives a `FIN`, responds with `FIN_ACK`, writes some more data, sends `FIN` and as a result of the above logic immediately closes the data channel. This final `FIN` may get lost and never makes it to the remote party.

Until the chromium bug is fixed, I think we have to accept that some channels may be flagged as "closed prematurely" because we receive a `close` event without having received a `FIN` flag.

sukunrt: >Chromium receives a FIN, responds with FIN_ACK, writes some more data, sends FIN and as a result of the above logic immediately closes the data channel. This final FIN may get lost and never makes it to the remote party.

This is not the speced behaviour. The behaviour in the specs is: 
```
If a FIN flag is received but the node has not finished writing data, it SHOULD
delay sending a FIN_ACK response until it has finished writing data, sent a
FIN flag, and finally received a FIN_ACK in response.
```
Which implies, 
1. Chromium receives a FIN
2. writes some more data
3. writes a FIN
4. writes a FIN_ACK 
5. waits for FIN_ACK from peer

The peer on receiving the FIN_ACK will write a FIN_ACK and close the channel, but this FIN_ACK may get lost. This is fine because the Chromium node in question will be informed of the datachannel closing by the onClose call back on the datachannel. Note: both sides would have sent a FIN and received a FIN before the onClose event handler fires. 

I must say this is a rather clever solution to the problem. @achingbrain 🙇 

Considering the discussion above, I think we should change the recommendation for delaying the FIN_ACK to MUST

webrtc/README.md
```diff
@@ -161,6 +165,50 @@ MUST pass an empty string. When receiving an `RTCDataChannel` via
...
+When a node wishes to close a stream for writing, it SHOULD send a message with
+the `FIN` flag set, then proceed to wait for a `FIN_ACK` message from the remote
+node.
+
+If a `FIN` flag is received and the node has finished writing data, it SHOULD
+respond with a `FIN_ACK` immediately.
+
+If a `FIN` flag is received but the node has not finished writing data, it SHOULD
+delay sending a `FIN_ACK` response until it has finished writing data, sent a
+`FIN` flag, and finally received a `FIN_ACK` in response.
```
sukunrt, line 185: Why do we need this `SHOULD`? 
The simpler alternative is: 

If a `FIN` flag is received the node should reply with a `FIN_ACK`. If the node has received a `FIN` from the peer and a `FIN_ACK` for the `FIN` it had sent, it SHOULD close the stream. 

```
A -> B: FIN
B -> A: FIN_ACK
B -> A: Data
B -> A: FIN
A -> B: FIN_ACK
```
A closes the stream. if B doesn't receive a FIN_ACK because A is on chromium it's okay, since it has all the information to infer that both sides have closed the stream. It has sent and received a FIN. 

Simultaneous Close case: 

```
A -> B: FIN
B -> A: FIN
A -> B: FIN_ACK
B -> A: FIN_ACK
```
In this case it doesn't matter which side closes the stream on FIN_ACK. if A closes after receiving FIN_ACK from B, at that point B has acked A's FIN and A has received B's FIN. 

cc: @thomaseizinger

webrtc/README.md
sukunrt: maybe the simpler solution works: see https://github.com/libp2p/specs/pull/582#discussion_r1343630055

webrtc/README.md
thomaseizinger: > Why do we need this `SHOULD`?

I agree that we should not delay the `FIN_ACK`. In my eyes, `FIN_ACK` says nothing about whether you are closing the stream or not, it just acknowledges that you've received the other node's `FIN` which gives the other node that guarantee that they've flushed all their data out of the buffer.

webrtc/README.md
achingbrain: It's true if the remote delays sending the `FIN_ACK` until it's received a `FIN_ACK`, otherwise yes, it's not enough to allow closing the channel.  I think this is hard to see on first reading so the language should be clearer.

I've removed the text that implies what you can do after receiving the flag, leaving that to the "Closing an `RTCDataChannel`" section.

webrtc/README.md
achingbrain: This seems better, no need to delay sending the `FIN_ACK` and it seems to work locally.  I've updated the PR text to remove mentions of delaying sending.

webrtc/README.md
achingbrain: I've simplified the text here leaving the interpretation to the "Closing an RTCDataChannel" section.

webrtc/README.md
```diff
@@ -161,6 +165,51 @@ MUST pass an empty string. When receiving an `RTCDataChannel` via
...
+sent after a datachannel is closed.  Other implementations maintain separate
+outgoing message and transport queues, the status of which may not be visible
+to the user. Consequently we must add an additional layer of signaling to ensure
+reliable data delivery.
+
+When a node wishes to close a stream for writing, it SHOULD send a message with
+the `FIN` flag set, then proceed to wait for a `FIN_ACK` message from the remote
+node.
+
+If a `FIN` flag is received the node SHOULD respond with a `FIN_ACK`.
```
thomaseizinger, line 180: If we flip the order of these two, it is easier to understand why we wait for a `FIN_ACK` after sending `FIN`.

webrtc/README.md
```diff
@@ -161,6 +165,51 @@ MUST pass an empty string. When receiving an `RTCDataChannel` via
...
+
+Some WebRTC implementations do not guarantee that any queued messages will be
+sent after a datachannel is closed.  Other implementations maintain separate
+outgoing message and transport queues, the status of which may not be visible
+to the user. Consequently we must add an additional layer of signaling to ensure
+reliable data delivery.
+
+When a node wishes to close a stream for writing, it SHOULD send a message with
+the `FIN` flag set, then proceed to wait for a `FIN_ACK` message from the remote
+node.
```
thomaseizinger, line 178:
```suggestion
When a node wishes to close a stream for writing, it SHOULD send a message with
the `FIN` flag set. A node SHOULD only consider its write-half closed once it received a `FIN_ACK`.
```

webrtc/README.md
```diff
@@ -161,6 +165,51 @@ MUST pass an empty string. When receiving an `RTCDataChannel` via
...
+
+If a `FIN` flag is received the node SHOULD respond with a `FIN_ACK`.
+
+When a `FIN_ACK` and a `FIN` have been received, a node may close the datachannel.
+
+The node MAY close the datachannel without receiving a `FIN_ACK`, for example in
+the case of a timeout, but there will be no guarantee that all previously sent
+messages have been received.
```
thomaseizinger, line 186:
```suggestion
messages have been received by the remote.
```

webrtc/README.md
```diff
@@ -161,6 +165,51 @@ MUST pass an empty string. When receiving an `RTCDataChannel` via
...
+finishes writing.
+
+```mermaid
+sequenceDiagram
+    NodeA->>NodeB: DATA
+    NodeA->>NodeB: FIN
+    NodeB->>NodeA: FIN_ACK
+    NodeB->>NodeA: DATA
+    NodeB->>NodeA: FIN
+    NodeA->>NodeB: FIN_ACK
```
thomaseizinger, line 204:
```suggestion
    A->>B: DATA
    A->>B: FIN
    B->>A: FIN_ACK
    B->>A: DATA
    B->>A: FIN
    A->>B: FIN_ACK
```

Slightly less noisy.

webrtc/README.md
```diff
@@ -161,6 +165,51 @@ MUST pass an empty string. When receiving an `RTCDataChannel` via
...
+```
+
+After NodeA has received the `FIN` it is free to close the datachannel since it
+has previously received a `FIN_ACK`. If NodeB receives the `FIN_ACK` before this
+it may close the channel since it previously received a `FIN`.
+
+This way the channel can be closed from either end without data loss.
```
thomaseizinger, line 211:
```suggestion
After _A_ has received the `FIN` it is free to close the datachannel since it
has previously received a `FIN_ACK`. If _B_ receives the `FIN_ACK` before this
it may close the channel since it previously received a `FIN`.

This way the channel can be closed from either end without data loss.
```

I think other parts in this document just use italic for the identifiers.

thomaseizinger (approved):
Great work! Some final comments.

webrtc/README.md
achingbrain: I've split the paragraph up and changed the order, hopefully the sequence of events is clearer now?

sukunrt (approved):
This looks great! 🚀

thomaseizinger (approved):
Great work!

No more concerns from my end.

achingbrain:
Awesome stuff everyone!

PR #590, opened 2023-10-16
Ambient peer discovery protocol

thomaseizinger:
This PR attempts to specify an ambient peer discovery protocol for libp2p. Lots of feedback from the discussion at https://github.com/libp2p/specs/discussions/587 has already been incorporated. Thanks to everyone for getting involved.

I am putting this up for a more structured review.

### DIFF
--- a/ambient-peer/README.md
+++ b/ambient-peer/README.md
@@ -37,14 +37,15 @@ Ambient peer discovery allows the web app to inquire for further nodes from the 
 1. Node _B_ chooses a subset of at most 5 known peer records received from other peers.
    1. The chosen peer records SHOULD at least have one address that share the same transport technology as the the connection between node _A_ and node _B_.
       For example, if node _A_ and node _B_ are connected via WebRTC, node _B_ SHOULD select 5 peer records where each one of them has at least one WebRTC address.
-   1. Node _B_ MUST NOT be currently connected to any of these nodes.
+   1. Node _B_ SHOULD NOT be currently connected to any of these nodes.
 1. Node _B_ writes these peer records onto the stream in their [protobuf encoding](https://github.com/libp2p/specs/blob/master/RFC/0003-routing-records.md#address-record-format), each record being length-prefixed using an unsigned varint and closes the stream after the last one.
 1. Node _A_ reads peer records from the stream until EOF or 5 have been received, whichever comes earlier.
 
 ## Security considerations
 
 Revealing even just some of your peers has serious privacy and security implications for a network.
-As a result, this specification forbids nodes to return peer records of nodes they are connected to.
+By default, implementations MUST NOT share records of peers they are currently connected to.
+Implementations MAY add a configuration flag that allows users to override this.
 
 <!-- @vyzo to add more text here -->
 
@@ -68,6 +69,12 @@ Implementations MAY group transports as follows:
 3. **Anything Web:** If a peer connects over `/webrtc`, `/webrtc-direct`, `/webtransport` or `/ws`, chances are they are a browser node.
    As such, nodes MAY assume that any peer record with one of these is useful.
 4. **IPv4 & IPv6**: Nodes MAY assume that the requesting peer is capable of dialing either version of IP, regardless of which one was used to make the connection.
+
+### Separating networks
+
+Libp2p is used across a range of networks and many of them may not actually have a useful overlap in compatible protocols.
+To avoid sharing addresses of peers that don't support useful protocols, implementations SHOULD allow configuration of the protocol identifier.
+For example, instead of `/libp2p/ambient-peers` a node may use `/my-cool-p2p-network/ambient-peers`.
 
 ## Prior art
 
--- a/ambient-peer/README.md
+++ b/ambient-peer/README.md
@@ -2,7 +2,7 @@
 
 | Lifecycle Stage | Maturity      | Status | Latest Revision |
 |-----------------|---------------|--------|-----------------|
-| 1A              | Working Draft | Active | r0, 2023-10-08  |
+| 1A              | Working Draft | Active | r0, 2023-10-16  |
 
 Authors: [@thomaseizinger]
 
--- a/ambient-peer/README.md
+++ b/ambient-peer/README.md
@@ -21,8 +21,7 @@ See the [lifecycle document][lifecycle-spec] for context about the maturity leve
 ## Overview
 
 The ambient peer discovery protocol allows peers to share some of their ambient peers with each other.
-Ambient in this case simply means "known peers".
-There is no definition of distance like in kademlia and neither do peers have to be connected to consider another peer as ambient.
+Ambient in this case means "peers I used to be connected to".
 
 ## Usecase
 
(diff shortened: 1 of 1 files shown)


ambient-peer/README.md
```diff
@@ -0,0 +1,109 @@
...
+The connection will succeed but because both nodes are browsers, further discovery of nodes via e.g. kademlia is not possible.
+Ambient peer discovery allows the web app to inquire for further nodes from the new connection.
+
+The protocol is designed to compliment other discovery mechanism like kademlia.
+It features a very small resource footprint and can thus also be used by lite-clients within browser or mobile environments.
+
+## Protocol
+
+1. Node _A_ opens a new stream to node _B_ with the protocol name `/libp2p/ambient-peers`.
+1. Node _B_ chooses a subset of at most 5 known peer records received from other peers.
```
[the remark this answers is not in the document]
thomaseizinger:
```suggestion
1. Node _B_ MUST choose a subset of known peer records received from other peers.
    Node _B_ SHOULD limit the subset to a maximum of 5 peers.
```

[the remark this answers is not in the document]
thomaseizinger: But how do you know, if the other side configured a different number? Unless we make this a runtime parameter of the protocol, I think enforcing a number is easier. A node can always run the same protocol again to get more peers.

ambient-peer/README.md
```diff
@@ -0,0 +1,109 @@
...
+   1. The chosen peer records SHOULD at least have one address that share the same transport technology as the the connection between node _A_ and node _B_.
+      For example, if node _A_ and node _B_ are connected via WebRTC, node _B_ SHOULD select 5 peer records where each one of them has at least one WebRTC address.
+   1. Node _B_ SHOULD NOT be currently connected to any of these nodes.
```
marten-seemann, line 42: 1. This is phrased in a misleading way.
2. Returning stale peers seems less useful.

ambient-peer/README.md
```diff
@@ -0,0 +1,109 @@
...
+Ambient peer discovery allows the web app to inquire for further nodes from the new connection.
+
+The protocol is designed to compliment other discovery mechanism like kademlia.
+It features a very small resource footprint and can thus also be used by lite-clients within browser or mobile environments.
+
+## Protocol
+
+1. Node _A_ opens a new stream to node _B_ with the protocol name `/libp2p/ambient-peers`.
+1. Node _B_ chooses a subset of at most 5 known peer records received from other peers.
+   1. The chosen peer records SHOULD at least have one address that share the same transport technology as the the connection between node _A_ and node _B_.
```
marten-seemann, line 40: Making the supported transports explicit seems like a strictly superior way.

ambient-peer/README.md
```diff
@@ -0,0 +1,109 @@
...
+Implementations MAY add a configuration flag that allows users to override this.
+
+<!-- @vyzo to add more text here -->
+
+## Implementation considerations
+
+### Bound local peer storage
+
+This protocol requires nodes to store records of peers they used to be connected to.
+This is useful independently of this protocol to e.g. reconnect to a peer you've once been connected to.
```
marten-seemann, line 59: This adds a lot of complexity to implementations.

marten-seemann (changes requested):
I'm not sure if @vyzo's concerns have been properly addressed here.

I'd also see a clear use case for this before continuing work on this proposal. We shouldn't go around and specify protocols without a _very clear_ need.

ambient-peer/README.md
thomaseizinger: Can you elaborate on why this adds complexity? I appreciate that the various implementations differ greatly in their internal design but fundamentally, this can easily be satisfied with an LRU cache. Really, all we are doing is accessing the ["peer store"](https://docs.libp2p.io/concepts/fundamentals/peers/#peer-store). I can make that clearer if you want?

ambient-peer/README.md
thomaseizinger: > Returning stale peers seems less useful.

Agreed that it is less useful but it is also a lot safer. We landed on this compromise because there was a lot of strong feedback that returning your _current_ peers is unacceptable from a privacy PoV.

thomaseizinger:
> I'd also see a clear use case for this before continuing work on this proposal. We shouldn't go around and specify protocols without a _very clear_ need.

Agreed!

The work was primarily motivated by universal-connectivity which as far as I understand, is the flagship of the whole connectivity story of libp2p. At the moment however, it is quite limited in that when you connect to another browser peer using it, there is no libp2p-native way to discover further nodes from it. This proposal aims to solve this.

I've also updated the PR description with the issues that have been created to fill this and similar needs.

It is also listed on the [roadmap](https://github.com/libp2p/specs/blob/master/ROADMAP.md#-peer-exchange-protocol).

mxinden:
In general, I think a minimal peer discovery protocol like the one proposed here is useful.

> I'd also see a clear use case for this before continuing work on this proposal. We shouldn't go around and specify protocols without a _very clear_ need.

Agreed. Without a clear use-case in mind, I don't think we will design a good protocol.

> The work was primarily motivated by universal-connectivity which as far as I understand, is the flagship of the whole connectivity story of libp2p.

I see the universal-connectivity app as a showcase application only. I don't think it suffices as a clear use-case to justify a new libp2p protocol. To me a simple hack, e.g. one Gossipsub topic where everyone periodically sends their address, is good enough for universal-connectivity.

All that said, in case there is a real-world use-case out there and they are willing to drive this effort, let's get a first implementation in place and then iterate on the specification.

thomaseizinger:
> > The work was primarily motivated by universal-connectivity which as far as I understand, is the flagship of the whole connectivity story of libp2p.
> 
> I see the universal-connectivity app as a showcase application only. I don't think it suffices as a clear use-case to justify a new libp2p protocol. To me a simple hack, e.g. one Gossipsub topic where everyone periodically sends their address, is good enough for universal-connectivity.

I agree that it is a showcase. But it also clearly surfaces a limitation of the current libp2p protocol suite: In a browser-to-browser setting, application developers need to roll their own solutions for discovering new peers.

> All that said, in case there is a real-world use-case out there and they are willing to drive this effort, let's get a first implementation in place and then iterate on the specification.

Given the simplicity of the protocol, how about we implement it within universal-connectivity as a first state? I am happy to implement the Rust version of this protocol. Perhaps @achingbrain would be interested in prototyping a JS implementation?

I also want to highlight that this spec is marked as [**Working Draft**](https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md#level-1-working-draft). It is by no means meant to be perfect and / or complete but I do think it is in a coherent state and something that is interesting to experiment with.

Having it as a module in universal-connectivity would:

- Allow us to gain experience in how well the current proposal works.
- Allow users to have something to copy-paste into their own projects.
- Allow users to give us feedback on the idea on whether or not it works for them.
- Clearly communicate that this isn't something the libp2p project is committed to supporting long-term because it is not part of the respective libp2p implementations.

If I understand the spec life-cycle correctly, then this is what working drafts are for. I'd like to ask to taking this into consideration when reviewing the proposal.

2color:
> I see the universal-connectivity app as a showcase application only. I don't think it suffices as a clear use-case to justify a new libp2p protocol. 

Why not? It demonstrates a pretty common hub-spoke network topology with browsers being ephemeral peers that come and go and one or two Rust/Go peers are constantly running and act like boostrappers that assist browser peers with establishing p2p connectivity. 

It's the use case laid out here: https://github.com/libp2p/universal-connectivity/issues/95

SgtPooki:
FYI: usecases that this spec resolves is asked about in ip-js, and other filecoin slack channels, often. 

> If I wanted to use WebRTC in a browser-to-browser app, do I need to setup my own relay node(s)? Or is there a public list of relay IDs?

also https://github.com/shovelers/network-monorepo is one specific app that comes to mind as a usecase for this.

---

I don't understand how there's confusion about the value of making it easier to discover peers in a (_the_) p2p lib..

PR #595, opened 2023-11-12
docs: update spelling in GossipSub-v1.1

rob-maron:
Just some small things I saw while reading through the spec


vyzo (approved):
thank you!

Issue #203, opened 2019-08-08
Discuss: Move away from semver for protocol versioning


raulk:
We have discussed this in various contexts. I believe the discussion deserves larger visibility.

I documented my rationale in the Ethereum 2.0 networking spec. Copying (with editorial fixes):

> ### Why do we version protocol strings with ordinals instead of semver?
> 
> Using semver for network protocols is confusing. It is never clear what a change in a field, even if backwards compatible on deserialisation, actually implies. Network protocol agreement should be explicit. Imagine two peers:
> 
> -  Peer A supporting v1.1.1 of protocol X.
> -  Peer B supporting v1.1.2 of protocol X.
> 
> These two peers should never speak to each other because the results can be unpredictable. This is an oversimplification: imagine the same problem with a set of 10 possible > versions. We now have 10^2 (100) possible outcomes that peers need to model for. The resulting complexity is unwieldy.
> 
> For this reason, we rely on negotiation of explicit, verbatim protocols. In the above case, peer B could provide backwards compatibility by supporting and advertising both v1.1.1 and v1.1.2 of the protocol.
> 
> Under this scenario, semver would be relegated to convey expectations at the human level, and it wouldn't do a good job there either, because it's unclear if "backwards-compatibility" and "breaking change" apply only to wire schema level, to behaviour, state machine, etc.
> 
> Therefore, we remove semver out of the picture and replace it with ordinals that require explicit agreement, and do not mandate a specific policy for changes.

I'm personally _for_ this change. Curious to hear from others.

marten-seemann:
I agree, semver seems to be a pretty bad fit for wire protocols.

One thing we might want to consider is dealing with extensions. Imagine you have a base protocol that gives peers the option to implement different extensions on top of this protocol. If you want to announce support for extensions upfront (as opposed to negotiating extensions within the protocol itself), you'd have to announce the supported extensions in some way. Even with a modest number of extensions, the number of combinations becomes huge, so giving each combination a unique name (version number) is not a solution.

aschmahmann:
@raulk your argument applies against the use of semver in general in distributed systems.  Not necessarily a good or bad thing, but just pointing out that there's very little uniqueness to protocol versioning.

>Under this scenario, semver would be relegated to convey expectations at the human level, and it wouldn't do a good job there either, because it's unclear if "backwards-compatibility" and "breaking change" apply only to wire schema level, to behaviour, state machine, etc.

[...]

PR #598, opened 2023-11-18
webrtc: introduce versioning for signalling protocol id

maschad:
The js-libp2p implementation of webrtc includes versioning for the [signalling protocol id](https://github.com/libp2p/js-libp2p/blob/9c67c5b3d0ab63c7a1a62f363ae732b300ef6b87/packages/transport-webrtc/src/private-to-private/transport.ts#L25) therefore existing nodes that support private-to-private would be utilizing a versioned protocol id. Whilst this may have been a mistake, it could be more prudent to introduce this versioning at the spec level as opposed to forcing existing nodes to synchronize upgrades.

1) There is already a precedent for versioning protocol ids across a variety of the protocols
2) The added benefit of future-proofing this protocol.

There is also the argument that we would be adjusting the spec in order to adapt to an implementation mistake, in most cases I would agree here but I think this is a more innocuous change than other situations.

Related:

https://github.com/libp2p/js-libp2p/issues/2238
https://github.com/libp2p/js-libp2p/pull/2239

### DIFF
--- a/webrtc/webrtc.md
+++ b/webrtc/webrtc.md
@@ -21,7 +21,7 @@ On a historical note, this specification replaces the existing [libp2p WebRTC st
 
 1. _B_ advertises support for the WebRTC browser-to-browser protocol by appending `/webrtc` to its relayed multiaddr, meaning it takes the form of `<relayed-multiaddr>/webrtc/p2p/<b-peer-id>`.
 
-2. Upon discovery of _B_'s multiaddress, _A_ learns that _B_ supports the WebRTC transport and knows how to establish a relayed connection to _B_ to run the `/webrtc-signaling` protocol on top.
+2. Upon discovery of _B_'s multiaddress, _A_ learns that _B_ supports the WebRTC transport and knows how to establish a relayed connection to _B_ to run the `/webrtc-signaling/0.0.1` protocol on top.
 
 3. _A_ establishes a relayed connection to _B_.
    Note that further steps depend on the relayed connection to be authenticated, i.e. that data sent on the relayed connection can be trusted.
@@ -68,7 +68,7 @@ It is not necessary for _A_ and _B_ to use the same STUN server when establishin
 
 ## Signaling Protocol
 
-The protocol id is `/webrtc-signaling`.
+The protocol id is `/webrtc-signaling/0.0.1`.
 Messages are sent prefixed with the message length in bytes, encoded as an unsigned variable length integer as defined by the [multiformats unsigned-varint spec][uvarint-spec].
 
 ``` protobuf


sukunrt:
I am fine either way since js-libp2p implementation is already out. 

The versioning we do in protocols doesn't make much sense right now. We can agree to do semantic versioning of sorts on multistream select, but that'd require significant changes. Currently, the version number especially the semantic versioning style(0.0.1) is pointless. multistream will not pick up the highest PATCH or MINOR version magically. So this doesn't really change anything except use a different string for the protocol id.

thomaseizinger:
> I am fine either way since js-libp2p implementation is already out.
> 
> The versioning we do in protocols doesn't make much sense right now. We can agree to do semantic versioning of sorts on multistream select, but that'd require significant changes. Currently, the version number especially the semantic versioning style(0.0.1) is pointless. multistream will not pick up the highest PATCH or MINOR version magically. So this doesn't really change anything except use a different string for the protocol id.

It is not entirely pointless because it allows developers to merge the codepaths and treat the two protocols as equal.

It is unfortunate though that `0.0.1` was chosen because I think it means that even `0.0.2` is not semver-compatible.

I understand that it is annoying to ship a breaking change to users but couldn't you just release it as such (i.e. minor version bump of `js-libp2p`)?

Does npmjs have the ability to yank a release? If yes, then you could yank the old release.

sukunrt:
@thomaseizinger Can you elaborate on?

> It is not entirely pointless because it allows developers that implement the protocol to merge the codepaths and treat the two protocols as equal.

thomaseizinger:
> It is not entirely pointless because it allows developers that implement the protocol to merge the codepaths and treat the two protocols as equal.

Assume the current protocol is versioned with `/1.0.0`. Assume we make a backwards-compatible change to the protocol by adding a new optional field to the payload and name it `/1.1.0`.

If I go ahead and implement "support" for it in `rust-libp2p`, I know from the version string that it is a backwards-compatible change and just accept both protocols (`/1.0.0` and `/1.1.0`) and have them execute the same codepath (modulo whatever feature was introduced by `/1.1.0`) instead of implementing an entire new module.

Arguably though, that is almost useless because the same thing could be done without modifying the protocol string and e.g. detecting the presence of a field in a protobuf struct. That is because protobufs allow for fields to be added in a backwards-compatible way which are a better mechanism for evolving protocols than versioning.

I think what I am trying to say is that the versions in the protocol string can be a _communication_ tool for humans, indicating that the various versions of the protocol are backwards-compatible with each other. On a technical level, there may be better tools like forwards-compatible wire-formats which e.g. don't require multiple round-trips during protocol negotiation.

achingbrain:
> Does npmjs have the ability to yank a release? If yes, then you could yank the old release.

Not after 72 hours - https://docs.npmjs.com/policies/unpublish

On versioning, I like the idea but I'm not sure we've ever made an effort to support semver programatically?

Most protocols only have a single version, the only one that springs to mind that doesn't is bitswap and even then the message format is different even though the version numbers are semver-compatible so we take care to not to send a `1.1.0` message to a `1.0.0` peer.

achingbrain (approved):

mxinden (approved):
Thank you for debugging @SgtPooki and @sukunrt.

Change looks good to me. Fine to merge given that js-libp2p is the only implementation today.

Additional context on the discussion of semver compliance of protocol names: https://github.com/libp2p/specs/issues/203

As mentioned above, I don't think it makes sense to act on the protocol version programmatically, e.g. multistream-select choosing the highest compatible version. I treat the version as a hint for developers, similar to what @thomaseizinger described above.

sukunrt (approved):

maschad:
I think we have consensus here, going ahead with merge.

thomaseizinger:
> It is unfortunate though that `0.0.1` was chosen because I think it means that even `0.0.2` is not semver-compatible.
> 
> I understand that it is annoying to ship a breaking change to users but couldn't you just release it as such (i.e. minor version bump of `js-libp2p`)?

This kind of fell through here. Why is it not an option to address this in `js-libp2p`? Note that `0.0.1` and e.g. `0.0.2` are **not** semver compatible to my understanding.

achingbrain:
> Why is it not an option to address this in js-libp2p? 

The problem is it's an option that would mean shipping a breaking change and forcing everyone to update/redeploy/etc for what seems like not a huge amount of benefit.

The actual version number is annoying but it's far less disruptive to just add the version number to the spec then force everyone to upgrade, given that our protocols tend to have versions and we tend to treat them like opaque identifiers rather than something with any semantic meaning.

thomaseizinger:
> > Why is it not an option to address this in js-libp2p?
> 
> The problem is it's an option that would mean shipping a breaking change and forcing everyone to update/redeploy/etc for what seems like not a huge amount of benefit.
> 
> The actual version number is annoying but it's far less disruptive to just add the version number to the spec then force everyone to upgrade, given that our protocols tend to have versions and we tend to treat them like opaque identifiers rather than something with any semantic meaning.

All things considered, I agree with you.

PR #601, opened 2023-12-18
muxer: correct inline muxer negotiation choice

thewoodfish:
Based on this
```
The multiplexer to use is determined by picking the first item from the initiator's list that both parties support.
```
The chosen multiplexer to use will be `yamux/1.0.0`


thomaseizinger (approved):
Good catch! I am not sure who else wants to review this, perhaps @MarcoPolo ?

MarcoPolo (approved):

PR #606, opened 2024-03-05
webtransport: add message framing to allow graceful stream closing

achingbrain:
When closing streams some implementations [do not guarantee all data has been sent](https://issues.chromium.org/issues/326887753) before closing the stream.

If this sounds familiar it's because we observed [exactly the same behaviour with WebRTC](https://issues.chromium.org/issues/40072842#comment5).

This PR adds opt-in message framing to WebTransport streams that lets us introduce a similar `FIN`/`FIN_ACK` mechanism that guarentees the remote has received all data sent on a stream before we close it.

It uses a Noise extension to signal to the remote that we will be framing all outgoing messages, with a recommendation that the framing is omitted if the remote does not signal the same in response.

This is an attempt to make this a backwards-compatible change which will be a lot less disruptive, even if the status quo is a lot more unsafe.

We're still waiting a definitive answer from the Chromium team as to whether the data loss on closing is a bug or a feature but I thought I'd open this early so we can move quickly if they confirm it's an oversight.

### DIFF
--- a/webtransport/README.md
+++ b/webtransport/README.md
@@ -69,3 +69,39 @@ In order to verify end-to-end encryption of the connection, the peers need to es
 On receipt of the `webtransport_certhashes` extension, the client MUST verify that the certificate hash of the certificate that was used on the connection is contained in the server's list. If the client was willing to accept multiple certificate hashes, but cannot determine which certificate was actually used to establish the connection (this will commonly be the case for browser clients), it MUST verify that all certificate hashes are contained in the server's list. If verification fails, it MUST abort the handshake.
 
 For the client, the libp2p connection is fully established once it has sent the last Noise handshake message. For the server, processing of that message completes the handshake.
+
+## Message framing
+
+Upon closing a WebTransport stream, some implementations do not wait for all outstanding stream data to be sent over the wire before freeing up stream resources and making queue memory available for other uses.
+
+To allow cleanly closing streams without the loss of data using these implementations it's necessary to adopt similar semantics to the WebRTC transport when [closing datachannels](../webrtc/README.md#closing-an-rtcdatachannel).
+
+During the [security handshake](#security-handshake), a `webtransport_message_framing` boolean value may be supplied as part of the Noise extension block.
+
+If present, all messages sent over this connection will framed within a protobuf wrapper with the following structure:
+
+```proto
+syntax = "proto3";
+
+message Message {
+  enum Flag {
+    // The sender will no longer send messages on the stream.
+    FIN = 0;
+    // The sender acknowledges receipt of a FIN
+    FIN_ACK = 1;
+  }
+
+  optional Flag flag = 1;
+  optional bytes message = 2;
+}
+```
+
+If a node sends the `webtransport_message_framing` flag but does not receive one from the remote, it MUST fall back to sending unframed messages.
+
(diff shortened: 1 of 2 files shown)


achingbrain:
Opened as a draft pending a response from the Chromium team.

webtransport/README.md
```diff
@@ -69,3 +69,39 @@ In order to verify end-to-end encryption of the connection, the peers need to es
...
+
+If present, all messages sent over this connection will framed within a protobuf wrapper with the following structure:
+
+```proto
+syntax = "proto3";
+
+message Message {
+  enum Flag {
+    // The sender will no longer send messages on the stream.
+    FIN = 0;
```
achingbrain, line 89: Not sure if this is an issue, but the `flag` field is optional and the `0` value starts closing the stream.

If you were to write a data message following `proto3` semantics which let you omit the `flag` field, then read the message using `proto2` semantics, the `flag` field would get initialised to it's default value of `0` and the receiver would think the sender wanted to close the stream.

We can work around this by having a `NIL` flag as the `0` value, which would be ignored by any receiver when read by a `proto2` or `proto3` decoder.

Or we expect implementations to apply `proto3` semantics because that's what the definition here says.

MarcoPolo:
This must be a bug in the Chromium implementation. The W3C spec says that the `[WebTransportSendStream](https://www.w3.org/TR/webtransport/#webtransportsendstream)` implements `[WritableStream](https://streams.spec.whatwg.org/#writablestream)` which flushes all queued data on `.close()`. If the writer intends to discard queued data they should use `.abort()` instead.

I would prefer this fixed in Chromium instead of changing the spec.

What does the Firefox implementation do?

mxinden:
> What does the Firefox implementation do?

Maybe helpful as an entrypoint:

https://github.com/mozilla/neqo/blob/a56e092cedad14039a8ff1e72155f099dfa489c0/neqo-http3/src/connection.rs#L1095-L1112

achingbrain:
> I would prefer this fixed in Chromium instead of changing the spec.

I completely agree, and it's why this PR is in draft pending the outcome of the [related Chromium issue](https://issues.chromium.org/issues/326887753).

Unfortunately, what we've learnt from the [similar issue with Chromium's WebRTC implementation](https://issues.chromium.org/issues/40072842) is that even if they agree that it's a bug, there's no guarantee it'll be fixed in a timely fashion or at all, so the pragmatic thing to do is to work around it and move on.

The nice thing here (I hope) is that the proposed solution is opt-in and implementations will not frame messages if the remote doesn't agree to do it too, so if this ever gets fixed in Chromium, once enough people have upgraded their browsers we can just stop sending the `webtransport_message_framing` flag and deprecate it in the spec.

> What does the Firefox implementation do?

A good question and one that's been hard to answer observationally.

Support for the `serverCertificateHashes` option has only just made it into a nightly build and doesn't appear to work with either [go](https://github.com/quic-go/webtransport-go/issues/116) or [js](https://github.com/fails-components/webtransport/issues/265) WebTransport servers just yet so some more investigation is needed there.

marten-seemann:
I don’t share your pessimism regarding getting this bug fixed in Chrome. Other than WebRTC, WebTransport is still under active development. Also keep in mind that there will be WebTransport use cases (e.g. MoQ) that will probably surface this bug.

achingbrain:
Well, we'll see. It usually pays to plan for the worst and hope for the best.

PR #618, opened 2024-06-15
chore: The iff to if

elecbug:
The PubSub document's typo is corrected


vyzo (changes requested):
Not a typo -- iff = if and only if.

elecbug:
Oh I understand. I don't alive near English culture. Thanks.

PR #622, opened 2024-07-16
yamux: add error codes on stream reset

sukunrt:
Required for #479 

As I understand, all the implementations discard the data when receiving a frame with the `RST` flag set. 
@MarcoPolo @achingbrain can you confirm this for rust-yamux and js-yamux. I think this is what both the implementations are doing.

We send the error code as a 4byte integer following the header. We cannot use the length field like in a goaway frame, because all the implementations look at the length field and read length bytes of data even with the RST flag set.

### DIFF
--- a/yamux/README.md
+++ b/yamux/README.md
@@ -128,7 +128,7 @@ This does a half-close indicating the sender will send no further data.
 
 Once both sides have closed the connection, the stream is closed.
 
-Alternatively, if an error occurs, the RST flag can be used to hard close a stream immediately.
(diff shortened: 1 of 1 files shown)


sukunrt:
We cannot send an error code in Data frames because the body of Data frames is subject to window flow control. Using the body of the Data frame will require us to wait for resetting a stream when the receive window on the peer is full. 

We can use the Window Update frame to reset the stream and encode the error code in the Length field just like a GoAway does. This has no issues as it's not subject to flow control, and the Length field has no meaning for existing implementations as the stream is Reset on receiving a frame with RST flag set.

MarcoPolo (approved):
Nice 👍 

Let's wait to merge until we have a proper implementation though.

Also before merge:
- [ ] @achingbrain to confirm this is backwards compatible with js-libp2p
- [ ] @jxs / @MarcoPolo to confirm this is backwards compatible with rust-libp2p

PR #623, opened 2024-07-16
error codes: provide error codes on stream reset and connection close

sukunrt:


error-codes/README.md
```diff
@@ -0,0 +1,97 @@
...
+`STOP_SENDING` frame using the `Application Protocol Error Code` field as per
+the frame definitions in the
+[RFC](https://www.rfc-editor.org/rfc/rfc9000.html#name-reset_stream-frames).
+
+For Connection Close, the error code MUST be sent on the CONNECTION_CLOSE frame
+using the Error Code field as defined in the
+[RFC](https://www.rfc-editor.org/rfc/rfc9000.html#section-19.19-6.2.1).
+
+### Yamux
+Yamux has no `STOP_SENDING` frame, so there's no way to signal an error on
```
MarcoPolo, line 64: I wonder if there's an addition we could add to the spec to support this in a backwards compatible way. Not worth doing now, but maybe could be added in the future.

error-codes/README.md
```diff
@@ -0,0 +1,97 @@
+# Error Codes
```
MarcoPolo, line 1: We're missing the usual prologue found in other specs.

MarcoPolo (commented):
Seems reasonable, let's do it!

error-codes/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Introduction
+In the event that a node detects violation of a protocol or is unable to
+complete the necessary steps required for the protocol, it's useful to provide a
+reason for disconnection to the other end. This error code can be sent on both
+Connection Close and Stream Reset. Its purpose is similar to http response
+status. A server, on receiving an invalid request can reset the stream providing
+a `BAD_REQUEST` error code, when it's busy handling too many requests can
```
yiannisbot, line 9:
```suggestion
a `BAD_REQUEST` error code, or when it's busy handling too many requests it can
```

error-codes/README.md
```diff
@@ -0,0 +1,97 @@
...
+provide a `RATE_LIMITED` error code, etc. An error code doesn't always indicate
+an error condition. For example, a connection may be closed prematurely because
+a connection to the same peer on a better transport is available.
```
yiannisbot, line 12:
```suggestion
of a connection to the same peer on a better transport is available. 
```

error-codes/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Semantics
+Error codes are unsigned 32bit integers. The range 0 to 10000 is reserved for
+libp2p errors. Application specific errors can be defined by protocols from
+integers outside of this range. Implementations supporting error codes MUST
+provide the error code provided by the other end to the application.
+
+Error codes provide a best effort guarantee that the error will be propogated to
+the applciation layer. This provides backwards compatibility with older nodes,
+allows smaller implementations and using transports that don't provide a
```
yiannisbot, line 22: Do we mean "less popular" here maybe?
```suggestion
allows less popular implementations that are using transports that don't provide a
```

error-codes/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Connection Close and Stream Reset Error Codes
+Error codes are defined separately for Connection Close and Stream Reset. Stream
+Reset errors are from the range 0 to 5000 and Connection Close errors are from
+the range 5001 to 10000. Having separate errors for Connection Close and stream
+reset requires some overlap between the error code definitions but provides more
+information to the receiver. Receiving a `Bad Request: Connection Closed` error
+on reading from a stream also tells the receiver that no more streams can be
+started on the same connection. Implementations MUST provide the Connection
+Close error code on streams that are reset as a result of remote closing the
```
yiannisbot, line 37:
```suggestion
Close error code on streams that are reset as a result of the remote closing the
```

error-codes/README.md
```diff
@@ -0,0 +1,97 @@
...
+For Connection Close, the error code MUST be sent on the CONNECTION_CLOSE frame
+using the Error Code field as defined in the
+[RFC](https://www.rfc-editor.org/rfc/rfc9000.html#section-19.19-6.2.1).
+
+### Yamux
+Yamux has no `STOP_SENDING` frame, so there's no way to signal an error on
+closing the read side of the stream.
+
+For Connection Close, the 32bit Length field is to be interpreted as the error
+code. The error code MUST be sent as an  Big Endian unsigned 32 bit integer.
```
yiannisbot, line 68:
```suggestion
code. The error code MUST be sent as a Big Endian unsigned 32 bit integer. 
```

error-codes/README.md
```diff
@@ -0,0 +1,97 @@
...
+Note: On TCP connections with `SO_LINGER` set to 0, the error code on connection
+close may not be delivered.  
+
+### WebRTC
+Since WebRTC doesn't provide reliable delivery for frames that are followed by
+closing of the underlying data channel, there's no simple way to introduce error
+codes in the current implementation. Consider the most common use case of
+resetting both read and write sides of the stream and closing the data channel.
+The chrome implementation will not deliver the final `RESET_STREAM` msg and
+while the go implementation will delivery the `RESET_STREAM` frame and then
```
yiannisbot, line 83:
```suggestion
while the go implementation will deliver the `RESET_STREAM` frame and then
```

yiannisbot (approved):
Mostly cosmetic nits. Thanks for putting this together.

error-codes/README.md
```diff
@@ -0,0 +1,97 @@
...
+an error condition. For example, a connection may be closed prematurely because
+a connection to the same peer on a better transport is available. 
+
+## Semantics
+Error codes are unsigned 32bit integers. The range 0 to 10000 is reserved for
+libp2p errors. Application specific errors can be defined by protocols from
+integers outside of this range. Implementations supporting error codes MUST
+provide the error code provided by the other end to the application.
+
+Error codes provide a best effort guarantee that the error will be propogated to
```
achingbrain, line 20: Typo:

```suggestion
Error codes provide a best effort guarantee that the error will be propagated to
```

error-codes/README.md
```diff
@@ -0,0 +1,97 @@
...
+the applciation layer. This provides backwards compatibility with older nodes,
```
achingbrain, line 21: Typo:

```suggestion
the application layer. This provides backwards compatibility with older nodes,
```

error-codes/README.md
achingbrain: I think this reads more naturally as-is.  If you add the `of` you'd also need to move the `is available` bit so it'd be something like:

> ..closed prematurely because of the existence of a connection to the same peer over a better transport.

..which to me reads clunkier than the original since now we have jammed the adjective between two prepositions and the noun is dangling off the end.

That said, `over a better transport` might be better than `on a better transport` but it's the tiniest of tiny nits.

error-codes/README.md
achingbrain: I'm not sure about popularity here - maybe "simpler" instead of "smaller"?

Either way this needs an Oxford comma:

```suggestion
allows smaller implementations, and using transports that don't provide a
```

error-codes/README.md
```diff
@@ -0,0 +1,97 @@
...
+started on the same connection. Implementations MUST provide the Connection
+Close error code on streams that are reset as a result of remote closing the
+connection. 
+
+For stream resets, when the underlying transport supports it, implementations
+SHOULD allow sending an error code on both closing the read side of the stream
+and resetting the write side of the stream. 
+
+## Libp2p Error Codes
+TODO!
```
achingbrain, line 45: Is this section going to contain a list of canonical error codes that map to well defined error conditions?

I think that would be useful.

error-codes/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+Note: On TCP connections with `SO_LINGER` set to 0, the error code on connection
+close may not be delivered.  
+
+### WebRTC
+Since WebRTC doesn't provide reliable delivery for frames that are followed by
+closing of the underlying data channel, there's no simple way to introduce error
+codes in the current implementation. Consider the most common use case of
+resetting both read and write sides of the stream and closing the data channel.
+The chrome implementation will not deliver the final `RESET_STREAM` msg and
```
achingbrain, line 82: Is this a reference to [this Chrome bug](https://issues.chromium.org/issues/40072842)?

It's been fixed so Chrome's datachannel message delivery should be reliable now.

sukunrt: Yes it is. I didn't know chrome fixed it. We can introduce error codes in webrtc then.

error-codes/README.md
```diff
@@ -0,0 +1,97 @@
+# Error Codes
+
+## Introduction
+In the event that a node detects violation of a protocol or is unable to
+complete the necessary steps required for the protocol, it's useful to provide a
+reason for disconnection to the other end. This error code can be sent on both
```
marten-seemann, line 6: It's broader than this. This doesn't only apply to protocol violations (which should be rare), but also to common events like running into resource limits, connections being pruned by the connection manager, etc.

error-codes/README.md
marten-seemann: I made some suggestions for the wire encoding in https://github.com/libp2p/specs/issues/479. You'd have to check if existing implementations handle this gracefully.

marten-seemann (changes requested):
You probably want to create a registry, so application protocols can register ranges of values. Otherwise, it's likely that you'll end up with collisions.

error-codes/README.md
sukunrt: @marten-seemann do you mean suggestions for sending an error code? There's this PR for yamux for that https://github.com/libp2p/specs/pull/622

sukunrt: @MarcoPolo I think we can add this, there are 16 flag bits, so we do have space for a new flag and implementations only look at the flag bits they care about. I created https://github.com/libp2p/specs/issues/627

error-codes/README.md
```diff
@@ -0,0 +1,97 @@
...
+[RFC](https://www.rfc-editor.org/rfc/rfc9000.html#section-19.19-6.2.1).
+
+### Yamux
+Yamux has no `STOP_SENDING` frame, so there's no way to signal an error on
+closing the read side of the stream.
+
+For Connection Close, the 32bit Length field is to be interpreted as the error
+code. The error code MUST be sent as an  Big Endian unsigned 32 bit integer. 
+
+For Stream Resets, the error code is sent as the first 4 bytes of the Data Frame
```
MarcoPolo, line 70: Linking this here since this will change: https://github.com/libp2p/specs/pull/622#issuecomment-2314399167

error-codes/README.md
sukunrt: Do we need this?

error-codes/README.md
sukunrt: I've rephrased this and removed the specific reference to protocol errors.

sukunrt:
@marten-seemann 
>You probably want to create a registry, so application protocols can register ranges of values. Otherwise, it's likely that you'll end up with collisions.

Do we need to? Application protocols can have conflicting error codes. From the application context it's clear which is the relevant error. I am planning to reserve a space for libp2p to use for its errors and let applications do whatever they like with the application error codes space. 
This is similar to how QUIC does it. 
For example, 
DNS over QUIC: https://www.rfc-editor.org/rfc/rfc9250.html#transaction-cancellation
RTP over QUIC: https://www.ietf.org/archive/id/draft-ietf-avtcore-rtp-over-quic-11.html#name-error-handling
have conflicting error codes.

marten-seemann:
> This is similar to how QUIC does it. For example,
> DNS over QUIC: https://www.rfc-editor.org/rfc/rfc9250.html
> RTP over QUIC: https://www.ietf.org/archive/id/draft-ietf-avtcore-rtp-over-quic-11.html
> have conflicting error codes.

They're not conflicting, since the application protocol is negotiated during the QUIC handshake.

> From the application context, it's clear which is the relevant error.

This is not correct, thanks to multistream.

sukunrt:
I see. All connections are just libp2p connections and they can speak multiple application protocols any of which may close the underlying connection on error.

error-codes/README.md
sukunrt: We can discuss this in #627.

error-codes/README.md
```diff
@@ -0,0 +1,105 @@
...
+error codes over WebTransport later.
+
+### HTTP
+Protocols that work over http MUST use the response header `Libp2p-Error-Code` to send the error code. The grammar for the field is similar to `Content-Length`
+```
+Libp2p-Error-Code: 1*DIGIT
+```
+
+### Multistream Select
+Multistream-Select is used to negotiate Security protocol for TCP connections before a stream muxer has been selected. There's only one error code defined for such cases, `PROTOCOL_NEGOTIATION_FAILED`. To encode this error, send the string `101` prefixed with the length and close the TCP connection.
```
sukunrt, line 102: This is not too useful. The most common case here is a server rejecting a connection because the handling it exceeds some resource limit. In such cases Writing on the connection and closing will cause the write to be dropped because almost certainly there will be unread data in the read buffer.

error-codes/README.md
```diff
@@ -0,0 +1,105 @@
...
+Implementations MUST provide the Connection Close error code on streams that are reset as a result of remote closing the connection. 
+
+Libp2p streams are reset unilaterally, calling `Reset` on a stream resets both the read and write end of a stream. For transports, like QUIC, which support cancelling the read and write ends of the stream separately, implementations MAY provide the ability to signal error codes separately on resetting either end. 
+
+## Error Codes Registry
+Libp2p connections are shared by multiple applications. The same connection used in the dht may be used for gossip sub, or for any other application. Any of these applications can close the underlying connection on an error, resetting streams used by the other applications. To correctly distinguish which application closed the connection, Connection Close error codes are allocated to applications from a central registry. 
+
+For simplicity, we manage both Connection Close and Stream Reset error codes from a central registry. The libp2p error codes registry is at: https://github.com/libp2p/error-codes/
```
MarcoPolo, line 38: Is there much value in having this be a separate repo? Why not have it in the specs repo in a sub folder of error-codes?

error-codes/README.md
```diff
@@ -0,0 +1,105 @@
...
+For Connection Close, the 32-bit Length field is interpreted as the error code.
+
+For Stream Resets, the error code is sent in the `Window Update` frame, with the
+32-bit Length field interpreted as the error code. See [yamux spec
+extension](https://github.com/libp2p/specs/pull/622).
+
+Connection Close error code delivery to the other end depends on the OS TCP implementation and the TCP options used for the socket. In particular, when `SO_LINGER` TCP option is set to 0 and the implementation closes the connection immediately after writing the error code containing frame, the error code may not be delivered.
+
+### WebRTC
+There is no way to provide any information on closing a peer connection in
```
MarcoPolo, line 77: The wording here is confusing.

error-codes/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Connection Close and Stream Reset Error Codes
+Error codes are defined separately for Connection Close and Stream Reset. The namespace doesn't overlap as it is clear from the context whether the stream was reset by the other end, or it was reset as a result of a connection close. 
+Implementations MUST provide the Connection Close error code on streams that are reset as a result of remote closing the connection. 
+
+Libp2p streams are reset unilaterally, calling `Reset` on a stream resets both the read and write end of a stream. For transports, like QUIC, which support cancelling the read and write ends of the stream separately, implementations MAY provide the ability to signal error codes separately on resetting either end. 
+
+## Error Codes Registry
+Libp2p connections are shared by multiple applications. The same connection used in the dht may be used for gossip sub, or for any other application. Any of these applications can close the underlying connection on an error, resetting streams used by the other applications. To correctly distinguish which application closed the connection, Connection Close error codes are allocated to applications from a central registry.
```
MarcoPolo, line 36: We should outline a process to allocate these error codes. e.g. something as simple as:

- A user requests a certain number of error codes.
- We ensure it's a reasonable request.
  - Verify the number is reasonable (e.g. not 16k errors)
  - Verify there is no conflict.
- Allocate that in a high number space.

The process exists so that we can make it easier for future reviewers.

error-codes/README.md
sukunrt: It'll require reviewing PRs for new allocations here. We can start here and then move to a new repo if needed.

error-codes/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Error Codes Registry
+Libp2p connections are shared by multiple applications. The same connection used in the dht may be used for gossip sub, or for any other application. Any of these applications can close the underlying connection on an error, resetting streams used by the other applications. To correctly distinguish which application closed the connection, Connection Close error codes are allocated to applications from a central registry. 
+
+For simplicity, we manage both Connection Close and Stream Reset error codes from a central registry. The libp2p error codes registry is at: https://github.com/libp2p/error-codes/
+
+### Libp2p Reserved Error Codes
+Error code 0 signals that no error code was provided. Implementations MUST handle closing a connection with error code 0 as they handle closing a connection with no error code, and resetting a stream with error code 0 as they handle resetting a stream with no error. 
+
+Error codes from 1 to 100 are reserved for transport errors. These are used by the transports to terminate connections on transport errors.
```
MarcoPolo, line 43: Can we make these ranges that fit neatly into a bit mask? It'll look and feel nicer to say transport errors are error codes between 0x01 and 0x7f. Same for the libp2p reserved error codes.

error-codes/README.md
sukunrt: updated.

error-codes/README.md
MarcoPolo: To be clear, does this end with a new line or no newline?

This also sounds like it would require spec changes for multistream select (similar to the wording here).

With the point you bring up, maybe it's not worth adding error codes to MSS right now.

error-codes/README.md
sukunrt: Done.

PR #624, opened 2024-08-13
docs: add WebSockets spec

### LINKED ISSUE #594: Missing WebSockets (/ws, /wss) Transport Specification
This is a placeholder issue, as we seem to be missing specification for `/ws` and `/wss`.

Which codebase/package would be the best "source of truth" to start documenting things?
@mxinden @marten-seemann @sukunrt

achingbrain:
Specifies the WebSockets transport with notes on certificates, multiplexing, etc.

It tries to document the current state of WebSocket transport with the exception of the `http-path` which I don't think is implemented anywhere.

A follow up PR to this PR will attempt to address the double-encryption problem as it's worthy of it's own discussion.

Please suggest an edit to the Interest Group if you want to be added.

Fixes #594

### DIFF
--- a/websockets/README.md
+++ b/websockets/README.md
@@ -1,0 +1,69 @@
+# libp2p WebSockets
+
+| Lifecycle Stage | Maturity                 | Status | Latest Revision |
+|-----------------|--------------------------|--------|-----------------|
+| 2A              | Candidate Recommendation | Active | r0, 2022-10-12  |
+
+Authors: [@achingbrain]
+
+Interest Group: [@MarcoPolo]
+
+[@achingbrain]: https://github.com/achingbrain
+[@MarcoPolo]: https://github.com/MarcoPolo
+
+See the [lifecycle document](../00-framework-01-spec-lifecycle.md) for context about maturity level
+and spec status.
+
+## Introduction
+
+[WebSockets](https://websockets.spec.whatwg.org/) are a way for web applications to maintain bidirectional communications with server-side processes.
+
+All major browsers have shipped WebSocket support and the implementations are both robust and well understood.
+
+A WebSocket request starts as a regular HTTP request, which is renegotiated as a WebSocket connection using the [HTTP protocol upgrade mechanism](https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism).
+
+## Drawbacks
+
+WebSockets suffer from [head of line blocking](https://en.wikipedia.org/wiki/Head-of-line_blocking) and provide no mechanism for stream multiplexing, encryption or authentication so additional features must be added by the developer or by libp2p.
+
+In practice they only run over TCP so are less effective with [DCuTR Holepunching](../relay/DCUtR.md).
+
+## Certificates
+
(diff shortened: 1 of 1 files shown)


achingbrain:
One improvement we could make here is to use the [protocols](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket#protocols) arg to specify the connection encrypter (e.g. `"noise"`), then use [noise early data](https://github.com/libp2p/specs/blob/master/connections/inlined-muxer-negotiation.md#multiplexer-negotiation-over-noise) to select a stream multiplexer?

This would save a bunch of multistream-select round trips.

websockets/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+Examples:
+
+* `/ip4/192.0.2.0/tcp/1234/ws` (an insecure address with a TCP port)
+* `/ip4/192.0.2.0/tcp/1234/tls/ws` (a secure address with a TCP port)
+* `/ip4/192.0.2.0/ws` (an insecure address that defaults to TCP port 80)
+* `/ip4/192.0.2.0/tls/ws` (a secure address that defaults to TCP port 443)
+* `/ip4/192.0.2.0/wss` (`/tls` may be omitted when using `/wss`)
+* `/dns/example.com/wss` (a DNS address)
+* `/dns/example.com/wss/http-path/path%2Fto%2Fendpoint` (an address with a path)
```
SgtPooki, line 69: We should call out whether WS/WSS/TCP can all listen at the same port. This may be a WS/TCP-specific thing that others already know, but this has come up several times.

SgtPooki (commented):
Reads fine to me.. one comment

websockets/README.md
achingbrain: This doesn't feel like a natural place for something like that to me.

Specs are all `WS/WSS/TCP MUST listen on the same port` or `WS/WSS/TCP SHOULD listen on the same port` - this sounds like a `WS/WSS/TCP MAY listen on the same port` which kind of makes it feel like an implementation detail?

I'd imagine not every TCP socket/web framework would support this either?

SgtPooki: yea, that makes sense. my bad!

websockets/README.md
```diff
@@ -0,0 +1,69 @@
+# libp2p WebSockets
+
+| Lifecycle Stage | Maturity                 | Status | Latest Revision |
+|-----------------|--------------------------|--------|-----------------|
+| 2A              | Candidate Recommendation | Active | r0, 2022-10-12  |
```
p-shahi, line 5:
```suggestion
| 3A              | Recommendation | Active | r0, 2022-10-12  |
```

Should this be a Recommendation instead of candidate? Given that more than there more than 2 reference implementations of this spec?

achingbrain: Sounds good, have merged the suggestion

achingbrain:
It's been a couple of months now so If there are no further comments here, I'm going to merge this at the end of the week.

websockets/README.md
```diff
@@ -0,0 +1,69 @@
+# libp2p WebSockets
+
+| Lifecycle Stage | Maturity                 | Status | Latest Revision |
+|-----------------|--------------------------|--------|-----------------|
+| 3A              | Recommendation | Active | r0, 2024-09-10  |
```
achingbrain, line 5:
```suggestion
| 3A              | Recommendation           | Active | r0, 2024-10-23  |
```

PR #628, opened 2024-09-02
Update WebRTC message size limit

achingbrain:
The browser bugs that made a 16KiB message size limit necessary have long been fixed.

Increase the max size to 256KiB which has been verified experimentally between Firefox/Chrome and update the link to the Mozilla blog entry that discusses the limit and the relevant WebRTC issue that enables 256KiB messages.

### DIFF
--- a/webrtc/README.md
+++ b/webrtc/README.md
@@ -88,9 +88,9 @@ QUIC. The libp2p WebRTC message framing is not concerned with flow-control and
 thus does not need the `RESET_STREAM` frame to be send in reply to a
 `STOP_SENDING` frame.
 
-Encoded messages including their length prefix MUST NOT exceed 16kiB to support
-all major browsers. See ["Understanding message size
-limits"](https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Using_data_channels#understanding_message_size_limits).
+Encoded messages including their length prefix MUST NOT exceed 256kiB to support
+all major browsers. See ["Large Data Channel Messages"](https://blog.mozilla.org/webrtc/large-data-channel-messages/)
+and [WebRTC#40644524](https://issues.webrtc.org/issues/40644524).
 Implementations MAY choose to send smaller messages, e.g. to reduce delays
 sending _flagged_ messages.
 


webrtc/README.md
```diff
@@ -88,9 +88,9 @@ QUIC. The libp2p WebRTC message framing is not concerned with flow-control and
...
 `STOP_SENDING` frame.
 
-Encoded messages including their length prefix MUST NOT exceed 16kiB to support
-all major browsers. See ["Understanding message size
-limits"](https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Using_data_channels#understanding_message_size_limits).
+Encoded messages including their length prefix MUST NOT exceed 256kiB to support
+all major browsers. See ["Large Data Channel Messages"](https://blog.mozilla.org/webrtc/large-data-channel-messages/)
+and [WebRTC#40644524](https://issues.webrtc.org/issues/40644524).
 Implementations MAY choose to send smaller messages, e.g. to reduce delays
 sending _flagged_ messages.
```
achingbrain, line 95: Should we account for forwards compatibility with future implementations that aren't arbitrarily limited like browsers currently?

```suggestion

Implementations MAY choose to send smaller messages, e.g. to reduce delays
sending _flagged_ messages, and they MAY choose to accept larger messages
to allow for forwards compatibility with more capable implementations in the
future.
```

A "be liberal in what you accept, conservative in what you emit" sort of thing.

webrtc/README.md
```diff
@@ -88,9 +88,9 @@ QUIC. The libp2p WebRTC message framing is not concerned with flow-control and
 thus does not need the `RESET_STREAM` frame to be send in reply to a
 `STOP_SENDING` frame.
 
-Encoded messages including their length prefix MUST NOT exceed 16kiB to support
-all major browsers. See ["Understanding message size
-limits"](https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Using_data_channels#understanding_message_size_limits).
+Encoded messages including their length prefix MUST NOT exceed 256kiB to support
+all major browsers. See ["Large Data Channel Messages"](https://blog.mozilla.org/webrtc/large-data-channel-messages/)
+and [WebRTC#40644524](https://issues.webrtc.org/issues/40644524).
```
lidel, line 93: I believe we've decided to keep it at 16KiB and we need to figure out how to negotiate higher size limit per connection? (https://developer.mozilla.org/en-US/docs/Web/API/RTCSctpTransport/maxMessageSize) ?

lidel (changes requested):

webrtc/README.md
achingbrain: I’m doing some experiments in the test plans repo to validate that we do in fact see a throughput increase with larger messages.

Marking this as a draft pending the outcome of that.

2color:
I'm curious, what was the conclusion with this PR and the potential throughput increase?

achingbrain:
The result was a roughly 15-25% speed increase when going from 16KB to 265KB chunks.

https://observablehq.com/@libp2p-workspace/performance-dashboard?branch=feat%2Fadd-circuit-relay#branch

I think this is still worth doing but we need the go/rust implementations to accept larger chunks - I think rust-libp2p will still reject them.

p-shahi:
Here's the draft pr @sukunrt created for go-libp2p: https://github.com/libp2p/go-libp2p/pull/2949
The rust-libp2p pr is waiting for the spec to be merged before shipping: https://github.com/libp2p/rust-libp2p/pull/5589

Is everyone agreed that we want to increase the message size? If so, we can proceed with merging the go & rust pr
@achingbrain @sukunrt

achingbrain:
@p-shahi I think we should increase the message size. WebRTC is by far the slowest libp2p transport but also one of the most useful since it lets browsers dial each other directly, so anything that speeds it up is very welcome.

webrtc/README.md
achingbrain: See linked performance graphs here: https://github.com/libp2p/specs/pull/628#issuecomment-2451491070

sukunrt (commented):
I think there's a way to make this backwards compatible for `/webrtc-direct` using the datachannel 0 that we use for handshaking. 

1. Both server and client use SDPs that indicate that peer supports 256kB messages. 

2. sever and client use the noise handshake early data extension to send the max message size. 

3. Depending on the max message size received in the handshake we the server and client decide whether to actually send 256kB messages or 16 kB messages.

PR #630, opened 2024-09-05
feat: add signed peer record to identify spec

2color:


identify/README.md
```diff
@@ -133,3 +134,10 @@ clients only support initiating requests while some servers (only) support
 responding to requests. To prevent clients from initiating requests to other
 clients, which given them being clients they fail to respond, clients should not
 advertise `foo` in their `protocols` list.
+
+### signedPeerRecord
+
+This is a serialized [SignedEnvelope](https://github.com/libp2p/go-libp2p/blob/master/core/record/pb/envelope.proto) containing a [PeerRecord](https://github.com/libp2p/go-libp2p/blob/master/core/peer/pb/peer_record.proto),
```
achingbrain, line 140: Can this link to specs rather than implementations?  If it must link to implementations, it should use the [permalink](https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files) format that has the commit hash embedded so future refactorings don't break links

e.g:

```
https://github.com/libp2p/go-libp2p/blob/d55bed5f78fad923353b92c7d1db64c758bcbf9c/core/record/pb/envelope.proto
```

identify/README.md
```diff
@@ -133,3 +134,10 @@ clients only support initiating requests while some servers (only) support
...
+signed by the sending node. It contains the same addresses as the `listenAddrs` field, but in a form that lets us share authenticated addrs with other peers.
+
+This field was introduced in a backwards compatible manner (meaning that it is sent along with the `listenAddrs` field), therefore, it is optional and may be omitted by older implementations.
```
achingbrain, line 143: I think we should add a paragraph like:

> If the signed peer record is present, implementations MUST use the data contained within it and ignore duplicated fields in the main identify message.

2color: Besides `listenAddrs`, are there are any other fields that may be present in the signed envelope?

achingbrain: Not right now but the [routing records](https://github.com/libp2p/specs/blob/master/RFC/0003-routing-records.md) calls out it's future extensibility, for example the list of supported protocols.

identify/README.md
```diff
@@ -81,6 +81,7 @@ message Identify {
   repeated bytes listenAddrs = 2;
   optional bytes observedAddr = 4;
   repeated string protocols = 3;
+  optional bytes signedPeerRecord = 8;
```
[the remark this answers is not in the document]
2color: I don't know. But it's consistent. And there's no hard requirement for [monotonically increasing field numbers according to the protobuf docs](https://protobuf.dev/programming-guides/proto3/#assigning)

It's the same for the protobuf definitions in js and go-libp2p:
js-libp2p: https://github.com/libp2p/js-libp2p/blob/5214dec4a0b7e7cb82056b9a681f1c77e82d34a2/packages/protocol-identify/src/pb/message.proto#L29
go-libp2p https://github.com/libp2p/go-libp2p/blob/d55bed5f78fad923353b92c7d1db64c758bcbf9c/p2p/protocol/identify/pb/identify.proto#L37

Interestingly, rust-libp2p doesn't seem to implement `signedPeerRecord`:
https://github.com/libp2p/rust-libp2p/blob/3837e33cd4c40ae703138e6aed6f6c9d52928a80/protocols/identify/src/generated/structs.proto

identify/README.md
```diff
@@ -133,3 +134,14 @@ clients only support initiating requests while some servers (only) support
...
+### signedPeerRecord
+
+This is a serialized [SignedEnvelope][envelope-rfc] containing a [PeerRecord][peer-record-rfc],
+signed by the sending node. It contains the same addresses as the `listenAddrs` field, but in a form that lets us share authenticated addrs with other peers.
+
+This field was introduced in a backwards compatible manner (meaning that it is sent along with the `listenAddrs` field), therefore, it is optional and may be omitted by older implementations. If the `signedPeerRecord` is present, implementations MUST use the data contained within it and ignore duplicated fields present in the main identify message
+
+
+[envelope-rfc]: ../RFC/0002-signed-envelopes.md#wire-format
+[peer-record-rfc]: ../RFC/0003-routing-records.md#address-record-format
```
sukunrt, line 147: I'd prefer to make this a separate spec `peer-record` spec. There we should add peer-record bits from https://github.com/libp2p/specs/blob/master/RFC/0003-routing-records.md and signed peer record bits from https://github.com/libp2p/specs/blob/master/RFC/0002-signed-envelopes.md

We can then reference it from places like: https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md

sukunrt (changes requested):

identify/README.md
```diff
@@ -133,3 +134,14 @@ clients only support initiating requests while some servers (only) support
 responding to requests. To prevent clients from initiating requests to other
 clients, which given them being clients they fail to respond, clients should not
 advertise `foo` in their `protocols` list.
+
+### signedPeerRecord
+
+This is a serialized [SignedEnvelope][envelope-rfc] containing a [PeerRecord][peer-record-rfc],
+signed by the sending node. It contains the same addresses as the `listenAddrs` field, but in a form that lets us share authenticated addrs with other peers.
+
+This field was introduced in a backwards compatible manner (meaning that it is sent along with the `listenAddrs` field), therefore, it is optional and may be omitted by older implementations. If the `signedPeerRecord` is present, implementations MUST use the data contained within it and ignore duplicated fields present in the main identify message
```
sukunrt, line 143:
```suggestion
This field was introduced in a backwards compatible manner (meaning that it is sent along with the `listenAddrs` field), therefore, it is optional and may be omitted by older implementations. If the `signedPeerRecord` is present, implementations SHOULD use the data contained within it and ignore duplicated fields present in the main identify message
```

If a particular implementation is keen on forcing the peer to use addresses from signed peer record instead of `listenAddrs` field it can leave the `listenAddrs` field blank.

identify/README.md
2color: Not sure I understand. What do you think should be the scope of the separate spec?

sukunrt: peer record definition and signed peer record.

2color: Is there anything missing in the two RFCs? Or do they just need to be ratified into a spec?

sukunrt: I think the second one. I'm not sure why we ever did RFCs. @MarcoPolo thoughts? 

While I dislike the fact that those 3 documents are RFCs and everything else in the specs is not, the real problem is that those documents are very dated:
Consider: https://github.com/libp2p/specs/blob/master/RFC/0003-routing-records.md#address-record-format

>A peer SHOULD only include addresses that it believes are routable via the public internet, ideally having confirmed that this is the case via some external mechanism such as a successful AutoNAT dial-back.

I see no reason why we should do this. And go-libp2p doesn't. Depends on what you're using them for.

There's some information that we don't need, like this discussion on Routing State

>To produce a "self-certified" address, a peer will construct a RoutingState containing their listen addresses and serialize it to a byte array using a protobuf encoder. The serialized records will then be wrapped in a [signed envelope](https://github.com/libp2p/specs/blob/master/RFC/0002-signed-envelopes.md), which is signed with the libp2p peer's private host key. The corresponding public key MUST be included in the envelope's public_key field.

What is a `RoutingState`? 

or the go-libp2p API suggestion elsewhere in RFC-0003. 

Most importantly,

it doesn't mention the most important field for a signed peer record, the domain string is `libp2p-peer-record`.

achingbrain: It would be very useful to have a canonical reference that defines a peer record.  The RFCs read like a work-in-progress which isn't helpful for implementers.

identify/README.md
achingbrain: The point is what to do if the addresses in `listenAddrs` and the signed peer record are different.

Do you merge them or favour one over the other?  If we specify `MUST` it makes the behaviour unequivocal.

achingbrain: > On a second reading, I realised that if it's a MUST it would mean that rust-libp2p is technically not following the spec.

Sort of?  It's just following an earlier version of the spec.

I think this is fine, see also #502 which adds a new optional field with a `MUST` in the description - this would also make all current implementations non-compliant.

identify/README.md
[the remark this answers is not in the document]
2color: For future reference, it was https://github.com/libp2p/specs/pull/502/files that introduced field number 7.

identify/README.md
```diff
@@ -133,3 +134,14 @@ clients only support initiating requests while some servers (only) support
 responding to requests. To prevent clients from initiating requests to other
 clients, which given them being clients they fail to respond, clients should not
 advertise `foo` in their `protocols` list.
+
+### signedPeerRecord
+
+This is a serialized [SignedEnvelope][envelope-rfc] containing a [PeerRecord][peer-record-rfc],
+signed by the sending node. It contains the same addresses as the `listenAddrs` field, but in a form that lets us share authenticated addrs with other peers.
+
+This field was introduced in a backwards compatible manner (meaning that it is sent along with the `listenAddrs` field), therefore, it is optional and may be omitted by older implementations. If the `signedPeerRecord` is present and the implementation supports SignedEnvelope, it MUST use the data contained within it and ignore duplicated fields present in the main identify message
```
achingbrain, line 143: Ah sorry - I [meant](https://github.com/libp2p/specs/pull/630#discussion_r1753553493) following an earlier version of the spec was fine.

Suggested edit:

```suggestion
This field was introduced in a backwards compatible manner (meaning that it is sent along with the `listenAddrs` field), therefore it is optional. If the `signedPeerRecord` is present the implementation MUST use the data contained within it and ignore duplicated fields present in the main identify message
```

- The field is optional so it feels redundant to explicitly call out that it can be omitted
- In this version of the spec acting on the field if it is present is not optional, that's the point of the `MUST`
- Don't forget to rev the `Latest Revision` field in the header..

PR #633, opened 2024-09-07
Create LICENSE (MIT)

### LINKED ISSUE #632: What is the license for the project?
Hello,

what is the license for the specifications?  I noticed that there's no `COPYING` or 'LICENSE` file in the repository, not even a mention of  the license in `README.md`.

Maybe you could explicitly use some kind of Creative Commons license, or maybe GNU FDL if it is possible?

Thanks!

-avp

MarcoPolo:
closes #632

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2024 libp2p
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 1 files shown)


vyzo (approved):

vyzo (changes requested):
You know, lets discuss the license with the community.

MIT is not appropriate for specifications i think.
Creative Commons or a license similar to what the IETF does would be more appropriate

marten-seemann:
Can you just impose a license on contributions after the fact?

People have already made [license-free](https://en.wikipedia.org/wiki/License-free_software) contributions for close to a decade now. ianal, but I'd be very surprised if you could put all of their work under a license without first getting explicit consent from each and every contributor.

vyzo:
Closing this, @marten-seemann is right, we can't do that without explicit permission from contributors.

MarcoPolo:
You’re right. Let’s discuss on the issue

PR #644, opened 2024-11-21
docs: clarify FIND_NODE response for DHT-client nodes

achingbrain:
If a node receives a `FIND_NODE` RPC call with a peer ID, it should return only closer DHT servers, unless it knows the peer ID in question, in which case it should include it, even if it is not a DHT server.

### DIFF
--- a/kad-dht/README.md
+++ b/kad-dht/README.md
@@ -490,7 +490,9 @@ These are the requirements for each `MessageType`:
 
 * `FIND_NODE`: In the request `key` must be set to the binary `PeerId` of the
   node to be found. In the response `closerPeers` is set to the `k` closest
-  `Peer`s.
+  `Peer`s.  Each closer peer will be a DHT server, unless `key` is an exact for
+  a peer known to the node, in which case the closer peers will include the
+  matching, non-server peer.
 
 * `GET_VALUE`: In the request `key` is an unstructured array of bytes. `record`
   is set to the value for the given key (if found in the datastore) and


kad-dht/README.md
```diff
@@ -490,7 +490,9 @@ These are the requirements for each `MessageType`:
 
 * `FIND_NODE`: In the request `key` must be set to the binary `PeerId` of the
   node to be found. In the response `closerPeers` is set to the `k` closest
-  `Peer`s.
+  `Peer`s.  Each closer peer will be a DHT server, unless `key` is an exact match
+  for a peer known to the node, in which case the closer peers will include the
+  matching peer, even if it is not a DHT server.
```
achingbrain, line 495: Should other close peers be included even if the node knows the peer being searched for?

```suggestion
  for a peer known to the node, in which case the closer peers will only include the
  matching peer, even if it is not a DHT server.
```

guillaumemichel:
There is so far no mention that DHT clients must be routable (in the case they are looked for), and whether DHT clients should be known by (the closest) DHT servers. This addition isn't required for clients to function correctly in the case clients only request to the network, and don't use `ADD_PROVIDER`. So I am not sure it would be relevant to add this to the spec.

However, since [go-libp2p-kad-dht](https://github.com/libp2p/go-libp2p-kad-dht) is doing it, and it may be expected from all nodes in the Amino DHT, it would be good to spec this somewhere, along with how DHT clients routing is expected to work. We may have to create an Amino DHT spec at some point to add specific DHT protocol improvements such as the ones described [here](https://github.com/libp2p/go-libp2p-kad-dht/blob/master/optimizations.md).

achingbrain:
> There is so far no mention that DHT clients must be routable

The spec makes explicit reference to publicly and non-publicly routable nodes in the [Client and server mode section](https://github.com/libp2p/specs/tree/master/kad-dht#client-and-server-mode), all this PR does is try to clarify when they are or are not returned.  Unless you mean something different?

kad-dht/README.md
guillaumemichel: I would say that including the `k` closest peers in important, even for the `FIND_NODE` RPC when the target is known since `FIND_NODE` can also be used to find the closest peers to a given peer.

Eg. when refreshing the routing table, nodes look for the closest nodes to themselves. If the DHT servers only respond with the requester's address, it will be harder for the requester to discover its closest peers.

guillaumemichel:
I meant that the spec describes how DHT servers should keep track of each other ([routing table](https://github.com/libp2p/specs/tree/master/kad-dht#kademlia-routing-table)), but not how DHT clients should be tracked. Let's assume:
* `C` is a DHT Client
* `S` is the DHT Server that is the closest to `C`
* There is no guarantee that `S` knows about `C`, because DHT servers aren't expected to store DHT Client addresses

Returning a DHT Client's addresses along with the closest DHT server peers is a hack to make DHT Clients routable/discoverable, but there is no guarantee that it will work. Amino is making use of this hack, but other libp2p networks don't necessarily need to have their DHT Clients routable/discoverable.

PR #646, opened 2024-12-02
docs(dht): Update default concurrency parameter

cokemine:
In this [PR](https://github.com/libp2p/go-libp2p-kad-dht/pull/605), the default value of the concurrency parameter is updated


guillaumemichel (approved):
LGTM, though I am not sure whether this constant should be spec'ed in the first place. Implementations with different alpha concurrency parameters can coexist in the same network. I would argue this constant is an implementation detail.

We can also state a recommended number in the spec, even though it isn't required.

PR #652, opened 2024-12-13
Announcesub spec: pubsub protocol with no duplicates

ppopth:
**DEPRECATED**

Announcesub is a new pubsub protocol over libp2p, heavily based on gossipsub.

The protocol is designed in a way to reduce the number of duplicates to zero, trading off with more latency.

It is actually the same as Gossipsub except that instead of sending the messages right away to mesh peers, it sends IANNOUNCE instead and waits for INEED before sending the actual messages.

This guarantees that each node will receives exactly one copy of messages.

---

Authored by: @ppopth, @nisdas, @chirag-parmar

I made this PR as a draft first just to gain some visibility. We need to do simulations and further analysis to compare it with Gossipsub.

Our next step would be to implement it in go-libp2p-pubsub and do simulations.

### DIFF
--- a/pubsub/announcesub/announcesub-v1.0.md
+++ b/pubsub/announcesub/announcesub-v1.0.md
@@ -1,0 +1,356 @@
+# announcesub v1.0: pubsub protocol with no duplicates
+
+| Lifecycle Stage | Maturity       | Status | Latest Revision |
+|-----------------|----------------|--------|-----------------|
+| 1A              | Working Draft  | Active | r0, 2024-12-04  |
+
+Authors: [@ppopth], [@nisdas], [@chirag-parmar]
+
+[@ppopth]: https://github.com/ppopth
+[@nisdas]: https://github.com/nisdas
+[@chirag-parmar]: https://github.com/chirag-parmar
+
+See the [lifecycle document][lifecycle-spec] for context about the maturity level
+and spec status.
+
+[lifecycle-spec]: https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md
+
+---
+
+## Overview
+
+This is the specification for a new pubsub protocol over libp2p, heavily based on gossipsub. The protocol is designed in a way to reduce the number of duplicates to zero, trading off with more latency.
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Background](#background)
+  - [Problem With Gossipsub](#problem-with-gossipsub)
+- [Enter Announcesub](#enter-announcesub)
+- [Dependencies](#dependencies)
+- [Parameters](#parameters)
+- [Router State](#router-state)
+  - [Peering State](#peering-state)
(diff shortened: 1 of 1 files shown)


vyzo:
I think it might make more sense to propose this as a gossipsub extension and not an entorely new protocol.

One way it could work is to have two levels of mesh, one for eager propagation and one for announce based lazy propagation.

Set the eager degree to 0, and you get pure announcesub, set the lazy degree to 0, and you get current gossipsub.

nisdas:
@vyzo Initially it was supposed to be an extension to gossipsub but we figured that because it was a big change to gossipsub having it as an entirely new protocol would be better. 

> One way it could work is to have two levels of mesh, one for eager propagation and one for announce based lazy propagation.

Yeah, this was something we discussed where there would be a hybrid approach of having both fast and slow links to your mesh rather than a fully announcement based protocol. We can take a look on how to make it an extension rather a new protocol.

vyzo:
The thing is, this is really where we have been going for gossipaub v2.0; the interplay between eager and lazy.

Having a tunable parameter that can adjust the bandwidth/latency profile would be a big win imo. And on the same time we can further explore dynamic bandwidth tuning schemes that allow a peer (or group of peers and so on) to dynamically adjust (similar to the episub idea).

So really, announcesub seems like a very natural extension of gossipsub.
So I would very  much like to see this a gossipsub v2.0 than a separate protocol. This would also solve deployment issues, by maintaining backwards compatibility.

ppopth:
Alright, I'm writing the spec for gossipsub v2.0

ppopth:
> this will help a lot with deployment

For Ethereum, we have hard forks anyway which require all the clients to update their software, how about other applications?

vyzo:
mixed environments are quite common in non blockchain environments i beleieve.

raulk:
Taking a step back, I think there's opportunity to have gossipsub v2 embrace modularity and extensibility head-on by turning it into a _framework_.

The _system space_ of gossipsub v2 would be in charge of maintaining one or many _named_ meshes, discovering peers through active or passive mechanisms, tracking metadata and metrics (which would be fed into the scoring engine), tracking subscriptions, etc.

The _user space_ would create and drive mesh instances, implement specialized dissemination policies, evaluate peers as they're being offered from the system, drive scoring based on system input, exchange specialised control messages, adjust system parameters dynamically, etc.

Gossipsub could even become multi-tenant, i.e. a single app could have multiple gossipsub instances/domains (reusing underlays) to cater to the needs of various subsystems co-existing **within** a process, eg. in an Eth client: mempool, PeerDAS, attestations, Beam-related flows, etc. Each domain can have its specialized policies while reusing the gossipsub system plumbing.

IMO, it's the right time to re-imagine gossipsub under the lens of a framework. This kind of architectural layering is consistent with the original modular vision. Executing on it now creates new degrees of freedom for experimentation and innovation at the app layer, which various teams may welcome to accelerate their roadmap. After all, no one size fits all, and the proliferation of parameters/knobs in the base layer to support all possible variants of everything introduces too much complexity and slows everyone down.

ppopth:
Closed as superseded by https://github.com/libp2p/specs/pull/653

PR #659, opened 2025-01-31
gossipsub: Preventing routing loops with timestamps and expirations

Stebalien:
See https://github.com/libp2p/go-libp2p-pubsub/issues/573 for some context.

I'm proposing that we solve the "routing loops" issue, an issue where messages linger in the network indefinitely (or at least for long periods of time), by repurposing the `seqno` fields of messages as a message timestamp. All messages would expire 2 minutes (likely configurable on a per-topic basis) after they're sent. 

This is a rough-draft and I'm happy to address any feedback and/or consider alternative approaches.

### DIFF
--- a/pubsub/gossipsub/gossipsub-v1.N.md
+++ b/pubsub/gossipsub/gossipsub-v1.N.md
@@ -1,0 +1,62 @@
+I'm proposing that we solve the "routing loops" issue, an issue where messages linger in the network indefinitely (or at least for long periods of time), by repurposing the `seqno` fields of messages as a message timestamp. All messages would expire 2 minutes (likely configurable on a per-topic basis) after they're sent.
+
+## Context
+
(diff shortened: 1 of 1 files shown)


Stebalien:
@vyzo any opinions on this approach? Would you prefer to add an additional expiration field?

vyzo:
they are pretty much the same functionality wise. Expiry might be easier to work with, but timestamp is more robust i think as the decision is at the reciever. Si we cant have senders set their expiry at 2262.

vyzo:
I also dont think we need additional field, and it can continue to function as a seqno for older clients, given sufficient resolution. We can use nanos, but micros should also be ok.

MarcoPolo:
Reviewed this again today. It makes sense. I would suggest making this an extension that, when set, reinterprets the seqno as defined here. This also gives you an explicit signal for when a peer has been upgraded to this new logic.

Issue #553, opened 2023-06-15
Tracking issue: mplex deprecation


marten-seemann:
Things left to do:
- [ ] https://github.com/ipfs/kubo/issues/9958
- [ ] https://github.com/libp2p/test-plans/issues/196
- [ ] archive https://github.com/libp2p/go-mplex
- [ ] remove it from go-libp2p transport suite
- [ ] mark it as deprecated in the specs

@mxinden @thomaseizinger Anything left to do on the Rust side? Feel free to add to above list.

Menduist:
Eth2 still specifies `mplex` as a "must have" multiplexer: https://github.com/ethereum/consensus-specs/blob/9e50c74a372c0d8349764ef411b22cfa71f2a656/specs/phase0/p2p-interface.md?plain=1#L164

And some eth2 clients still don't have a prod-ready yamux (nimbus doesn't, last time I talked to lodestar they didn't)

achingbrain:
> Anything left to do on the JS side?

- [ ] Add a deprecation notice to the mplex readme
- [ ] Deprecate the mplex module on npm

People can still configure js-libp2p with mplex and continue to use it if they wish, they'll just get a warning when they do an `npm install`.

marten-seemann:
> Eth consensus still specifies `mplex` as a "must have" multiplexer: https://github.com/ethereum/consensus-specs/blob/9e50c74a372c0d8349764ef411b22cfa71f2a656/specs/phase0/p2p-interface.md?plain=1#L164

Thank you @Menduist, I wasn't aware. What would be the process to get that spec changed?

Menduist:
The first step would be for every client to have yamux enabled by default (which requires stable support in their underlying libp2ps, and proper testing)

Once that is done & we are confident in all yamuxes, we can update the specs to remove mandatory support

thomaseizinger:
> @mxinden @thomaseizinger Anything left to do on the Rust side?

Nope. It is already "annoying" to use because you need to add a separate dependency.

A next step could be to remove it from the mono-repo and add it to a community maintained one for people that still want to use it.

marten-seemann:
Kubo removed mplex support: https://github.com/ipfs/kubo/issues/9958.

BigLep:
Per https://github.com/ipfs/kubo/issues/10069, Kubo is actually going to disable mplex by default for the next Kubo release and then followup with removing mplex in the next release.

[...]

PR #661, opened 2025-02-25
Deprecate mplex

MarcoPolo:
We've talked a lot about deprecating mplex for many reasons, but I don't think we've done a good job of documenting the reasons. A lot of the reasons are spread out in various issues, repos, and [docs](https://docs.libp2p.io/concepts/multiplex/mplex/). My goal here is to summarize the reasoning and document why we want to deprecate mplex.

Related to https://github.com/libp2p/specs/issues/553

### DIFF
--- a/mplex/README.md
+++ b/mplex/README.md
@@ -2,9 +2,9 @@
 
 > The spec for the friendly Stream Multiplexer (that works in 3 languages!)
 
-| Lifecycle Stage | Maturity       | Status | Latest Revision |
-|-----------------|----------------|--------|-----------------|
-| 3A              | Recommendation | Active | r0, 2018-10-10  |
+| Lifecycle Stage | Maturity       | Status     | Latest Revision |
+|-----------------|----------------|------------|-----------------|
+| 3A              | Recommendation | Deprecated | r0, 2018-10-10  |
 
 Authors: [@daviddias], [@Stebalien], [@tomaka]
 
@@ -25,16 +25,17 @@ and spec status.
 ## Table of Contents
 
 - [mplex](#mplex)
-    - [Table of Contents](#table-of-contents)
-    - [Overview](#overview)
-    - [Message format](#message-format)
-        - [Flag Values](#flag-values)
-    - [Protocol](#protocol)
-        - [Opening a new stream](#opening-a-new-stream)
-        - [Writing to a stream](#writing-to-a-stream)
-        - [Closing a stream](#closing-a-stream)
-        - [Resetting a stream](#resetting-a-stream)
-    - [Implementation notes](#implementation-notes)
+  - [Table of Contents](#table-of-contents)
+  - [Overview](#overview)
+  - [Deprecation Notice](#deprecation-notice)
+  - [Message format](#message-format)
+    - [Flag Values](#flag-values)
+  - [Protocol](#protocol)
+    - [Opening a new stream](#opening-a-new-stream)
+    - [Writing to a stream](#writing-to-a-stream)
+    - [Closing a stream](#closing-a-stream)
+    - [Resetting a stream](#resetting-a-stream)
+  - [Implementation notes](#implementation-notes)
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


mplex/README.md
```diff
@@ -50,6 +51,41 @@ Implementations in:
...
+  - No way for a reader to backpressure a sender.
+  - No good solution for slow readers.
+    - Implementations generally do some combination of these mitigations:
+      - Reset the stream once we reach a certain amount of unread buffered data.
+        [source](https://github.com/libp2p/rust-libp2p/blob/1c9b3ca355aecffa0bcf83d2495cd4cc1019425b/muxers/mplex/src/config.rs#L118)
+      - Block operations until the full stream is read from.
+        [source](https://github.com/libp2p/rust-libp2p/blob/1c9b3ca355aecffa0bcf83d2495cd4cc1019425b/muxers/mplex/src/config.rs#L130)
+      - Block up to a certain amount of time, then reset the stream.
+        [source](https://github.com/libp2p/go-mplex/blob/ad9bfb922974b5875cc48c6e7492c4987c0cb94a/multiplex.go#L35-L37)
+- Head of line blocking between streams
```
ppopth, line 77: As long as you use TCP, there is no way to resolve this, right? So I think it's a good idea to note that only using QUIC will resolve this and yamux won't.

MarcoPolo: You're right, you can only really solve head of line blocking of packets with QUIC. However, there's another head of line blocking issue that I'm referencing here, that Yamux can avoid due to stream flow control.

Say you have two streams, and one of them writes a 10 MB payload, and the other one then writes a 1 KB message.

In Yamux, you would send the first 256KB of the first stream, then the next 1KB of the other message. You would not send any more data on the first stream until you receive a window update message.

In mplex, you would try to send the full 10MB payload before you make any progress on the 1KB payload.

jxs (approved):
Thanks for this Marco, LGTM! Can we get this merged?

mplex/README.md
sukunrt: This is just stream level flow control, no? You're mentioning the initial stream flow credits of 256kB.

mplex/README.md
```diff
@@ -50,6 +51,41 @@ Implementations in:
...
+  - For example, No way to interleave data from streams if one stream makes a
+    big write, and another stream has a small write
+  - A single slow reader of a single stream can stall the whole application and TCP connection.
+- No way of propagating errors.
+  - Errors that could explain why a peer reset the stream.
+- No way to signal to a peer that you will not read any more data (e.g. QUIC's
+  STOP_SENDING frame).
+- Both sides can open streams with the same ID, differing only by who opened the
+  stream. which may lead to confusion.
+- stream names are relatively unused (go-libp2p does not use them. I don't think rust-libp2p or js-libp2p uses them either)
```
sukunrt, line 87: Aren't all of these fixable? We're deprecating this because of the lack of stream flow control.

mplex/README.md
```diff
@@ -50,6 +51,41 @@ Implementations in:
...
+
+**mplex is deprecated** for applications requiring resiliency and liveness.
+Users should **prefer QUIC or Yamux**.
+
+**Core limitation: lack of stream flow control**
+
+mplex does not support stream-level flow control, preventing receivers from
+applying backpressure. This leads issues varying from easy to exploit DoS
+vulnerabilities due to unbounded sender behavior to hard to debug application
+	  stalls caused by slow a single slow stream receiver.
```
sukunrt, line 64:
```suggestion
mplex does not support stream-level flow control, preventing receivers from
applying backpressure. This leads to issues varying from easy to exploit DoS
vulnerabilities due to unbounded sender behavior to hard to debug application
	  stalls caused by a single slow stream receiver.
```

sukunrt (approved):

mplex/README.md
MarcoPolo: Would you rather put this as a sub-bullet under `- No stream-level flow control.` ?

sukunrt: I think thats better. Not a strong opinion though.

MarcoPolo:
let's merge this :)

PR #662, opened 2025-03-04
docs: add ALPN section to TLS and link to inlined muxer negotiation

### LINKED ISSUE #496: tls: specify ALPN
> Unless I am missing something this is not true: the TLS spec does not specify what ALPN to use.

Good catch. Both Go and Rust do this. @marten-seemann can you add it to the TLS spec? Either in this pull request or a separate one?

_Originally posted by @mxinden in https://github.com/libp2p/specs/pull/446#discussion_r1046356481_

### LINKED ISSUE #607: TLS spec doesn't mention `ALPN` `libp2p` used by implementations
See https://grep.app/search?q=alpn&filter[repo.pattern][0]=libp2p.
Implementations use the `libp2p` ALPN which allows to route TLS streams to libp2p and non libp2p connections on the same listener (or libp2p over quic vs libp2p over webtransport for quic).

I don't know if current implementations can handle arbitrary ALPNs and this is just a nice to have client sends or if it's actually required for it to function.

A quick `CTRL + F` `ALPN` in https://github.com/libp2p/specs/blob/master/tls/tls.md yields no results.
Same for looking at all the occurrences of the `libp2p` string in the same file.

achingbrain:
Adds note about which ALPN protocol to use and links to spec for inlining muxer negotiation.

Avoids things like https://github.com/libp2p/jvm-libp2p/issues/404

Fixes #607
Fixes #496

### DIFF
--- a/tls/tls.md
+++ b/tls/tls.md
@@ -128,7 +128,7 @@ How the public key is encoded into the `Data` bytes depends on the Key Type.
 
 "libp2p" is used as the application protocol for [ALPN](https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation).
 
-The handshake MUST be aborted if both peers can't agree on the application protocol.
+The server MUST abort the handshake if it doesn't support any of the application protocols offered by the client.
 
 ## Inlined Muxer Negotiation
 
--- a/tls/tls.md
+++ b/tls/tls.md
@@ -27,6 +27,8 @@ and spec status.
 - [Handshake Protocol](#handshake-protocol)
 - [Peer Authentication](#peer-authentication)
   - [libp2p Public Key Extension](#libp2p-public-key-extension)
+- [ALPN](#alpn)
+- [Inlined Muxer Negotiation](#inlined-muxer-negotiation)
 - [Test vectors](#test-vectors)
   - [1. Valid certificate authenticating an ED25519 Peer ID](#1-valid-certificate-authenticating-an-ed25519-peer-id)
   - [2. Valid certificate authenticating an ECDSA Peer ID](#2-valid-certificate-authenticating-an-ecdsa-peer-id)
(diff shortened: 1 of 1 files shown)


tls/tls.md
```diff
@@ -122,6 +124,16 @@ How the public key is encoded into the `Data` bytes depends on the Key Type.
 - `Secp256k1`: Only the compressed form of the public key. 33 bytes.
 - The rest of the keys are encoded as a [SubjectPublicKeyInfo structure](https://www.rfc-editor.org/rfc/rfc5280.html#section-4.1) in PKIX, ASN.1 DER form.
 
+## ALPN
+
+"libp2p" is used as the application protocol for [ALPN](https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation).
+
+The handshake MUST be aborted if both peers can't agree on the application protocol.
```
marten-seemann, line 131: This is underspecified. You should say: The server MUST abort if...

achingbrain: How about:

> The server MUST abort the handshake if the client does not send a supported application protocol.

marten-seemann: Sounds a bit like you're blaming the client for misbehaving, why not the slightly less accusatory:
> The server MUST abort the handshake if it doesn't support any of the application protocols offered by the client.

achingbrain: Well the client did send the wrong thing 😜 

The amendment sounds fine though.

```suggestion
The server MUST abort the handshake if it doesn't support any of the application protocols offered by the client.
```

tls/tls.md
```diff
@@ -122,6 +124,16 @@ How the public key is encoded into the `Data` bytes depends on the Key Type.
 - `Secp256k1`: Only the compressed form of the public key. 33 bytes.
 - The rest of the keys are encoded as a [SubjectPublicKeyInfo structure](https://www.rfc-editor.org/rfc/rfc5280.html#section-4.1) in PKIX, ASN.1 DER form.
 
+## ALPN
+
+"libp2p" is used as the application protocol for [ALPN](https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation).
```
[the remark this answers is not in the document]
achingbrain: I've created https://github.com/libp2p/specs/issues/663 to make sure this doesn't get lost

MarcoPolo (approved):

achingbrain:
Refs https://github.com/ipshipyard/roadmaps/issues/16

PR #664, opened 2025-03-17
Generalised Gossipsub

AgeManning:
# Overview

This an idea for modularizing gossipsub to allow for greater optimizations while minimizing any engineering effort for implementations that current support gossipsub. 

Fundamentally this is just a re-factor of code, with the addition of two new control messages:
CHOKE/UNCHOKE

The added control messages are the only thing that needs to be implemented to support this change. Any variation of these control messages would also work, I think we just need this degree of freedom to adjust mesh broadcasts without having to adjust the mesh sizes. 

## tldr

We shift current gossipsub logic into conceptual modules called strategies:
- Broadcast strategy - Controls how we send messages, can tune amount of gossip vs direct message sending
- Mesh strategy - Controls how we form networks. How big is the mesh, who we want to connect to, how it changes over time
- Scoring - Different scoring algorithms. 

Each of these can be applied per-topic (and dynamically if needed), allowing a complex network with competing constraints to choose different strategies per different applications (per-topic) or under different scenarios (i.e bandwidth limitations). 

## Benefits (Intended at least)
- Researches/Developers can fine-tune networks without having to modify the spec (they do this via stating/specing new strategies)
- Strategies can be applied dynamically and per topic, allowing a greater configuration space for complex use cases
- Implementations can mix/match or change dynamically if the network scenario changes, giving more flexibility to handle changing network conditions
- I'm hoping testing becomes easier. Researches can add/replace strategies and test them against a baseline a bit more easily.

### DIFF
--- a/generalized-gossipsub/README.md
+++ b/generalized-gossipsub/README.md
@@ -137,8 +137,8 @@ message ControlMessage {
  repeated ControlIWant iwant = 2;
  repeated ControlGraft graft = 3;
  repeated ControlPrune prune = 4;
- repeated ControlChoke prune = 5;
- repeated ControlUnChoke prune = 6;
+ repeated ControlChoke choke = 5;
+ repeated ControlUnChoke unchoke = 6;
 }
 
 message ControlIHave {
--- a/generalized-gossipsub/README.md
+++ b/generalized-gossipsub/README.md
@@ -1,0 +1,344 @@
+# Generalized Gossipsub
+
+This is a generalization of the original [gossipsub protocol](../gossipsub/README.md)
+that makes the protocol extensible through additional modules called strategies.
+These strategies can be employed per-topic and allow the protocol to be
+fine-tuned under a range of scenarios and circumstances and applied either
+dynamically or statically on a per-application (topic) basis.
+
+NOTE: Any current gossipsub implementation can be fully compatible with this
+protocol by simply implementing the CHOKE/UNCHOKE control messages. This protocol is
+fully backwards compatible with original gossipsub implementations.
+
+## Table of Contents
+
+- [Generalized Gossipsub](#generalized-gossipsub)
+  - [Motivation](#motivation)
+  - [Overview](#overview)
+  - [Strategies](#strategies)
+  - [The RPC](#the-rpc)
+  - [Deviations from Gossipsub](#deviations-from-gossipsub)
+    - [Protocol State](#protocol-state)
+    - [Topic Membership](#topic-membership)
+    - [Control Messages](#control-messages)
+    - [Message Propagation](#message-propagation)
+
+## Motivation
+
+[Gossipsub](../pubsub/gossipsub/README.md) is designed to be a configurable protocol that allows users to adjust
+configuration parameters to make trade-offs between resiliency, latency and
+bandwidth. Over the years, these parameters have been found too constraining
+resulting in a number of proposals to modify the specification to achieve better
+results under specific circumstances.
+
+There are three major areas that are constrained by the gossipsub specification, that are a
+prime targets to free from the specification and allow users to implement freely, these
+are:
+
+- **Scoring** - A system used to score peers for security and performance.
+- **Mesh construction/peer sampling** (and therefore network topologies).
+- **Message dissemination** - The choice of how to send messages either direct or
+  via gossip and the timing.
+
+A further constraint is that gossipsub requires that any given choice of these
+areas is enforced across all topics, which constrains the solution space for
+users to one design choice for potentially different applications.
+
+There are mesh construction strategies that can select mesh's in such a way that
+can form efficient broadcast trees. We could think of a topology that forms a
+single tree spanning all nodes on the network. This would be very efficient, but
+not very resilient. Another example could be small mesh sizes, which sacrifices
+resiliency for bandwidth, which may be applicable to applications implementing
+schemes like erasure coding which have redundancy built in.
+There are different broadcast strategies that could favour gossip over direct
+sending. This could be a trade-off between latency and bandwidth. In some
+cases these strategies would be better suited for nodes under bandwidth
+constraints.
+
+The generalization proposed here aims to specify these strategies and allow the core protocol
+to select each strategy per topic. This allows for scenarios where an
+application may have a topic where resiliency is not very important, so a
+low-bandwidth strategy could be chosen, (i.e low-mesh, sparse topology) and at
+the same time have a topic where resiliency is important so chooses a
+high-bandwidth strategy (i.e high-mesh, dense topology).
+
+A node that is heavily resource constrained, might also wish to switch to a
+combination of strategies that is known to perform better under those
+conditions.
+
+The goal of this protocol is into increase the degrees of freedom in fine-tuning
+p2p message dissemination in a general way that doesn't require specification
+changes to apply. It also aims to minimize engineering overhead for
+implementations that already have gossipsub.
+
+## Overview
+
+The protocol defined here is an extension/modification of the [original
+gossipsub](../pubsub/gossipsub/README.md) protocol. It is aimed to increase the
+configurability of the protocol to allow for a wider set of applications and
+performance tunings.
+
+As this is based on the original gossipsub specification, this document does not compile all past versions of the specification here,
+rather lists the main departures from the original specification and consolidates all the protobuf and control messages.
+
+## Strategies
+
+There are three strategies that can be selected at any given time on a per-topic
+bases:
+
+- **Scoring** - How to score peers based on their behaviour/performance. This can
+  also be global (not necessarily per-topic)
+- **Mesh** - Construction/Maintenance - This strategy, if combined across all nodes,
+  fundamentally controls the topology of the network.
+- **Broadcast** - This strategy can dictate tradeoffs between resilience, bandwidth
+  and latency by deciding how aggressively to directly send messages vs
+  gossiping them.
+
+Each of these are detailed in the [Strategies](./strategies.md) section.
+
+Examples strategies are also provided:
+
+- [Original Gossipsub Broadcast](./strategies/broadcast/original-gossipsub.md)
+- [Original Gossipsub Mesh](./strategies/mesh/original-gossipsub.md)
+- [Random Choke Mesh](./strategies/mesh/random-choke.md)
+
+## The RPC
+
+All communication between peers happens in the form of exchanging protobuf RPC
+messages between participating peers.
+
+The `RPC` protobuf is as follows:
+
+```protobuf
+syntax = "proto2";
+message RPC {
+ repeated SubOpts subscriptions = 1;
+ repeated Message publish = 2;
+
+ message SubOpts {
+  optional bool subscribe = 1; // subscribe or unsubscribe
+  optional string topic_id = 2;
+ }
+
+ optional ControlMessage control = 3;
+}
+
+message Message {
+ optional bytes from = 1;
+ optional bytes data = 2;
+ optional bytes seqno = 3;
+ required string topic = 4;
+ optional bytes signature = 5;
+ optional bytes key = 6;
+}
+
+message ControlMessage {
+ repeated ControlIHave ihave = 1;
+ repeated ControlIWant iwant = 2;
+ repeated ControlGraft graft = 3;
+ repeated ControlPrune prune = 4;
+ repeated ControlChoke prune = 5;
+ repeated ControlUnChoke prune = 6;
+}
+
+message ControlIHave {
+ optional string topic_id = 1;
+ repeated bytes message_ids = 2;
+}
+
+message ControlIWant {
+ repeated bytes message_ids= 1;
+}
+
+message ControlGraft {
+ optional string topic_id = 1;
+}
+
+message ControlPrune {
+ optional string topic_id = 1;
+ repeated PeerInfo peers = 2; // gossipsub v1.1 PX
+ optional uint64 backoff = 3; // gossipsub v1.1 backoff time (in seconds)
+}
+
+message ControlChoke {
+    required string topicID = 1;
+}
+message ControlUnChoke {
+    required string topicID = 1;
+}
+
+message PeerInfo {
+ optional bytes peer_id = 1;
+ optional bytes signed_peer_record = 2;
+}
+```
+
+## Deviations from Gossipsub Logic
+
+Logic has been split from what we call "core" protocol into strategies. Here the
+distinction between the split is made.
+
+### Parameters
+
+The parameters that remain in the protocol are:
+
+| Parameter            | Purpose                                               | Reasonable Default |
+| -------------------- | ----------------------------------------------------- | ------------------ |
+| `heartbeat_interval` | Time between [heartbeats](#heartbeat)                 | 1 second           |
+| `fanout_ttl`         | Time-to-live for each topic's fanout state            | 60 seconds         |
+| `mcache_len`         | Number of history windows in message cache            | 5                  |
+| `mcache_gossip`      | Number of history windows to use when emitting gossip | 3                  |
+| `seen_ttl`           | Expiry time for cache of seen message ids             | 2 minutes          |
+
+### Protocol State
+
+The core protocol still handles the control messages such as
+SUBSCRIBE/UNSUBSCRIBE/GRAFT/PRUNE/CHOKE/UNCHOKE and therefore must still
+maintain a state of peers that form a mesh (how this is formed is up to any
+given strategy), and keep track of which peers are subscribed to which topic.
+
+### Topic Membership
+
+The [original gossipsub](../pubsub/gossipsub/README.md) specification describes
+how a router behaves when peers join or leave topics. The terminology used is
+`JOIN(topic)` and `LEAVE(topic)`. The new logic of this behaves is described
+here.
+
+When the application invokes `JOIN(topic)`, the router will form a topic mesh by
+asking the selected [mesh strategy][./mesh-strategy.md] for the given topic.
+
+For all new peers that have formed the mesh, the router will inform them that they have been added to the
+mesh by sending them a `GRAFT` control message.
+
+The application can invoke `LEAVE(topic)` to unsubscribe from a topic. The
+router will inform the peers in `mesh[topic]` by sending them a `PRUNE` control
+message, so that they can remove the link from their own topic mesh.
+
+After sending `PRUNE` messages, the router will forget `mesh[topic]` and delete
+it from its local state.
+
+### Control Messages
+
+Control messages are exchanged to maintain topic meshes and emit gossip. This
+section lists the control messages in the core gossipsub protocol
+
+#### SUBSCRIBE
+
+When a subscribe message is received, the router records that the new peer is
+now subscribed to the topic and informs the mesh strategy. The mesh strategy may
+choose to GRAFT this peer on this topic.
+
+#### GRAFT
+
+The `GRAFT` message grafts a new link in a topic mesh. The `GRAFT` informs a peer
+that it has been added to the local router's mesh view for the included topic id.
+
+#### PRUNE
+
+The `PRUNE` message prunes a mesh link from a topic mesh. `PRUNE` notifies a
+peer that it has been removed from the local router's mesh view for the
+included topic id.
+
+#### IHAVE
+
+The `IHAVE` message provides the
+remote peer with a list of messages that were recently seen by the local router.
+
+The remote peer may then request the full message content with an `IWANT` message.
+
+The speed and degree at which these messages are sent are entirely left to the
+broadcast strategy.
+
+#### IWANT
+
+The `IWANT` message requests the full content of one or more messages whose IDs
+were announced by a remote peer in an `IHAVE` message.
+
+#### The CHOKE Message
+
+Upon receiving a `CHOKE` message, the router MUST no longer forward messages to
+the peer that sent the `CHOKE` message, while it is still in the mesh. Instead
+it MUST always send an IHAVE message (provided there are messages to send and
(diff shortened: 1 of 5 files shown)


vyzo:
Just a small structural comment, can we put those under pubsub/ at the very least?

Ideally it should go into the pubsub/gossipsub/ ditectory

AgeManning:
Yeah sure. This is still very much a rough draft. Can move location and fix a lot of this up, if there is any interest in this

MarcoPolo (commented):
I've done an initial pass on these documents. Thank you for writing this! I like the general idea and want to clarify a couple high level points before diving into some of the details.  

1. How can new strategies be sure to be compatible with other strategies and their scoring methods?
    1. Specifically, how can new strategies stay backwards compatible with the expectations of the existing gossipsub network?
    2. Generally, I wonder if we need to define rules of what peers MUST NOT do in order to remain in good standing.
    3. This might be the trickiest part of this proposal.
2. Do we need a way to add new messages to the protocol?
    1. It seems to me that Choke is needed here for the "random choke" strategy. I can imagine new strategies which needing new messages as well.
        1. example: A variant on choke that allows for new original messages to be sent, while still avoiding forwarded messages.
    2. My concern is overfitting for these two strategies while claiming a generic interface.
        1. Maybe drafting a third strategy would be enough to address this concern.
3. Is the expectation that within a given topic different peers may implement different strategies depending on their own concerns?

vyzo:
CHOKE is also needed for implementing the announcrsub proposal, where we then use eager IHAVE instead of IANNOUNCE.

AgeManning:
@MarcoPolo Thanks for the comments. 

1. From a pure interop perspective, as long as they are sending the spec'd messages, any strategy is "compatible" with any other. This is entirely backwards compatible. Implementations that don't have choke/unchoke, we just don't send them these messages. If we just use the original gossipsub strategy, there is no difference. 

However, I think you're asking more or less how the general network would look, and whether different scoring strategies can work together. In general they can't. I imagine, this to be spec'd outside of libp2p. For example, if Ethereum wanted to use a specific strategy, they would would specify that so that all of their nodes used one strategy. There is no versioning, you are right. Perhaps that might be something to add, but in the Ethereum case, we can upgrade strategies on hard forks to avoid partial upgrades. 
ii. The problem as I see it, is that gossipsub attempts to be too general. What I mean by this, is that its a protocol thats used by many implementations and many networks. If someone wants to make a modification, it cannot be just that its good for one application, but it must be good for every single application on every network that uses gossipsub. Therefore in the Ethereum case, we can't engineer something that works just for Ethereum and put it into the specs, if it doesn't work for everyone else. So the definition of what peers must do to be in good standing, I think each network/application should define themselves outside of the gossipsub spec.

2. No, we don't. The reason for adding it, seems to be that attempts at optimizing current gossipsub involve changing how the mesh works. It adds an extra degree of freedom to do this. I added it, because I think it solves two proposed solutions to improve gossipsub (https://github.com/libp2p/specs/pull/653) (the thing I called random choke) and "episub" (https://github.com/libp2p/specs/pull/413). It adds a degree of freedom to do what people are suggesting by modifying the mesh. I agree with you, it could be over-fitting, but I've not seen other proposals to improve gossipsub that require other messages. In either case, I think adding new protocol messages, would be simpler in this case. We define what they do, how to handle them in the spec, then whatever strategy can contain the more complex logic and implementations that require it can implement the strategy or not. 

3. The idea is to shift the burden of this decision to outside the libp2p specs, so every use case can decide for itself. Some applications may have static strategies for all topics (as is now), but if people want to shift and change them and experiment, they are able to, without having to make a spec change and enforce it on everyone else. Also, if it decided that it doesn't work in practice, it can be easily reverted.

generalized-gossipsub/strategies.md
```diff
@@ -0,0 +1,79 @@
...
+balance between lean broadcast trees and heavily connected networks by selecting how many and who will form mesh peers.
+
+A host of network-specific strategies can be made to produce a range of varying overlay networks which can be distinct per-topic.
+
+### Interface
+
+This section defines the basic logical interface that defines a mesh strategy. The exact API of how a mesh strategy interfaces with the core
+protocol is left to each specific implementation.
+
+#### JOIN(Topic)
```
ppopth, line 24: What is the difference between `JOIN` and `SUBSCRIBE`?

generalized-gossipsub/strategies.md
```diff
@@ -0,0 +1,79 @@
...
+## Broadcast Strategies
+
+These strategies define how messages are broadcast through the network. They set
+how eagerly we send messages, whether we direct send them or gossip them and how
+often. These strategies can effectively provide tradeoffs between latency,
+bandwidth and redundancy on a per-topic, per-time basis.
+
+### Interface
+
+This section defines the basic logical interface that defines a mesh strategy. The exact API of how a mesh strategy interfaces with the core
```
ppopth, line 56:
```suggestion
This section defines the basic logical interface that defines a mesh strategy. The exact API of how a broadcast strategy interfaces with the core
```

generalized-gossipsub/strategies/broadcast/original-gossipsub.md
```diff
@@ -0,0 +1,57 @@
...
+| --------- | -------------------------------------------------- | ------------------ |
+| `D_lazy`  | (Optional) the outbound degree for gossip emission | `D`                |
+
+Note that `D_lazy` is used to control the outbound
+degree when [emitting gossip](#gossip-emission), which may be tuned separately
+than the degree for eager message propagation.
+
+## Interface Implementation
+
+### Publish(Topic) and Forward(Topic)
```
ppopth, line 24:
```suggestion
### Publish(Topic)
```

ppopth:
No, your random choke is not gossipsubv2. They are completely different things. Saying that random choke is gossipsubv2 is very misleading.

In gsv2, when I receive the first IANNOUNCE, I send INEED. If the timeout is triggered, I send INEED to the peer who sent the second IANNOUNCE. An example timeline is as follows.

Let's say the timeout is 500ms
0.0001s - Receive IANNOUNCE from peer1
0.0001s - Send INEED to peer1
0.0100s - Receive IANNOUNCE from peer2
0.5001s - Hasn't received the message from peer1, so the timeout is triggered and send INEED to peer2
0.5101s - Receive the message from peer1
0.5201s - Receive the message from peer2

While in your random choke, the timeline will be different
0.0001s - Receive IHAVE from peer1
0.0001s - Send IWANT to peer1
0.0100s - Receive IHAVE from peer2
0.0100s - Send IWANT to peer2
0.0200s - Receive the message from peer2
0.5101s - Receive the message from peer1

Please also note that with your current generalization, gossipsubv2 strategies cannot be implemented at all.

@AgeManning I encouraged you to read the [Gossipsub v2 spec](https://github.com/libp2p/specs/pull/653) to get familiar with the protocol.

MarcoPolo:
After rereading gsv2 and random choke, it seems like the difference is whether the source or destination controls the push vs pull semantics. In gsv2, the source decides to push the message or rely on the peer to pull the message following an `IANNOUNCE`. In random choke, the destination decides which peers are choked, and thus which peers will send an `IHAVE` (pull semantics) rather than push the message.

On the wire, `IHAVE/IWANT` is roughly equivalent to `IANNOUNCE/INEED`. The difference is in the semantics of how peers treat these messages. My understanding of this proposal is that it would allow networks to define their own semantics of these messages.

ppopth:
> the difference is whether the source or destination controls the push vs pull semantics

That's not the only difference. Another one is explained here https://github.com/libp2p/specs/pull/664#issuecomment-2741388528. The idea is that gsv2 has the timeout which is the core of the improvement.

> My understanding of this proposal is that it would allow networks to define their own semantics of these messages.

The problem is that this PR doesn't allow me to do the timeout.

MarcoPolo:
My understanding is that the timeout could be implemented by the broadcast strategy. From this PR:

> On receiving an IHAVE(ids) message, the router will check its seen cache. If the IHAVE message contains message IDs that have not been the router will ask the [broadcast strategy](https://github.com/libp2p/specs/blob/generalised-gossipsub/generalized-gossipsub/strategies.md) whether it should request it via an IWANT message.

That makes it seem that the broadcast strategy could track these `IHAVE`, request the message immediately, and configure a timeout to try another peer.

ppopth:
@MarcoPolo 
>  0.0100s - Receive IANNOUNCE from peer2
0.5001s - Hasn't received the message from peer1, so the timeout is triggered and send INEED to peer2

That would make sense if the generalization allows me to delay answering that question. You can see that I received IANNOUNCE at 0.0100s but decided to send INEED at 0.5001s.

MarcoPolo:
@AgeManning apologies if this is obvious, but what are the benefits of this spec versus say starting from scratch with knowledge of GossipSub? I don't yet have a strong opinion here. This is not a leading question.

To play devils advocate, in many ways the strategy for mesh and broadcasting is a big part of the solution. If we take that out what is the valuable thing being spec'd here? Is it the control messages? What does an application gain by constraining itself to only these control messages? Is it the interface for the strategies? Is the hope that by defining the interface, applications can build against that and share code with other applications? Are we trying to define high level primitives for any pubsub-like system?

MarcoPolo:
> @MarcoPolo
> 
> > 0.0100s - Receive IANNOUNCE from peer2
> > 0.5001s - Hasn't received the message from peer1, so the timeout is triggered and send INEED to peer2
> 
> That would make sense if the generalization allows me to delay answering that question. You can see that I received IANNOUNCE at 0.0100s but decided to send INEED at 0.5001s.

Ah, I think I see your point. Your point is that "random choke" as defined in this PR does not implement this delay. That's true. However, I don't see anything in the "generalized" part that would prohibit a strategy from implementing this delay.

MarcoPolo:
> Disclaimer: I assumed that you cannot arbitrarily send IWANT any time you want. The only way to send it is to tell the core protocol to send it for you through the IHAVE(ids) function.

I think this is where we interpret the document differently. This highlights the need for more precise language 😄 . I assumed the opposite, that a a strategy could arbitrarily send IWANTs anytime.

generalized-gossipsub/README.md
```diff
@@ -0,0 +1,344 @@
...
+
+### Message Processing
+
+#### Message
+
+When a message is received that is valid and was not published by the router itself, the router informs the broadcast module via the
+Forward(Topic) interface.
+
+The broadcast module will return which peers to forward the message to and which
+to gossip to (if any).
```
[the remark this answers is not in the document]
AgeManning: Yes, if we want the choke action, then a broadcast module cannot send a message to a peer that has requested us to choke them (this would defeat the purpose of the choking mechanism). 

A broadcast module can still decide who to choke, which non choked peers to send a message to, which peers to send IHAVE's to. 

Here are some examples of different broadcast strategies that demonstrate the point of the broadcast module:
- Floodsub-like: Send the direct message to all known peers that are non-choked
- Episub-Like: Send direct message to all mesh peers that are non-choked
- Gossipsubv2-like: Pick a random set of mesh peers and send IHAVEs to them, send direct messages to non-choked peers.
- No Duplicates: Only send IHAVE messages to mesh peers.

AgeManning:
Thanks guys. 

Keep in mind, this was just a quick writeup to explain the concept, I wouldn't consider it ready to merge and we can change it to adapt to peoples needs.

@pop
> No, your random choke is not gossipsubv2. They are completely different things. Saying that random choke is gossipsubv2 is very misleading.

I never said that random choke it is gossipsub v2 (thats why I named it differently), here is what I wrote:

> This is an example strategy that resembles the gossipsub v2 proposal (mainly as
an example strategy).

And I do think it resembles gossipsub v2 (which primarily makes a trade-off between direct sending messages in the mesh and gossiping them. It decides this on a per-message basis, with a random factor. As does random-choke). 

> @AgeManning I encourage you to read the https://github.com/libp2p/specs/pull/653 to get familiar with the protocol.

I have read it, that's why I'm suggesting this (and our long telegram chat). 

> Let's say the timeout is 500ms
0.0001s - Receive IANNOUNCE from peer1
0.0001s - Send INEED to peer1
0.0100s - Receive IANNOUNCE from peer2
0.5001s - Hasn't received the message from peer1, so the timeout is triggered and send INEED to peer2
0.5101s - Receive the message from peer1
0.5201s - Receive the message from peer2

> While in your random choke, the timeline will be different
0.0001s - Receive IHAVE from peer1
0.0001s - Send IWANT to peer1
0.0100s - Receive IHAVE from peer2
0.0100s - Send IWANT to peer2
0.0200s - Receive the message from peer2
0.5101s - Receive the message from peer1

Not sure why the message from peer2 is coming earlier here. But as I explained in the telegram chat, IHAVE's also have a timeout that is configurable. It's required from the gossipsub 1.1 spec. If you set that to 500ms, then the IWANT will fail also. If you want to write an implementation that re-requests an IHAVE when one expires but from another peer, there is nothing in the specifications from preventing you doing that. If it is a major concern, we can also have an API in a broadcast strategy that decides how to handle failed IWANTs.

>That's not the only difference. Another one is explained here https://github.com/libp2p/specs/pull/664#issuecomment-2741388528. The idea is that gsv2 has the timeout which is the core of the improvement.

Honestly I'm surprised that the core improvement is re-requesting IWANTs. I didn't really think this would have an impact so didn't directly focus on it, I was focused more on the trade-off between the direct vs gossip send in the meshes. 

> Disclaimer: I assumed that you cannot arbitrarily send IWANT any time you want. The only way to send it is to tell the core protocol to send it for you through the IHAVE(ids) function.

I didn't think this would be such a big issue. This is just handling "broken promises" (i.e an IWANT response didn't hit its deadline). As I mentioned, anyone can implement this, but if people want it in the spec, I'd suggest just adding an API to handle broken promises to the broadcast module. 


> Please also note that with your current generalization, gossipsubv2 strategies cannot be implemented at all.

I think the main principle of gossipsubv2 can be implemented. If it is just re-requesting IHAVE's that you're concerned about, we can do that. If it's specifically the IANNOUNCE/INEED messages, we can use those instead of CHOKE/UNCHOKE, but I dont think they gain us much and have the following issue:
- Currently, if you have a set of mesh peers, you would currently expect them to give you messages directly. If you were getting messages slowly, you might consider cycling your mesh peers because they are not performing. It allows you to locally optimize your mesh, and the 1.1 scoring have mechanisms to encourage this. With IANNOUNCE/INEED, you can no longer know if any of your mesh peers are just going to send you messages slowly or if you are randomly unlucky etc. The uncertainty in the mesh, makes it hard to optimize. The CHOKE/UNCHOKE is similar (I won't say the same), but as @MarcoPolo points out, it pushes the knowledge to our peers to gossip rather than direct send. This has the added benefit that, we know at any given time that all our mesh peers will directly send us messages and we know which peers want us to gopssip to. 

@MarcoPolo 
> After rereading gsv2 and random choke, it seems like the difference is whether the source or destination controls the push vs pull semantics. 

Exactly.

AgeManning:
@MarcoPolo 

> @AgeManning apologies if this is obvious, but what are the benefits of this spec versus say starting from scratch with knowledge of GossipSub? I don't yet have a strong opinion here. This is not a leading question.

> To play devils advocate, in many ways the strategy for mesh and broadcasting is a big part of the solution. If we take that out what is the valuable thing being spec'd here? Is it the control messages? What does an application gain by constraining itself to only these control messages? Is it the interface for the strategies? Is the hope that by defining the interface, applications can build against that and share code with other applications? Are we trying to define high level primitives for any pubsub-like system?

Yeah good question, I guess I didn't explain it explicitly. There has been out-of-band conversation about this, which is where this originated from. 

This adds little to no value in terms of specification (apart from the CHOKE/UNCHOKE, but I'm not entirely convinced we should add them). If you take those messages out, this is exactly the same spec as current gossipsub (as we include current gossipsub broadcast and mesh strategies). So I'm trying to make as little change to gossipsub as possible to avoid building anything new from scratch. 

I consider this more of an engineering refactor of code, where we clearly split up parts of the spec that all implementations MUST agree on for the protocol to work, and the other parts where implementations could currently have wildly wrong (they build the mesh incorrectly (not to spec) or send IWANTs for every message etc), or nodes have different configurations (huge mesh degrees) etc and other nodes on the network wouldn't know the difference. 

I don't intend this to constrain gossipsub to only these control messages. More control messages can and probably should be added, just as we do with current gossipsub. Even the interface to strategies is unimportant here, because we're not spec'ing cross-client strategies. I don't even expect code sharing. 

The value here I think, is to clearly label what MUST be implemented to be part of a gossipsub network (i.e control messages) and what implementers can tweak and modify for their own individual use cases for optimisations and different scenarios. 

The thing that has motivated this, was the gossipsubv2 spec, which trades latency for bandwidth. This is an optimisation which has a specific use case. The choice of making this trade-off will be application and network specific. It will be good in some circumstances and bad in others. In general there's a lot more trade-offs that can be made and I'd like us to be able to make other trade-offs and not go too deep down one path. 

This PR is proposing to shift the decision of which trade-off to make to be outside of this spec. So every interested party is free to make their own trade-offs without effecting each other. Also if one turns out not to work as expected, its easy to revert and try something else.

For example, an optimization designed for Ethereum doesn't now necessarily affect the filecoin network. Each team are free to choose. Also, it's applied per-topic which allows trade-offs per use (which is important in Ethereum). 

The main value in this is deciding what is common in gossipsub across every single party, and what can we split out to so we can individually make our optimisations.

AgeManning:
I was curious about the re-sending of the IWANTs, as I thought I mis-judged the importance of this. 

Here is a node running gossipsub on Ethereum mainnet. 

It is sending about 1700 IWANT messages every 12 seconds:

iwant

Of those 1700 IWANT messages, barely any are failed to be delivered in time:

promises

As such I've always considered failed IWANT messages to be a low issue, not making a re-send worth it.

ufarooqstatus:
> I was curious about the re-sending of the IWANTs, as I thought I mis-judged the importance of this.
> 
> Here is a node running gossipsub on Ethereum mainnet.
> 
> It is sending about 1700 IWANT messages every 12 seconds:

This happens because sending a large message can take considerable time. During this period, peers already receiving this message may also send IWANT requests (multiple IWANTS in v1.1) for the same message.

The same is the limitation faced by IDONTWANT announcements. A receiver can send IDONTWANTs only after it finishes downloading the entire message. For a large download time, the receiver will likely start receiving the same message from other mesh members. 

[Message preamble](https://github.com/libp2p/specs/pull/654) is crucial in eliminating these duplicates.

generalized-gossipsub/README.md
```diff
@@ -0,0 +1,344 @@
...
+}
+
+message ControlMessage {
+ repeated ControlIHave ihave = 1;
+ repeated ControlIWant iwant = 2;
+ repeated ControlGraft graft = 3;
+ repeated ControlPrune prune = 4;
+ repeated ControlChoke choke = 5;
+ repeated ControlUnChoke unchoke = 6;
+}
```
ufarooqstatus, line 142: There are additional message proposals that can be valuable for a broader range of GossipSub use cases. For example,

1) [PREAMBLE/IMRECEIVING](https://github.com/libp2p/specs/pull/654)
Message preamble enables receivers to learn about ongoing message receptions, and IMRECEIVING requests mesh members to defer sending the same message we are already receiving. 

2) [OBSERVE](https://github.com/libp2p/specs/blob/3c0085ef514b82618ef6239212bec2e1518702d8/pubsub/gossipsub/gossipsub-v1.3.md#observing-a-topic) can allow observing a topic without having to download entire messages. Analogous to CHOKE, but allows observing non-mesh members as well

3) [INEED](https://github.com/libp2p/specs/pull/653) already discussed here

generalized-gossipsub/README.md
```diff
@@ -0,0 +1,344 @@
...
+high-bandwidth strategy (i.e high-mesh, dense topology).
+
+A node that is heavily resource constrained, might also wish to switch to a
+combination of strategies that is known to perform better under those
+conditions.
+
+The goal of this protocol is into increase the degrees of freedom in fine-tuning
+p2p message dissemination in a general way that doesn't require specification
+changes to apply. It also aims to minimize engineering overhead for
+implementations that already have gossipsub.
```
ufarooqstatus, line 72: A broader use case encompassing different implementations may still require negotiating around either mesh/forwarding/scoring strategies or the appropriate GossipSub versions. 
However, generalizing the GossipSub core and facilitating adaptiveness in mesh/forwarding/scoring strategies would probably encourage the adoption of new techniques with minimal engineering effort.

More importantly, remaining open to incorporating new messages and techniques into the GossipSub core is crucial for the continued growth and evolution of the broader ecosystem.

generalized-gossipsub/README.md
AgeManning: Yep, like the current gossipsub, these are not excluded and can be added as spec changes.

generalized-gossipsub/README.md
AgeManning: Yep, I think we agree here.

Nashatyrev:
Interesting proposal. 
Absolutely support independent topics flexible configuration. 
Not sure about introducing new abstraction level. Probably need to read it more thoughtfully. 

BTW we still have multiple topics per message 'feature'. It could complicate the things. 
TBH if it's possible I would eliminate this 'flexibility' and leave just single topic per message. It's usually so much hustle from both spec and implementation perspectives. 
I know it would be backward incompatible but maybe... 
@vyzo by any chance are you aware of real world use cases when this feature makes sense?

vyzo:
Yeah, we should. It doesnt make much sense in any real world applications, and it complicates implementation as well... R.I.P.

AgeManning:
I think this is mainly an implementation design to allow us to experiment more readily. It was something we may have started work on in rust-libp2p. 

The landscape seems to be changing now. We have extensions (which helps) and at least for our use case it looks like we may be moving more towards custom implementations for our needs and shifting from gossipsub. 

Therefore, we are unlikely to commit any further work down this path and so I'm closing this PR.

PR #676, opened 2025-04-30
docs: adds an expires field to the bearer token response

### LINKED ISSUE #674: Include expiry date alongside bearer token in HTTP PeerId Auth
The server in the [HTTP PeerId Auth scheme](https://github.com/libp2p/specs/blob/master/http/peer-id-auth.md) defines an opaque bearer token that the client should include in every request.  It's common to encode an expiry date in the token so that the server can expire tokens after a set amount of time.

When this is the case the client doesn't know when the token will expire, so it needs to handle fetching a new token if the current one starts causing 401s.

This is fine for GET/HEAD/OPTIONS/DELETE requests but when sending request bodies, the body will be consumed by the request that causes the 401 and the user will have to recreate the body to use in a subsequent request with an updated token.  Easy if it's a string, less so if it's a stream.

The browser fetch API also doesn't let us read any part of the response until the complete request has been sent so we can't just send the request headers and then read the response headers before deciding whether or not to continue sending data.

A workaround for this is to send an OPTIONS preflight-style request for every POST/PUT/etc request but this increases the amount of requests that need to be made, and it introduces a race condition whereby the OPTIONS request might succeed but the following POST fails due to the token expiring between the two requests.

How about including an additional `expires` key in the `Authentication-Info` header that contains the bearer token?

Then the client can step through the authentication process some arbitrary time before their current token expires.

cc @MarcoPolo

achingbrain:
Where a bearer token is returned to the client for future use, optionally include an `expires` field which indicates the latest time that it will be treated as valid.

The server is free to expire tokens at any time so this field is purely advisory.

Fixes #674

### DIFF
--- a/http/peer-id-auth.md
+++ b/http/peer-id-auth.md
@@ -177,7 +177,7 @@ protocol operates as follows:
    format:
 
    ```
-   Authentication-Info: libp2p-PeerID sig="<base64-signature-bytes>" bearer="<base64-encoded-opaque-blob>" expires="<ISO-8601-formatted-date-string>"
+   Authentication-Info: libp2p-PeerID sig="<base64-signature-bytes>" bearer="<base64-encoded-opaque-blob>" expires="<RFC-3339-formatted-date-string>"
    ```
 
    Note that the `expires` field is only advisory, the server may expire the
@@ -271,7 +271,7 @@ The client initiated handshake is as follows
    format:
 
    ```
-   Authentication-Info: libp2p-PeerID bearer="<base64-encoded-opaque-blob>" expires="<ISO-8601-formatted-date-string>"
+   Authentication-Info: libp2p-PeerID bearer="<base64-encoded-opaque-blob>" expires="<RFC-3339-formatted-date-string>"
    ```
 
    Note that the `expires` field is only advisory, the server may expire the
--- a/http/peer-id-auth.md
+++ b/http/peer-id-auth.md
@@ -173,8 +173,7 @@ protocol operates as follows:
    ```
 
    The server MAY include an `expires` field which contains the expiry time of
-   the bearer token in [ISO-8601](https://www.iso.org/iso-8601-date-and-time-format.html)
-   format:
+   the bearer token in [RFC3339](https://www.rfc-editor.org/rfc/rfc3339) format:
 
    ```
    Authentication-Info: libp2p-PeerID sig="<base64-signature-bytes>" bearer="<base64-encoded-opaque-blob>" expires="<ISO-8601-formatted-date-string>"
@@ -267,8 +266,7 @@ The client initiated handshake is as follows
    - The token creation date (to allow tokens to expire).
   
    The server MAY include an `expires` field which contains the expiry time of
-   the bearer token in [ISO-8601](https://www.iso.org/iso-8601-date-and-time-format.html)
-   format:
+   the bearer token in [RFC3339](https://www.rfc-editor.org/rfc/rfc3339) format:
 
    ```
    Authentication-Info: libp2p-PeerID bearer="<base64-encoded-opaque-blob>" expires="<ISO-8601-formatted-date-string>"
--- a/http/peer-id-auth.md
+++ b/http/peer-id-auth.md
@@ -173,7 +173,8 @@ protocol operates as follows:
    ```
 
    The server MAY include an `expires` field which contains the expiry time of
-   the bearer token in [RFC3339](https://www.rfc-editor.org/rfc/rfc3339) format:
+   the bearer token in [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339)
+   format:
 
    ```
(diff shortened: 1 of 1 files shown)


http/peer-id-auth.md
```diff
@@ -186,7 +197,7 @@ protocol operates as follows:
    - The `hostname` parameter.
    - The token creation date (to allow tokens to expire).
 
-5. The client MUST verify the signature. After verification the client has
+6. The client MUST verify the signature. After verification the client has
```
achingbrain, line 200:
```suggestion
5. The client MUST verify the signature. After verification the client has
```

http/peer-id-auth.md
```diff
@@ -254,8 +265,19 @@ The client initiated handshake is as follows
...
+
+   ```
+   Authentication-Info: libp2p-PeerID bearer="<base64-encoded-opaque-blob>" expires="<ISO-8601-formatted-date-string>"
+   ```
+
+   Note that the `expires` field is only advisory, the server may expire the
+   token at any time.
 
-5. The client SHOULD send the `bearer` token for future Peer ID authenticated
+6. The client SHOULD send the `bearer` token for future Peer ID authenticated
```
achingbrain, line 280:
```suggestion
5. The client SHOULD send the `bearer` token for future Peer ID authenticated
```

http/peer-id-auth.md
```diff
@@ -172,6 +172,17 @@ protocol operates as follows:
    Authentication-Info: libp2p-PeerID sig="<base64-signature-bytes>" bearer="<base64-encoded-opaque-blob>"
    ```
 
+   The server MAY include an `expires` field which contains the expiry time of
+   the bearer token in [ISO-8601](https://www.iso.org/iso-8601-date-and-time-format.html)
```
MarcoPolo, line 176: I think we should use [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339) instead. It's a subset of ISO 8601 with a focus on [simplicity](https://www.rfc-editor.org/rfc/rfc3339#section-5.5) and use over internet.

In JS, I believe `toISOString()` is RFC 3339 compliant.

http/peer-id-auth.md
```diff
@@ -254,6 +265,17 @@ The client initiated handshake is as follows
    - The client's Peer ID.
    - The `hostname` parameter.
    - The token creation date (to allow tokens to expire).
+  
+   The server MAY include an `expires` field which contains the expiry time of
+   the bearer token in [ISO-8601](https://www.iso.org/iso-8601-date-and-time-format.html)
+   format:
```
achingbrain, line 271:
```suggestion
   the bearer token in [RFC3339](https://www.rfc-editor.org/rfc/rfc3339) format:
```

http/peer-id-auth.md
```diff
@@ -172,6 +172,16 @@ protocol operates as follows:
    Authentication-Info: libp2p-PeerID sig="<base64-signature-bytes>" bearer="<base64-encoded-opaque-blob>"
    ```
 
+   The server MAY include an `expires` field which contains the expiry time of
+   the bearer token in [RFC3339](https://www.rfc-editor.org/rfc/rfc3339) format:
```
achingbrain, line 176:
```suggestion
   the bearer token in [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339)
   format:
```

http/peer-id-auth.md
```diff
@@ -254,6 +264,16 @@ The client initiated handshake is as follows
    - The client's Peer ID.
    - The `hostname` parameter.
    - The token creation date (to allow tokens to expire).
+  
+   The server MAY include an `expires` field which contains the expiry time of
+   the bearer token in [RFC3339](https://www.rfc-editor.org/rfc/rfc3339) format:
```
achingbrain, line 269:
```suggestion
   the bearer token in [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339)
   format:
```

http/peer-id-auth.md
achingbrain: I think that's what I meant.  I'm all in favour of simplicity.

http/peer-id-auth.md
```diff
@@ -254,6 +265,17 @@ The client initiated handshake is as follows
    - The client's Peer ID.
    - The `hostname` parameter.
    - The token creation date (to allow tokens to expire).
+  
+   The server MAY include an `expires` field which contains the expiry time of
+   the bearer token in [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339)
+   format:
+
+   ```
+   Authentication-Info: libp2p-PeerID bearer="<base64-encoded-opaque-blob>" expires="<ISO-8601-formatted-date-string>"
```
achingbrain, line 274:
```suggestion
   Authentication-Info: libp2p-PeerID bearer="<base64-encoded-opaque-blob>" expires="<RFC-3339-formatted-date-string>"
```

MarcoPolo (approved):

p-shahi (commented):
Seems good to merge once the revision's updated

PR #680, opened 2025-05-24
datagrams: Add initial datagram draft

MarcoPolo:
Initial text for specifying how datagrams will work on libp2p.

cc @raulk @sukunrt @jxs

### DIFF
--- a/datagrams/datagrams.md
+++ b/datagrams/datagrams.md
@@ -1,0 +1,75 @@
+# Datagrams in libp2p
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+| --------------- | ------------- | ------ | --------------- |
+| 1A              | Working Draft | Active | r0, 2025-05-23  |
+
+Authors: [@marcopolo]
+
+Interest Group: [@sukunrt], [@jxs], [@raulk]
+
+[@marcopolo]: https://github.com/marcopolo
+[@sukunrt]: https://github.com/sukunrt
+[@jxs]: https://github.com/jxs
+[@raulk]: https://github.com/raulk
+
+See the [lifecycle document][lifecycle-spec] for context about the maturity level
+and spec status.
+
+[lifecycle-spec]: https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md
+
+## Table of Contents
+
+- [Introduction](#introduction)
+- [Native support at the transport level](#native-support-at-the-transport-level)
+- [Encoding Datagrams](#encoding-datagrams)
+- [Datagrams on top of a multiplexed stream on top of TCP](#datagrams-on-top-of-a-multiplexed-stream-on-top-of-tcp)
+- [Datagrams with WebRTC](#datagrams-with-webrtc)
+
+## Introduction
+
+This specification defines the datagrams on libp2p, which enable the
+transmission of MTU-sized, unreliable, and low latency messages between peers.
+
+TODO write more about the tradeoffs here as well as the kinds of applications that would benefit from using datagrams.
+
+## Native support at the transport level
+
+Some libp2p transports natively support datagrams, such as QUIC and WebTransport. libp2p implementations MUST use the native datagram support to send datagrams.
+
+QUIC has an Unreliable Datagram Extension [RFC 9221]
+WebTransport supports Datagrams though HTTP Datagrams [WebTransport draft RFC](https://www.ietf.org/archive/id/draft-ietf-webtrans-http3-12.html#name-datagrams)
+
+## Encoding Datagrams
+
+The libp2p datagram encoding introduces minimal framing overhead to application payloads. Each datagram MUST be encoded with the following structure to ensure proper protocol identification and version compatibility.
+
+The datagram frame consists of the following components:
+
+| Field Name         | Data Type        | Length (bits) | Description                                          |
+| ------------------ | ---------------- | ------------- | ---------------------------------------------------- |
+| version            | unsigned integer | 8             | Protocol version identifier                          |
+| protocol id length | [uvarint]        | 8..72         | Length of the protocol identifier field              |
+| protocol id        | byte sequence    | variable      | Protocol identifier as defined by protocol id length |
+| application data   | byte sequence    | variable      | Application payload                                  |
+
+The complete datagram frame is constructed as follows:
+
+```
+version || protocol id length || protocol id || application data
+```
+
+Where "||" denotes byte concatenation.
+
+The version field provides extensibility for future encoding modifications while maintaining backward compatibility. Implementations MUST verify the version field before processing the remainder of the datagram.
+
+## Datagrams on top of a multiplexed stream on top of TCP
+
+This is currently not specified.
+
+## Datagrams with WebRTC
+
+This is currently not specified.
+
+[uvarint]: https://github.com/multiformats/unsigned-varint
+[RFC 9221]: https://www.rfc-editor.org/rfc/rfc9221
--- a/datagrams/datagrams.md
+++ b/datagrams/datagrams.md
@@ -50,7 +50,7 @@ Each datagram flow MUST be associated with a control stream. A datagram flow is
 defined as a logical flow of datagrams related to a specific application
 protocol. A control stream is a QUIC bidirectional stream that has negotiated
 the libp2p datagram control stream protocol ID `/dg/1`. The initiator MUST send
-the related application protocol ID related to the datagram flow after
+the related application protocol ID with a [uvarint] length prefix after
 negotiating the control stream. The control stream MUST stay open for the
 duration of the datagram flow. Implementation MAY create the control stream and
 start sending datagrams at once. There is currently no other use for the control
@@ -98,3 +98,4 @@ This is currently not specified.
 
 [RFC 9221]: https://www.rfc-editor.org/rfc/rfc9221
 [QUIC varint]: https://www.rfc-editor.org/rfc/rfc9000.html#name-variable-length-integer-enc
+[uvarint]: https://github.com/multiformats/unsigned-varint
--- a/datagrams/datagrams.md
+++ b/datagrams/datagrams.md
@@ -47,21 +47,18 @@ send datagrams.
 ## Datagrams on QUIC
 
 Each datagram flow MUST be associated with a control stream. A datagram flow is
-defined as a logical flow of datagrams for a specific application protocol. A
-control stream is a QUIC bidirectional stream that has negotiated the application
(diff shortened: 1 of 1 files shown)


marten-seemann (commented):
Have you considered tying a datagram flow to a stream, similar to what HTTP DATAGRAMS (RFC 9297) does?

This would have the following benefits:
1. Shorter datagram header (especially important given MTU restrictions).
2. Both peers get the opportunity to cancel a datagram flow (by closing / resetting the associated stream).
3. A reliable control channel, e.g. for application protocols that need to send an occasional control message next to a datagram flow.

johannamoran:
@jxs - I've reached out to zig, nim, py and dotnet for feedback to get this moving.

datagrams/datagrams.md
```diff
@@ -0,0 +1,101 @@
...
+| Control StreamID | [QUIC varint] | 8..64         | The stream ID of the associated control stream for this datagram flow |
+| Application Data | byte sequence | variable      | Application payload                                                   |
+
+## Datagrams on WebTransport
+
+libp2p datagrams on WebTransport behave the same as libp2p datagrams on QUIC.
+However, the application data may be further limited by the overhead of HTTP
+Datagram's Quarter Stream ID field.
+
+In the future, a separate spec may be able to remove this overhead.
```
richard-ramos, line 89: the quic encoding uses a stream id as the control stream identifier, but browser webtransport doesn't expose quic stream ids. If browser implementations are expected to support this, how is the control stream identified? would some specific webtransport flow identifier be needed?

datagrams/datagrams.md
```diff
@@ -0,0 +1,101 @@
...
+defined as a logical flow of datagrams related to a specific application
+protocol. A control stream is a QUIC bidirectional stream that has negotiated
+the libp2p datagram control stream protocol ID `/dg/1`. The initiator MUST send
+the related application protocol ID with a [uvarint] length prefix after
+negotiating the control stream. The control stream MUST stay open for the
+duration of the datagram flow. Implementation MAY create the control stream and
+start sending datagrams at once. There is currently no other use for the control
+stream besides negotiating the application protocol ID. Receipt of a QUIC
+DATAGRAM frame whose payload is too short to allow parsing the Control Stream ID
+field MUST be treated as a connection error of type PROTOCOL_VIOLATION (0x1003).
```
richard-ramos, line 59: Perhaps something like this so the name does not clash with QUIC's own PROTOCOL_VIOLATION (0x0a) error?  (also on line 71:72)
```suggestion
field MUST be treated as a connection error of type DG_PROTOCOL_VIOLATION (0x1003).
```

datagrams/datagrams.md
```diff
@@ -0,0 +1,101 @@
...
+
+If a libp2p datagram is received and its Control Stream ID field maps to a
+stream that has not yet been created, the receiver SHALL either drop that
+datagram silently or buffer it temporarily (on the order of a round trip) while
+awaiting the creation of the corresponding stream.
+
+If a libp2p datagram is received and its Control Stream ID field maps to a
+stream that cannot be created due to client-initiated bidirectional stream
+limits, it MUST be treated as a connection error of type PROTOCOL_VIOLATION
+(0x1003).
```
richard-ramos, line 72: i'm a bit confused by the reference to "client-initiated bidirectional stream limits". Can control streams be opened by either peer?

datagrams/datagrams.md
```diff
@@ -0,0 +1,101 @@
...
+- WebTransport supports Datagrams though HTTP Datagrams [WebTransport draft RFC](https://www.ietf.org/archive/id/draft-ietf-webtrans-http3-12.html#name-datagrams)
+
+## Datagrams on QUIC
+
+Each datagram flow MUST be associated with a control stream. A datagram flow is
+defined as a logical flow of datagrams related to a specific application
+protocol. A control stream is a QUIC bidirectional stream that has negotiated
+the libp2p datagram control stream protocol ID `/dg/1`. The initiator MUST send
+the related application protocol ID with a [uvarint] length prefix after
+negotiating the control stream. The control stream MUST stay open for the
```
richard-ramos, line 54: i think it would help to clarify the receiver behavior if the application protocol id is unsupported or malformed. (i.e. close stream / reset?). Is this going to use multistream-select?

datagrams/datagrams.md
```diff
@@ -0,0 +1,101 @@
...
+duration of the datagram flow. Implementation MAY create the control stream and
+start sending datagrams at once. There is currently no other use for the control
```
richard-ramos, line 56: What should happen to datagrams that arrive before control stream has finished `/dg1` processing and application protocol negotiation? should they be dropped or buffered?

jxs: I suggest discarding them, buffering them introduces a possible attack vector

datagrams/datagrams.md
jxs: I'd say probably be closed

datagrams/datagrams.md
```diff
@@ -0,0 +1,101 @@
...
+- [Datagrams on QUIC](#datagrams-on-quic)
+  - [Encoding](#encoding)
+- [Datagrams on WebTransport](#datagrams-on-webtransport)
+- [Datagrams on top of a multiplexed stream on top of TCP](#datagrams-on-top-of-a-multiplexed-stream-on-top-of-tcp)
+- [Datagrams with WebRTC](#datagrams-with-webrtc)
+
+## Introduction
+
+This specification defines libp2p datagrams, which enable the transmission of
+MTU-sized, unreliable, and low latency messages between peers.
```
gmelodie, line 34: > the transmission of ... unreliable ... messages between peers

It's not very clear to me what these are. If the definition is the same of QUIC's unreliable datagrams, maybe it's worth it to be explicit here?

Come to think of it... are not all datagrams unreliable? This is a different discussion but the name seems confusing.

datagrams/datagrams.md
```diff
@@ -0,0 +1,101 @@
...
+
+TODO write more about the tradeoffs here as well as the kinds of applications that would benefit from using datagrams.
+
+## Native support at the transport level
+
+Some libp2p transports natively support datagrams, such as QUIC and
+WebTransport. libp2p implementations MUST use the native datagram support to
```
gmelodie, line 41: > *such as* QUIC and WebTransport

But then what follows is a detailed description of how datagrams in QUIC are supposed to be handled. At least to me this seems to telegraph that this document will hold an extensive list of the supported underlying protocols, and how to handle each of them.

In other words, reading this makes me think that following there will be a list of all supported protocols and how to handle them, otherwise they're not supported at all.

So I guess if we'll have particular cases such as QUIC, we must also have a non-particular (i.e. default) case that hopefully most of the underlying protocols will follow

In other words, sth like this:

> such as quic and webtransport
> 
> # Datagrams
> for most protocols this is the case...
> ## Excepctions
> ### QUIC Datagrams
> blabla
> ### another exception

The alternative would be to have an extensive list of supported underlying protos, but to me that would make it harder to extend and the spec would need constant maintenance

datagrams/datagrams.md
```diff
@@ -0,0 +1,101 @@
...
+start sending datagrams at once. There is currently no other use for the control
+stream besides negotiating the application protocol ID. Receipt of a QUIC
+DATAGRAM frame whose payload is too short to allow parsing the Control Stream ID
+field MUST be treated as a connection error of type PROTOCOL_VIOLATION (0x1003).
+libp2p datagrams MUST NOT be sent unless the control stream's send side is open.
+If a datagram is received after the corresponding stream's receive side is
+closed, the received datagrams MUST be silently dropped.
+
+If a libp2p datagram is received and its Control Stream ID field maps to a
+stream that has not yet been created, the receiver SHALL either drop that
```
gmelodie, line 65:
```suggestion
stream that has not yet been created, the receiver MUST either drop that
```

datagrams/datagrams.md
```diff
@@ -0,0 +1,101 @@
...
+awaiting the creation of the corresponding stream.
+
+If a libp2p datagram is received and its Control Stream ID field maps to a
+stream that cannot be created due to client-initiated bidirectional stream
+limits, it MUST be treated as a connection error of type PROTOCOL_VIOLATION
+(0x1003).
+
+### Encoding
+
+Each libp2p datagram SHALL be encoded with the following structure.
```
gmelodie, line 76:
```suggestion
Each libp2p datagram MUST be encoded with the following structure.
```

datagrams/datagrams.md
```diff
@@ -0,0 +1,101 @@
...
+send datagrams.
+
+- QUIC has an Unreliable Datagram Extension [RFC 9221].
+- WebTransport supports Datagrams though HTTP Datagrams [WebTransport draft RFC](https://www.ietf.org/archive/id/draft-ietf-webtrans-http3-12.html#name-datagrams)
+
+## Datagrams on QUIC
+
+Each datagram flow MUST be associated with a control stream. A datagram flow is
+defined as a logical flow of datagrams related to a specific application
+protocol. A control stream is a QUIC bidirectional stream that has negotiated
```
gmelodie, line 51: > A datagram flow is defined as a logical flow of datagrams related to a specific application protocol. A control stream is a QUIC bidirectional stream that has negotiated...

These could be in a *Terminology* section to make it easier to understand when only reading parts of the document.

datagrams/datagrams.md
gmelodie, line 59: nit: wording
```suggestion
stream besides negotiating the application protocol ID. If a DATAGRAM frame's payload cannot be parsed as a Control Stream ID followed by application data, the endpoint MUST close the connection with a PROTOCOL_VIOLATION (0x1003) error.
```

datagrams/datagrams.md
```diff
@@ -0,0 +1,101 @@
...
+the related application protocol ID with a [uvarint] length prefix after
+negotiating the control stream. The control stream MUST stay open for the
+duration of the datagram flow. Implementation MAY create the control stream and
+start sending datagrams at once. There is currently no other use for the control
+stream besides negotiating the application protocol ID. Receipt of a QUIC
+DATAGRAM frame whose payload is too short to allow parsing the Control Stream ID
+field MUST be treated as a connection error of type PROTOCOL_VIOLATION (0x1003).
+libp2p datagrams MUST NOT be sent unless the control stream's send side is open.
+If a datagram is received after the corresponding stream's receive side is
+closed, the received datagrams MUST be silently dropped.
```
gmelodie, line 62: nit: wording
```suggestion
An endpoint MUST NOT send libp2p datagrams unless the send side of the associated control stream is open. An endpoint that receives a datagram after the receive side of the corresponding control stream is closed MUST silently drop it.
```

PR #682, opened 2025-06-05
Add AutoTLS client spec

gmelodie:
This is an attempt to formalize the spec for AutoTLS client as announced in https://blog.libp2p.io/autotls/

**Note:** @kaiserd noted that AutoTLS might be using [DANE](https://www.rfc-editor.org/rfc/rfc6698), but I need some input since I couldn't find any work explicitly linking the two.

### DIFF
--- a/tls/autotls-client.md
+++ b/tls/autotls-client.md
@@ -157,7 +157,10 @@ while "broker" and "AutoTLS broker", which are used interchangeably, refer to th
 
     **Note:** `varint` is a protobuf [varint](https://protobuf.dev/programming-guides/encoding/#varints) field that encodes the length of each of the `key=value` string.
 
-    **Note:** The node SHOULD include only multiaddresses containing public IPv4 addresses in `multiaddrs`.
+    **Note:** The node MUST only include only publicly reachable IP addresses in `multiaddrs`.
+    IPv4 private address ranges MUST be excluded.
+    IPv6 private address ranges and NAT64 translated addresses MUST be excluded.
+    Addresses containing `/p2p-circuit` MUST be excluded.
 	4. Node sends a POST request to `/v1/_acme-challenge` endpoint using `payload` as HTTP body and `headers` as HTTP headers.
 	5. Node SHOULD save the `bearer` token from the `authentication-info` response header, and use it for following requests to the AutoTLS broker.
 


SgtPooki (commented):
just some basic errors found while skimming

tls/autotls-client.md
```diff
@@ -0,0 +1,396 @@
...
+    **Note:** `varint` is a protobuf [varint](https://protobuf.dev/programming-guides/encoding/#varints) field that encodes the length of each of the `key=value` string.
+
+    **Note:** The node SHOULD include only multiaddresses containing public IPv4 addresses in `multiaddrs`.
+	4. Node sends a POST request to `/v1/_acme-challenge` endpoint using `payload` as HTTP body and `headers` as HTTP headers.
+	5. Node SHOULD save the `bearer` token from the `authentication-info` response header, and use it for following requests to the AutoTLS broker.
+
+
+
+## Signalling challenge completion to ACME server
+1. Node SHOULD query DNS records (`TXT _acme-challenge.{b36peerid}.libp2p.direct` and `A dashed-public-ip-address.{b36peerid}.libp2p.direct`) until they are set by the AutoTLS broker.
```
lidel, line 167: no point in querying `A` if `TXT` is not published by the registration endpoint:

```suggestion
1. Node SHOULD query DNS records (`TXT _acme-challenge.{b36peerid}.libp2p.direct`) until they are set by the AutoTLS broker using exponential backoff (double the interval after each failed attempt).
```

tls/autotls-client.md
```diff
@@ -0,0 +1,396 @@
...
+
+        payload = {
+            "value": keyAuthorization,
+            "addresses": multiaddrs
+        }
+	```
+
+    **Note:** `varint` is a protobuf [varint](https://protobuf.dev/programming-guides/encoding/#varints) field that encodes the length of each of the `key=value` string.
+
+    **Note:** The node SHOULD include only multiaddresses containing public IPv4 addresses in `multiaddrs`.
```
lidel, line 160: iirc there is more this (IPv6, filtering):

- The node MUST include only publicly reachable addresses
- For IPv4: exclude private ranges
- For IPv6: exclude private ranges and NAT64 translated addresses
- Exclude relay addresses (containing `/p2p-circuit`)

tls/autotls-client.md
```diff
@@ -0,0 +1,396 @@
...
+
+## Signalling challenge completion to ACME server
+1. Node SHOULD query DNS records (`TXT _acme-challenge.{b36peerid}.libp2p.direct` and `A dashed-public-ip-address.{b36peerid}.libp2p.direct`) until they are set by the AutoTLS broker.
+
+**Note:** Here, `dashed-public-ip-address` is the public IPv4 address on in which the node received the confirmation dial from the broker.
+For example, if the node has two public IPv4 addresses `1.1.1.1` and `8.8.8.8`, and the broker dialed it through `1.1.1.1`, then the node SHOULD query the `A 1-1-1-1.{b36peerid}.libp2p.direct`.
+
+**Note:** The node SHOULD NOT send more than `max_dns_retries` DNS requests.
+After `max_dns_timeout`, the communication is considered failed.
+What to do after `max_dns_timeout` has passed is left as an implementation decision.
```
lidel, line 174: nit: imo the spec should clearly say that at this point the certificate request flow SHOULD be aborted and retried later.
 
(Do not attempt to ask ACME for cert if DNS TXT record is not confirmed to exist – we don't want to hammer Let's Encrypt with obviously failing requests)

tls/autotls-client.md
```diff
@@ -0,0 +1,396 @@
...
+## Downloading certificate
+1. Node finalizes the certificate request:
+    1. Generate CSR for the `*.{b36peerid}.libp2p.direct` domain.
+    2. Encode the CSR with URL safe base 64 (`b64CSR`).
+    3. Send a `kid` signed POST request to `finalizeUrl` with JSON HTTP body of `{"csr": b64CSR}`.
+2. Node MUST poll ACME server by sending GET requests to `orderUrl` until the ACME server's response contains a `status` field with a value different than `processing`.
+
+**Note:** The node SHOULD NOT send more than `max_acme_poll_retries` poll requests to the ACME server.
+After `max_acme_timeout`, the communication has failed.
+What to do after `max_acme_timeout` has passed is left as an implementation decision.
```
lidel, line 196: Similar ask: spec should clearly note that there should be exponential backoff for retry attempts.

tls/autotls-client.md
```diff
@@ -0,0 +1,396 @@
...
+6. Node authenticates with AutoTLS broker at `"https://registration.libp2p.direct/v1/_acme-challenge"` (refer to the [peer ID authentication spec](https://github.com/libp2p/specs/blob/master/http/peer-id-auth.md) for guidance and examples) and sends the following JSON body:
+    ```json
+    {
+      "value": "jP5hwrZwCbP_qeeET_qAa9pgG0YulNaR0ivruESzCrE",
+      "addresses": [
+        "/ip4/142.93.194.175/tcp/49309"
+      ]
+    }
+    ```
+
```
lidel, line 282:
```suggestion

**User-Agent Header**: Implementations SHOULD include a descriptive User-Agent header that identifies the client library name and version or the host application (e.g., "kubo/0.17.0", "helia/2.0.0"). 

```

tls/autotls-client.md
```diff
@@ -0,0 +1,396 @@
...
+	4. Node sends a POST request to `/v1/_acme-challenge` endpoint using `payload` as HTTP body and `headers` as HTTP headers.
+	5. Node SHOULD save the `bearer` token from the `authentication-info` response header, and use it for following requests to the AutoTLS broker.
+
+
+
+## Signalling challenge completion to ACME server
+1. Node SHOULD query DNS records (`TXT _acme-challenge.{b36peerid}.libp2p.direct` and `A dashed-public-ip-address.{b36peerid}.libp2p.direct`) until they are set by the AutoTLS broker.
+
+**Note:** Here, `dashed-public-ip-address` is the public IPv4 address on in which the node received the confirmation dial from the broker.
+For example, if the node has two public IPv4 addresses `1.1.1.1` and `8.8.8.8`, and the broker dialed it through `1.1.1.1`, then the node SHOULD query the `A 1-1-1-1.{b36peerid}.libp2p.direct`.
```
lidel, line 170:
```suggestion
For example, if the node has two public IPv4 addresses `1.1.1.1` and `8.8.8.8`, and the broker dialed it through `1.1.1.1`, then the node SHOULD query the `A 1-1-1-1.{b36peerid}.libp2p.direct`.
For IPv6: Replace colons with dashes, handle leading/trailing colons as zeros (`2001:db8::1` → `2001-db8--1`, `2001:db8::` → `2001-db8--0`), and query `AAAA 2001-db8--1.{b36peerid}.libp2p.direct`.
```

tls/autotls-client.md
```diff
@@ -0,0 +1,396 @@
...
+6. ACME server checks challenge in broker.
+7. Node sends CSR to finalize certificate request.
+8. Node polls ACME server until certificate is ready for download.
+9. Node downloads certificate.
+
+## Parameters
+
+| Parameter                | Description                                                      | Reasonable Default |
+|--------------------------|------------------------------------------------------------------|--------------|
+| `max_dns_retries` | The maximum number of DNS queries that the node SHOULD make before giving up | ???  |
```
lidel, line 61: `p2p-forge/client` uses time-based timeout (3 minutes) with exponential backoff rather than fixed retry counts

tls/autotls-client.md
```diff
@@ -0,0 +1,396 @@
...
+| `max_dns_timeout` | The maximum number of seconds a node SHOULD wait for DNS records to be set | ???  |
```
lidel, line 62: iirc `p2p-forge/client` gives up waiting for TXT record after 3min I think -- at this point something went wrong and the entire registration needs to be resumed

tls/autotls-client.md
```diff
@@ -0,0 +1,396 @@
...
+| `max_acme_poll_retries` | The maximum number of GET requests that the node SHOULD issue to ACME server before giving up | ???  |
```
lidel, line 63: ACME specs recommend polling until timeout rather than fixed retry counts

tls/autotls-client.md
```diff
@@ -0,0 +1,396 @@
...
+| `max_acme_timeout` | The maximum number of seconds a node SHOULD wait for an ACME resource status to change | ???  |
```
lidel, line 64: Let's Encrypt recommendation for challenge completion is ~10minutes, p2p-forge client aborts sooner, around 3 minutes I think

lidel (changes requested):
@gmelodie   Do we have a third-party implementation of this spec other than two Shipyard ones? 
- JS: https://www.npmjs.com/package/@libp2p/auto-tls
- GO: https://github.com/ipshipyard/p2p-forge/tree/main/client

Note that Shipyard did not write this PR. Personally I would prefer to wait until 
- someone actually tries to implement the client independently, by only following this document
- there is at least one server implementation with clear licensing situation (https://github.com/ipshipyard/p2p-forge has no license, only client library has https://github.com/ipshipyard/p2p-forge/blob/main/client/LICENSE.md)

As for the spec proposed in this document, some missing things:
- guidance on sane timeouts and exponential backoff
- IPv6 

(details inline)

ps. AutoTLS is not using DANE (spec does not claim that, but the PR description mentions it).

gmelodie:
> Do we have a third-party implementation of this spec other than the two created by [Shipyard](https://ipshipyard.com/) team?

Yes! There's the [`nim-libp2p` implementation](https://github.com/vacp2p/nim-libp2p/tree/master/libp2p/autotls)

> (bare minimum) someone actually tries to implement AutoTLS client independently, by only following this document, providing feedback if spec is complete

We actually wrote the document based on the implementation that we did.

> ps. AutoTLS is not using DANE (spec does not claim that, but the PR description mentions it).

Thanks for clarifying!

@lidel can you check again please?

seetadev:
Hi @lidel,

Hope you are doing well. We’ve been discussing AutoTLS in the recent libp2p spec meetings. As a follow-up, @gmelodie has updated the example implementation based on our discussions.

Please share your review and feedback whenever convenient.

Thanks you for your review and feedback. Appreciate it.

aschmahmann:
Hi folks, thanks for your patience on this.

Unfortunately, Shipyard’s available time for libp2p maintenance work is quite limited these days (see our blog post: https://ipshipyard.com/blog/2025-libp2p-maintenance-update/).

# Does the Client Spec Look OK?
We don’t have the capacity for the kind of in-depth review we’d normally allocate to spec work in IPFS or our prior work in libp2p. That said, the spec appears to be generally sound and if it enables two additional AutoTLS clients to obtain certificates via libp2p.direct, that would be a net positive outcome.

# Does It Make Sense to Merge This PR?
In my view, it’s a bit awkward to include a libp2p spec that, in practice, is just an HTTP API for working with libp2p.direct through the registration.libp2p.direct endpoint. I’d be supportive of improved documentation for the registration endpoint, perhaps linked from p2p-forge/README, to help others implement their own AutoTLS clients. But I’m skeptical that this belongs in libp2p/specs, given the current context. That being said I have no objections to merging if you want so I'm giving an approval with that it mind. If you want it merged and no one else has permissions ping me and I'll hit the button.

# Important Context on AutoTLS / libp2p.direct
With the changes at Shipyard, our ability to support broader libp2p use cases is now largely opportunistic, rather than being part of a coordinated, project-wide mandate as it was previously.

seetadev:
@aschmahmann : Thank you so much, Adin — appreciate you taking the time to review this and share such a thoughtful perspective, especially given Shipyard’s current bandwidth and focus areas. 

Also, wish to thank @lidel for sharing his detailed feedback and pointers.

PR #684, opened 2025-06-23
Gossipsub v1.3: Extensions control Message

MarcoPolo:
This lets us try more extensions and changes without requiring a new Gossipsub version. Hopefully it makes future extensions easier to iterate on, use, and deploy. More details in the doc

There is a lot of prior art here as many modern protocols have something like this, QUIC's transport parameters and H{3,2} SETTINGS frames to name a few.

### DIFF
--- a/pubsub/gossipsub/gossipsub-v1.3.md
+++ b/pubsub/gossipsub/gossipsub-v1.3.md
@@ -1,0 +1,137 @@
+# gossipsub v1.3: Extensions Control Message
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+| --------------- | ------------- | ------ | --------------- |
+| 1A              | Working Draft | Active | r0, 2025-06-23  |
+
+Authors: [@marcopolo]
+
+Interest Group: TODO
+
+[@marcopolo]: https://github.com/marcopolo
+
+See the [lifecycle document][lifecycle-spec] for context about the maturity level
+and spec status.
+
+[lifecycle-spec]: https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md
+
+## Overview
+
+This version specifies a way to for gossipsub peers to describe their
+characteristics to each other without requiring a new protocol ID per extension.
+
+This spec MUST be updated upon introducing a new extension, either canonical or
+experimental, to the network.
+
+## Motivation
+
+This version makes Gossipsub easier to extend by allowing applications to
+selectively make use of the extensions it would benefit from. It removes the
+need to make Gossipsub extensions follow a strict ordering. Finally, it allows
+extensions to iterate independently from Gossipsub's versioning.
+
+## The Extensions Control Message
+
+If a peer supports any extension, the Extensions control message MUST be
+included in the first message on the stream. An Extensions control message MUST
+NOT be sent more than once. If a peer supports no extensions, it may omit
+sending the Extensions control message.
+
+Extensions are not negotiated; they describe characteristics of the sending peer
+that can be used by the receiving peer. However, a negotiation can be implied:
+each peer uses the Extensions control message to advertise a set of supported
+values. The specification of an extension describes how each peer combines the
+two sets to define its behavior.
+
+Peers MUST ignore unknown extensions.
+
+Extensions that modify or replace core protocol functionality will be difficult
+to combine with other extensions that modify or replace the same functionality
+unless the behavior of the combination is explicitly defined. Such extensions
+SHOULD define their interaction with previously defined extensions modifying the
+same protocol components.
+
+## Protocol ID
+
+The Gossipsub version for this specification is `v1.3.0`. The protocol ID is
+`/meshsub/1.3.0`.
+
+## Process to add a new extensions to this spec
+
+### Canonical Extensions
+
+A Canonical Extension is an extension that is well defined, has multiple
+implementations, has shown to be useful in real networks, and has rough
+consensus on becoming a canonical extension. The extension specification MUST be
+defined in the `libp2p/specs` GitHub repo. After an extension meets the stated
+criteria, this specification MUST be updated to include the extension in the
+`ControlExtensions` protobuf with a link to the extension's specification doc in
+a comment. The extension SHOULD use the next lowest available field number.
+
+Any new messages defined by the extension MUST be added to `RPC` message
+definition in this spec. Extensions SHOULD minimize the number of new messages
+they introduce here. Try to introduce a single new message, and use that message
+as a container for more messages similar to the strategy used by the
+ControlMessage in the RPC.
+
+All extension messages MUST be an `optional` field.
+
+### Experimental Extensions
+
+In contrast with a Canonical Extension, an Experimental Extension is still being
+evaluated and iterated upon. Adding an experimental extension to this spec lets
+others see what is being tried, and ensure there are no misinterpretations of
+messages on the wire. A patch to this spec is not needed if experimenting with
+an extension in a controlled environment. A patch to this spec is also not
+needed if you are not using the `/meshsub/v1.3.0` protocol ID.
+
+If the extension is being tested on a live network, a PR MUST be created that
+adds the extension to the `ControlExtensions` protobuf with a link to the
+extension's specification. Experimental extensions MUST use a large field number
+randomly generated to be at least 4 bytes long when varint encoded. The
+extension author MUST ensure this field number does not conflict with any
+existing field.
+
+New messages defined by this extension should follow the same guidelines as new
+messages for canonical extensions. Field numbers MUST be randomly generated and
+be at least 4 bytes long when varint encoded.
+
+Maintainers MUST check that the extension is well specified, in the experimental
+range, and that the extension will be tested on a live network. If so,
+maintainers SHOULD merge the change.
+
+## Protobuf
+
+```protobuf
+syntax = "proto2";
+
+message ControlExtensions {
+  // Initially empty. Future extensions will be added here along with a
+  // reference to their specification.
+
+  // Experimental extensions must use field numbers larger than 0x200000 to be
+  // encoded with at least 4 bytes
+}
+
+message ControlMessage {
+	repeated ControlIHave ihave = 1;
+	repeated ControlIWant iwant = 2;
+	repeated ControlGraft graft = 3;
+	repeated ControlPrune prune = 4;
+	repeated ControlIDontWant idontwant = 5;
+	ControlExtensions extensions = 6;
+}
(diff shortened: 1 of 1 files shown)


pubsub/gossipsub/gossipsub-v1.3.md
```diff
@@ -0,0 +1,137 @@
+# gossipsub v1.3: Extensions Control Message
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+| --------------- | ------------- | ------ | --------------- |
+| 1A              | Working Draft | Active | r0, 2025-06-23  |
+
+Authors: [@marcopolo]
+
+Interest Group: TODO
```
MarcoPolo, line 9: Please comment here if you'd like to be included in this interest group

cortze: ✋🏽

AgeManning:
I like the concept and the idea, I have some questions though. 

Thinking out-loud there is a bit of engineering complexity with this. The new control message MUST be sent as the first message and MUST only be sent once. As the protobuf allows these to be sent at any time, there is a bit of annoyance that we have to check its the first message, hasn't been sent a second time and reject all messages that may contain this extra control message. 

The reason I bring this up, is that in rust-libp2p, we already have all of this machinery for protocol-id. It can already only be sent once and we keep of track of each peer and what they support. Peers can send us a list of protocol-ids. 

Can't we have the same functionality where a peer supports say:
- `/meshsub/1.1.0`
- `/meshsub/1.0.0`
- `/extension/17`
- `/extension/52`

I guess the obvious downside is that multistream select chooses the "best" protocol-id to match on. So it would limit interactions between extensions, but a workaround would be choosing a different extension number that represents the interaction of multiple extensions, i.e extension 3 could mean support for extension 1 and 2. 

I only raise this, because I think this mentioned approach requires 0 engineering effort on our behalf and we dont have to add a control message. 

I'm not against adding this, just wanted to raise this as potentially a simpler possibility.

pubsub/gossipsub/gossipsub-v1.3.md
cskiraly: ✋🏽

vyzo:
We can put the extensions supported in the hello RPC, that should be backwards compatible (since it is protobuf, older clients can ignore).

AgeManning:
> I'm not opposed to continue discussing using the protocol ID as maybe I've missed something. But I also want to hear more about what makes this proposal difficult to implement.

I dont think this is difficult to implement. Just thought maybe we could get away with doing nothing by piggybacking the protocol-id. It seems like we cant. 

Curious if we have some extensions we might want to try out already?

MarcoPolo:
> I still think we should stop adding version numbers in PR titles. It is confusing and versioning should not be done this way. Having said that, this would be good for 1.3, as it unlocks many other improvements.

Generally agree. I initially included it as a signal on how I think the various extensions should be prioritized. I've removed it to separate the versioning discussion.

> Finally, some nitpicking: `Extensions are not negotiated` : I would phrase this differently to avoid misunderstanding. E.g. `Extension negotiation is out of scope for this specification`.

What is the misunderstanding? I think the text is accurate. This is the same language as what is used in the HTTP/3 SETTINGS RFC (which I liberally copied from). That RFC states:

> SETTINGS parameters are not negotiated; they describe characteristics of the sending peer that can be used by the receiving peer. However, a negotiation can be implied by the use of SETTINGS: each peer uses SETTINGS to advertise a set of supported values. The definition of the setting would describe how each peer combines the two sets to conclude which choice will be used. SETTINGS does not provide a mechanism to identify when the choice takes effect.

https://datatracker.ietf.org/doc/html/rfc9114#section-7.2.4-4

BoomchainLabs (approved):

cskiraly:
> What is the misunderstanding? I think the text is accurate. This is the same language as what is used in the HTTP/3 SETTINGS RFC (which I liberally copied from). That RFC states:
> 
> > SETTINGS parameters are not negotiated; they describe characteristics of the sending peer that can be used by the receiving peer. However, a negotiation can be implied by the use of SETTINGS: each peer uses SETTINGS to advertise a set of supported values. The definition of the setting would describe how each peer combines the two sets to conclude which choice will be used. SETTINGS does not provide a mechanism to identify when the choice takes effect.
> 
> https://datatracker.ietf.org/doc/html/rfc9114#section-7.2.4-4

I see. You are right, it is hard to misunderstand (where the misunderstanding I wanted to avoid is to see this an an explicit ban of negotiating the use of extensions, which is clarified by the "however ... " part).

pubsub/gossipsub/gossipsub-v1.3.md
ppopth: me too please

pubsub/gossipsub/gossipsub-v1.3.md
```diff
@@ -0,0 +1,137 @@
...
+## Motivation
+
+This version makes Gossipsub easier to extend by allowing applications to
+selectively make use of the extensions it would benefit from. It removes the
+need to make Gossipsub extensions follow a strict ordering. Finally, it allows
+extensions to iterate independently from Gossipsub's versioning.
+
+## The Extensions Control Message
+
+If a peer supports any extension, the Extensions control message MUST be
```
raulk, line 35: We should future-proof this in case we move away from the single-stream (per direction) design. For context: we speculate that more granular streams on top of QUIC can alleviate transport HoL blocking and can enable more fluid cancellation. We've discussed:

- one stream per subscription (+ control stream)
- one stream per message (+ control stream)

pubsub/gossipsub/gossipsub-v1.3.md
```diff
@@ -0,0 +1,137 @@
...
+range, and that the extension will be tested on a live network. If so,
+maintainers SHOULD merge the change.
+
+## Protobuf
+
+```protobuf
+syntax = "proto2";
+
+message ControlExtensions {
+  // Initially empty. Future extensions will be added here along with a
```
raulk, line 109: Let's describe field expectations and provide some examples. For example, some extensions may simply advertise themselves as boolean flags, while others may send preamble data/config.

pubsub/gossipsub/gossipsub-v1.3.md
```diff
@@ -0,0 +1,137 @@
...
+experimental, to the network.
+
+## Motivation
+
+This version makes Gossipsub easier to extend by allowing applications to
+selectively make use of the extensions it would benefit from. It removes the
+need to make Gossipsub extensions follow a strict ordering. Finally, it allows
+extensions to iterate independently from Gossipsub's versioning.
+
+## The Extensions Control Message
```
raulk, line 33: Why not advertise the extensions during the `HELLO` exchange?

pubsub/gossipsub/gossipsub-v1.3.md
MarcoPolo: I don't think we need any text change here. We'll always have an initial stream and the extensions MUST be in the first message on that stream.

The best next optimization for multiple streams would be to open a stream per message. I don't think multiple control streams are worth it because it is simpler if control messages are serialized and control messages are small.

Aside, it would be great to use bidirectional streams as bidirectional. There is implementation complexity in correctly treating two separate streams as a single bidirectional stream (the current go implementation has bugs here).

pubsub/gossipsub/gossipsub-v1.3.md
MarcoPolo: We have a real example with the TestExtension. We'll also have another one soon with Partial messages.

pubsub/gossipsub/gossipsub-v1.3.md
MarcoPolo: Hello is an implementation detail, there's no "HELLO" protobuf message. It is shorthand for sending the topics we are subscribed to. If an implementation wants to include it in their "hello packet", that's fine. But there's no need to reference the hello packet here. The requirement is simply that this MUST be the first message.

pubsub/gossipsub/gossipsub-v1.3.md
raulk: Those won't be linked from here. What I'm saying is that this spec defines a framework for extensions, and it feels incomplete without an explanation of how extensions would _actually_ represent themselves, other than "here's a schema where they can drop something". Some stereotypical examples would help frame this right.

MarcoPolo: It will be linked here. As well as having the code points defined here.

raulk: Can you explain your motivation for wanting to inventorize all eventual extensions here? The convention is to have the spec define the mechanism for the **extension point**, while managing the assignment in a separate resource (registry).

This separation between mechanism and registry is true for: [multicodec](https://github.com/multiformats/multicodec/blob/master/table.csv), [multiaddr](https://github.com/multiformats/multiaddr/blob/master/protocols.csv), and major industry standards like CBOR (tags), TLS (params), MIME (media types), etc. where the base spec is an IETF RFC, extensions are separate RFCs, and their ID assignment is tracked by IANA registries.

I'm happy to be convinced that deviating from convention is appropriate here, but if there isn't a clear case, I'd say we have this spec establish the registry (e.g. a local csv file), together with the test extension (#686), and the v1 of the registry containing the latter.

pubsub/gossipsub/gossipsub-v1.3.md
raulk: That makes sense; the current wording works for me. I think there's opportunity to formalize the initial exchange a bit better—but that's outside the scope of this PR.

pubsub/gossipsub/gossipsub-v1.3.md
MarcoPolo: I don't have a strong preference here. We can create a separate file to serve as registry. (although it should be a `.proto` rather than a `.csv`, as it's a more direct mapping.)

My motivation to keep it in one file is to keep things simple. There are tradeoffs in maintaining a separate registry. IANA works well because an RFC is a published document that is more or less complete, and the considerations to adopting a new codepoint need to hashed out elsewhere (the WG might not even still exist). The downside is that there is more overhead to a separate registry (we can mitigate that with a file within this repo, such that changes can be atomic to a PR).

While I don't have a strong preference, I don't see much benefit in having a separate registry file for this use case. Can you elaborate on what you expect the benefits to be?

For some prior art in this repo, we have [Noise Extensions](https://github.com/libp2p/specs/tree/master/noise#the-libp2p-handshake-payload) that does not maintain a separate registry.

PR #685, opened 2025-06-25
Gossipsub: Partial Message Extension

MarcoPolo:
All the info is in the spec.

### DIFF
--- a/pubsub/gossipsub/extensions/extensions.proto
+++ b/pubsub/gossipsub/extensions/extensions.proto
@@ -25,6 +25,11 @@ message RPC {
   message SubOpts {
     optional bool subscribe = 1; // subscribe or unsubcribe
     optional string topicid = 2;
+
+    // Used with Partial Messages extension.
+    // If set, the receiver of this message MUST send partial messages to the
+    // sender instead of full messages.
+    optional bool partial = 3;
   }
 
   repeated SubOpts subscriptions = 1;
@@ -40,4 +45,25 @@ message RPC {
   optional TestExtension testExtension = 6492434;
 
   optional PartialMessagesExtension partial = 16418754;
+}
+
+message PartialMessagesExtension {
+  optional bytes topicID = 1;
+  optional bytes groupID = 2;
+
+  optional PartialMessage message = 3;
+  optional PartialIWANT iwant = 4;
+  optional PartialIHAVE ihave = 5;
+}
+
+message PartialMessage {
+  optional bytes data = 1;
+}
+
+message PartialIWANT {
+  optional bytes metadata = 1;
+}
+
+message PartialIHAVE {
+  optional bytes metadata = 1;
 }
--- a/pubsub/gossipsub/extensions/extensions.proto
+++ b/pubsub/gossipsub/extensions/extensions.proto
@@ -51,19 +51,9 @@ message PartialMessagesExtension {
   optional bytes topicID = 1;
   optional bytes groupID = 2;
 
-  optional PartialMessage message = 3;
-  optional PartialIWANT iwant = 4;
-  optional PartialIHAVE ihave = 5;
-}
+  // An encoded partial message
+  optional bytes partialMessage = 3;
 
-message PartialMessage {
-  optional bytes data = 1;
-}
-
-message PartialIWANT {
-  optional bytes metadata = 1;
-}
-
-message PartialIHAVE {
-  optional bytes metadata = 1;
+  // An encoded representation of the parts a peer has and wants.
+  optional bytes partsMetadata = 4;
 }
--- a/pubsub/gossipsub/extensions/extensions.proto
+++ b/pubsub/gossipsub/extensions/extensions.proto
@@ -25,9 +25,13 @@ message RPC {
     optional string topicid = 2;
 
     // Used with Partial Messages extension.
-    // If set, the receiver of this message MUST send partial messages to the
-    // sender instead of full messages.
-    optional bool partial = 3;
+    // If set to true, signals to the receiver that the sender prefers partial
+    // messages.
+    optional bool requestsPartial = 3;
+    // If set to true, signals to the receiver that the sender supports sending
+    // partial messages on this topic. If requestsPartial is true, this is
+    // assumed to be true.
+    optional bool supportsSendingPartial = 4;
   }
 
   repeated SubOpts subscriptions = 1;
--- a/pubsub/gossipsub/extensions/extensions.proto
+++ b/pubsub/gossipsub/extensions/extensions.proto
@@ -8,6 +8,8 @@ message ControlExtensions {
   // encoded with at least 4 bytes
 
   optional bool testExtension = 6492434;
+
+  optional bool partialMessages = 16418754;
 }
 
 message ControlMessage {
@@ -36,4 +38,6 @@ message RPC {
   // bytes
 
   optional TestExtension testExtension = 6492434;
+
+  optional PartialMessagesExtension partial = 16418754;
 }
--- a/pubsub/gossipsub/extensions/extensions.proto
+++ b/pubsub/gossipsub/extensions/extensions.proto
@@ -1,15 +1,13 @@
 syntax = "proto2";
 
 message ControlExtensions {
-  // Initially empty. Future canonical extensions will be added here along with
-  // a reference to their specification.
+    optional bool partialMessages = 10;
 
   // Experimental extensions must use field numbers larger than 0x200000 to be
   // encoded with at least 4 bytes
 
   optional bool testExtension = 6492434;
 
-  optional bool partialMessages = 16418754;
 }
 
 message ControlMessage {
@@ -37,6 +35,7 @@ message RPC {
   optional ControlMessage control = 3;
 
   // Canonical Extensions should register their messages here.
+  optional PartialMessagesExtension partial = 10;
 
   // Experimental Extensions should register their messages here. They
   // must use field numbers larger than 0x200000 to be encoded with at least 4
@@ -44,7 +43,6 @@ message RPC {
 
   optional TestExtension testExtension = 6492434;
 
-  optional PartialMessagesExtension partial = 16418754;
 }
 
 message PartialMessagesExtension {
--- a/pubsub/gossipsub/partial-messages.md
+++ b/pubsub/gossipsub/partial-messages.md
@@ -80,51 +80,27 @@ then partial messages would be the same as smaller messages.
 
 The following section specifies the semantics of each new protocol message.
 
-### PartialIWANT
+### partialMessage
 
-A `PartialIWANT` signal to a receiver that the sending peer only wants a part of
-some message.
+The `partialMessage` field encodes one or more parts of the full message. The
+encoding is application defined.
 
-The message to which the peer is requesting a part of is identified by the
-`groupID` identifier. This is similar to a complete message's `MessageID`, but,
-in contrast to a content-based message id, does not require the full message to
-compute. For example, in the Ethereum use case, this could simply be the hash of
-the signed block header.
+### partsMetadata
 
-The `topicID` references the Gossipsub topic a message, and thus its parts,
-belong to.
+The `partsMetadata` field encodes the parts a peer has and wants. The encoding
+is application defined. An unset value carries no information besides that the
+peer did not send a value.
 
-The `metadata` field is opaque application defined metadata associated with this
-request. This can be a bitmap, a list of ranges, or a bloom filter. The
-application generates this and consumes this.
+Upon receiving a `partsMetadata` a node SHOULD respond with only parts the peer
+wants.
 
-A later `PartialIWANT` serve to refine the request of prior a prior `PartialIWANT`.
+A later `partsMetadata` replaces a prior one.
 
-Nodes SHOULD assume a `PartialIWANT` implies a `IDONTWANT` for the full message.
-
-### PartialIHAVE
-
-A `PartialIHAVE` allows nodes to signal HAVE information before receiving all
-segments, unlocking the use of `PartialIWANT` in more contexts.
-
-In the context of partial messages, it is more useful than IHAVE as it includes
-the group ID. In contrast, an IHAVE does not. A receiving peer has no way to
-link an IHAVE's message ID with a group ID, without having the full message.
-
-A `PartialIHAVE` message can be used both in the context of lazy push, notifying
-peers about available parts, and in the context of heartbeats as a replacement
-to IHAVEs.
-
-The structure of `PartialIHAVE` is analogous to that of `PartialIWANT`.
-
-The metadata, as in a `PartialIWANT`, is application defined. It is some encoding
-that represents the parts the sender has.
+`partsMetadata` can be used during heartbeat gossip to inform non-mesh topic
+peers about parts this node has.
 
 Implementations are free to select when to send an update to their peers based
 on signaling bandwidth tradeoff considerations.
-
-Receivers MUST treat a `PartialIHAVE` as a signal that the peer does not want
-the indicated part.
 
 ### Changes to `SubOpts` and interaction with the existing Gossipsub mesh.
 
@@ -144,24 +120,23 @@ The partial field value MUST be ignored when a peer sends an unsubscribe message
 
 ## Application Interface
 
-Message contents are application defined. Thus splitting a message must be
+This specific interface is not intended to be normative to implementations, it
+is only an example of one possible API.
+
+Message contents are application defined, thus splitting a message must be
 application defined. Applications should provide a Partial Message type that
 supports the following operations:
 
 1. `.GroupID() -> GroupID: bytes`
-2. `.PartialMessageBytesFromMetadata(metadata: bytes) -> Result<(EncodedPartialMessage: bytes, metadata: bytes), Error>` (When responding to a `PartialIWANT` or eagerly pushing a partial message)
-  a. The returned metadata represents the still missing parts. For example, if a
-     peer is only able to fulfill a part of the the request, the returned
-     metadata represents the parts it couldn't fulfill.
-3. `.ExtendFromEncodedPartialMessage(data: bytes) -> Result<(), Error>` (When receiving a `PartialMessage`)
-4. `.MissingParts() -> Result<metadata: bytes, Error>` (For `PartialIWANT`)
-5. `.AvailableParts() -> Result<metadata: bytes, Error>` (For `PartialIHAVE`)
+2. `.PartialMessageBytes(partsMetadata: bytes) -> Result<(EncodedPartialMessage: bytes, newPartsMetadata: bytes), Error>`
+  a. The method should return an encoded partial message with just the parts the
+     peer requested.
+  b. The returned `newPartsMetadata` can be used to track parts that could not
+     be fulfilled. This allows the GossipSub library to avoid sending duplicate
+     parts to the same peer.
+3. `.PartsMetadata() -> bytes`
 
 Gossipsub in turn provides a `.PublishPartial(PartialMessage)` method.
-
-Note that this specific interface is not intended to be normative to
-implementations, rather, it is high level summary of what each layer should
-provide.
 
 ## Protobuf
 
--- a/pubsub/gossipsub/partial-messages.md
+++ b/pubsub/gossipsub/partial-messages.md
@@ -143,12 +143,5 @@ Gossipsub in turn provides a `.PublishPartial(PartialMessage)` method.
 
 Refer to the protobuf registry at `./extensions/extensions.proto`
 
-## Open Questions
-
-- Do we want to add a TTL to PartialIWANTs? This would allow us to cancel them after some time.
-- Should we rename the metadata bytes to iwant and ihave?
-- In the bitmap usecase, iwant/ihave are simply inverses of each other. Do we need to send them both?
-- There's a bit of extra complexity around assuming opaque metadata, is it worth it?
-
 [1]: https://ethresear.ch/t/is-data-available-in-the-el-mempool/22329
 [2]: https://ethresear.ch/t/fulldas-towards-massive-scalability-with-32mb-blocks-and-beyond/19529#possible-extensions-13
--- a/pubsub/gossipsub/partial-messages.md
+++ b/pubsub/gossipsub/partial-messages.md
@@ -186,6 +186,39 @@ parts or forwarding the message. Both are done with `.PublishPartial`.
 Gossipsub MUST forward all messages to the application, not just messages from
 peers.
 
+## Upgrading a topic to use partial messages
+
+Rolling out partial messages on an existing topic allows for incremental
+migration with backwards compatibility. The steps are as follows:
+
+1. Deploy nodes that support partial messages, but do not request them for the
+   target topic. The goal is to seed support for partial messages before making
+   the switch. Nodes signal their support for partial messages by setting the
+   subscribe option `supportsSendingPartial` to true.
+2. Slowly deploy and monitor nodes that request (and implicitly support) partial
+   messages. These nodes should find peers that send them partial messages from
+   the previous step. Nodes request partial messages by setting the subscribe
+   option `requestPartial` to true.
+
+### Supporting both full and partial messages for a topic
+
+Partial messages use the same mesh as "full" messages. Supporting both is
+straightforward. If a peer subscribes to a topic with a "requestPartial", the
+node SHOULD send the peer partial messages. Otherwise, send the node full
+messages.
+
+On the receiving side, if the node is in a mixed network of partial and full
+messages, and it requests partial messages, the node MUST support receiving full
+messages.
+
+## Creating a topic to only use partial messages
+
+There is currently no mechanism to specify a topic should only be used for
+partial messages. A future extension may define this.
+
+With this extension nodes can choose to only graft peers that support partial
+messages, and prune those that do not.
+
 ## Protobuf
 
 Refer to the protobuf registry at `./extensions/extensions.proto`
--- a/pubsub/gossipsub/partial-messages.md
+++ b/pubsub/gossipsub/partial-messages.md
@@ -162,19 +162,16 @@ application defined. Applications should provide a Partial Message type that
 supports the following operations:
 
 1. `.GroupID() -> GroupID: bytes`
-2. `.PartialMessageBytes(partsMetadata: bytes) -> Result<(EncodedPartialMessage: bytes, newPartsMetadata: bytes), Error>`
+2. `.PartialMessageBytes(partsMetadata: bytes) -> Result<(EncodedPartialMessage: bytes), Error>`
    1. The method should return an encoded partial message with just the parts the
       peer requested.
-   2. The returned `newPartsMetadata` can be used to track parts that could not
-      be fulfilled. This allows the GossipSub library to avoid sending duplicate
-      parts to the same peer.
 3. `.PartsMetadata() -> bytes`
 
 Gossipsub in turn provides a `.PublishPartial(PartialMessage, PartialPublishOptions)` method.
 
 The `PartialPublishOptions` contains:
 
-1. Any eager data that should be pushed to peers who haven't sent us a bitmap yet.
+1. Optional eager data that should be pushed to peers who haven't sent us a bitmap yet.
 2. An optional list of peers to publish to instead of the topic mesh peers.
     1. This is useful for responding to peers who are not in the node's mesh, but
       sent the node a PartialMessage (e.g similar to Gossipsub's `IHAVE`)
--- a/pubsub/gossipsub/partial-messages.md
+++ b/pubsub/gossipsub/partial-messages.md
@@ -126,6 +126,22 @@ on signaling bandwidth tradeoff considerations.
 Receivers MUST treat a `PartialIHAVE` as a signal that the peer does not want
 the indicated part.
 
+### Changes to `SubOpts` and interaction with the existing Gossipsub mesh.
+
+Partial Messages uses the same mesh as normal Gossipsub messages. It is a
+replacement to "full" messages. A node requests a peer to use partial messages
+for a specific topic by setting the `partial` field in the `SubOpts` message.
+The `SubOpts` message is how a peer subscribes to a topic.
+
+If a node receives a subscribe request with the `partial` field set to true, it
+MUST send partial messages instead of full messages.
+
+It is an error to set the partial field true if the peer does not support
+partial extensions.
+
+The partial field value MUST be ignored when a peer sends an unsubscribe message
+`SubOpts.subscribe=false`.
+
 ## Application Interface
 
 Message contents are application defined. Thus splitting a message must be
@@ -149,31 +165,7 @@ provide.
 
 ## Protobuf
 
-```protobuf
-syntax = "proto2";
-
-message PartialMessagesExtension {
-  optional bytes topicID = 1;
-  optional bytes groupID = 2;
-
-  optional PartialMessage message = 3;
-  optional PartialIWANT iwant = 4;
-  optional PartialIHAVE ihave = 5;
-}
-
-message PartialMessage {
-  optional bytes data = 1;
-}
-
-message PartialIWANT {
-  optional bytes metadata = 1;
-}
-
-message PartialIHAVE {
-  optional bytes metadata = 1;
-}
-
-```
+Refer to the protobuf registry at `./extensions/extensions.proto`
 
 ## Open Questions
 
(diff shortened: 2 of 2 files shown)


cskiraly:
@MarcoPolo proposed improvements to the partial message spec here 
https://github.com/cskiraly/libp2p-specs/tree/csaba/partial-message

In my version I had both partialIHAVE and partialIWANT.

I think I see your point in focusing on the IWANT part, as it might allow us to delegate all the logic to the upper layer, similar to how you can do partial requests in other protocols where you need pieces only (e.g. HTTP Accept-Ranges for resume, and similar use cases). We still need the other node to send the IHAVE (the normal, full IHAVE, and we still would need the logic in our node not to fetch the whole message, so either a hook before IWANT or some call to say "for this I already have something, please send a PartialIWANT with this extra metadata".

If instead we also define the Partial IHAVE, we can do more. In the FullDAS implementation I'm doing these partial IHAVEs to some extent already, because I'm segmenting the column to cell-level messages. We can also give a bitmap encoding to these, as it was proposed there.

Regarding keeping the metadata opaque, I think it is an interesting idea, but it seems to me it requires a more complex API. At the end of the day we do need a protocol layer that is aware of the underlying segment structure. We can try to have this layering of the structure aware and structure unaware part, but I feel this will lead to inefficiencies at some point.

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,189 @@
...
+
+The `topicID` references the Gossipsub topic a message, and thus its parts,
+belong to.
+
+The `metadata` field is opaque application defined metadata associated with this
+request. This can be a bitmap, a list of ranges, or a bloom filter. The
+application generates this and consumes this.
+
+
+Nodes SHOULD assume a `partialIWANT `implies a `IDONTWANT `for the full message.
```
raulk, line 103:
```suggestion
Nodes SHOULD assume a `partialIWANT` implies a `IDONTWANT `for the full message.
```

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,189 @@
...
+then partial messages would be the same as smaller messages.
+
+
+## Protocol Messages
+
+The following section specifies the semantics of each new protocol message.
+
+### PartialIWANT
+
+Partial IWants signal to a receiver that the sending peer only wants a part of
```
raulk, line 86:
```suggestion
`PartialIWANT`s signal to a receiver that the sending peer only wants a part of
```

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,189 @@
...
+Implementations SHOULD NOT send a PartialIHAVE to a peer with parts that the
+peer has previously sent a PartialIDONTWANT for.
+
+### PartialIHAVE
+
+PartialIHave allows nodes to signal HAVE information before receiving all
+segments, unlocking the use of partialIWANT in more contexts.
+
+In the context of partial messages, it is more useful than IHAVE as it tells the
+peer the group ID. In contrast, a IHAVE only includes a message ID that is
```
raulk, line 121:
```suggestion
peer the group ID. In contrast, an IHAVE only includes a message ID that is
```

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,189 @@
...
+knowing the full message, so it can not simply be a hash of the full message.
+
+**Group of Partial Messages**: A set of partial messages all belonging to the
+same group. The group may be complete, meaning that all parts are available. Or
+it may be incomplete, meaning some parts are missing.
+
+## Motivation
+
+The main motivation for this extension is optimizing Ethereum's Data
+Availability Sampling (DAS) protocol. In Ethereum's upcoming fork, Fusaka,
```
raulk, line 46: Technically this doesn't benefit the Sampling part of things. Maybe just generalize this to Ethereum's DA protocols.

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,189 @@
...
+custodied data is laid out in a matrix, where the rows represent user data
```
raulk, line 47:
```suggestion
custodied data is laid out in a matrix per block, where the rows represent user data
```

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,189 @@
...
+(called blobs), and the columns represent a slice across all blobs (each blob
```
raulk, line 48:
```suggestion
(called blobs), and the columns represent a slice across all blobs included in the block (each blob
```

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,189 @@
...
+  optional PartialMessage message = 1;
+  optional PartialIWANT iwant = 2;
+  optional PartialIDONTWANT idontwant = 3;
+  optional PartialIHAVE ihave = 4;
+}
+
+message PartialMessage {
+  optional bytes topicID = 1;
+  optional bytes data = 2;
+}
```
raulk, line 166: Would the partial message benefit some contextual metadata to indicate which parts are actually being delivered? Or do we expect the data to encode this (e.g. with null elements inside a list, if that's the shape)? Or does the `PartialMessage` refer to a single part?

Also: this schema only allows for one inflight partial request per peer per topic. I guess that's not a problem and we (Ethereum) are fine with that limitation?

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,189 @@
...
+Implementations are free to select when to send an update to their peers based
+on signaling bandwidth tradeoff considerations.
+
+## Application Interface
+
+Message contents are application defined. Thus splitting a message must be
+application defined. Here are a list of operations an application is expected to
+provide to Gossipsub to enable partial message delivery.
+
+1. Split a full message into a group of partial messages.
```
raulk, line 143: We may not have a **full** message, but still benefit from splitting what we have into pieces that can be trasmitted as parts right? (e.g. the common case for PeerDAS columns)

Also, is the term "partial message" is being used ambiguously in this spec? Here it refers to parts, but below it might refer to a collection of parts in response to a PARTIALIWANT?

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,189 @@
...
+2. Given a group of partial messages, encode a request for the missing parts. (Used by `PartialIWANT`)
```
raulk, line 144: This will require some higher level contextual information also, right? In our case, in addition to the actual partial messages generated in (1), we need to know the total number of parts and the positions our local parts occupy within the series, so we can generate the bitmap.

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,191 @@
...
+See the [lifecycle document][lifecycle-spec] for context about the maturity level
+and spec status.
+
+[lifecycle-spec]: https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md
+
+## Overview
+
+Partial Messages Extensions allow users to transmit only a small part of a
+message rather than a full message. This is especially useful in cases where
+there is a large messages and a peer is missing only a small part of the
```
jxs:
```suggestion
there is a large message and a peer is missing only a small part of the
```

sukunrt (commented):
I want to understand more about how this will interact with the normal(non
partial) message flow on the topic. 

As I understand from:
https://github.com/ethereum/consensus-specs/pull/4558/files we want to disable
normal gossipsub for peers that support partial message extension and rely
completely on partial messages to reconstruct the columns. However we want to
keep the old (non partial) system for peers that don't support partial messages.

If this is the case, implementation wise would it be better to make this a
topic level property where only peers with partial message extension can
participate on a partial message specific topic.

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,191 @@
...
+all the blobs from its mempool, but in cases where it doesn't (~38%[1]) have
+_all_ of the blobs it almost always has _most_ of the blobs (today, it almost
+always has all but one [1]).
+
+This extension would allow nodes to only request the column message part
+belonging to the missing blob. Reducing the network resource usage
+significantly. As an example, if there are 32 blob cells in a column and the
+node has all but one cell, this would result in a transfer of 2KiB rather than
+64KiB per column. and since nodes custody at least 8 columns, the total savings
+per slot is around 500KiB, or 4 Megabits per slot.
```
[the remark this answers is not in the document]
MarcoPolo: 500KiB*8=4mb. Bits since we are talking about bandwidth, but we can just remove that

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,191 @@
...
+2. `.PartialMessageBytesFromMetadata(metadata: bytes) -> Result<(EncodedPartialMessage: bytes, metadata: bytes), Error>` (When responding to a `PartialIWANT` or eagerly pushing a partial message)
+  a. The returned metadata represents the still missing parts. For example, if a
+     peer is only able to fulfill a part of the the request, the returned
+     metadata represents the parts it couldn't fulfill.
+3. `.ExtendFromEncodedPartialMessage(data: bytes) -> Result<(), Error>` (When receiving a `PartialMessage`)
+4. `.MissingParts() -> Result<metadata: bytes, Error>` (For `PartialIWANT`)
+5. `.AvailableParts() -> Result<metadata: bytes, Error>` (For `PartialIHAVE`)
+
+Gossipsub in turn provides a `.PublishPartial(PartialMessage)` method.
+
```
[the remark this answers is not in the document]
MarcoPolo: the implicit `to` is the peers in a topic mesh, although having a specific to is also useful (the Go implementation supports both implicit of all topic peers and explicit peers)

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,191 @@
...
+Message contents are application defined. Thus splitting a message must be
+application defined. Applications should provide a Partial Message type that
+supports the following operations:
+
+1. `.GroupID() -> GroupID: bytes`
+2. `.PartialMessageBytesFromMetadata(metadata: bytes) -> Result<(EncodedPartialMessage: bytes, metadata: bytes), Error>` (When responding to a `PartialIWANT` or eagerly pushing a partial message)
+  a. The returned metadata represents the still missing parts. For example, if a
+     peer is only able to fulfill a part of the the request, the returned
+     metadata represents the parts it couldn't fulfill.
```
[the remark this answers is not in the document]
MarcoPolo: Suggestion for text to make this clearer?

MarcoPolo:
> If this is the case, implementation wise would it be better to make this a topic level property where only peers with partial message extension can participate on a partial message specific topic.

Currently, in the Go implementation, applications themselves decide if a topic should support partial messages by setting a topic option.

Thinking about this again, maybe it would be better to have an explicit flag in the `SubOpts` message that requests the receiver to use partial messages.

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,178 @@
...
+`SubOpts.subscribe=false`.
+
+## Application Interface
+
+Message contents are application defined. Thus splitting a message must be
+application defined. Applications should provide a Partial Message type that
+supports the following operations:
+
+1. `.GroupID() -> GroupID: bytes`
+2. `.PartialMessageBytesFromMetadata(metadata: bytes) -> Result<(EncodedPartialMessage: bytes, metadata: bytes), Error>` (When responding to a `PartialIWANT` or eagerly pushing a partial message)
```
sukunrt, line 152: NIT: I prefer `repr` over `metadata` as it is the IWant byte representation, not some meta info about it.

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,178 @@
...
+partial extensions.
+
+The partial field value MUST be ignored when a peer sends an unsubscribe message
+`SubOpts.subscribe=false`.
+
+## Application Interface
+
+Message contents are application defined. Thus splitting a message must be
+application defined. Applications should provide a Partial Message type that
+supports the following operations:
```
sukunrt, line 149: I'd move the non normative disclaimer here.
 
For the rest of the text, I see two implementation strategies: 
1. Discard all gossipsub infra except sending and receiving messages, and topic mesh handling, and do all the partial message stuff in the application. So all you need from the application is `PublishTo(msg, to)` and `ReceiveRPC(from, rpc)`

2. The one that go-libp2p does where the application periodically does a PublishMessage(msg) providing its view of a particular partial message(topicID, groupID) to the gossipsub layer. The gs layer keeps track of the want and have lists per peer as opaque metadata and updates it based on the methods provided below. 

You may rephrase the text above or use something similar. When reading the document I assumed we'd do 1, and so found this section confusing, but reviewing https://github.com/libp2p/go-libp2p-pubsub/pull/631/files#diff-d45a19dbc01044648070c29c0fc7b2b63edd3deaccf85b797314d5ef991f4db6R48-R51 I can see we do 2, which clears up doubts on my initial readings, like https://github.com/libp2p/specs/pull/685/files#r2379080632

pubsub/gossipsub/partial-messages.md
sukunrt, line 152: NIT: 
```suggestion
2. `.PartialMessageBytesFromMetadata(IWantMetadata: bytes) -> Result<(EncodedPartialMessage: bytes, newIWantMetadata: bytes), Error>` (When responding to a `PartialIWANT` or eagerly pushing a partial message)
```

pubsub/gossipsub/partial-messages.md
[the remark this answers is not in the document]
MarcoPolo: PublishPartial is idempotent. If a peer has sent us a partial iwant, this publish call will respond to it with the relevant parts.

If you call PublishPartial twice with no changes in between, the second time does nothing.

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Protobuf
+
+Refer to the protobuf registry at `./extensions/extensions.proto`
+
+## Open Questions
+
+- Do we want to add a TTL to PartialIWANTs? This would allow us to cancel them after some time.
+- Should we rename the metadata bytes to iwant and ihave?
+- In the bitmap usecase, iwant/ihave are simply inverses of each other. Do we need to send them both?
```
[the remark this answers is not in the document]
MarcoPolo: Done in 8e42e179.

I think this simplifies the implementation quite a bit as well.

pubsub/gossipsub/partial-messages.md
MarcoPolo: We mostly do 2. We try to do as much as we can in the gossipsub layer without making too many assumptions.

pubsub/gossipsub/partial-messages.md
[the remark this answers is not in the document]
MarcoPolo: I rephrased this, is it clearer?

pubsub/gossipsub/partial-messages.md
```diff
@@ -104,20 +104,53 @@ on signaling bandwidth tradeoff considerations.
...
-it supports the partial message extension, it MUST send partial messages instead
-of full messages.
+Partial Messages uses the same mesh as normal Gossipsub messages. It is a
+replacement to "full" messages. A node requests a peer to send partial messages
+for a specific topic by setting the `requestsPartial` field in the `SubOpts`
+message to true. A node signals support for sending partial messages on a given
+topic by setting the `supportsPartial` field in `SubOpts` to true. A node can
+support sending partial messages without wanting to receive them.
+
+If a node requests partial messages, it MUST support partial messages.
```
sukunrt:
```suggestion
If a node requests partial messages, it MUST support sending partial messages.
```

sukunrt (commented):
Can we add a section for safe rollout of partial messages on existing topics? This would be just an implementation/rollout guideline.

MarcoPolo:
We need to also add a section for interacting with non-mesh peers.

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,227 @@
+# Partial Messages Extension
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+| --------------- | ------------- | ------ | --------------- |
+| 1A              | Working Draft | Active | r0, 2025-06-23  |
+
+Authors: [@marcopolo, @cskiraly]
```
jxs:
```suggestion
Authors: [@marcopolo], [@cskiraly]
```

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,227 @@
...
+
+Interest Group: TODO
```
jxs:
```suggestion
Interest Group: [@jxs], [@dknopik]
```

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,227 @@
...
+
+[@marcopolo]: https://github.com/marcopolo
+[@cskiraly]: https://github.com/cskiraly
+
```
jxs, line 17:
```suggestion
[@jxs]: https://github.com/jxs
[@dknopik]: https://github.com/dknopik
```

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,224 @@
...
+1. Optional eager data that should be pushed to peers who haven't sent us a bitmap yet.
+2. An optional list of peers to publish to instead of the topic mesh peers.
+    1. This is useful for responding to peers who are not in the node's mesh, but
+      sent the node a PartialMessage (e.g similar to Gossipsub's `IHAVE`)
+
+When Gossipsub receives a partial message it MUST forward it to the application.
+The application decides if it should act on the message by either requesting
+parts or forwarding the message. Both are done with `.PublishPartial`.
+
+Gossipsub MUST forward all messages to the application, not just messages from
```
pawanjay176: is this not just from mesh peers?

MarcoPolo: yes, thank you.

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,178 @@
...
+The structure of `PartialIHAVE` is analogous to that of `PartialIWANT`.
+
+The metadata, as in a `PartialIWANT`, is application defined. It is some encoding
+that represents the parts the sender has.
+
+Implementations are free to select when to send an update to their peers based
+on signaling bandwidth tradeoff considerations.
+
+Receivers MUST treat a `PartialIHAVE` as a signal that the peer does not want
+the indicated part.
```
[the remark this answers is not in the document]
MarcoPolo: I think this is covered now.

MarcoPolo:
@sukunrt, @jxs could I trouble you all for a review? I think this is ready for merging.

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,297 @@
...
+is application defined. An unset value carries no information besides that the
+peer did not send a value.
+
+Upon receiving a `partsMetadata` a node SHOULD respond with only parts the peer
+wants.
+
+A later `partsMetadata` replaces a prior one.
+
+`partsMetadata` can be used during heartbeat gossip to inform non-mesh topic
+peers about parts this node has.
```
jxs:
```suggestion
During heartbeat gossip, partsMetadata can be used to inform a random subset of non-mesh topic peers about the parts held by this node, similar to full message IHAVE gossip.
```

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,297 @@
...
+The application decides if it should act on the message by either requesting
+parts or forwarding the message. Both are done with `.PublishPartial`.
+
+Gossipsub MUST forward all messages to the application, not just messages from
+mesh peers.
+
+### Fanout and Gossip messages
+
+Fanout and Gossip messages by definition come from non-mesh peers. Partial
+messages, without eager data, require an exchange of bitmaps before parts are
```
jxs:
```suggestion
messages without eager data, require an exchange of bitmaps before parts are
```

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,297 @@
...
+node SHOULD send the peer partial messages. Otherwise, send the node full
+messages.
+
+On the receiving side, if the node is in a mixed network of partial and full
+messages, and it requests partial messages, the node MUST support receiving full
+messages.
+
+## Creating a topic to only use partial messages
+
+There is currently no mechanism to requre that a topic only be used for partial
```
jxs:
```suggestion
There is currently no mechanism to require that a topic only be used for partial
```

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,297 @@
...
+   the switch. Nodes signal their support for partial messages by setting the
+   subscribe option `supportsSendingPartial` to true.
+2. Slowly deploy and monitor nodes that request (and implicitly support) partial
+   messages. These nodes should find peers that send them partial messages from
+   the previous step. Nodes request partial messages by setting the subscribe
+   option `requestPartial` to true.
+
+### Supporting both full and partial messages for a topic
+
+Partial messages use the same mesh as "full" messages. Supporting both is
```
jxs: should it be
```suggestion
Partial messages use the same mesh as full messages. Supporting both is
```
instead?
as we already define above the concept of full messages

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,297 @@
...
+
+1. `.GroupID() -> GroupID: bytes`
+2. `.PartialMessageBytes(partsMetadata: bytes) -> Result<(EncodedPartialMessage: bytes), Error>`
+   1. The method should return an encoded partial message with just the parts the
+      peer requested.
+3. `.EagerPartialMessageBytes() -> Result<(EncodedPartialMessage: bytes, partsMetadata: bytes), Error>`
+   1. The method should return an encoded partial message of eager data that
+      should be sent, along with the partsMetadata a peer would have after
+      receiving the encoded partial message.
+4. `.PartsMetadata() -> bytes`: The parts this partial message has.
```
jxs: this API is now different in the `rust-libp2p`  [implementation](https://github.com/jxs/rust-libp2p/blob/gossipsub-partial-messages/protocols/gossipsub/src/extensions/partial_messages.rs#L72-L93).

jxs (approved):
left some suggestions Marco but overall LGTM

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,297 @@
...
+encoding is application defined.
+
+### partsMetadata
+
+The `partsMetadata` field encodes the parts a peer has and wants. The encoding
+is application defined. An unset value carries no information besides that the
+peer did not send a value.
+
+Upon receiving a `partsMetadata` a node SHOULD respond with only parts the peer
+wants.
```
sukunrt:
```suggestion
Upon receiving a `partsMetadata` a node SHOULD respond with only parts the peer
doesn't have.
```

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,297 @@
...
+topic by setting the `supportsSendingPartial` field in `SubOpts` to true. A node can
+support sending partial messages without wanting to receive them.
+
+If a node requests partial messages, it MUST support sending partial messages.
+
+A node uses a peer's `supportsSendingPartial` setting to know if it can send a
+partial message request to a peer. It uses its `requestsPartial` setting to know
+whether to send the peer a full message or a partial message.
+
+If a peer supports partial messages on a topic but did not request them, a node
```
sukunrt, line 133:
```suggestion
If a peer supports sending partial messages on a topic but did not request them, a node
```

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,297 @@
...
+
+| SubOpts.supportsSendingPartial | behavior of receiver that did not request partial messages for the topic |
+| ------------------------ | ------------------------------------------------------------------------ |
+| \*                       | The receiver expects full messages                                       |
+
+
+## Partial Message Gossip
+
+Partial Messages can replace Gossipsub's IHAVE/IWANT with a message that
+provides more context (via the Group ID) and allows for partial responses.
```
sukunrt:
```suggestion
Partial Messages SHOULD replace Gossipsub's IHAVE/IWANT with a message that
provides more context (via the Group ID) and allows for partial responses.
```

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,297 @@
...
+responded with parts, the node can assume the peer's view of itself is the union
+of the its previously sent `partsMetadata `and the remote peer's
+`partsMetadata`. The node does not need to send this peer its updated
+`partsMetadata` if its `partsMetadata` has not changed since the previously sent
+`partsMetadata`.
+
+On the other side, when a node responds to a peer's request for parts, it should
+update the peer's `partsMetadata` with the union of its `partsMetadata` and the
+peers `partsMetadata`.
+
```
sukunrt, line 195: I think this isn't true anymore. This spec doesn't enforce or even suggest sending all the parts that the node has and the peer doesn't have to the peer. So the union op is incorrect.

pubsub/gossipsub/extensions/extensions.proto
```diff
@@ -23,17 +23,39 @@ message RPC {
...
   // must use field numbers larger than 0x200000 to be encoded with at least 4
   // bytes
 
   optional TestExtension testExtension = 6492434;
+
+}
+
+message PartialMessagesExtension {
+  optional bytes topicID = 1;
+  optional bytes groupID = 2;
```
vladopajic, line 54: why are these fields optional?

vladopajic: since `groupID` identifies a full logical message, it only makes sense for it to always be specified. feels to me like this should be a `MUST`.

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,297 @@
...
+peer in order to update its view of the peer's `partsMetadata`. This allows the
+node to avoid sending duplicate data in the case that a peer concurrently sends
+a `partsMetadata` that requests data already eagerly sent.
+
+### Reacting to `IHAVE`
+
+If a node requests partial messages and is connected partial message capable
+peers, it MAY prefer to delay reacting to a peer's `IHAVE` message in order to
+give the opportunity for a partial message request to finish and provide the
+missing message more efficiently.
```
sukunrt:
```suggestion
If a node is reconstructing a message with partial message extension, it MAY prefer to delay reacting to a peer's `IHAVE` message in order to
give the opportunity for a partial message request to finish and get the message more efficiently.
```

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,297 @@
...
+4. `.PartsMetadata() -> bytes`: The parts this partial message has.
+
+Gossipsub in turn provides a `.PublishPartial(PartialMessage)` method. This
+method should be idempotent. The application should be able to call it multiple
+times, and the library should make sure not to send redundant or duplicate
+messages to peers.
+
+When Gossipsub receives a partial message it MUST forward it to the application.
+The application decides if it should act on the message by either requesting
+parts or forwarding the message. Both are done with `.PublishPartial`.
```
sukunrt: Should this be outside of the implementation guidelines heading?

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,297 @@
...
+
+Gossipsub MUST forward all messages to the application, not just messages from
+mesh peers.
```
sukunrt: Same comment on this one. Should this be outside implementation guidelines?

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,297 @@
...
+
+### Fanout and Gossip messages
+
```
sukunrt: Same comment here. Move it outside implementation recommendations.

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,297 @@
...
+Fanout and Gossip messages by definition come from non-mesh peers. Partial
+messages, without eager data, require an exchange of bitmaps before parts are
+transferred. In order for fanout and gossip messages to be useful, the Gossipsub
+implementation SHOULD include the peers that sent them in the list of peers to
+publish partial messages to. This allows the application to simply call
+`PublishPartial` to respond to both mesh and non mesh peers.
```
sukunrt: NIT
```suggestion
Fanout and Gossip messages by definition come from non-mesh peers. After receiving partial messages from Fanout and Gossip peers, the Gossipsub
implementation SHOULD include these non mesh peers that sent them in the list of peers to
publish partial messages to. This allows the application to simply call
`PublishPartial` to respond to both mesh and non mesh peers.
```

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,297 @@
...
+## Terms and Definitions
+
+**Full Message**: A Gossipsub Message.
+
+**Message Part**: The smallest verifiable part of a message.
+
+**Partial Message**: A group of one or more message parts.
+
+**Group ID**: An identifier to some Full Message. This must not depend on
+knowing the full message, so it can not simply be a hash of the full message.
```
sukunrt, line 39: Can you add Eager Data here?

sukunrt (approved):
Mostly nits and rephrasings.

pubsub/gossipsub/extensions/extensions.proto
MarcoPolo: The protocol requires these, but in protobuf terms these are optional. "Semantics for required fields should be implemented at the application layer instead." from https://protobuf.dev/programming-guides/proto2/#required-deprecated

vladopajic: got it

pubsub/gossipsub/partial-messages.md
MarcoPolo: you're right. I'll remove this section.

pubsub/gossipsub/partial-messages.md
MarcoPolo: This is up to the implementation, right?

pubsub/gossipsub/partial-messages.md
MarcoPolo: I'm not sure, fanout and gossip and mesh peers are identical from the protocol's perspective. How this is presented is up to the implementation. What do you think should go outside the implementation guidelines section?

pubsub/gossipsub/partial-messages.md
sukunrt: this specific part isn't. If the partsMetadata is opaque, the implementation MUST necessarily invoke the application to see what to do with the incoming partial message.

pubsub/gossipsub/partial-messages.md
sukunrt: on this point + https://github.com/libp2p/specs/pull/685#discussion_r2789068407 combined:

I would prefer calling out that a partial message irrespective of the sender (gossip, mesh, fanout) MUST be pushed to the application. There's nothing gossipsub layer can do here.

pubsub/gossipsub/partial-messages.md
```diff
@@ -0,0 +1,297 @@
...
+replacement to "full" messages. A node requests a peer to send partial messages
+for a specific topic by setting the `requestsPartial` field in the `SubOpts`
+message to true. A node signals support for sending partial messages on a given
+topic by setting the `supportsSendingPartial` field in `SubOpts` to true. A node can
+support sending partial messages without wanting to receive them.
+
+If a node requests partial messages, it MUST support sending partial messages.
+
+A node uses a peer's `supportsSendingPartial` setting to know if it can send a
+partial message request to a peer. It uses its `requestsPartial` setting to know
```
ackintosh: What does "partial message request" mean? IIUC, it means sending `partsMetadata` to request missing parts from a peer. If so, I'd suggest making this explicit: 


```suggestion
A node uses a peer's `supportsSendingPartial` setting to know if it can send `partsMetadata` to a peer. It uses its `requestsPartial` setting to know
```

pubsub/gossipsub/partial-messages.md
MarcoPolo: I'll rework this section from our latest learnings.

pubsub/gossipsub/partial-messages.md
MarcoPolo: Thanks, updated in the latest commit.

MarcoPolo:
This should be good to go. I'll merge this first thing tomorrow unless there's an objection.

Recent changes have been editorial, while the protocol itself has been fairly stable.

PR #686, opened 2025-07-03
Gossipsub: test extension

MarcoPolo:
for interop testing

### DIFF
--- a/pubsub/gossipsub/extensions/experimental/test-extension.md
+++ b/pubsub/gossipsub/extensions/experimental/test-extension.md
@@ -1,0 +1,37 @@
+# Test Extension
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+| --------------- | ------------- | ------ | --------------- |
+| 1A              | Working Draft | Active | r0, 2025-06-23  |
+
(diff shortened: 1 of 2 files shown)


raulk (commented):
Is there a reason this cannot be a test vector under the [extensions spec](https://github.com/libp2p/specs/pull/684)?

MarcoPolo:
> Is there a reason this cannot be a test vector under the https://github.com/libp2p/specs/pull/684?

My understanding of test vectors comes from using them in deterministic functions. For example, given some input a certain hash function should return some output. What would a test vector look like for this?

Another point is that the Extensions spec doesn't need to know about this extension at all (except as a registry of code points). This is the first example of an extension, and we don't need to treat it differently than any future extension.

raulk:
My point is that this isn't useful as a standalone spec. And if the only reason it exists is for testing another spec, it should be a test harness under that spec, accompanied by the relevant test vectors.

MarcoPolo:
Is the difference whether this is inlined in the extensions file or a separate file? I'd argue it's a better example if it's a separate file like any other extension would be.

PR #688, opened 2025-07-14
gossipsub: Add table to track implementation status

MarcoPolo:
of Gossipsub versions and extensions.


MarcoPolo:
added. Along with WFR Gossip reference

MarcoPolo:
merging as this is informational and not normative.

PR #701, opened 2025-11-13
multistream-select-v2: Initial draft

### LINKED ISSUE #700: Protocol ID abbreviation to reduce overhead
## Context

Applications using long string-based protocol IDs can benefit from automatic compression/abbreviation during stream opening.

One concrete example is Ethereum's Req/Resp domain. Protocol strings like the following create inefficiencies when opening per-request streams.

```
/eth2/beacon_chain/req/beacon_blocks_by_range/2 [47 bytes]
```
Combined with the fixed `/multistream/1.0.0` preamble and length-prefixing varints, each request incurs in ~70 bytes of overhead that can be reduced to single digits.

## Goals

Design a dictionary-based compression mechanism for protocol IDs during stream opening that is simple, symmetric, dynamic yet deterministic, and easily debuggable.

### multistream 2 definition

We introduce the `/multistream/2` protocol, with compact [multicodec](https://github.com/multiformats/multicodec) `0x41`. We reserve the `0x40-0x4f` range for multistream versioning, where `0x40` aliases `/multistream/v1.0.0` (currently unused).

`/multistream/v1.0.0` continues using the fixed byte string `<varint encoding of 19> || /multistream/v1.0.0` for itself. This byte string is:

```
0x132f6d756c746973747265616d2f76312e302e30
```

When the first byte on a stream is `0x41`, the protocol selection is processed by `/multistream/2`.

### Identify

We add a new Protobuf field in the payload of identify and identify push:

```protobuf
// Maximum version of multiselect supported by the peer.
// Allowable range: [1-16], encoding as a single byte.
optional uint32 max_multiselect_version = 9;
```

Absence of this field indicates v1-only support. Presence indicates the peer supporting v1 through to the specified version.

New multistream protocols can only be used after receiving the peer's identify payload. Both ends the effective version via `min(self.max_multiselect_version, peer.max_multiselect_version)`, producing identical results.

Peers MUST wait to start new streams until receiving their peer's Identify. Most applications already follow this behaviour; go-multistream even includes explicit waiting logic.

For peers supporting multistream > v1, v1 usage is effectively limited to first first identify protocol stream during connection bootstrap.

### Protocol abbreviation with multistream 2

Identify enumerates all protocol strings supported by the peer. Abbreviations are calculated by:

1. Hashing each string with SHA-256/BLAKE3-256 (TBD).
2. Populating the abbreviation table by using the first byte of each digest. On collisions, expand ambiguous values by selecting additional bytes one by one until no ambiguity exists.

When peer A informs of new protocol strings via identify push, new collisions may arise with extant protocols that require its expansion. Existing streams will remain undisturbed, and new streams will use the newly extended selector. However, Peer A MUST retain old mappings for affected protocols to avoid races if peer P is opening streams while the update is in transit.

Example:

**Old state (Peer A)**
- `protocol/a` => `0x01`
- `protocol/b` => `0xa9`
- `protocol/c` => `0x7d`

**New state (Peer A)**
- `protocol/a` => `0x01`
- `protocol/b` => `0xa9`
- `protocol/c` => `0x7daf` (expanded because prefix conflicts)
- `protocol/d` => `0x7d88` (expanded because prefix conflicts)
- `protocol/c` => `0x7d` (retained for session with P)

**Optimization:** Peer P will only selects protocols it actually supports, so remembering mappings for unopenable protocols is unnecessary. Applications may supply all possible outbound protocol strings to libp2p ahead of time (and subsequently update this knowledge), enabling efficient data structures and lookups by tracking only relevant mappings.

### Protocol selection with multistream 2

Protocols are now selected by sending:

```
varint length || 0x41 (multicodec) || varint length || protocol abbreviation
```

Example. To select `/eth2/beacon_chain/req/beacon_blocks_by_range/2`:

With v1:

```
0x132f6d756c746973747265616d2f76312e302e302f2f657468322f626561636f6e5f636861696e2f7265712f626561636f6e5f626c6f636b735f62795f72616e67652f32
Breakdown:
    - varint(19) + "/multistream/v1.0.0" = 20 bytes
    - varint(47) + "/eth2/beacon_chain/req/beacon_blocks_by_range/2" = 48 bytes
    - total: 68 bytes
```

With v2:

```
0x0141010c
Breakdown:
    - BLAKE3-256 hash: 0x0c210f5e4f3bd7f720e887684c77c48f9d65b8d2065a3f2b83b0d47496e5bb29
    - varint(1) + 0x41 = 2 bytes
    - varint(1) + first byte of BLAKE3-256 hash = 2 bytes [assuming no collisions]
    - total: 4 bytes (94% saving)
```

MarcoPolo:
closes #700 

An initial pass at specifying the abbreviation tree. There's still more to define like the wire format, how these protocol strings can be exchanged, and how to know if your peer supports multistream select v2.

### DIFF
--- a/connections/multistream-select-v2.md
+++ b/connections/multistream-select-v2.md
@@ -1,0 +1,265 @@
+# Multistream Select V2
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+| --------------- | ------------- | ------ | --------------- |
+| 1A              | Working Draft | Active | r1, 2025-11-13  |
+
+Authors: [@raulk], [@marcopolo]
+
+Interest Group: [@ppopth]
+
+[@marcopolo]: https://github.com/marcopolo
+[@ppopth]: https://github.com/ppopth
+[@raulk]: https://github.com/raulk
+
+## Terminology
+
+Server: The endpoint advertising its supported protocol strings.
+
+Client: The endpoint receiving the advertisement and using protocol string
+abbreviations when opening streams.
+
+Abbreviation Tree: The tree data structure that determines the abbreviation to
+use for a given protocol string.
+
+Note, A peer can behave as both a client and server. For the purpose of defining
+this protocol, it's useful to focus on the client/server interaction.
+
+## Abbreviation Tree
+
+A list of protocol strings are abbreviated by creating a minimal hash prefix
+tree.
+
+### Construction
+
+Each protocol string is hashed and inserted into the tree as shallowly as
+possible. Each node in the tree has 256 leaf branches representing a byte of the
+hash (2^8). If multiple protocol strings share a common byte prefix, they are
+distinguished the next level down the tree.
+
+Each protocol string in the tree has a tombstone bit associated with it. This is
+set to true if the protocol is currently not supported (but was previously).
+
+A node in the tree may contain a value as well as children. This only happens
+when introducing a new supported protocol introduces a conflict. This does not
+happen on initial construction.
+
+### Example
+
+```txt
+(root)
+  |
+  +-- 0xaa -> "/some-protocol/a/v1"
+  |
+  +-- 0xbb -> "/some-protocol/b/v1"
+  |
+  +-- 0xcc
+  |   |
+  |   +-- 0x01 -> "/some-protocol/c/v1"
+  |   |
+  |   +-- 0x02 -> "/some-protocol/c'/v1"
+  |
+  +-- 0xdd -> "/some-protocol/d/v1" tombstone=true
+```
+
+### Inserting a New Protocol String
+
+The protocol string is inserted into the tree as shallowly as possible. If there
+is a conflict, the protocol string should be inserted one level down and
+duplicate the conflicting protocol string to the new level if it is not
+tombstoned. The conflicting protocol string MUST not removed from its original
+position as it may still be referenced.
+
+#### Example
+
+Using the initial example as the initial state and we add three new protocol
+strings:
+
+1. `"/new-protocol/foo"` that hashes to `0xaa02`. This example highlights a
+   conflict with an existing protocol string.
+2. `"/new-protocol/bar"` that hashes to `0xee...`. This example highlights no
+   conflicts and inserting shallowly.
+3. `"/new-protocol/d"` that hashes to `0xdd02`. This example highlights a
+   conflict with a tombstoned protocol string.
+
+```txt
+(root)
+  |
+  +-- 0xaa -> "/some-protocol/a/v1"
+  |   |
+  |   +-- 0x01 -> "/some-protocol/a/v1"
+  |   |
+  |   +-- 0x02 -> "/new-protocol/foo"
+  |
+  +-- 0xbb -> "/some-protocol/b/v1"
+  |
+  +-- 0xcc
+  |   |
+  |   +-- 0x01 -> "/some-protocol/c/v1"
+  |   |
+  |   +-- 0x02 -> "/some-protocol/c'/v1"
+  |
+  +-- 0xdd -> "/some-protocol/d/v1" tombstone=true
+  |   |
+  |   +-- 0xdd01 -> "/new-protocol/d"
+  |
+  +-- 0xee -> "/new-protocol/bar"
+```
+
+### Removing a Protocol String
+
+For each instance of the protocol string in the tree, the tombstone bit is set
+to true. The protocol string MUST not be removed from the tree as that could
+lead to inconsistencies if a new protocol string is introduced with the same
+prefix hash.
+
+#### Example
+
+Using the initial example as the initial state and we remove protocol
+string `"/some-protocol/b/v1"`.
+
+```txt
+(root)
+  |
+  +-- 0xaa -> "/some-protocol/a/v1"
+  |
+  +-- 0xbb -> "/some-protocol/b/v1" tombstone=true
+  |
+  +-- 0xcc
+  |   |
+  |   +-- 0x01 -> "/some-protocol/c/v1"
+  |   |
+  |   +-- 0x02 -> "/some-protocol/c'/v1"
+  |
+  +-- 0xdd -> "/some-protocol/d/v1" tombstone=true
+```
+
+### Reintroducing a previously removed protocol string
+
+For each instance of the protocol string in the tree, the tombstone bit should
+be set to false. The protocol string should then be inserted into the tree as if
+inserting a new protocol string. If the protocol string is already present as a
+leaf node, no changes are required.
+
+#### Example
+
+Using the example for removing a protocol string as the initial state. We
+reintroduce protocol string `"/some-protocol/b/v1"`.
+
+```txt
+(root)
+  |
+  +-- 0xaa -> "/some-protocol/a/v1"
+  |
+  +-- 0xbb -> "/some-protocol/b/v1"
+  |
+  +-- 0xcc
+  |   |
+  |   +-- 0x01 -> "/some-protocol/c/v1"
+  |   |
+  |   +-- 0x02 -> "/some-protocol/c'/v1"
+  |
+  +-- 0xdd -> "/some-protocol/d/v1" tombstone=true
+```
+
+#### Example with a new leaf node
+
+If the tombstoned protocol string is no longer at a leaf position we need to
+insert a new protocol string, as well as untombstoning the existing protocol
+string.
+
+Initial State:
+
+```txt
+(root)
+  |
+  +-- 0xaa -> "/some-protocol/a/v1"
+  |
+  +-- 0xbb -> "/some-protocol/b/v1" tombstone=true
+  |   |
+  |   +-- 0x02 -> "/some-protocol/b'/v1"
+  |
+  +-- 0xcc
+  |   |
+  |   +-- 0x01 -> "/some-protocol/c/v1"
+  |   |
+  |   +-- 0x02 -> "/some-protocol/c'/v1"
+  |
+  +-- 0xdd -> "/some-protocol/d/v1" tombstone=true
+```
+
+After reintroducing `"/some-protocol/b/v1"`:
+
+```txt
+(root)
+  |
+  +-- 0xaa -> "/some-protocol/a/v1"
+  |
+  +-- 0xbb -> "/some-protocol/b/v1"
+  |   |
+  |   +-- 0x01 -> "/some-protocol/b/v1"
+  |   |
+  |   +-- 0x02 -> "/some-protocol/b'/v1"
+  |
+  +-- 0xcc
+  |   |
+  |   +-- 0x01 -> "/some-protocol/c/v1"
+  |   |
+  |   +-- 0x02 -> "/some-protocol/c'/v1"
+  |
+  +-- 0xdd -> "/some-protocol/d/v1" tombstone=true
+```
+
+## Limits
+
+Implementations SHOULD limit the number of protocol strings they track
+(including tombstoned protocol strings).
+
+## Mapping Algorithm
+
+A mapping algorithm maps a protocol string to byte array. This byte array is
+used for the abbreviation tree.
+
+SHA256 MUST be used for the mapping algorithm.
+
+TODO: Do we want to consider faster keyed hashes? Risks complicating
+implementations.
+
+TODO: Do we need a hash function at all? Would an index into the list of a
+server's protocols work instead?
+
+## Initial Server Client Exchange
+
+When a client connects to a server, the server shares all the state necessary to
+create an abbreviation tree it can use to communicate with the server. Namely:
+the list of supported protocol strings, and the list of previously supported
+protocol strings (tombstoned protocol strings).
+
+The client's abbreviation tree will differ from the server's abbreviation tree
+only in that protocol strings will only exist at leaf nodes in the client's
+abbreviation tree. This is because non-leaf protocol strings only occur when
+adding a protocol string to an existing tree.
+
+# Multistream Select V2
+
+## Wire Format
+
+TODO define the wire format of multistream select v2.
+
+## Client opening a new stream
+
+A client identifies the protocol string it wishes to use on a stream by the
+minimal hash prefix as determined by the abbreviate stream.
+
+## Server accepting a new stream
+
+The server identifies the protocol string by looking up the minimal hash prefix
+in the tree.
+
+## Version Negotiation
+
+Multistream Select V2 does not support negotiating different protocols on a
+single stream like Multistream Select V1 does. A client specifies what protocol
+they would like to use for a stream and MAY start sending protocol immediately.
+If the server does not support this protocol, it MUST close the stream with
+error code (TODO specify this).


connections/multistream-select-v2.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Initial Server Client Exchange
+
+When a client connects to a server, the server shares all the state necessary to
+create an abbreviation tree it can use to communicate with the server. Namely:
+the list of supported protocol strings, and the list of previously supported
+protocol strings (tombstoned protocol strings).
+
+The client's abbreviation tree will differ from the server's abbreviation tree
+only in that protocol strings will only exist at leaf nodes. This is because
```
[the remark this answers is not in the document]
MarcoPolo: I'm not sure I follow. Can you expand please?

MarcoPolo:
Do we need the hash function at all? The hash function acts as a mapping from a protocol string to a byte string with high likelihood of distinct short prefixes. But we could also imagine just using the index of the protocol string from the server's list instead.

If a server returned protocols `["/a", "/b", "/c"]` this would be mapped to abbreviations `[0x00, 0x01, 0x02]`. The tombstones would solve the remove-reintroduce problem without increasing the list size. New protocols are added to the end. The implementation would also be simpler as it's just a list.

marten-seemann (commented):
Is it really necessary to support removing strings?

MarcoPolo:
Whether to remove support for a protocol is an application concern. At this level, MSSv1 does support it, so it makes sense that the next version also supports. 

That said, If supporting removing strings makes the solution much more complicated, that's a tradeoff we should discuss. It would be good to understand the complexity of supporting it first.

connections/multistream-select-v2.md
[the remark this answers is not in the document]
haxxpop: I mean the client's abbreviation tree will never have non-leaf protocol strings, right?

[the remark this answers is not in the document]
MarcoPolo: yes. I've reworded it. Is it clearer now?

dozyio:
Think I prefer an index based solution for simplicity of implementation. Presume the server would maintain the index.

Would it make sense to move some of this to an identify / identify push v2?

MarcoPolo:
For the index based version, the server returns a list of protocol strings which implies the index. It also returns a list (or bitmap) of which protocol strings are tombstoned. The latter part is new to identify, and would have to be introduced.

The fewer changes to identify the better.

haxxpop:
I think in the hash-based approach, you also don't need to maintain a per-peer state on the server side, right? You only need to maintain one tree.

In the index-based approach, you have to remember the order or indices of the protocol strings sent to each peer, unless you make sure you always send the same order of protocol strings to peers.

sukunrt:
@MarcoPolo 
>For the index based version, the server returns a list of protocol strings which implies the index. It also returns a list (or bitmap) of which protocol strings are tombstoned. The latter part is new to identify, and would have to be introduced.

For a minimal identify change here: 
The server can return another list along with the protocol list which is the int value of the corresponding protocol in the protocol list. 

`identify.Protocols[i]` can be used with mss-v2 with `identify.ProtocolIndex[i]`

It's up to the server to do the tombstoning correctly.

connections/multistream-select-v2.md
```diff
@@ -0,0 +1,265 @@
...
+Server: The endpoint advertising its supported protocol strings.
+
+Client: The endpoint receiving the advertisement and using protocol string
+abbreviations when opening streams.
+
+Abbreviation Tree: The tree data structure that determines the abbreviation to
+use for a given protocol string.
+
+Note, A peer can behave as both a client and server. For the purpose of defining
+this protocol, it's useful to focus on the client/server interaction.
```
gmelodie, line 26: Nit
```suggestion
- **Server**: the endpoint advertising its supported protocol strings.
- **Client*: the endpoint receiving the advertisement and using protocol string
abbreviations when opening streams.
- **Abbreviation Tree*: a trie over the hashed protocol strings, in which each string's depth is the length of the shortest hash prefix that uniquely identifies it; that prefix is the string's abbreviation.

**Note**: a peer can behave as both a client and server. For the purpose of defining
this protocol, it's useful to focus on the client/server interaction.
```

connections/multistream-select-v2.md
```diff
@@ -0,0 +1,265 @@
...
+
+## Terminology
+
+Server: The endpoint advertising its supported protocol strings.
+
+Client: The endpoint receiving the advertisement and using protocol string
+abbreviations when opening streams.
+
+Abbreviation Tree: The tree data structure that determines the abbreviation to
+use for a given protocol string.
```
gmelodie, line 23: here, a non-circular definition
```suggestion
Abbreviation Tree: a trie over the hashed protocol strings, in which each string's depth is the length of the shortest hash prefix that uniquely identifies it; that prefix is the string's abbreviation.
```

connections/multistream-select-v2.md
```diff
@@ -0,0 +1,265 @@
...
+A mapping algorithm maps a protocol string to byte array. This byte array is
+used for the abbreviation tree.
+
+SHA256 MUST be used for the mapping algorithm.
+
+TODO: Do we want to consider faster keyed hashes? Risks complicating
+implementations.
+
+TODO: Do we need a hash function at all? Would an index into the list of a
+server's protocols work instead?
```
gmelodie, line 229: I don't think hashing adds anything for us here, it only makes the idea more cumbersome and less human-readable

connections/multistream-select-v2.md
```diff
@@ -0,0 +1,265 @@
...
+tree.
+
+### Construction
+
+Each protocol string is hashed and inserted into the tree as shallowly as
+possible. Each node in the tree has 256 leaf branches representing a byte of the
+hash (2^8). If multiple protocol strings share a common byte prefix, they are
+distinguished the next level down the tree.
+
+Each protocol string in the tree has a tombstone bit associated with it. This is
```
gmelodie, line 40: Maybe add `Tombstone bit` to the terminology section?

connections/multistream-select-v2.md
```diff
@@ -0,0 +1,265 @@
...
+  +-- 0xbb -> "/some-protocol/b/v1"
+  |
+  +-- 0xcc
+  |   |
+  |   +-- 0x01 -> "/some-protocol/c/v1"
+  |   |
+  |   +-- 0x02 -> "/some-protocol/c'/v1"
+  |
+  +-- 0xdd -> "/some-protocol/d/v1" tombstone=true
+```
```
gmelodie, line 63:
````suggestion
### Example

The hash values in this and all following examples are illustrative and
chosen for readability. Real implementations use the mapping algorithm
defined below (SHA256); no real protocol string hashes to these values.

Protocol strings are inserted in the following order, with these (example)
hashes:

| Step | Protocol string        | Example hash | Result                                                      |
|------|------------------------|--------------|-------------------------------------------------------------|
| 1    | `/some-protocol/a/v1`  | `0xaa…`      | No conflict, inserted at `0xaa`.                            |
| 2    | `/some-protocol/b/v1`  | `0xbb…`      | No conflict, inserted at `0xbb`.                            |
| 3    | `/some-protocol/c1/v1` | `0xcc01…`    | No conflict, inserted at `0xcc`.                            |
| 4    | `/some-protocol/c2/v1` | `0xcc02…`    | Conflicts with step 3 at `0xcc`. Both move one level down: `c1/v1` to `0xcc01`, `c2/v1` to `0xcc02`. |
| 5    | `/some-protocol/d/v1`  | `0xdd…`      | No conflict, inserted at `0xdd`.                            |
| 6    | remove `/some-protocol/d/v1` |        | Tombstone bit set; the entry stays at `0xdd` so the prefix remains reserved. |

Resulting tree (node labels are the full prefix, i.e. the bytes sent on the wire):

```txt
(root)
  |
  +-- 0xaa   -> "/some-protocol/a/v1"
  |
  +-- 0xbb   -> "/some-protocol/b/v1"
  |
  +-- 0xcc
  |   |
  |   +-- 0xcc01 -> "/some-protocol/c1/v1"
  |   |
  |   +-- 0xcc02 -> "/some-protocol/c2/v1"
  |
  +-- 0xdd   -> "/some-protocol/d/v1" tombstone=true
```

Note that step 4 moves `c1/v1` down during initial construction. This is the
only case in which an existing entry moves: once the tree is in use, entries
are never removed or relocated (see "Inserting a New Protocol String").
````

File: connections/multistream-select-v2.md [hunk 7]
```diff
@@ -0,0 +1,265 @@
...
+  |   +-- 0x01 -> "/some-protocol/c/v1"
+  |   |
+  |   +-- 0x02 -> "/some-protocol/c'/v1"
+  |
+  +-- 0xdd -> "/some-protocol/d/v1" tombstone=true
+  |   |
+  |   +-- 0xdd01 -> "/new-protocol/d"
+  |
+  +-- 0xee -> "/new-protocol/bar"
+```
```
gmelodie, line 107:
````suggestion
#### Example

Starting from the tree in the construction example above, three protocol
strings are inserted, in this order. Note that `/some-protocol/a/v1`'s full hash is `0xaa01…`; its first byte alone was
enough to place it originally.

| Step | Protocol string      | Example hash | Result |
|------|----------------------|--------------|--------|
| 1    | `/new-protocol/foo`  | `0xaa02…`    | Conflicts with `a/v1` at `0xaa`. `foo` is inserted one level down at `0xaa02`. `a/v1` is not tombstoned, so it is duplicated to `0xaa01`; its original entry at `0xaa` MUST stay, since clients that built their tree earlier still refer to `a/v1` as `0xaa`. |
| 2    | `/new-protocol/bar`  | `0xee…`      | No conflict, inserted at `0xee`. |
| 3    | `/new-protocol/d`    | `0xdd02…`    | Conflicts with the tombstoned `d/v1` at `0xdd`. `new-protocol/d` is inserted one level down at `0xdd02`. `d/v1` is tombstoned, so it is not duplicated; its entry at `0xdd` stays so that a stale client sending `0xdd` is rejected rather than misrouted to `new-protocol/d`. |

Resulting tree:

```txt
(root)
  |
  +-- 0xaa   -> "/some-protocol/a/v1"
  |   |
  |   +-- 0xaa01 -> "/some-protocol/a/v1"
  |   |
  |   +-- 0xaa02 -> "/new-protocol/foo"
  |
  +-- 0xbb   -> "/some-protocol/b/v1"
  |
  +-- 0xcc
  |   |
  |   +-- 0xcc01 -> "/some-protocol/c1/v1"
  |   |
  |   +-- 0xcc02 -> "/some-protocol/c2/v1"
  |
  +-- 0xdd   -> "/some-protocol/d/v1" tombstone=true
  |   |
  |   +-- 0xdd02 -> "/new-protocol/d"
  |
  +-- 0xee   -> "/new-protocol/bar"
```

After step 1, `0xaa` is a node that holds a value and has children. A client
that connected before step 1 sends `0xaa` for `a/v1`; a client that connects
after step 1 builds its tree from the advertised list and sends `0xaa01`. The
server accepts both.
````

File: connections/multistream-select-v2.md [hunk 8]
```diff
@@ -0,0 +1,265 @@
...
+  +-- 0xbb -> "/some-protocol/b/v1" tombstone=true
+  |
+  +-- 0xcc
+  |   |
+  |   +-- 0x01 -> "/some-protocol/c/v1"
+  |   |
+  |   +-- 0x02 -> "/some-protocol/c'/v1"
+  |
+  +-- 0xdd -> "/some-protocol/d/v1" tombstone=true
+```
```
gmelodie, line 135:
````suggestion
#### Example

Starting from the tree in the construction example above, `/some-protocol/b/v1`
is removed.

`b/v1` appears once in the tree, at `0xbb`. Its tombstone bit is set to true.

Resulting tree:

```txt
(root)
  |
  +-- 0xaa   -> "/some-protocol/a/v1"
  |
  +-- 0xbb   -> "/some-protocol/b/v1" tombstone=true
  |
  +-- 0xcc
  |   |
  |   +-- 0xcc01 -> "/some-protocol/c1/v1"
  |   |
  |   +-- 0xcc02 -> "/some-protocol/c2/v1"
  |
  +-- 0xdd   -> "/some-protocol/d/v1" tombstone=true
```

If the removed protocol string appears at more than one position (see the
insertion example, where `a/v1` exists at both `0xaa` and `0xaa01`), the
tombstone bit is set on every occurrence.
````

File: connections/multistream-select-v2.md [hunk 9]
```diff
@@ -0,0 +1,265 @@
...
+  +-- 0xbb -> "/some-protocol/b/v1"
+  |
+  +-- 0xcc
+  |   |
+  |   +-- 0x01 -> "/some-protocol/c/v1"
+  |   |
+  |   +-- 0x02 -> "/some-protocol/c'/v1"
+  |
+  +-- 0xdd -> "/some-protocol/d/v1" tombstone=true
+```
```
gmelodie, line 163:
````suggestion
#### Example

Starting from the tree in the removal example above, `/some-protocol/b/v1` is
reintroduced.

`b/v1` appears once in the tree, at `0xbb`, with its tombstone bit set. The
bit is set back to false. `0xbb` is a leaf, so `b/v1` is already at the
shallowest position that identifies it and no insertion is needed. The tree
is identical to the one before the removal.

Resulting tree:

```txt
(root)
  |
  +-- 0xaa   -> "/some-protocol/a/v1"
  |
  +-- 0xbb   -> "/some-protocol/b/v1"
  |
  +-- 0xcc
  |   |
  |   +-- 0xcc01 -> "/some-protocol/c1/v1"
  |   |
  |   +-- 0xcc02 -> "/some-protocol/c2/v1"
  |
  +-- 0xdd   -> "/some-protocol/d/v1" tombstone=true
```
````

File: connections/multistream-select-v2.md [hunk 10]
```diff
@@ -0,0 +1,265 @@
...
+  |   +-- 0x02 -> "/some-protocol/b'/v1"
+  |
+  +-- 0xcc
+  |   |
+  |   +-- 0x01 -> "/some-protocol/c/v1"
+  |   |
+  |   +-- 0x02 -> "/some-protocol/c'/v1"
+  |
+  +-- 0xdd -> "/some-protocol/d/v1" tombstone=true
+```
```
gmelodie, line 211:
````suggestion
#### Example: reintroducing a protocol string that is no longer a leaf

Starting from the tree in the withdrawal example above, two things happen in
order, using illustrative hashes as before:

| Step | Operation                        | Example hash | Result |
|------|----------------------------------|--------------|--------|
| 1    | insert `/some-protocol/b2/v1`    | `0xbb02…`    | Conflicts with the tombstoned `b/v1` at `0xbb`. `b2/v1` is inserted at `0xbb02`. `b/v1` is tombstoned, so it is not duplicated down. `0xbb` is now a non-leaf node holding a tombstoned value. |
| 2    | reintroduce `/some-protocol/b/v1` | `0xbb01…`   | See below. |

State after step 1:

```txt
(root)
  |
  +-- 0xaa   -> "/some-protocol/a/v1"
  |
  +-- 0xbb   -> "/some-protocol/b/v1" tombstone=true
  |   |
  |   +-- 0xbb02 -> "/some-protocol/b2/v1"
  |
  +-- 0xcc
  |   |
  |   +-- 0xcc01 -> "/some-protocol/c1/v1"
  |   |
  |   +-- 0xcc02 -> "/some-protocol/c2/v1"
  |
  +-- 0xdd   -> "/some-protocol/d/v1" tombstone=true
```

Step 2 does two things:

1. The tombstone bit on the existing `b/v1` entry at `0xbb` is set back to
   false. Clients that built their tree before the withdrawal still send
   `0xbb` for `b/v1`, and those requests must be accepted again.
2. `b/v1` is inserted as if it were a new protocol string. `0xbb` is no longer
   a leaf, so the shallowest unambiguous position is `0xbb01`. A client that
   connects after this point builds its tree from the advertised list
   `{a/v1, b/v1, b2/v1, c1/v1, c2/v1}` and will send `0xbb01` for `b/v1`; the
   server must recognise that prefix too.

Resulting tree:

```txt
(root)
  |
  +-- 0xaa   -> "/some-protocol/a/v1"
  |
  +-- 0xbb   -> "/some-protocol/b/v1"
  |   |
  |   +-- 0xbb01 -> "/some-protocol/b/v1"
  |   |
  |   +-- 0xbb02 -> "/some-protocol/b2/v1"
  |
  +-- 0xcc
  |   |
  |   +-- 0xcc01 -> "/some-protocol/c1/v1"
  |   |
  |   +-- 0xcc02 -> "/some-protocol/c2/v1"
  |
  +-- 0xdd   -> "/some-protocol/d/v1" tombstone=true
```

`b/v1` now exists at both `0xbb` and `0xbb01`, and the server accepts either.
````

**Review by gmelodie — CHANGES_REQUESTED** [review]
I'm not convinced that shortening ~70 bytes per connection will make a real difference in performance, when noise/identify themselves use the order of KBs. Besides, this will require implementations to follow more rules they can get wrong and likely introduce more compatibility issues. I think we need more hard evidence of the performance benefits of this approach before moving forward.

Suggestion for the examples: do the following instead of redrawing the tree many times in different sections depending on previous sections.

````md
### Worked Example

All hashes below are illustrative and chosen for readability; real
implementations use the mapping algorithm defined below (SHA256). Node
labels are full prefixes, i.e. the bytes a client sends on the wire.

**Construction.** Five protocol strings are inserted in order:

| Step | Protocol string        | Example hash | Result |
|------|------------------------|--------------|--------|
| 1    | `/some-protocol/a/v1`  | `0xaa01…`    | Inserted at `0xaa`. |
| 2    | `/some-protocol/b/v1`  | `0xbb01…`    | Inserted at `0xbb`. |
| 3    | `/some-protocol/c1/v1` | `0xcc01…`    | Inserted at `0xcc`. |
| 4    | `/some-protocol/c2/v1` | `0xcc02…`    | Conflicts with `c1/v1` at `0xcc`. During construction entries may move: `c1/v1` → `0xcc01`, `c2/v1` → `0xcc02`. |
| 5    | `/some-protocol/d/v1`  | `0xdd01…`    | Inserted at `0xdd`. |

```txt
(root)
  |
  +-- 0xaa   -> "/some-protocol/a/v1"
  |
  +-- 0xbb   -> "/some-protocol/b/v1"
  |
  +-- 0xcc
  |   |
  |   +-- 0xcc01 -> "/some-protocol/c1/v1"
  |   |
  |   +-- 0xcc02 -> "/some-protocol/c2/v1"
  |
  +-- 0xdd   -> "/some-protocol/d/v1"
```

**Mutations.** Once the tree is in use, entries are never removed or moved;
clients that built their tree earlier may still reference any existing
prefix. The following operations are applied in order:

| Step | Operation                          | Example hash | Result |
|------|------------------------------------|--------------|--------|
| 6    | withdraw `/some-protocol/d/v1`     |              | Tombstone bit set at `0xdd`. Node stays. |
| 7    | insert `/new-protocol/foo`         | `0xaa02…`    | Conflicts with live `a/v1` at `0xaa`. `foo` → `0xaa02`. `a/v1` is not tombstoned, so it is *duplicated* to `0xaa01`; its entry at `0xaa` stays. |
| 8    | insert `/new-protocol/d`           | `0xdd02…`    | Conflicts with tombstoned `d/v1` at `0xdd`. `new-protocol/d` → `0xdd02`. `d/v1` is tombstoned, so it is *not* duplicated. Its entry at `0xdd` stays so a stale client sending `0xdd` is rejected rather than misrouted. |
| 9    | insert `/new-protocol/bar`         | `0xee01…`    | No conflict. Inserted at `0xee`. |
| 10   | withdraw `/some-protocol/b/v1`     |              | Tombstone bit set at `0xbb`. Node stays. |
| 11   | insert `/some-protocol/b2/v1`      | `0xbb02…`    | Conflicts with tombstoned `b/v1` at `0xbb`. `b2/v1` → `0xbb02`, no duplication. `0xbb` is now a non-leaf holding a tombstoned value. |
| 12   | reintroduce `/some-protocol/b/v1`  | `0xbb01…`    | Tombstone bit at `0xbb` cleared, so pre-withdrawal clients sending `0xbb` are accepted again. Then `b/v1` is inserted as if new: `0xbb` is no longer a leaf, so it also goes at `0xbb01`, which is what a freshly connecting client will compute. |

Resulting tree:

```txt
(root)
  |
  +-- 0xaa   -> "/some-protocol/a/v1"
  |   |
  |   +-- 0xaa01 -> "/some-protocol/a/v1"
  |   |
  |   +-- 0xaa02 -> "/new-protocol/foo"
  |
  +-- 0xbb   -> "/some-protocol/b/v1"
  |   |
  |   +-- 0xbb01 -> "/some-protocol/b/v1"
  |   |
  |   +-- 0xbb02 -> "/some-protocol/b2/v1"
  |
  +-- 0xcc
  |   |
  |   +-- 0xcc01 -> "/some-protocol/c1/v1"
  |   |
  |   +-- 0xcc02 -> "/some-protocol/c2/v1"
  |
  +-- 0xdd   -> "/some-protocol/d/v1" tombstone=true
  |   |
  |   +-- 0xdd02 -> "/new-protocol/d"
  |
  +-- 0xee   -> "/new-protocol/bar"
```

Note that `a/v1` and `b/v1` each exist at two positions; the server accepts
either prefix. A tombstoned withdrawal that is reintroduced while still a
leaf (e.g. withdrawing and reintroducing `bar`) only flips the bit and does
not change the tree.
````

File: connections/multistream-select-v2.md [hunk 11]
```diff
@@ -0,0 +1,265 @@
...
+  |
+  +-- 0xcc
+  |   |
+  |   +-- 0x01 -> "/some-protocol/c/v1"
+  |   |
+  |   +-- 0x02 -> "/some-protocol/c'/v1"
+  |
+  +-- 0xdd -> "/some-protocol/d/v1" tombstone=true
+  |   |
+  |   +-- 0xdd01 -> "/new-protocol/d"
```
richard-ramos, line 104:
```suggestion
  |   +-- 0x02 -> "/new-protocol/d"
```
to match what's it's written in line 82

connections/multistream-select-v2.md
```diff
@@ -0,0 +1,265 @@
...
+
+Abbreviation Tree: The tree data structure that determines the abbreviation to
+use for a given protocol string.
+
+Note, A peer can behave as both a client and server. For the purpose of defining
+this protocol, it's useful to focus on the client/server interaction.
+
+## Abbreviation Tree
+
+A list of protocol strings are abbreviated by creating a minimal hash prefix
```
richard-ramos, line 30:
```suggestion
Protocol strings are abbreviated by creating a minimal hash prefix
```

connections/multistream-select-v2.md
```diff
@@ -0,0 +1,265 @@
...
+Each protocol string is hashed and inserted into the tree as shallowly as
+possible. Each node in the tree has 256 leaf branches representing a byte of the
+hash (2^8). If multiple protocol strings share a common byte prefix, they are
+distinguished the next level down the tree.
+
+Each protocol string in the tree has a tombstone bit associated with it. This is
+set to true if the protocol is currently not supported (but was previously).
+
+A node in the tree may contain a value as well as children. This only happens
+when introducing a new supported protocol introduces a conflict. This does not
```
richard-ramos, line 44:
```suggestion
when adding a new supported protocol creates a conflict. This does not
```

connections/multistream-select-v2.md
```diff
@@ -0,0 +1,265 @@
...
+  |
+  +-- 0xdd -> "/some-protocol/d/v1" tombstone=true
+```
+
+### Inserting a New Protocol String
+
+The protocol string is inserted into the tree as shallowly as possible. If there
+is a conflict, the protocol string should be inserted one level down and
+duplicate the conflicting protocol string to the new level if it is not
+tombstoned. The conflicting protocol string MUST not removed from its original
```
richard-ramos, line 70:
```suggestion
tombstoned. The conflicting protocol string MUST NOT be removed from its original
```

connections/multistream-select-v2.md
```diff
@@ -0,0 +1,265 @@
...
+```
+
+## Limits
+
+Implementations SHOULD limit the number of protocol strings they track
+(including tombstoned protocol strings).
+
+## Mapping Algorithm
+
+A mapping algorithm maps a protocol string to byte array. This byte array is
```
richard-ramos, line 220:
```suggestion
A mapping algorithm maps a protocol string to a byte array. This byte array is
```

connections/multistream-select-v2.md
```diff
@@ -0,0 +1,265 @@
...
+# Multistream Select V2
+
+## Wire Format
+
+TODO define the wire format of multistream select v2.
+
+## Client opening a new stream
+
+A client identifies the protocol string it wishes to use on a stream by the
+minimal hash prefix as determined by the abbreviate stream.
```
richard-ramos, line 252:
```suggestion
minimal hash prefix as determined by the abbreviation tree.
```

connections/multistream-select-v2.md
```diff
@@ -0,0 +1,265 @@
...
+## Server accepting a new stream
+
+The server identifies the protocol string by looking up the minimal hash prefix
+in the tree.
+
+## Version Negotiation
+
+Multistream Select V2 does not support negotiating different protocols on a
+single stream like Multistream Select V1 does. A client specifies what protocol
+they would like to use for a stream and MAY start sending protocol immediately.
```
richard-ramos, line 263:
```suggestion
they would like to use for a stream and MAY start sending protocol data immediately.
```

connections/multistream-select-v2.md
```diff
@@ -0,0 +1,265 @@
...
+If the server does not support this protocol, it MUST close the stream with
+error code (TODO specify this).
```
richard-ramos, line 265:
```suggestion
an error code (TODO specify this).
```

connections/multistream-select-v2.md
```diff
@@ -0,0 +1,265 @@
...
+
+The server identifies the protocol string by looking up the minimal hash prefix
+in the tree.
+
+## Version Negotiation
+
+Multistream Select V2 does not support negotiating different protocols on a
+single stream like Multistream Select V1 does. A client specifies what protocol
+they would like to use for a stream and MAY start sending protocol immediately.
+If the server does not support this protocol, it MUST close the stream with
```
richard-ramos, line 264: `close` or `reset`?

connections/multistream-select-v2.md
```diff
@@ -0,0 +1,265 @@
...
+| 1A              | Working Draft | Active | r1, 2025-11-13  |
+
+Authors: [@raulk], [@marcopolo]
+
+Interest Group: [@ppopth]
+
+[@marcopolo]: https://github.com/marcopolo
+[@ppopth]: https://github.com/ppopth
+[@raulk]: https://github.com/raulk
+
```
richard-ramos, line 14: Could a `Overview` section be added with the motivation/goal behind this spec?

connections/multistream-select-v2.md
```diff
@@ -0,0 +1,265 @@
...
+  +-- 0xdd -> "/some-protocol/d/v1" tombstone=true
+```
+
+### Inserting a New Protocol String
+
+The protocol string is inserted into the tree as shallowly as possible. If there
+is a conflict, the protocol string should be inserted one level down and
+duplicate the conflicting protocol string to the new level if it is not
+tombstoned. The conflicting protocol string MUST not removed from its original
+position as it may still be referenced.
```
richard-ramos, line 71: when would a client actually need an old abbreviation to keep working after the protocol set changes? could we just let that fail and retry with the updated set?

connections/multistream-select-v2.md
richard-ramos: yeah, this also makes me wonder about the following:

iiuc, the problem described in #700 is that once peers know each other's supported protocols, opening a stream shouldn't require sending the full protocol string again. we just need a compact mapping from protocol ids to abbreviations for that.

to solve this, this pr also defines how that mapping behaves across protocol additions, removals, readding protocols, and stale protocol sets, which introduces some complexity around tombstones and the abbreviation tree.

is all of this required to solve #700? or could something simpler, like a dictionary scoped to the connection, be enough, with peers learning about new protocols via identify push when enabled?

PR #706, opened 2026-03-04
add gossipsub v1.4 spec addressing reviewer feedback

bomanaps:
Continues PR https://github.com/libp2p/specs/pull/654 by @ufarooqstatus and @kaiserd with reviewer feedback from @Nashatyrev  and @MarcoPolo  addressed. Adds deferDurationMs field to IMRECEIVING for sender-computed timeouts, enforces one-shot rule per messageID, and reframes safety strategy as optional, Also corrects messageLength field type from int32 to uint64 in both ControlPreamble and ControlIMReceiving.

### DIFF
--- a/pubsub/gossipsub/extensions/extensions.proto
+++ b/pubsub/gossipsub/extensions/extensions.proto
@@ -53,7 +53,7 @@ message RPC {
 
   // PREAMBLE and IMRECEIVING extension for large message handling
   // See: ./experimental/preamble-imreceiving.proto
-  optional PreambleImreceivingExtension preambleImreceiving = 3847291;
+  // optional PreambleImreceivingExtension preambleImreceiving = 3847291;
 
 }
 
--- a/pubsub/gossipsub/extensions/extensions.proto
+++ b/pubsub/gossipsub/extensions/extensions.proto
@@ -8,6 +8,10 @@ message ControlExtensions {
 
   optional bool testExtension = 6492434;
 
+  // PREAMBLE and IMRECEIVING extension for large message handling
+  // See: ./experimental/preamble-imreceiving.md
+  optional bool preambleImreceiving = 3847291;
+
 }
 
 message ControlMessage {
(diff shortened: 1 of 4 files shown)


MarcoPolo:
This can be an extension right? You can reserve an experimental code point and deploy this if you want

bomanaps:
> This can be an extension right? You can reserve an experimental code point and deploy this if you want

I will  reframe as  v1.3 experimental extension with reserved code points, should I  add this to extensions.proto or create a separate extension spec doc? Will update the PR accordingly.

MarcoPolo:
Add the codepoint here: https://github.com/libp2p/specs/blob/bba73cb7b80f4c7f7d880f9c59d1c9898a33b9db/pubsub/gossipsub/extensions/extensions.proto#L10

And have a separate file for all of this extension protobufs. If this becomes a canonical extension we can import it into extensions.proto.

bomanaps:
> Add the codepoint here:
> 
> https://github.com/libp2p/specs/blob/bba73cb7b80f4c7f7d880f9c59d1c9898a33b9db/pubsub/gossipsub/extensions/extensions.proto#L10
> 
> And have a separate file for all of this extension protobufs. If this becomes a canonical extension we can import it into extensions.proto.

 Thanks for the guidance! I've added the codepoint and moved the protobufs to a separate file,  could you please help me review and let me know if everything looks right?

pubsub/gossipsub/extensions/extensions.proto
```diff
@@ -47,6 +51,10 @@ message RPC {
 
   optional TestExtension testExtension = 6492434;
 
+  // PREAMBLE and IMRECEIVING extension for large message handling
+  // See: ./experimental/preamble-imreceiving.proto
+  optional PreambleImreceivingExtension preambleImreceiving = 3847291;
```
MarcoPolo:
```suggestion
  // optional PreambleImreceivingExtension preambleImreceiving = 3847291;
```

Let's leave this commented out for now so that this is still a valid standalone file.

bomanaps: Done

PR #708, opened 2026-03-30
Fix broken libp2p implementations link in README.md

seetadev:
The hyperlink for libp2p implementations was pointing to a page that no longer exists.

This PR updates the link to the correct and currently active URL for libp2p implementations to ensure accuracy and improve navigation within the specs repository.

No other changes included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
 
 This repository contains the specifications for [`libp2p`](https://libp2p.io), a
 framework and suite of protocols for building peer-to-peer network applications.
-libp2p has several [implementations][libp2p_implementations], with more in development.
+libp2p has several [implementations](https://libp2p.io/projects/), with more in development.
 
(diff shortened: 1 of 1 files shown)


MarcoPolo:
You should change this line instead: https://github.com/libp2p/specs/blob/bba73cb7b80f4c7f7d880f9c59d1c9898a33b9db/README.md?plain=1#L120.

But arguably, https://libp2p.io/implementations should redirect to the new link instead.

seetadev:
@MarcoPolo: Thanks for your feedback and pointers. Appreciate it.

I’ve updated the reference at the suggested line so it correctly maps to the current implementations page for now.

Completely agree that having `https://libp2p.io/implementations` redirect to the new location is the better fix. We’ll coordinate on the website side to get that in place. Will open a new PR and redirect to `https://libp2p.io/implementations` once that is done.

Appreciate the review.

PR #709, opened 2026-04-08
Add signedPeerRecord and discussion of message sizes to identify

achingbrain:
Adds the missing `signedPeerRecord` field and an implementation notes section to discuss max message sizes and how to send/receive multiple identify messages during one transaction.

Also updates the language around receiving messages to make it clear that multiple messages may be sent.

### DIFF
--- a/identify/README.md
+++ b/identify/README.md
@@ -165,7 +165,7 @@ With the addition of signed peer records, Circuit Relay addresses, multiaddrs
 that contain certificate hashes, etc, it is now much more likely that this
 threshold will be exceeded.
 
-If an a modern implementation's Identify message would exceed this limit, it
+If a modern implementation's Identify message would exceed this limit, it
 should break it up into smaller chunks.
 
 For optimum backwards compatibility the first message SHOULD NOT exceed 2KB.
--- a/identify/README.md
+++ b/identify/README.md
@@ -5,7 +5,7 @@
 
 | Lifecycle Stage | Maturity Level | Status | Latest Revision |
 |-----------------|----------------|--------|-----------------|
-| 3A              | Recommendation | Active | r1, 2021-08-09  |
+| 3A              | Recommendation | Active | r2, 2026-04-08  |
 
 Authors: [@vyzo]
 
@@ -17,6 +17,7 @@ Interest Group: [@yusefnapora], [@tomaka], [@richardschneider], [@Stebalien], [@
 [@richardschneider]: https://github.com/richardschneider
 [@Stebalien]: https://github.com/Stebalien
 [@bigs]: https://github.com/bigs
+[@achingbrain]: https://github.com/achingbrain
 
 See the [lifecycle document][lifecycle-spec] for context about the maturity level
 and spec status.
@@ -26,17 +27,21 @@ and spec status.
 ## Table of Contents
 
 - [Identify v1.0.0](#identify-v100)
-    - [Table of Contents](#table-of-contents)
-    - [Overview](#overview)
-        - [`identify`](#identify)
-        - [`identify/push`](#identifypush)
-    - [The Identify Message](#the-identify-message)
-        - [protocolVersion](#protocolversion)
-        - [agentVersion](#agentversion)
-        - [publicKey](#publickey)
-        - [listenAddrs](#listenaddrs)
-        - [observedAddr](#observedaddr)
-        - [protocols](#protocols)
+  - [Table of Contents](#table-of-contents)
+  - [Overview](#overview)
+    - [`identify`](#identify)
+    - [`identify/push`](#identifypush)
+  - [The Identify Message](#the-identify-message)
+    - [protocolVersion](#protocolversion)
+    - [agentVersion](#agentversion)
+    - [publicKey](#publickey)
+    - [listenAddrs](#listenaddrs)
+    - [observedAddr](#observedaddr)
+    - [protocols](#protocols)
+    - [signedPeerRecord](#signedpeerrecord)
+  - [Implementation notes](#implementation-notes)
+    - [Message sizes and message splitting](#message-sizes-and-message-splitting)
+    - [Receiving multiple messages](#receiving-multiple-messages)
 
 
 ## Overview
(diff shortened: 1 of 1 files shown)


MarcoPolo (approved):

identify/README.md
```diff
@@ -133,3 +140,47 @@ clients only support initiating requests while some servers (only) support
...
+
+Early implementations of the Identify protocol limited the incoming message size
+to 2KB and would reject any messages larger than this. Others have since
+increased this limit to 4KB or 8KB.
+
+With the addition of signed peer records, Circuit Relay addresses, multiaddrs
+that contain certificate hashes, etc, it is now much more likely that this
+threshold will be exceeded.
+
+If an a modern implementation's Identify message would exceed this limit, it
```
dennis-tra:
```suggestion
If a modern implementation's Identify message would exceed this limit, it
```

identify/README.md
```diff
@@ -133,3 +140,47 @@ clients only support initiating requests while some servers (only) support
...
+
+With the addition of signed peer records, Circuit Relay addresses, multiaddrs
+that contain certificate hashes, etc, it is now much more likely that this
+threshold will be exceeded.
+
+If a modern implementation's Identify message would exceed this limit, it
+should break it up into smaller chunks.
+
+For optimum backwards compatibility the first message SHOULD NOT exceed 2KB.
+Subsequent messages SHOULD NOT exceed 4KB.
```
wjmelements, line 172: Can signed peer records be subdivided? I suspect they can themselves exceed 4 KB.

dozyio: One possible issue is signature size of signed peer records if/when libp2p supports post quantum signatures - 4KB would be too small for `ML-DSA-44` and `ML-DSA-65`, 8KB would be too small for `ML-DSA-87`

| Variant | Public key (bytes) | Signature (bytes) |
| --- | ---: | ---: | ---: |
| `ML-DSA-44` | 1312 | 2420 |
| `ML-DSA-65` | 1952 | 3309 |
| `ML-DSA-87` | 2592 | 4627 |

johannamoran:
@dozyio @tabcat - If you share your thoughts on this pls?

identify/README.md
```diff
@@ -133,3 +140,47 @@ clients only support initiating requests while some servers (only) support
...
 advertise `foo` in their `protocols` list.
+
+### signedPeerRecord
+
+This field contains a serialized [SignedEnvelope](../RFC/0002-signed-envelopes.md)
+containing a [PeerRecord](../RFC/0003-routing-records.md) signed by the sending
+node.
+
+It normally contains the same addresses as the `listenAddrs` field, but in a
+form that lets us share authenticated addresses with other peers.
```
gmelodie, line 151:
```suggestion
It normally contains the same addresses as the `listenAddrs` field, but in a
format that allows to share authenticated addresses with other peers.
```

identify/README.md
```diff
@@ -133,3 +140,47 @@ clients only support initiating requests while some servers (only) support
...
+
+### Message sizes and message splitting
+
+Early implementations of the Identify protocol limited the incoming message size
+to 2KB and would reject any messages larger than this. Others have since
+increased this limit to 4KB or 8KB.
+
+With the addition of signed peer records, Circuit Relay addresses, multiaddrs
+that contain certificate hashes, etc, it is now much more likely that this
+threshold will be exceeded.
```
gmelodie, line 166: I think this needs to reason about which is better in which case and why. As it's written rn it just seems like research on current state of implementations

identify/README.md
```diff
@@ -133,3 +140,47 @@ clients only support initiating requests while some servers (only) support
...
+
+If a modern implementation's Identify message would exceed this limit, it
+should break it up into smaller chunks.
```
gmelodie, line 169: i think the past 2 paragraphs and this one could all be replaced by:
> An Identify message that exceeds this limit SHOULD be broken up into smaller chunks [of size XYZ].

identify/README.md
```diff
@@ -133,3 +140,47 @@ clients only support initiating requests while some servers (only) support
...
+
+For optimum backwards compatibility the first message SHOULD NOT exceed 2KB.
+Subsequent messages SHOULD NOT exceed 4KB.
+
+Listen addresses SHOULD be ordered such that addresses that have a higher
+likelihood of being dialed successfully (e.g. public, non-NAT or Circuit Relay)
```
gmelodie, line 175: This spec should say which are more likely. This feels a bit too vague

identify/README.md
```diff
@@ -133,3 +140,47 @@ clients only support initiating requests while some servers (only) support
...
+are sent in the first Identify message and the `signedPeerRecord`, if used.
+
+### Receiving multiple messages
+
+When multiple messages are received, all fields SHOULD be respected, with later
+updates for fields with a cardinality of one taking priority.
```
gmelodie, line 181:
```suggestion
When multiple messages are received, later updates for fields with a cardinality of one SHOULD take priority.
```

identify/README.md
```diff
@@ -133,3 +140,47 @@ clients only support initiating requests while some servers (only) support
...
+
+Where repeated fields encountered in subsequent messages, they SHOULD be
+appended to the earlier occurrences of the field and deduplicated as necessary.
+
+A maximum of 10 Identify messages SHOULD be accepted.
```
gmelodie, line 186: Would be good to add a note on why 10 is the chosen number

identify/README.md
```diff
@@ -133,3 +140,47 @@ clients only support initiating requests while some servers (only) support
...
+likelihood of being dialed successfully (e.g. public, non-NAT or Circuit Relay)
+are sent in the first Identify message and the `signedPeerRecord`, if used.
+
+### Receiving multiple messages
+
+When multiple messages are received, all fields SHOULD be respected, with later
+updates for fields with a cardinality of one taking priority.
+
+Where repeated fields encountered in subsequent messages, they SHOULD be
+appended to the earlier occurrences of the field and deduplicated as necessary.
```
gmelodie, line 184:
```suggestion
Repeated fields encountered in subsequent messages SHOULD be deduplicated and
appended to earlier occurrences of the field.
```

identify/README.md
```diff
@@ -133,3 +140,47 @@ clients only support initiating requests while some servers (only) support
...
+### signedPeerRecord
+
+This field contains a serialized [SignedEnvelope](../RFC/0002-signed-envelopes.md)
+containing a [PeerRecord](../RFC/0003-routing-records.md) signed by the sending
+node.
+
+It normally contains the same addresses as the `listenAddrs` field, but in a
+form that lets us share authenticated addresses with other peers.
+
+When present the fields from a `PeerRecord` MUST be preferred over those in the
```
richard-ramos, line 153:
```suggestion
When present, the fields from a `PeerRecord` MUST be preferred over those in the
```

identify/README.md
richard-ramos, line 172: If a signed peer record alone exceeds 4kb what should the sender do? it seems that splitting a signed peer record would not work with the merge rules below, because it is a single bytes field, so the later parts would replace the earlier parts instead of joining them.

PR #710, opened 2026-04-11
RFC-0004: ML-DSA Key Support for libp2p Peer Identities

dozyio:
RFC to kick of some discussion around a post quantum strategy for libp2p

### DIFF
--- a/RFC/0004-mldsa-peer-ids.md
+++ b/RFC/0004-mldsa-peer-ids.md
@@ -138,3 +138,4 @@ recommendation:
   identity strategies.
 3. Whether to define mandatory test vectors in the peer-id spec for ML-DSA
   encodings.
+4. Canonical private key format: should it be FIPS 204 expanded form ([multiformats/multicodec#399](https://github.com/multiformats/multicodec/pull/399), W3C DI Quantum-Safe Cryptosuite)?
--- a/RFC/0004-mldsa-peer-ids.md
+++ b/RFC/0004-mldsa-peer-ids.md
@@ -51,6 +51,8 @@ This RFC supports the three standardized ML-DSA parameter sets:
 - `ML-DSA-65`
 - `ML-DSA-87`
 
+Parameter names and sizes are defined by [FIPS 204](#references).
+
 ### 3. Key serialization in protobuf `Data`
 
 For `PublicKey` and `PrivateKey` messages where `Type = MLDSA`, the `Data`
@@ -66,24 +68,27 @@ Where `variant-prefix` is one byte:
 - `0x02` = `ML-DSA-65`
 - `0x03` = `ML-DSA-87`
 
-Raw key lengths are the standard ML-DSA key sizes for each variant:
+Raw key and signature lengths are fixed per ML-DSA variant:
 
-- `ML-DSA-44`: public key `1312` bytes, expanded private key `2560` bytes
-- `ML-DSA-65`: public key `1952` bytes, expanded private key `4032` bytes
-- `ML-DSA-87`: public key `2592` bytes, expanded private key `4896` bytes
+| Variant | Public key (bytes) | Expanded private key (bytes) | Signature (bytes) |
+| --- | ---: | ---: | ---: |
+| `ML-DSA-44` | 1312 | 2560 | 2420 |
+| `ML-DSA-65` | 1952 | 4032 | 3309 |
+| `ML-DSA-87` | 2592 | 4896 | 4627 |
 
-Implementations MUST reject malformed key payloads (unknown prefix,
-missing prefix, or length mismatch).
+Implementations MUST reject malformed key payloads, including unknown
+`variant-prefix`, missing prefix, or length mismatch against the table above.
 
 ### 4. Signature semantics
 
 ML-DSA signatures are generated and verified using the standard ML-DSA
 algorithm for the corresponding parameter set.
 
-- Input to signing is the exact message bytes.
-- No additional pre-hash step is applied by libp2p at the key API boundary.
-- Verification is performed over the same message bytes with the same parameter
-  set.
+- Implementations MUST sign the exact message bytes.
+- Implementations MUST NOT apply an additional pre-hash at the libp2p key API
+  boundary.
+- Implementations MUST verify signatures over the exact same message bytes and
+  ML-DSA parameter set.
 
 ### 5. Peer ID derivation
 
@@ -121,7 +126,9 @@ Current language support for ML-DSA is still evolving:
 - Node.js runtime support is experimental.
 - Browser WebCrypto support is unavailable.
 - Go has an internal implementation for ML-DSA but no public API as of Go 1.26.
-  - Worth noting: `crypto/mlkem` went public in Go 1.24, so `crypto/mldsa` going public soon is likely. For now, `github.com/cloudflare/circl/sign/mldsa` is available.
+  - Worth noting: `crypto/mlkem` went public in Go 1.24, `crypto/mldsa` is in
+    proposal phase [golang/go#77626](https://github.com/golang/go/issues/77626).
+    For now, `github.com/cloudflare/circl/sign/mldsa` is available.
 
 This RFC specifies interoperability behavior independent of implementation
 maturity.
@@ -136,7 +143,14 @@ recommendation:
 1. Default variant selection for new key generation (`ML-DSA-44`, `-65`, or
   `-87`).
 2. Whether any libp2p subsystem should require dual-signature or hybrid
-  identity strategies.
+   identity strategies.
 3. Whether to define mandatory test vectors in the peer-id spec for ML-DSA
-  encodings.
-4. Canonical private key format: should it be FIPS 204 expanded form ([multiformats/multicodec#399](https://github.com/multiformats/multicodec/pull/399), W3C DI Quantum-Safe Cryptosuite)?
+   encodings.
+4. Canonical private key format: should it be FIPS 204 expanded form
+   [multiformats/multicodec#399](https://github.com/multiformats/multicodec/pull/399),
+   W3C DI Quantum-Safe Cryptosuite)?
+
+## References
+
+- [FIPS 204] NIST, *Module-Lattice-Based Digital Signature Standard* (ML-DSA),
+  https://csrc.nist.gov/pubs/fips/204/final
--- a/RFC/0004-mldsa-peer-ids.md
+++ b/RFC/0004-mldsa-peer-ids.md
@@ -121,6 +121,7 @@ Current language support for ML-DSA is still evolving:
 - Node.js runtime support is experimental.
 - Browser WebCrypto support is unavailable.
 - Go has an internal implementation for ML-DSA but no public API as of Go 1.26.
+  - Worth noting: `crypto/mlkem` went public in Go 1.24, so `crypto/mldsa` going public soon is likely. For now, `github.com/cloudflare/circl/sign/mldsa` is available.
 
 This RFC specifies interoperability behavior independent of implementation
 maturity.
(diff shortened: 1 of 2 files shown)


paschal533:
Hi @dozyio , great to see this as a formal RFC. I have been working on the handshake side of the PQC picture (XXhfs + X-Wing KEM,  https://github.com/ChainSafe/js-libp2p-noise/pull/665)
From my implementation and benchmarks, the combined overhead when XXhfs + ML-DSA-65 identity are both enabled on both sides is roughly:
 | Scenario                           | Approx. wire size |
 |------------------------------------|-------------------|
 | Classical XX + Ed25519             | ~500 B            |
 | XXhfs + Ed25519 (KEM upgrade only) | ~2,852 B          |
 | XXhfs + ML-DSA-65 both sides       | ~9,400 B          |
 
 The ML-DSA-65 identity cost (~6,600 B across both sides: 1,952 B public key + 3,309 B signature + overhead, per side) is 2.7x larger than the KEM overhead (~2,352 B). Worth factoring in when discussing whether to require hybrid signatures or allow a phased migration.

One argument for deferring hybrid signatures: the threat timelines are different. For the KEM, XXhfs hybrid makes sense today because of store-now-decrypt-later... someone recording traffic now can decrypt it later. For identity signatures, the threat only materialises once quantum computers actually exist, so a clean migration from Ed25519 to ML-DSA could be reasonable rather than carrying both indefinitely. That said, downgrade resistance is a real concern your security considerations section mentions
I have working test vectors from the JS XXhfs implementation that cover ML-DSA-65 signing and verification. Happy to contribute those if it helps progress

RFC/0004-mldsa-peer-ids.md
```diff
@@ -0,0 +1,140 @@
...
+
+- `0x01` = `ML-DSA-44`
+- `0x02` = `ML-DSA-65`
+- `0x03` = `ML-DSA-87`
+
+Raw key lengths are the standard ML-DSA key sizes for each variant:
+
+- `ML-DSA-44`: public key `1312` bytes, private key `2560` bytes
+- `ML-DSA-65`: public key `1952` bytes, private key `4032` bytes
+- `ML-DSA-87`: public key `2592` bytes, private key `4896` bytes
```
lidel: The stated sizes (`2560` / `4032` / `4896` bytes) are the FIPS 204 *expanded* private key. Both the JS implementation ([js-libp2p#3432](https://github.com/libp2p/js-libp2p/pull/3432)) and Go's `crypto/internal/fips140/mldsa` serialize the 32-byte **seed** (`ξ`, FIPS 204 §3.6) instead.

The two options are incompatible on the wire:

| Format | Size | Used by |
|--------|------|---------|
| Seed | 32 bytes (all variants) | current JS and Go prototype implementations |
| Expanded | 2560 / 4032 / 4896 bytes | [multiformats/multicodec#399](https://github.com/multiformats/multicodec/pull/399), W3C DI Quantum-Safe Cryptosuite |


Size alone is enough to guess what is stored, but probably better to be explicit in the spec,rather than implicit:


```suggestion
- `ML-DSA-44`: public key `1312` bytes, expanded private key `2560` bytes
- `ML-DSA-65`: public key `1952` bytes, expanded private key `4032` bytes
- `ML-DSA-87`: public key `2592` bytes, expanded private key `4896` bytes
```

RFC/0004-mldsa-peer-ids.md
```diff
@@ -0,0 +1,140 @@
...
+
+```
+<variant-prefix><raw-key-bytes>
+```
+
+Where `variant-prefix` is one byte:
+
+- `0x01` = `ML-DSA-44`
+- `0x02` = `ML-DSA-65`
+- `0x03` = `ML-DSA-87`
```
lidel, line 69: For **public keys**: length alone identifies the variant (1312 / 1952 / 2592 bytes are unique), so the prefix is redundant. iiuc the JS PoC implementation already falls back to length-based detection? 

For **private keys**: necessity depends on the format choice:
- Seed (32 bytes for all variants): prefix is **required** to identify the variant.
- Expanded (unique sizes per variant): prefix is redundant, consistent with the   multicodec approach.

I think if you go with expanded private key format for this RFC, then this prefix is redundant, all we need is length of the expanded priv key alone to tell which version it is

RFC/0004-mldsa-peer-ids.md
```diff
@@ -0,0 +1,140 @@
...
+
+### 4. Signature semantics
+
+ML-DSA signatures are generated and verified using the standard ML-DSA
+algorithm for the corresponding parameter set.
+
+- Input to signing is the exact message bytes.
+- No additional pre-hash step is applied by libp2p at the key API boundary.
+- Verification is performed over the same message bytes with the same parameter
+  set.
```
lidel: [FIPS 204 §5.2](https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.204.pdf) defines `ML-DSA.Sign(sk, M, ctx)` where `ctx` is a caller-supplied context string (0–255 bytes). The spec says nothing about what value libp2p uses for  `ctx`. Two compliant implementations using different values produce incompatible signatures.

Existing libp2p signing contexts use message-level string prefixes prepended
before calling `Sign()`, not algorithm-level context parameters:

- Noise: `"noise-libp2p-static-key:" || static_pub_key`   ([handshake.go](https://github.com/libp2p/go-libp2p/blob/master/p2p/security/noise/handshake.go#L24))
- TLS: `"libp2p-tls-handshake:" || cert_pub_key`  ([crypto.go](https://github.com/libp2p/go-libp2p/blob/master/p2p/security/tls/crypto.go#L26))
- Record envelope: `varint(len(domain)) || domain || varint(len(type)) || type || varint(len(payload)) || payload` ([envelope.go](https://github.com/libp2p/go-libp2p/blob/master/core/record/envelope.go#L268))

So.. iiuc to stay consistent with this pattern, the RFC should explicitly state what `ctx` is? `ctx = ""` ? `ctx = "libp2p-identity"`?


```suggestion
- Input to signing is the exact message bytes.
  - Implementations MUST pass an empty context string (`ctx = "??TODO DECIDE??"`) to `ML-DSA.Sign`. Domain separation is the responsibility of the caller via message prefixes, as with all other libp2p key types.
- No additional pre-hash step is applied by libp2p at the key API boundary.
- Verification is performed over the same message bytes with the same parameter
  set.
```

Note I'm not a cryptographer, but it may be safer to do belt-and-supenders and also require `ctx = "libp2p-identity"`.

RFC/0004-mldsa-peer-ids.md
```diff
@@ -0,0 +1,140 @@
...
+
+The key protobuf enum in the peer-id spec is extended with:
+
+```protobuf
+enum KeyType {
+  RSA = 0;
+  Ed25519 = 1;
+  Secp256k1 = 2;
+  ECDSA = 3;
+  MLDSA = 4;
```
lidel, line 42: `MLDSA = 4` covers all three parameter sets; the variant is only in `Data`. 

Multicodec assigns separate code points per variant ([multiformats/multicodec#392](https://github.com/multiformats/multicodec/pull/392)).

With this design, the security level is not visible from the protobuf envelope without parsing  the key data. 

The choice is defensible, but this is an opportunity to remove indirection for new key types and just use the same number as in [multicodec table](https://github.com/multiformats/multicodec/pull/399/changes):


Assuming we use expanded key format, we could do:

```suggestion
  MLDSA44 = 0x1317;
  MLDSA65 = 0x1318;
  MLDSA87 = 0x1319;
```

RFC/0004-mldsa-peer-ids.md
```diff
@@ -0,0 +1,140 @@
...
+- Deployments should continue evaluating algorithm maturity, implementation
+  quality, and runtime support in their target environments.
+
+## Implementation Status (Non-Normative)
+
+Current language support for ML-DSA is still evolving:
+
+- Node.js runtime support is experimental.
+- Browser WebCrypto support is unavailable.
+- Go has an internal implementation for ML-DSA but no public API as of Go 1.26.
```
lidel, line 128:
```suggestion
- Go has an internal implementation for ML-DSA but no public API as of Go 1.26.
  - Worth noting: `crypto/mlkem` went public in Go 1.24, so `crypto/mldsa` going public soon is likely. For now, `github.com/cloudflare/circl/sign/mldsa` is available.
```

RFC/0004-mldsa-peer-ids.md
```diff
@@ -0,0 +1,140 @@
...
+
+The following questions should be resolved before promoting this to a candidate
+recommendation:
+
+1. Default variant selection for new key generation (`ML-DSA-44`, `-65`, or
+  `-87`).
+2. Whether any libp2p subsystem should require dual-signature or hybrid
+  identity strategies.
+3. Whether to define mandatory test vectors in the peer-id spec for ML-DSA
+  encodings.
```
lidel:
```suggestion
  encodings.
4. Canonical private key format: should it be FIPS 204 expanded form ([multiformats/multicodec#399](https://github.com/multiformats/multicodec/pull/399), W3C DI Quantum-Safe Cryptosuite)?
```

RFC/0004-mldsa-peer-ids.md
MarcoPolo: The context string should be a unique based on what the signature is for. Similar to the context in `Ed25519ctx`. Applications should be be able to pass this in.

MarcoPolo:
A different tact would be to add a PKIX key type, and the data is the PKIX key encoding. Then we would get ML-DSA for free with https://www.rfc-editor.org/rfc/rfc9881.html#name-ml-dsa-public-keys-in-pkix

MarcoPolo:
> A different tact would be to add a PKIX key type, and the data is the PKIX key encoding. Then we would get ML-DSA for free with https://www.rfc-editor.org/rfc/rfc9881.html#name-ml-dsa-public-keys-in-pkix

To be concrete: https://github.com/libp2p/specs/pull/711

RFC/0004-mldsa-peer-ids.md
dozyio: There is a proposal for this https://github.com/golang/go/issues/77626 - so would hopefully in the next release or two

RFC/0004-mldsa-peer-ids.md
dennis-tra: Can't the type be inferred from the length of the data field? This would be similar to the RSA case, no?

**Edit:** I see that this is being discussed in another comment.

dennis-tra: If my understanding of the protobuf wire format is correct I would limit the enum values to a maximum of `7` (as long as there's space) because going beyond that would then require two bytes on the wire due to varint encoding.

MarcoPolo: I think you mean 7 bits, right? or values less than `0x80/128` https://protobuf.dev/programming-guides/encoding/#varints.

dennis-tra: Oh of course, my brain slipped. For some reason I thought the number seven requires seven bits (so that the maximum enum value is 7). The argument still holds because the proposed multicodec values are larger than 127 and thus would still require multiple bytes to encode

dennis-tra: One could argue that it doesn't matter because MLDSA keys are large anyway. One thing to consider could be how they will fit in MTUs (just to rule out that these values don't push the encoded data into an additional packet). At the same time they are not sent in isolation anyway. So nevermind...

RFC/0004-mldsa-peer-ids.md
dozyio: thanks, added

PR #712, opened 2026-04-14
Add support for COSE keys and key sets

MarcoPolo:
See https://github.com/multiformats/multicodec/pull/400 for a detailed motivation.

cc @lidel

### DIFF
--- a/peer-ids/peer-ids.md
+++ b/peer-ids/peer-ids.md
@@ -216,6 +216,10 @@ For backwards compatibility, if a key algorithm has a prior libp2p specific enco
 The Data field is defined by [Section 7 of RFC
 9052](https://www.rfc-editor.org/rfc/rfc9052#section-7).
 
+The `COSE_Key` MUST be encoded using [CBOR Core Deterministic Encoding (RFC 8949 §4.2)](https://www.rfc-editor.org/rfc/rfc8949#section-4.2). Implementations MUST reject non-deterministic encodings.
+
+Signing and verification for each member are defined by that member's COSE `alg`  parameter.
+
 #### COSE KeySet
 
 The Data field is defined by [Section 7 of RFC
--- a/peer-ids/peer-ids.md
+++ b/peer-ids/peer-ids.md
@@ -63,6 +63,9 @@ enum KeyType {
 	Ed25519 = 1;
 	Secp256k1 = 2;
 	ECDSA = 3;
+	
+	PKIX = 0x40;
+	PKCS8 = 0x41;
 }
 
 message PublicKey {
@@ -110,11 +113,13 @@ The second is for generating peer ids; this is discussed in the section below.
 
 ### Key Types
 
-Four key types are supported:
+The following key types are supported:
  - RSA
  - Ed25519
  - Secp256k1
  - ECDSA
+ - PKIX Encoded Public Keys
+ - PKCS#8 Encoded Private Keys
 
 Implementations MUST support Ed25519. Implementations SHOULD support RSA if they wish to
 interoperate with the mainline IPFS DHT and the default IPFS bootstrap nodes. Implementations MAY
@@ -184,6 +189,25 @@ We encode the private key using DER-encoded PKIX.
 To sign a message, we hash the message with SHA 256, and then sign it with the
 [ECDSA standard algorithm](https://tools.ietf.org/html/rfc6979), then we encode
 it using [DER-encoded ASN.1.](https://wiki.openssl.org/index.php/DER)
+
+#### PKIX Public Keys
+
+The PKIX key type only encodes public keys. The Data field is the [PKIX
+encoding](https://www.rfc-editor.org/rfc/rfc5280) of the public key. The public
+key and algorithm are identified by the [Subject Public Key
+Info](https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.7) field.
+
+Signature Verification is defined by the key algorithm used.
+
+For backwards compatibility, if a key algorithm has a prior libp2p specific encoding, implementers SHOULD prefer that.
+
+#### PKCS#8 Private Keys
+
+The PKCS8 key type only encodes private keys. The Data field is the [PKCS#8 encoding](https://www.rfc-editor.org/rfc/rfc5958.html) of the private key.
+
+Signing is defined by the key algorithm used.
+
+For backwards compatibility, if a key algorithm has a prior libp2p specific encoding, implementers SHOULD prefer that.
 
 ### Test vectors
 
--- a/peer-ids/peer-ids.md
+++ b/peer-ids/peer-ids.md
@@ -66,6 +66,8 @@ enum KeyType {
 	
 	PKIX = 0x40;
 	PKCS8 = 0x41;
+	COSE_KEY = 0x42;
+	COSE_KEY_SET = 0x43;
 }
 
 message PublicKey {
@@ -208,6 +210,16 @@ The PKCS8 key type only encodes private keys. The Data field is the [PKCS#8 enco
 Signing is defined by the key algorithm used.
 
 For backwards compatibility, if a key algorithm has a prior libp2p specific encoding, implementers SHOULD prefer that.
+
+#### COSE Key
+
+The Data field is defined by [Section 7 of RFC
+9052](https://www.rfc-editor.org/rfc/rfc9052#section-7).
+
+#### COSE KeySet
+
+The Data field is defined by [Section 7 of RFC
+9052](https://www.rfc-editor.org/rfc/rfc9052#section-7).
 
 ### Test vectors
 


peer-ids/peer-ids.md
```diff
@@ -209,6 +211,16 @@ Signing is defined by the key algorithm used.
 
 For backwards compatibility, if a key algorithm has a prior libp2p specific encoding, implementers SHOULD prefer that.
 
+#### COSE Key
+
+The Data field is defined by [Section 7 of RFC
+9052](https://www.rfc-editor.org/rfc/rfc9052#section-7).
```
lidel, line 217: Something I realized in the context of libp2p is that PeerID is a hash of this protobuf.

RFC 9052 `COSE_Key` is plain CBOR; no deterministic encoding is required. But this spec requires deterministic protobuf encoding so peer IDs are stable. 

Problem: Two conformant COSE encoders can produce  different `Data` bytes for the same key, yielding different peer IDs.

 Fix: unsure, maybe require [RFC 8949 §4.2 Core Deterministic Encoding](https://www.rfc-editor.org/rfc/rfc8949#section-4.2) for `COSEKEY*`, and reject non-deterministic encodings on decode?


```suggestion
#### COSE Key

The Data field is defined by [Section 7 of RFC
9052](https://www.rfc-editor.org/rfc/rfc9052#section-7).

The `COSE_Key` MUST be encoded using [CBOR Core Deterministic Encoding (RFC 8949 §4.2)](https://www.rfc-editor.org/rfc/rfc8949#section-4.2). Implementations MUST reject non-deterministic encodings.

Signing and verification for each member are defined by that member's COSE `alg`  parameter.
```

peer-ids/peer-ids.md
```diff
@@ -209,6 +211,16 @@ Signing is defined by the key algorithm used.
...
+
+#### COSE KeySet
+
+The Data field is defined by [Section 7 of RFC
+9052](https://www.rfc-editor.org/rfc/rfc9052#section-7).
```
lidel, line 226: PeerID semantics feel undefined?

Fix: spell out identity and verification rules, or drop `COSE_KEY_SET` from this PR until the use case is settled.

I think it may be  beneficial to keep it to support hybrid keys and signatures, for example when we have both ed25519+something post-quantum secure, and requiring both sigs/handshakes etc.  

so maybe:

```suggestion
#### COSE KeySet

The Data field is defined by [Section 7 of RFC
9052](https://www.rfc-editor.org/rfc/rfc9052#section-7).

`COSE_KeySet` is intended for hybrid keys and signatures, for example pairing Ed25519 with a post-quantum scheme so that handshakes and signatures combine both. The following rules apply:

- The `COSE_KeySet` MUST be encoded using [CBOR Core Deterministic Encoding (RFC 8949 §4.2)](https://www.rfc-editor.org/rfc/rfc8949#section-4.2). Implementations MUST reject non-deterministic encodings.
- Members MUST be ordered in ascending lexicographic order of their deterministic CBOR encoding. This ensures independent encoders produce identical bytes for the same set, so the resulting PeerID is stable.
- A `COSE_KeySet` MUST contain at least two members. A single-member set MUST be encoded as `COSE_Key` instead, so the same identity has exactly one on-the-wire representation.
- Each COSE `alg` (and key-type family) MUST appear at most once in the set, removing ambiguity about which member is expected to produce a given signature or handshake component.
- A signature or handshake over a `COSE_KeySet` identity is valid only if **every** member validates its corresponding component. Peers MUST NOT accept proofs that cover only a subset; doing so would defeat the purpose of the hybrid.

Signing and verification for each member are defined by that member's COSE `alg`  parameter.
```

peer-ids/peer-ids.md
```diff
@@ -66,6 +66,8 @@ enum KeyType {
 	
 	PKIX = 0x40;
 	PKCS8 = 0x41;
+	COSE_KEY = 0x42;
+	COSE_KEY_SET = 0x43;
```
lidel, line 70:
```suggestion
	COSE_Key = 0x42;
	COSE_KeySet = 0x43;
```

Other ones look like `Secp256k1` and RFC 9052 uses `COSE_KeySet`. 

Perhaps rename to match the RFC?

lidel (commented):
I believe we need to explicitly require deterministic CBOR variant (details inline)

peer-ids/peer-ids.md
MarcoPolo: Yes. 9052 has encoding restrictions, but they don't apply to the keys.

peer-ids/peer-ids.md
MarcoPolo: I think it makes sense to back out the cose keyset changes. It only gives us an array of cose keys. You could imagine having a array of these protobufs instead.

dennis-tra:
Doesn't this PR allow to e.g. represent an RSA key in three different ways? All of which would result in a differently encoded PeerID while all use the same key. The same is true when one peer uses COSE and the other PKIX with the same key.

Why is such flexibility needed? The last change to the PeerID spec was three years ago. The rate of change doesn't really signal that fast iterations with new key types is required.

peer-ids/peer-ids.md
lidel: How do you mean? Current wire format for `libp2p-key` does not have repeatable fields, we don't have a spec for `libp2p-key[]`, that is something you would have to invent, and add a new multicodec for it. 

RFC for `COSE_KeySet` already exists, and could provide a backward-compatible way for hardening things with multiple keys (e.g. hybrid ed25519 + post-quantum).

Of course this is just an option, we dont have to support o require  hybrid signatures this way, or at all, and can defer to future codes, but feels sensible to preregister `COSE_KeySet` here to allow libp2p ecosystem to experiment and evolve without gatekeepers.

MarcoPolo: I mean using multiple protobuf messages instead of a COSE array. Maybe this doesn't work and the COSE array is better. I'm not sure. We'd also need to define how to hash the key set properly.

peer-ids/peer-ids.md
MarcoPolo: I also found this RFC: https://datatracker.ietf.org/doc/html/rfc9679 which sounds like what we should use instead of trying to define this.

lidel:
The flexibility is a harmless side-effect. This is more about future-proofing stack and reducing gatekeeper role of the protobuf here. 

We could lock down spec noting the pre-existing key types are only allowed to be defined the old way, to remove formal ambiguity, not feeling strongly about it tho.

peer-ids/peer-ids.md
```diff
@@ -203,14 +205,26 @@ For backwards compatibility, if a key algorithm has a prior libp2p specific enco
...
+9052](https://www.rfc-editor.org/rfc/rfc9052#section-7).
+
+The `COSE_Key` MUST be encoded using [CBOR Core Deterministic Encoding (RFC 8949 §4.2)](https://www.rfc-editor.org/rfc/rfc8949#section-4.2). Implementations MUST reject non-deterministic encodings.
+
+Signing and verification for each member are defined by that member's COSE `alg`  parameter.
+
+#### COSE KeySet
+
+The Data field is defined by [Section 7 of RFC
+9052](https://www.rfc-editor.org/rfc/rfc9052#section-7).
```
richard-ramos, line 226: How do signing and verification operations work when using a COSE_KeySet? i.e. how to choose which member signs, which verifies? Reading https://www.rfc-editor.org/rfc/rfc9052#section-7 it seems to me that authentication is left unspecified.

peer-ids/peer-ids.md
```diff
@@ -203,14 +205,26 @@ For backwards compatibility, if a key algorithm has a prior libp2p specific enco
...
 For backwards compatibility, if a key algorithm has a prior libp2p specific encoding, implementers SHOULD prefer that.
 
+#### COSE Key
+
+The Data field is defined by [Section 7 of RFC
+9052](https://www.rfc-editor.org/rfc/rfc9052#section-7).
+
+The `COSE_Key` MUST be encoded using [CBOR Core Deterministic Encoding (RFC 8949 §4.2)](https://www.rfc-editor.org/rfc/rfc8949#section-4.2). Implementations MUST reject non-deterministic encodings.
+
+Signing and verification for each member are defined by that member's COSE `alg`  parameter.
```
richard-ramos, line 221: In this section, we are refering to a single key, so maybe this should look like this?
```suggestion
Signing and verification are defined by the key's COSE `alg`  parameter.
```

peer-ids/peer-ids.md
richard-ramos, line 202:
```suggestion
Signature verification is defined by the key's algorithm.
```

peer-ids/peer-ids.md
```diff
@@ -203,14 +205,26 @@ For backwards compatibility, if a key algorithm has a prior libp2p specific enco
...
-encoding](https://www.rfc-editor.org/rfc/rfc5958.html) of the private key.
+The PKCS8 key type only encodes private keys. The Data field is the [PKCS#8 encoding](https://www.rfc-editor.org/rfc/rfc5958.html) of the private key.
 
 Signing is defined by the key algorithm used.
 
 For backwards compatibility, if a key algorithm has a prior libp2p specific encoding, implementers SHOULD prefer that.
 
+#### COSE Key
+
+The Data field is defined by [Section 7 of RFC
```
richard-ramos, line 216:
```suggestion
The Data field contains a `COSE_Key` encoded as specified in [Section 7 of RFC
```

peer-ids/peer-ids.md
```diff
@@ -203,14 +205,26 @@ For backwards compatibility, if a key algorithm has a prior libp2p specific enco
...
+9052](https://www.rfc-editor.org/rfc/rfc9052#section-7).
+
+The `COSE_Key` MUST be encoded using [CBOR Core Deterministic Encoding (RFC 8949 §4.2)](https://www.rfc-editor.org/rfc/rfc8949#section-4.2). Implementations MUST reject non-deterministic encodings.
+
+Signing and verification for each member are defined by that member's COSE `alg`  parameter.
+
+#### COSE KeySet
```
richard-ramos, line 223:
```suggestion
#### COSE Key Set
```

PR #715, opened 2026-04-23
docs: webrtc-direct-v2

dozyio:
* Adds webrtc-direct-v2 spec

### DIFF
--- a/webrtc/webrtc-direct.md
+++ b/webrtc/webrtc-direct.md
@@ -2,7 +2,7 @@
 
 | Lifecycle Stage | Maturity                  | Status | Latest Revision |
 |-----------------|---------------------------|--------|-----------------|
-| 2A              | Candidate Recommendation  | Active | r1, 2023-04-12  |
+| 2A              | Candidate Recommendation  | Active | r2, 2026-04-23  |
 
 Authors: [@mxinden]
 
@@ -34,6 +34,10 @@ The TLS certificate fingerprint in `/certhash` is a
 [multibase](https://github.com/multiformats/multibase) encoded
 [multihash](https://github.com/multiformats/multihash).
 
+WebRTC Direct v1 and v2 both use the same `/webrtc-direct` multiaddr protocol.
+Version selection is done via the ICE username fragment prefix, not via a new
+multiaddr protocol.
+
 For compatibility implementations MUST support hash algorithm
 [`sha-256`](https://github.com/multiformats/multihash) and base encoding
 [`base64url`](https://github.com/multiformats/multibase). Implementations MAY
@@ -42,7 +46,7 @@ connect to all other nodes.
 
 ## Connection Establishment
 
-### Browser to public Server
+### Browser to public Server (v1, SDP munging)
 
 Scenario: Browser _A_ wants to connect to server node _B_ where _B_ is publicly
 reachable but _B_ does not have a TLS certificate trusted by _A_.
@@ -70,11 +74,9 @@ reachable but _B_ does not have a TLS certificate trusted by _A_.
 
 4. _A_ constructs _B_'s SDP answer locally based on _B_'s multiaddr.
 
-   _A_ generates a random string prefixed with "libp2p+webrtc+v1/". The prefix
-   allows us to use the ufrag as an upgrade mechanism to role out a new version
-   of the libp2p WebRTC protocol on a live network. While a hack, this might be
-   very useful in the future. _A_ sets the string as the username (_ufrag_ or _username fragment_)
-   and password on the SDP of the remote's answer.
+   _A_ generates a random string prefixed with `libp2p+webrtc+v1/` and sets it
+   as both the username fragment (`ice-ufrag`) and password (`ice-pwd`) in the
+   synthetic remote answer SDP.
 
    _A_ MUST set the `a=max-message-size:16384` SDP attribute. See reasoning
    [multiplexing] for rational.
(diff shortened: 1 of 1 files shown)


dozyio:
Draft JS implementation https://github.com/libp2p/js-libp2p/pull/3480

webrtc/webrtc-direct.md
```diff
@@ -149,10 +152,51 @@ reachable but _B_ does not have a TLS certificate trusted by _A_.
...
+
+5. _B_ infers _A_'s SDP offer as follows:
+   1. `ice-ufrag = client_ufrag`
+   2. `ice-pwd = <payload after libp2p+webrtc+v2/>`
+   3. set an arbitrary sha-256 remote fingerprint
+   4. set `c=IN <ip> <port>` from the incoming STUN packet source
+   5. set `a=max-message-size:16384`
+
+6. _B_ sets this inferred offer as remote description, generates an answer, and
+   sets it as local description.
```
tabcat: B's answer `ice-ufrag`/`ice-pwd` aren't specified here, but must be `libp2p+webrtc+v2/<client_pwd>` (the password A set for B in the synthetic answer) — otherwise the connection fails.

webrtc/webrtc-direct.md
```diff
@@ -149,10 +152,51 @@ reachable but _B_ does not have a TLS certificate trusted by _A_.
...
+   _A_ constructs _B_'s synthetic remote answer with:
+   - `a=ice-ufrag:libp2p+webrtc+v2/<client_pwd>`
+   - `a=ice-pwd:libp2p+webrtc+v2/<client_pwd>`
+
+   _A_ MUST set `a=max-message-size:16384` and set the remote answer via
+   [`RTCPeerConnection.setRemoteDescription()`](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/setRemoteDescription).
+
+4. _A_ starts sending STUN connectivity checks. _B_ parses STUN `USERNAME` as
+   `server_ufrag:client_ufrag` and validates that `server_ufrag` has the prefix
+   `libp2p+webrtc+v2/`.
```
tabcat: v1 has no equivalent prefix check, so a v1 server's handling of a `v2/` ufrag is undefined. Each version should validate its prefix and drop on mismatch.

tabcat:
@MarcoPolo

johannamoran:
Hi @dozyio , 

In reference to WebRTC-Direct, sending for your FYI a reference link shared by @lidel 

- https://github.com/libp2p/go-libp2p/pull/3512 

Thanks, 

Johanna

webrtc/webrtc-direct.md
lidel: Addressed in 3587339. Step 6 now says that before generating its answer, B MUST set its own `ice-ufrag` and `ice-pwd` to `server_ufrag` (the `libp2p+webrtc+v2/<client_pwd>` value A embedded). A signs its checks with that value and expects B's STUN responses under it, so ICE can't complete otherwise, which you flagged.

webrtc/webrtc-direct.md
lidel: Addressed in 3587339. Version selection is now explicit: B reads the version from the `server_ufrag` prefix, where `libp2p+webrtc+v1/` selects v1 and `libp2p+webrtc+v2/` selects v2. The `libp2p+webrtc+` namespace is reserved for versions, so B MUST reject a `server_ufrag` whose version it doesn't recognize (or that carries no recognized prefix) rather than assume v1. A `v2/` ufrag reaching a server that doesn't speak v2 is then dropped instead of undefined.

lidel (approved):
Thanks for getting this rolling, @dozyio! The v2 design here mostly lgtm.

An air-tight transport spec is hard to write without real code and a second implementation to test against, so I wrote a GO server and used it to find the spots where the v2 text still left room for two implementations to diverge. I pushed 3587339, which fills those gaps so an implementer can build an interoperable v2 server and browser dialer from the spec alone.

Implementations:
- go-libp2p: https://github.com/libp2p/go-libp2p/pull/3520 (new)
- js-libp2p: https://github.com/libp2p/js-libp2p/pull/3480 (by @dozyio, i only tested + pushed small hardening commits in https://github.com/libp2p/js-libp2p/pull/3480/commits/3c12a80b4c8646ebeaa2492a5432f7772e7ae0af) 

Spec gaps filled as much as possible, so LLM-driven implementations avoid footguns:

- **Server's own ICE credentials**: B must set its `ice-ufrag`/`ice-pwd` to `server_ufrag` (the `libp2p+webrtc+v2/<client_pwd>` value A embedded), or A rejects B's STUN responses and ICE never completes. This is the one that bit first.
- **USERNAME parsing and validation**: split on the first `:`, validate both halves as ICE ufrags (`ice-char`, length bounds), and reject the rest, which keeps `CR`/`LF` out of the SDP B builds.
- **Client password**: recover it by stripping the prefix from `server_ufrag`, read the browser's own value back from `localDescription`, abort if it's missing, and keep `server_ufrag` within the ICE length limit.
- **Version dispatch**: recognize the v1 and v2 prefixes explicitly and reject an unknown version instead of assuming v1.
- **ICE and DTLS roles**: B is ICE Lite (controlled); the answer is `a=setup:passive`, so A is the DTLS client and B the DTLS server.
- **`c=` line**: it carries an address family and IP, not a port (`c=IN <IP4|IP6> <ip>`); fixed in v1 too.
- **RFC references**: `ice-char` grammar and lengths (RFC 8839 section 5.4), USERNAME format (RFC 8445 section 7.2.2), randomness (RFC 8445 section 5.3).
- **Chrome context**: linked the WebRTC-NoSdpMangleUfrag trial, the WebRTC tracker issue, the discuss-webrtc notice, and the getStats side channel that motivated v2.
- **Guidance**: new connections SHOULD use v2; v1 will stop working in browsers, but servers SHOULD keep accepting it during migration.
- **Versioning note**: new versions ride the ufrag prefix; a new multiaddr is only needed for a change the prefix can't carry.

I also tested it end to end: the JS client in a real Google Chrome (Chrome-for-Testing, with the `WebRTC-NoSdpMangleUfrag` field trial on, so munging is disabled) dialing the GO server over v2. v2 connects and completes the Noise handshake, and v1 fails under the trial exactly as expected. 

So.. I'm confident the spec at 3587339 is good to go, but second set of eyes + landing both GO and JS first would be good before merging.

seetadev:
Thank you so much @dozyio and @lidel  for driving the WebRTC-direct v2 spec forward and pushing this important update.

Also appreciate @tabcat for the early protocol review and for catching critical edge cases around version handling and ICE credential behavior. Those comments helped tighten the spec significantly.

CCing @sumanjeet0012, @yashksaini-coder and @acul71 on following up on the new additions shared by @lidel.

@yashksaini-coder: Would like you to go through each of the additions shared by @lidel at https://github.com/libp2p/specs/pull/715#pullrequestreview-4535931594. We will discuss this in fine details this week. Appreciate your initiative and efforts. @sumanjeet0012 and @acul71 will help you in the transport interop testing efforts with go and js. 

Special thanks to @lidel for sharing review, implementation testing across both Go and JS, and especially the work on closing the spec gaps. The additional clarifications around ICE credentials, version dispatch, DTLS roles, RFC alignment, and interoperability details make the spec much stronger and far more implementation-safe.

CCing @johannamoran and @flcl42 from dotnet-libp2p.

webrtc/webrtc-direct.md
sukunrt: Can we use a ufrag and password for the server that's derived from the certificate hash? Can A and B both independently set the ufrag and pwd for B to `libp2p+webrtc+v2/<function of server cert hash>`? I think both the setups are equivalent so there's not much benefit changing this.

seetadev (approved):
+1, LGTM. Thanks @lidel, @tabcat, @johannamoran for the thorough review and feedback.  

@dozyio recommended some changes during a recent JS-libp2p maintainers' call. Once those are addressed, this should be good to merge.

johannamoran:
Hi @dozyio - Let us know if this has the thumbs up to move ahead.

webrtc/webrtc-direct.md
lidel: Yes, better to keep it as is: the certhash is public, so a pwd derived from it fails the "MUST be unguessable" requirement of [RFC 8445, section 5.3](https://www.rfc-editor.org/rfc/rfc8445.html#section-5.3), while `client_pwd` is fresh random per connection.

tabcat (approved):

johannamoran:
With 3 approvals, this looks good to merge. We will keep improving the WebRTC Direct SPEC. We would encourage the other libp2p implementations to use this SPEC PR in relation to WebRTC Direct. We will update the maintainers in the upcoming monthly call.

PR #717, opened 2026-04-30
docs: pkix

dozyio:
Thinking we should probably have libp2p defined algorithms for interop, but also allow PKIX for experimentation

### DIFF
--- a/peer-ids/peer-ids.md
+++ b/peer-ids/peer-ids.md
@@ -28,6 +28,8 @@ about maturity level and spec status.
         - [Ed25519](#ed25519)
         - [Secp256k1](#secp256k1)
         - [ECDSA](#ecdsa)
+        - [PKIX Public Keys](#pkix-public-keys)
+        - [PKCS#8 Private Keys](#pkcs%238-private-keys)
 - [Peer Ids](#peer-ids)
     - [String representation](#string-representation)
         - [Encoding](#encoding)
@@ -63,6 +65,9 @@ enum KeyType {
 	Ed25519 = 1;
 	Secp256k1 = 2;
 	ECDSA = 3;
+	
+	PKIX = 0x40;
+	PKCS8 = 0x41;
 }
 
 message PublicKey {
@@ -110,11 +115,13 @@ The second is for generating peer ids; this is discussed in the section below.
 
 ### Key Types
 
-Four key types are supported:
+The following key types are supported:
  - RSA
  - Ed25519
  - Secp256k1
  - ECDSA
+ - PKIX Encoded Public Keys
+ - PKCS#8 Encoded Private Keys
 
 Implementations MUST support Ed25519. Implementations SHOULD support RSA if they wish to
 interoperate with the mainline IPFS DHT and the default IPFS bootstrap nodes. Implementations MAY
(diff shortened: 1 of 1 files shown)


dennis-tra:
Thanks @dozyio, I think this is a thoughtful PKIX direction.

Following up on the concern I raised over in #711 that with a PKIX key type the same RSA key could be encoded two ways, once under the existing RSA type and once under PKIX, and end up with two different PeerIDs. The "MUST use the existing libp2p-specific encoding" language addresses exactly that, so that one's resolved for me. Thanks!

There's a related thing I've been turning over that I want to bring up, even if it might be a bit orthogonal. With a generic container, PeerID determinism leans on the DER encoding coming out identical across implementations. DER is canonical for a given ASN.1 value, but the algorithm profiles underneath can still leave room. RFC 5480, for instance, permits both compressed and uncompressed EC points for the same key, so "follow the RFC" alone doesn't necessarily pin a single encoding. We've actually navigated this already because ECDSA and RSA has been hashing DER PKIX into PeerIDs for years without trouble. And for ML-DSA specifically, RFC 9881 leaves essentially no ambiguity, so I don't think we'd run into issues there. Where it could bite is the open-ended part of a generic container where some future algorithm whose SPKI profile isn't as tight has two conformant encoders producing different bytes for the same key.

This makes me wonder about the overall shape. We could go with the generic container and accept that every algorithm needs a profile pinning its encoding plus test vectors which is this PR's direction, and it pays off most I think if we expect to add several algorithms without touching the protobuf each time (but we'd still need to pin profiles and I would question that there will be extraordinarily many new keys we want to support). The permissionless extension angle is also weakened if we still need to pin profiles.

Alternatively we could give the algorithm we actually want to support at the moment (ML-DSA) its own key type with `Data` spec'd as the RFC 9881 SPKI. Since we need a profile and test vectors either way, a dedicated type doesn't cost much beyond an enum value.

There's also Marco's earlier idea of keeping PKIX out of the protobuf and deriving the PeerID from a separate `pkix-hash` multicodec. I also inherits the "future algorithm whose SPKI profile isn't as tight" concern but could be accompanied with a similar profile pinning as in this PR.

Anyways, just my thoughts above. Curious to hear others. Generally, I'm much more comfortable with the direction overall. Mostly I'd like us to choose deliberately between the generic container and a dedicated type. The IETF Standards reuse is the part I'd care most about, and I think all preserve that.

peer-ids/peer-ids.md
```diff
@@ -192,24 +194,88 @@ it using [DER-encoded ASN.1.](https://wiki.openssl.org/index.php/DER)
 
 #### PKIX Public Keys
 
-The PKIX key type only encodes public keys. The Data field is the [PKIX
-encoding](https://www.rfc-editor.org/rfc/rfc5280) of the public key. The public
-key and algorithm are identified by the [Subject Public Key
-Info](https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.7) field.
+The PKIX key type only encodes public keys. The Data field MUST contain the
```
richard-ramos, line 197:
```suggestion
The PKIX key type only encodes public keys. The `Data` field MUST contain the
```

peer-ids/peer-ids.md
```diff
@@ -192,24 +194,88 @@ it using [DER-encoded ASN.1.](https://wiki.openssl.org/index.php/DER)
...
+- accepted `AlgorithmIdentifier` OID(s)
+- `AlgorithmIdentifier` parameter encoding rules
+- public-key encoding and validation rules
+- signature algorithm and verification rules
+- libp2p protocol contexts where the key can authenticate a peer
+- Peer ID test vectors
+
+Implementations MUST reject unsupported algorithms carried in SPKI that they
+do not explicitly support, including algorithms whose profiles are unknown to
+the implementation.
```
richard-ramos, line 236:
```suggestion
Implementations MUST reject unsupported algorithms carried in SPKI  unless they explicitly support the corresponding profiles.
```

peer-ids/peer-ids.md
```diff
@@ -192,24 +194,88 @@ it using [DER-encoded ASN.1.](https://wiki.openssl.org/index.php/DER)
...
+
+Implementations MUST NOT accept key-agreement-only algorithms as identity keys
+unless explicitly specified by a libp2p specification.
+
+Implementations MAY support additional algorithms carried in SPKI for
+experimental, private-network, or application-specific use.
+
+Implementations MAY accept DER-encoded SPKI as an import/export format for
+public keys, but before Peer ID derivation they MUST canonicalize the key to
+the existing libp2p-specific public key encoding.
```
richard-ramos, line 249:
```suggestion
Implementations MAY accept DER-encoded SPKI as an import/export format for
public keys. If the algorithm has an existing libp2p-specific public key encoding, 
implementations MUST use that encoding before deriving a Peer ID.
```

peer-ids/peer-ids.md
```diff
@@ -192,24 +194,88 @@ it using [DER-encoded ASN.1.](https://wiki.openssl.org/index.php/DER)
...
+for Peer ID derivation.
 
-Signing is defined by the key algorithm used.
+Implementations MUST NOT derive alternate Peer IDs by re-encoding such public
+keys under the generic PKIX `KeyType` when a canonical libp2p-specific public
+key encoding exists.
 
-For backwards compatibility, if a key algorithm has a prior libp2p specific encoding, implementers SHOULD prefer that.
+If the PKCS#8 structure includes a public key and it does not match the public
+key derived from the private key, implementations MUST reject the key.
```
richard-ramos, line 278:
```suggestion
If the PKCS#8 structure includes a public key, implementations MUST reject 
the key if that public key differs from the one derived from the private key.
```

Issue #163, opened 2019-04-10
DHT provider records: adopt sloppiness and backtracking from Coral


raulk:
# Context

_Provider records_ in our Kad DHT implementation are equivalent to the concept of _pointers_ as proposed in the Coral DHT [0].

Unfortunately they are implemented half-way in go-libp2p, and probably in other languages too. That makes them scale poorly.

The problem is that nodes relatively close to a popular hash `H` get swamped with provider records. Assuming zero network churn and deterministic crawls, holders of content `H` end up advertising at the same KValue=20 nodes.

There is no concept of backpressure or spill over. No limits, throttling policies or validations are enforced. Peers swallow any provider record they receive. This ends up causing all kinds of local negative effects.

* https://github.com/ipfs/go-ipfs/issues/5613#issuecomment-471284697
* https://github.com/libp2p/go-libp2p-kad-dht/issues/316

# Proposal

Adopt these features of Coral:

1. rejection/throttling,
2. sloppy hashing / spill over,
3. backtracking.

**Rejection/throttling:** nodes restrict the amount of provider records stored per hash, and the rate of registration. The limit can be locally enforced and need not be synchronised across the network. Attacks by injecting uncollaborative nodes (with low or zero local limits) are nullified by spill-over.

**Sloppy hashing/spill over:** on a rejection, the requester backtracks along the traversal path and tries storing the record in earlier nodes. This has the effect of expanding the radius of advertisement of popular hashes, such that subsequent demands for that content will result in faster resolutions.

However, depending on how lookups are implemented, this could end up overloading new registrants, and weaning traffic off elder providers. I feel this is an oversight in Coral.

**Backtracking:** the process by which we walk back our traversal trail hitting earlier nodes along the way, attempting to register our record further from the desired target, thus expanding the radius where the hash is advertised.

Coral's approach is insecure, insofar a single adversary along the path rejecting your record will derail you and cause you to backtrack, even if closer peers to the key had spare capacity.

To resist such attacks, we can continue traversing forward a few more steps to confirm the rejection, or we can traverse disjoint paths.

Additionally it does not play well with TTLs and disappearing content. If a hash has spilled over several layers, and early registrants fail to reprovide (if they're offline or they pruned the hash), farther nodes holding recent registrations who happen to be over-capacity could preclude us from finding closer nodes that now have vacancies due to expired and unrenewed earlier records. This could result in progressive expansion of the radius, with a hollow centre.

---

See these instructional videos for reference:
[...]

PR #719, opened 2026-05-13
Kademlia provider record spillover

gmelodie:
## Summary

Adds `kad-dht/provider-record-spillover.md`, a working-draft specification for an opt-in extension to the Kademlia DHT `ADD_PROVIDER` flow that addresses provider record hotspots on popular keys.

## Problem

Under the base `kad-dht` protocol, the `k` closest peers to a key unconditionally accumulate every `ADD_PROVIDER` record for it.

For popular keys, this creates an unbounded load concentration with no mechanism for a peer to decline new provider records or for advertisers to spill over to other peers. This has been a known scalability issue, including:

- `libp2p/go-libp2p-kad-dht#316`
- `ipfs/kubo#5613`

This proposal implements the “rejection + sloppy hashing / spillover” approach originally proposed in `libp2p/specs#163`, drawing from the Coral DHT paper:

> Freedman & Mazières, 2003

### Server-side — provider record limits

A node MAY enforce `maxProvidersPerKey`.

Once at capacity, it MUST reject `ADD_PROVIDER` requests from new providers for that key, while always accepting re-advertisements from existing providers so records can continue to be refreshed.

### Wire protocol

A new optional `providerStatus` field (`field 11`) is added to the `ADD_PROVIDER` response message with two possible values:

- `accepted (0)`
- `rejected (1)`

For backward compatibility:

- An absent `providerStatus` field MUST be treated as `accepted`.
- `providerStatus` is response-only.
- Advertisers MUST NOT set it in requests.
- Receivers MUST ignore it if present in a request.

### Client-side — spillover algorithm

The advertiser performs the normal iterative `FIND_NODE` lookup, then processes the sorted candidate list in chunks of size `α`, from closest to farthest.

After each chunk, it counts peers that:

- accepted the provider record, or
- did not respond

toward the replication target `k`.

If `k` has not yet been reached, the advertiser continues to the next chunk (a spillover round).

The process stops when either:

- `k` successful placements are reached, or
- all candidates are exhausted.

Non-responding peers count as accepted to preserve compatibility with nodes that predate this extension.

## Compatibility

This extension is fully backward compatible.

Nodes that do not implement the extension:

- never write `providerStatus`
- silently ignore `providerStatus` if received

No changes are made to:

- `GET_PROVIDERS`
- `FIND_NODE`
- any other message type

## Related

- https://github.com/libp2p/specs/issues/163
- https://github.com/libp2p/go-libp2p-kad-dht/issues/316

### DIFF
--- a/kad-dht/provider-record-spillover.md
+++ b/kad-dht/provider-record-spillover.md
@@ -1,0 +1,241 @@
+# Provider Record Spillover
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+|-----------------|---------------|--------|-----------------|
+| 1A              | Working Draft | Active | r0, 2026-05-11  |
+
+Authors: [@gmelodie]
+
+Interest Group: [@mxinden, @guillaumemichel, @MarcoPolo]
+
+[@gmelodie]: https://github.com/gmelodie
+[@mxinden]: https://github.com/mxinden
+[@guillaumemichel]: https://github.com/guillaumemichel
+[@MarcoPolo]: https://github.com/MarcoPolo
+
+See the [lifecycle document][lifecycle-spec] for context about the maturity level
+and spec status.
+
+[lifecycle-spec]: https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md
+
+---
+
+## Overview
+
+This document specifies an optional extension to the [libp2p Kademlia DHT
+specification][kad-spec] that addresses provider record hotspots. When a key is
+popular, the `k` closest nodes concentrate all `ADD_PROVIDER` traffic for it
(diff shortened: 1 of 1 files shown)


guillaumemichel:
Hi @gmelodie, thank you for submitting this PR! I like the idea of provider record spillover and think it could indeed be very useful for the ecosystem.

However, since this spec is about Kad-DHT provider records and provider records are mostly used in IPFS-land, I'd recommend opening the PR against [ipfs/specs](https://github.com/ipfs/specs) instead. The canonical IPFS DHT spec has moved there and is now published at [specs.ipfs.tech](https://specs.ipfs.tech/routing/kad-dht/).

gmelodie:
Closing in favor of https://github.com/ipfs/specs/pull/537

PR #726, opened 2026-07-02
gossipsub: add SPREAD extension spec

MatheusFranco99:
## Summary

_Acknowledgement:_ This work was supported by a [grant](https://forum.ssv.network/t/ssv-x-university-of-lisbon-grant-proposal-blockchain-research-p2p/1939) from the **SSV Network DAO** awarded to the University of Lisbon.

This PR adds **SPREAD (Secure Peer-to-Peer Relay for Efficient Anonymous Dissemination)** as a GossipSub extension spec, under `pubsub/gossipsub/spread/`.

SPREAD extends GossipSub to raise the bar against **sender deanonymization** while, at the same time, **improving dissemination efficiency** in geo-distributed networks. It is motivated by the well-documented result that GossipSub leaks enough timing and structural information for a small set of passive listeners to map validators to their network identities, and by the fact that existing anonymity proposals (e.g. Dandelion++) have been considered infeasible for latency-sensitive settings such as Ethereum's consensus layer.

## Design

SPREAD combines two mechanisms. First, an **intra-cluster random walk** (a cobra walk) restricted to a node's nearby peers provides sender anonymity without a large performance penalty. Second, a **topology-aware inter-cluster dissemination** provides efficient global reach, using virtual coordinates (Vivaldi, secured by Newton checks) to avoid redundant long-distance hops. Clusters are formed locally — each node treats the closest `t%` of its neighbors as its cluster — so the protocol stays fully decentralized.

## Extension Choice

SPREAD is opt-in and backwards compatible. It is advertised through the existing GossipSub control/handshake fields and becomes active on a connection only when both peers support it. SPREAD messages carry a marker in the standard RPC envelope (no new message type); peers that do not support the extension ignore the marker and fall back to standard GossipSub. This enables mixed deployments and an incremental adoption path, with partial anonymity and performance gains possible before network-wide adoption.

## Evaluation highlights

SPREAD is evaluated by running the real `go-libp2p-pubsub` implementation over the [`simnet`](https://github.com/marcopolo/simnet) packet simulator with real-world Internet latencies, against GossipSub and Dandelion++ configured for equal expected per-node bandwidth:

- **Anonymity:** substantially reduces first-timestamp deanonymization accuracy relative to GossipSub, sitting between GossipSub and Dandelion++ on the anonymity spectrum.
- **Performance:** roughly 23% lower mean stretch than GossipSub and 67% lower than Dandelion++, with an even larger reduction in the tail (99th percentile).

## Reference implementation

A reference implementation is available as an opt-in extension in `go-libp2p-pubsub`: libp2p/go-libp2p-pubsub#717.

EthResearch post: https://ethresear.ch/t/spread-extending-gossipsub-with-efficient-anonymous-dissemination/25343


seetadev:
Thank you, @MatheusFranco99, for this excellent contribution. SPREAD tackles one of the key challenges in peer-to-peer networking, improving resistance to sender deanonymization while maintaining the low latency and efficiency required by production systems. 

The combination of intra-cluster cobra walks for anonymity and topology-aware inter-cluster dissemination is a thoughtful approach, and it is great to see it backed by both a detailed specification and a reference implementation. The opt-in, backwards-compatible design also makes incremental adoption practical, which is especially valuable for the libp2p ecosystem.

One question we would be interested in discussing is whether SPREAD can be viewed as two complementary improvements to GossipSub 1.1, 1.2 and 1.3: **(1) topology-aware clustering for dissemination efficiency** and **(2) cobra walks for publisher anonymity**. If so, how much of the dissemination improvement comes from the clustering mechanism alone, and what additional performance tradeoffs are incurred to achieve the stronger anonymity guarantees? 

Understanding that separation would be particularly valuable for initiatives like Lean Ethereum, Ethereum Consensus clients, where both efficient propagation and stronger privacy are important. 

This is indeed an exciting direction for GossipSub and a great example of research translating into practical networking improvements. Looking forward to the discussion and experimentation across the broader libp2p community.

CCing @dozyio, @tabcat, @acul71, @johannamoran, @sumanjeet0012 and @Winter-Soren on this thread. 
We discussed on your spec proposal today with friends at py-libp2p. Will discuss at tomorrow's js-libp2p meeting with @dozyio and @tabcat.

MatheusFranco99:
Hey @seetadev ! Thanks for your comment.

> One question we would be interested in discussing is whether SPREAD can be viewed as two complementary improvements to GossipSub 1.1, 1.2 and 1.3: (1) topology-aware clustering for dissemination efficiency and (2) cobra walks for publisher anonymity. If so, how much of the dissemination improvement comes from the clustering mechanism alone, and what additional performance tradeoffs are incurred to achieve the stronger anonymity guarantees?

Good question about the numbers, would definitely be interesting to see.
But, yes, it absolutely could be two used as two different improvements. One way I like to frame it is that SPREAD can be anywhere in the region between GossipSub and Dandelion regarding anonymity, while keeping performance better than both. But, as you said, we can fully optimize efficiency, but at the cost of reducing anonymity.

About the numbers, I’m running now the experiments for it and will report here as soon as I have it :)

MatheusFranco99:
Hey @seetadev, following up with some numbers:
- **Optimizing exclusively for dissemination**: our best configuration improves on every performance metric against GossipSub (numbers in the table below), while suffering on the anonymity side with 41% of attacker accuracy (for when attackers control 10% of the network) vs. 38% from GossipSub.

| Metric | mean latency | p95 latency | p99 latency | mean stretch | p95 stretch | p99 stretch |
| --- | --- | --- | --- | --- | --- | --- |
| Lowered by | 24% | 26% | 30% | 38% | 58% | 60% |

- **For the "same or better anonymity than GossipSub"**: a tunned configuration can produce lower attacker accuracy (30% vs. GossipSub's 38%) while still improving considerably all performance metrics.

| Metric | mean latency | p95 latency | p99 latency | mean stretch | p95 stretch | p99 stretch |
| --- | --- | --- | --- | --- | --- | --- |
| Lowered by | 22% | 24% | 25% | 32% | 51% | 49% |

- **Dandelion-level anonymity**: we can also configure SPREAD to match Dandelion's anonymity, beating it slightly on performance, though being considerably worse against the performance-optimized case. The pareto curve:

| Configuration | Acc@10% | Mean Latency | Mean stretch |
| --- | --- | --- | --- |
| Best performance found, but worse than GossipSub on anonymity | 41% | 98 ms | 1.54 |
| Still high-performance, while beating GossipSub anonymity | 30% | 101 ms | 1.69 |
| Mid-point | 28% | 104 ms | 1.74 |
| Close to Dandelion | 20% | 143 ms | 2.87 |
| Dandelion-level anonymity, but worst performance | 11% | 207 ms | 4.76 |

I ran these simulations overnight, but I plan to run these for a few more days to try to find even better configurations and increase the sample size.

seetadev:
> Hey @seetadev, following up with some numbers:

Thanks a lot, @MatheusFranco99. This is very much valuable, and I appreciate how quickly you followed up with concrete numbers. This is fantastic.

The results are quite encouraging. What stood out to me most is that the "same or better anonymity than GossipSub" configuration still improves latency and stretch across the board while reducing attacker accuracy. That is a very compelling point on the tradeoff curve and suggests there may be practical deployment configurations that provide meaningful privacy improvements without sacrificing network performance. 

We will definitely experiment with SPREAD in **py-libp2p** and **js-libp2p**. It will be interesting to compare behavior across implementations and see how the results translate beyond the current reference implementation.

A few questions came to mind while reading the latest results:

* Wish to ask if you have noticed any "sweet spot" where relatively small parameter changes produce large gains in anonymity with only minimal impact on latency?
* How stable are these results under higher churn? We are curious whether the topology-aware clustering remains effective as peers join and leave more frequently.
* Do we expect the dissemination improvements to come primarily from the clustering mechanism itself, or is there a meaningful interaction between clustering and the cobra walks that contributes to the overall performance?
* Longer term, do we envision SPREAD exposing a few recommended operating profiles (for example, performance-focused, balanced, and privacy-focused), or do we see applications tuning these parameters directly based on their requirements?

Impressive work overall. It is great to see a proposal that combines a thoughtful design, a reference implementation, and quantitative evaluation. Looking forward to seeing the larger experimental dataset, and we are excited to continue the discussion as we begin experimenting with it in the other libp2p implementations too.

CCing @johannamoran, @Winter-Soren and @sumanjeet0012.

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+| --------------- | ------------- | ------ | --------------- |
+| 1A              | Working Draft | Active | r0, 2026-02-13  |
+
+Authors: Diogo Cardoso, Matheus Franco, Rodrigo Rodrigues
+
+## Overview / Summary
+
+GossipSub is the key communication infrastructure underlying societal-critical protocols in ecosystems such as Ethereum. Several characteristics of gossip protocols make this an interesting communication substrate, namely robustness, scalability, and simplicity. However, the important property of anonymity (hiding the true source of a message), is one that is sometimes cited as being desirable to avoid targeted attacks against the sender, but is in practice well-known that it cannot be attained by GossipSub.
```
gmelodie, line 12:
```suggestion
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+SPREAD consists of two components: a local random walk to preserve anonymity, and a geographically directed propagation for efficient global dissemination, which gives preference to using nearby nodes as a stepping stone, to avoid costly long-distance hops. We are implementing SPREAD as a protocol extension on top of the go-libp2p-pubsub infrastructure. This document presents this proposal, detailing SPREAD's motivation, rationale, and protocol design.
```
gmelodie, line 16:
```suggestion
This document specifies SPREAD (Secure Peer-to-Peer Relay for Efficient Anonymous Dissemination), an extension to the GossipSub protocol. SPREAD raises the cost of sender deanonymization while improving dissemination efficiency in geo-distributed deployments. Both properties are relevant to systems such as Ethereum, where targeted attacks may prevent a node from meeting time-bound duties (e.g., fulfilling a duty within its assigned slot), and where low-latency delivery is required by protocols layered on top of the gossip substrate. SPREAD does not provide an absolute guarantee of source indistinguishability: sender anonymity is a quantitative property dependent on the adversary's observation capabilities, and the goal is therefore to raise the bar against deanonymization rather than to prevent it entirely.

SPREAD consists of two mechanisms:

1. a local random walk, which obscures the message origin prior to broad propagation; and
2. geographically directed propagation, which prefers nearby peers as relay hops in order to avoid costly long-distance transmissions during global dissemination.

SPREAD is implemented as a protocol extension on top of the go-libp2p-pubsub library. The remainder of this document details SPREAD's motivation, rationale, and protocol design.
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Terms and definitions
+
+Gossip Protocol - A distributed communication protocol in which nodes repeatedly forward information to a subset of neighboring peers, leading to probabilistic but eventually broadcast of a given message.
```
gmelodie, line 20:
```suggestion
Gossip Protocol - A distributed communication protocol in which nodes repeatedly forward information to a subset of neighboring peers, resulting in probabilistic, eventual delivery to all nodes.
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+Curious Nodes (Honest-but-Curious Observers) - Nodes that follow the protocol correctly but attempt to infer additional information (e.g., the originator of a given message) from observed traffic patterns.
+
+Fanout - The number of peers to which a node forwards a message during a dissemination step.
+
+Random Walk - A forwarding strategy in which each node forwards a message to a single randomly selected peer (possibly with probabilistic branching).
```
gmelodie, line 26: the propabilistic branching is already explained in "Cobra Walk"
```suggestion
Random Walk - A forwarding strategy in which each node forwards a message to a single randomly selected peer.
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+Virtual Coordinates - Latency-estimated coordinates assigned to nodes in a synthetic geometric space, allowing estimation of network distance without direct measurement.
+
+Bernoulli Trial - A probabilistic decision mechanism with two outcomes (success/failure), parameterized by the respective probability value.
```
gmelodie, line 30:
```suggestion
Bernoulli Trial - A random experiment with exactly two outcomes (success/failure), where success occurs with a fixed probability `p`.
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+Gossip Protocol - A distributed communication protocol in which nodes repeatedly forward information to a subset of neighboring peers, leading to probabilistic but eventually broadcast of a given message.
+
+Curious Nodes (Honest-but-Curious Observers) - Nodes that follow the protocol correctly but attempt to infer additional information (e.g., the originator of a given message) from observed traffic patterns.
+
+Fanout - The number of peers to which a node forwards a message during a dissemination step.
+
+Random Walk - A forwarding strategy in which each node forwards a message to a single randomly selected peer (possibly with probabilistic branching).
+
+Virtual Coordinates - Latency-estimated coordinates assigned to nodes in a synthetic geometric space, allowing estimation of network distance without direct measurement.
```
gmelodie, line 28:
```suggestion
Virtual Coordinates - Synthetic coordinates assigned to nodes in a geometric space such that distances between coordinates approximate pairwise network latency, allowing latency estimation without direct measurement.
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+Bernoulli Trial - A probabilistic decision mechanism with two outcomes (success/failure), parameterized by the respective probability value.
+
+Stretch - A performance metric defined as the ratio between the actual end-to-end delivery latency and the direct (usually optimal) communication latency between sender and receiver.
+
+Deanonymization Accuracy - The fraction of times an adversary correctly infers the original sender of a message, based on timing observations across attacker-controlled nodes.
```
gmelodie, line 34: this includes a specific attack into a definition, maybe make it more general
```suggestion
Deanonymization Accuracy - The fraction of times an adversary correctly infers the original sender of a message, based on observations available to attacker-controlled nodes.
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+SPREAD consists of two components: a local random walk to preserve anonymity, and a geographically directed propagation for efficient global dissemination, which gives preference to using nearby nodes as a stepping stone, to avoid costly long-distance hops. We are implementing SPREAD as a protocol extension on top of the go-libp2p-pubsub infrastructure. This document presents this proposal, detailing SPREAD's motivation, rationale, and protocol design.
+
+## Terms and definitions
```
gmelodie, line 18: I would reorder this section to be alphabetical and only break this ordering for codepentent fields such as random wak and cobra walk:
```
Bernoulli Trial
Curious Nodes (Honest-but-Curious Observers)
Deanonymization Accuracy
Fanout
Gossip Protocol
Random Walk
Cobra Walk (Coalescing-Branching Random Walk)
Stretch
Virtual Coordinates
Cluster
Voronoi Diagram (Dirichlet Tessellation)
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+To address this challenge, we propose SPREAD, a new gossip protocol to be implemented as a protocol extension, that raises the bar against sender deanonymization compared to GossipSub while actually providing a more efficient dissemination. Our approach rests on two principles: anonymity via a random walk that obfuscates the message origin, and low latency via topology-aware hop selection. We separate the random-walk hops, which prioritize low latency, from wide-area hops, which try to communicate with nearby nodes as a stepping stone, whenever it is possible to avoid costly long-distance paths. This design suits applications requiring both low latency and sender privacy, including blockchain validator messaging, anonymous communication systems, and censorship-resistant platforms.
```
gmelodie, line 52: already said in introduction
```suggestion
SPREAD is designed to address this compound challenge, as detailed in the following sections.
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
...pace. This allows each node to locally determine the subset of its peers that comprise the neighboring members of its own cluster. Additionally, it is possible to approximate the idealized inter-cluster routing in a fully decentralized way leveraging virtual coordinates. The idea, inspired by natural navigation, is to always avoid a direct jump to a more distant neighbor whenever a closer neighbor exists within the set of non-cluster nodes that are reasonably well-aligned with the distant node (which, intuitively, means there is a closer nearby destination that can serve as a stepping stone). In this case, well-aligned means that the angle bearing to the closer node is within a configurable angular interval of the bearing of the distant node (where this angle bearing is given by the virtual coordinates in the Euclidean space). This, again, allows each node to locally split its set of non-cluster overlay peers into two groups: those that have a closer “stepping stone” member of the set, and therefore should not be used for inter-cluster or any type of forwarding (denoted occluded_remote), and those that do not, and therefore are eligible for inter-cluster hops (unobstructed_remote).
```
gmelodie, line 69:
```suggestion
SPREAD, however, is fully decentralized: no single node holds a global view of the clusters or their connections. Instead, each node approximates the idealized routing locally using virtual coordinates, a scheme that securely assigns each node Euclidean coordinates whose distances estimate pairwise latency [Vivaldi, Newton]. With these coordinates, a node defines its cluster as the closest t% of its overlay neighbors, and classifies each remaining (non-cluster) neighbor by whether a closer neighbor lies in roughly the same direction, i.e., within a configurable angular interval of its bearing in the coordinate space. Neighbors with such a closer, aligned alternative are occluded: the closer node can serve as a stepping stone, so they are excluded from forwarding (occluded_remote). The rest are eligible for inter-cluster hops (unobstructed_remote).
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
...ropagates the same message multiple times. The reason why the algorithm does not include some logic to stop such duplicated behavior is to guarantee anonymity. In particular, for protocols in which nodes propagate a message only once, [Bellet et al.](https://drops.dagstuhl.de/entities/document/10.4230/LIPIcs.DISC.2020.8) have shown that the attacker would be more likely to identify the source by tracking communication timestamps. However, for real-world applications that continuously generate new messages, termination must be provided to avoid network congestion. A simple approach is to let the node propagate the same message (which can arrive via different neighbors as the message is routed) a fixed number of times, after which it stops. Usually, this parameter controls the reliability of the protocol, and small values are commonly sufficient for real-world network sizes, as described by [Kermarrec et al.](https://doi.org/10.1109/TPDS.2003.1189583). Still, extra reliability mechanisms can be added and in fact are already present in frameworks such as GossipSub, such as heartbeat messages that advertise a list of seen messages and “pull” protocol requests to fetch missing messages.
```
gmelodie, line 106:
```suggestion
Each protocol iteration performs two steps: intra-cluster and inter-cluster propagation (lines 12–13). Intra-cluster propagation follows the cobra walk (coalescing-branching random walk) algorithm ([Dutta et al.](https://doi.org/10.1145/2817830)): a random walk that branches according to a local Bernoulli trial with parameter ρintra (line 15). On failure, the node forwards the message to a single peer selected uniformly at random from cluster_i, continuing the random walk (line 16); on success, it forwards to fanoutintra such peers, accelerating dissemination within the cluster (line 18). Inter-cluster propagation drives global dissemination and is governed by a second Bernoulli trial with parameter ρinter (line 20). On failure, the node keeps all communication within its cluster; on success, it forwards the message to fanoutinter peers selected uniformly at random from unobstructed_remote_i (line 21).

A node may propagate the same message multiple times, and the algorithm deliberately includes no deduplication: [Bellet et al.](https://drops.dagstuhl.de/entities/document/10.4230/LIPIcs.DISC.2020.8) have shown that when nodes propagate a message only once, timestamp tracking makes the source easier to identify. In practice, however, termination is required to avoid congestion in systems that continuously generate messages. A simple mechanism suffices: a node propagates a given message (which may arrive repeatedly via different neighbors) a fixed number of times and then stops. This parameter effectively controls the reliability of the protocol, and small values suffice for realistic network sizes ([Kermarrec et al.](https://doi.org/10.1109/TPDS.2003.1189583)). Additional reliability mechanisms can complement this, and GossipSub already provides them: heartbeat messages advertising seen messages, and pull requests to fetch missing ones.
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
...x.org/conference/usenixsecurity25/presentation/heimbach); [Rhea](https://ethresear.ch/t/packetology-validator-privacy/7547)). This is possible despite GossipSub’s use of randomized forwarding to obscure message paths because a validator’s immediate peers (i.e., the validator's direct peers in the gossip overlay) consistently receive and propagate its messages noticeably earlier than others. As such, by deploying a few tens of listener nodes it becomes highly likely that, after a few epochs, one of the listener nodes will become a direct peer (and remain so for several subsequent epochs). Therefore, by keeping track of the first listener node to hear a message over multiple consensus epochs (and which network address that message came from), the coordinator is eventually able to reliably map validators to their network identities with high confidence. Once identified, validators can be selectively targeted (e.g., through denial-of-service), leading to slashing due to missed duties or economic attacks. Crucially, this deanonymization does not rely on any privileged access and can be mounted only by listening to traffic at a small number of well-behaved (honest but curious) observers.
```
gmelodie, line 46:
```suggestion
Despite the general understanding that gossip protocols can aid in protecting the anonymity of the original sender of any given message, it was not designed and is not able to provide such guarantees. In particular, a simple attack on the GossipSub layer is possible by observing message timing at a small set of listener nodes, enabling a centralized coordinator to correlate timings and infer the true source of gossip messages. This class of passive, timing-based attack was first demonstrated on Bitcoin, where transactions were linked to the IP addresses that originated them from a small number of supernodes ([Biryukov et al.](https://doi.org/10.1145/2660267.2660379), [Fanti and Viswanath](https://proceedings.neurips.cc/paper_files/paper/2017/file/6c3cf77d52820cd0fe646d38bc2145ca-Paper.pdf)), and it was later shown to map Ethereum validators to their peer IDs and IP addresses by monitoring attestation propagation over a few epochs (Sharma et al.; [Heimbach et al.](https://www.usenix.org/conference/usenixsecurity25/presentation/heimbach); [Rhea](https://ethresear.ch/t/packetology-validator-privacy/7547)). This is possible despite GossipSub’s use of randomized forwarding to obscure message paths because a validator’s immediate peers (i.e., the validator's direct peers in the gossip overlay) consistently receive and propagate its messages noticeably earlier than others. As such, by deploying a few tens of listener nodes it becomes highly likely that, after a few epochs, one of the listener nodes will become a direct peer (and remain so for several subsequent epochs). Therefore, by keeping track of the first listener node to hear a message over multiple consensus epochs (and which network address that message came from), the coordinator is eventually able to reliably map validators to their network identities with high confidence. Once identified, validators can be selectively targeted (e.g., through denial-of-service), leading to slashing due to missed duties or economic attacks. Crucially, this deanonymization does not rely on any privileged access and can be mounted only by listening to traffic at a small number of well-behaved (honest but curious) observers.
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+This problem is rooted at a critical tension in the design of gossip protocols: increasing fanout, and more generally improving dissemination speed, reduces anonymity, while limiting exposure through low fanout improves privacy at the cost of propagation speed. Unfortunately, GossipSub fails to strike a sensible balance: it leaks sufficient structural information to enable deanonymization, yet remains inefficient due to latency-insensitive paths that amplify delay and overhead.
```
gmelodie, line 48:
```suggestion
This vulnerability stems from a fundamental tension in the design of gossip protocols: increasing fanout, and more generally dissemination speed, reduces sender anonymity, whereas limiting exposure through low fanout improves privacy at the cost of propagation delay. GossipSub occupies an unfavorable point in this tradeoff: it leaks enough structural information to enable deanonymization, yet disseminates inefficiently, since its latency-oblivious forwarding paths amplify both delay and overhead.
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+Prior research has proposed defenses against this class of attack, most notably [Dandelion](https://doi.org/10.1145/3084459) and [Dandelion++](https://doi.org/10.1145/3224424), which provide stronger, formally analyzed sender-anonymity guarantees by beginning dissemination with a random-walk obfuscation phase. However, these guarantees come at a significant latency cost, which has proven to be a fundamental barrier to their adoption in latency-sensitive settings. In fact, the Ethereum community concluded that, "because of latency constraints [...] this proposal [Dandelion++] is infeasible for the Ethereum consensus layer (at least not for any strong anonymity guarantees)" ([EthResearch discussion](https://ethresear.ch/t/ethereum-consensus-layer-validator-anonymity-using-dandelion-and-rln-conclusion/12698)). This tension is bound to intensify: with the planned reduction of Ethereum's slot time from 12 to 6 seconds ([EIP-7782](https://eips.ethereum.org/EIPS/eip-7782)), the gossip layer faces even stricter performance requirements, resulting in a compound challenge where today's deployments must provide both the required performance and the desired sender-anonymity defenses.
```
gmelodie, line 50:
```suggestion
Prior work has proposed defenses against this class of attack, most notably [Dandelion](https://doi.org/10.1145/3084459) and [Dandelion++](https://doi.org/10.1145/3224424), which achieve formally analyzed sender-anonymity guarantees by prepending a random-walk obfuscation phase to dissemination. These guarantees, however, carry a latency cost that has prevented adoption in latency-sensitive settings: the Ethereum community concluded that latency constraints make Dandelion++ infeasible for the consensus layer, at least under strong anonymity guarantees ([EthResearch discussion](https://ethresear.ch/t/ethereum-consensus-layer-validator-anonymity-using-dandelion-and-rln-conclusion/12698)). This tension tends to intensify with the planned reduction of Ethereum's slot time from 12 to 6 seconds ([EIP-7782](https://eips.ethereum.org/EIPS/eip-7782)), imposing even stricter timing requirements on the gossip layer. Thus, deployments must provide the required performance and sender-anonymity defenses simultaneously.
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+
+## Design of SPREAD: insights and overview
+
+The recent formal work of [Guerraoui et al.](https://arxiv.org/abs/2308.02477) showed the need for gossip protocols designed for anonymity to include a strong random walk component, where each node will often forward a message to a single overlay neighbor, to make the identity of the source difficult to determine. This technique is employed, for example, by the [Dandelion family of protocols](https://arxiv.org/abs/1805.11060), which start with a random walk-based anonymity phase and then probabilistically switch to a dissemination phase with a high fanout.
```
gmelodie, line 57:
```suggestion
Formal work by [Guerraoui et al.](https://arxiv.org/abs/2308.02477) showed that anonymity-preserving gossip requires a strong random walk component, in which nodes often forward a message to a single overlay neighbor, so that the source remains difficult to identify. The [Dandelion family of protocols](https://arxiv.org/abs/1805.11060) applies this technique by starting with a random-walk phase and probabilistically switching to high-fanout dissemination.
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+However, this raises the problem that, during the random walk phase, where messages are sent to at most one peer per time step, there is a chance that an individual step can be unlucky and cross a slow or distant path. This is problematic since a single slow hop is sufficient to noticeably hurt the average end-to-end performance (as measured by the stretch, or the ratio between the overlay and a direct message connection). This problem is then amplified by the fact that blockchains such as Ethereum layer their multi-step protocols on top of the gossip substrate. Furthermore, the countermeasures to improve the performance, namely to anticipate the switch to the more efficient mode where messages are sent to multiple peers in parallel, are not only still vulnerable to a single long hop in the initial phase, but are also in direct tension with the intended anonymity guarantees.
```
gmelodie, line 59:
```suggestion
The random walk, however, is where the latency cost concentrates: each hop is chosen without regard to topology, so a single unlucky step across a slow or distant path is enough to degrade end-to-end stretch, a penalty that is further multiplied when systems such as Ethereum layer multi-step protocols on top of the gossip substrate. Switching to the high-fanout phase earlier does not help: it remains exposed to one long initial hop, and it directly weakens the intended anonymity guarantees.
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+To navigate this tradeoff effectively, we leverage the insight that the geographic distribution of nodes tends to naturally form clusters corresponding to the world’s most densely populated and economically developed regions (e.g., US East and West Coasts, Europe, Asia, etc.) This allows the design of SPREAD to organize nodes into clusters with the characteristic that the network latency between nodes in the same cluster is low, thus allowing for fast multi-hop dissemination inside a cluster. With this design decision in place, we can then leverage the intra-cluster communication to conduct random walks that allow for building privacy without a significant performance penalty, while occasionally turning to inter-cluster communication for global dissemination.
```
gmelodie, line 61:
```suggestion
SPREAD navigates this tradeoff by exploiting the fact that nodes naturally cluster in some regions (e.g., the US coasts, Europe, Asia). Since latency within such a cluster is low, random walks can be confined to it, providing anonymity at little performance cost, while occasional inter-cluster hops handle global dissemination.
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+A second challenge is that the inter-cluster message steps can also be very penalizing for the end-to-end performance if not managed carefully. For instance, we would like to avoid having to send a message from Europe to the East Coast of the US via an overlay hop through Asia or the Middle East. To avoid this, we need to devise an efficient wide area dissemination, since this step is so crucial for performance, and our protocols already achieve anonymity through the intra-cluster communication. To this end, SPREAD attempts that inter-cluster message hops are conducted with nodes in adjacent clusters, so that they can be used as a stepping stone to reach more distant ones. In an idealized routing scenario, there is a single global view of the clusters and how they connect to each other through a high level overlay, which corresponds to a Dirichlet tessellation, dividing the space into regions according to the set of centroids of the clusters. This allows the idealized routing to only send inter-cluster messages to overlay peers in neighboring clusters according to the Dirichlet tessellation.
```
gmelodie, line 63:
```suggestion
Inter-cluster hops pose a second challenge: chosen carelessly, they are just as penalizing, e.g., routing a message from Europe to the US East Coast via an overlay hop through Asia. Since anonymity is already provided by the intra-cluster walks, wide-area dissemination can be optimized purely for efficiency. SPREAD therefore directs inter-cluster hops toward adjacent clusters, using them as stepping stones to reach more distant ones. Ideally, with a global view, each cluster would be represented by its center, the space would be divided into regions according to which center is closest (a Voronoi diagram, as shown in Figure 1), and messages would cross cluster boundaries only between neighboring regions.
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
+
+# SPREAD: Extending GossipSub with Efficient Anonymous Dissemination
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+| --------------- | ------------- | ------ | --------------- |
+| 1A              | Working Draft | Active | r0, 2026-02-13  |
+
+Authors: Diogo Cardoso, Matheus Franco, Rodrigo Rodrigues
+
+## Overview / Summary
```
gmelodie, line 10:
```suggestion
## Overview
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+
+## Protocol overview
+
+The protocol has two components that run in parallel: an algorithm for maintaining a set of overlay peers (or neighbors) and their respective secure virtual coordinates, and the main protocol for sending and propagating gossip messages. The overlay neighbors of node i are automatically partitioned into three subsets: cluster_i,  occluded_remote_i, and unobstructed_remote_i, according to the criteria described before.
```
gmelodie, line 74:
```suggestion
SPREAD comprises two components running in parallel: a maintenance algorithm that manages a node's overlay peers (neighbors) and their secure virtual coordinates, and the dissemination protocol that publishes and propagates gossip messages. Using the criteria described above, the maintenance algorithm partitions the neighbors of node i into three subsets: cluster_i, occluded_remote_i, and unobstructed_remote_i.
```

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+With this overlay state in place, we can now define the protocol for broadcasting messages in a simple manner, based on the intuition of combining intra-cluster random walks for ensuring anonymity with inter-cluster efficient dissemination through the unobstructed remote neighbors. The decision to incur in each of these two alternatives is made upon receiving a message to be propagated, simply by flipping a coin that is biased according to a system parameter. The pseudocode of the protocol is explained next, focusing on how to broadcast a message once the cluster information and overlay neighbor formation is in place.
```
gmelodie, line 76:
```suggestion
Given this overlay state, the dissemination protocol combines intra-cluster random walks, which provide anonymity, with inter-cluster forwarding through unobstructed remote neighbors, which provides efficient global dissemination. Upon receiving or publishing a message, a node decides between these two behaviors by biased coin flips, with biases set by system parameters. The pseudocode below specifies how a message is broadcast once the overlay state is established.
```

gmelodie (commented):
There's some nitpicking, so feel free to ignore where you see fit, but I found a lot of the wording to be excessively long and convoluted. Treid to make it more to the point.

Also, consider using the following structure instead:

1. Overview
2. Motivation
3. Terms and definitions
4. Design rationale
5. Protocol specification (incl. extension negotiation, coordinate maintenance)
6. Security considerations
7. Evaluation

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+This message pull mechanism is also valuable for robustness in two additional scenarios. First, under churn, a succession of node failures or departures can prevent the direct propagation from being effective, and the heartbeat-and-pull mechanism lets nodes recover the messages they missed. Second, it helps defend against Byzantine nodes: while simple cryptography prevents such nodes from tampering with message contents, they can still deliberately delay or refuse to forward messages, endangering progress, which the heartbeats and pulls counter.
+
+## Using protocol extensions and coexistence with GossipSub peers
+
+We implement SPREAD as an extension of GossipSub rather than as a separate protocol. In this case, it's an optional feature that a node may choose to support, without requiring agreement from the rest of the network. Each node advertises the extensions it supports through dedicated control fields that are already exchanged when two peers connect, and SPREAD becomes active on a given connection only when both endpoints advertise support for it. Nodes that do not support the extension, or that pair with a peer that does not, simply continue to use standard GossipSub.
```
richard-ramos, line 112: The usage of this extension would require StrictNoSign, right? Should this be mentioned?

richard-ramos, line 112: If this is meant to be included as an extension to [gossipsub](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.3.md) , this PR should extend the protobuffer defined in https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/extensions/extensions.proto similar to how it was done with the partial messages extension.

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+To keep its footprint minimal, SPREAD triggers its dissemination without introducing a new message type. A published message uses the standard GossipSub envelope, with an additional field that marks it as a SPREAD message. When a SPREAD-capable node receives such a message, it applies the SPREAD propagation rules and copies the marker onto the messages it relays, so that SPREAD behavior is preserved along the path; a node that does not support the extension simply ignores the marker and forwards the message using standard GossipSub. This forward compatibility is what enables mixed deployments and a smooth, incremental adoption path: a SPREAD message benefits from its privacy-enhancing and efficient dissemination until it reaches a node that does not support the extension, which means that partial anonymity and performance gains are already possible before network-wide adoption.
```
richard-ramos, line 114: If it's not meant to use extensions, the additional field the protobuffer field used for SPREAD messages should be defined. (otherwise at the moment the spec is not clear enough on how should this field be added to the message)

However i think it makes sense to use a new gossipsub extension (with the message type specific for the extension) instead of a new field due to semantics (as well as the advantage of extension negotiation being a mechanism that already exists in implementations), I'd structure roughly like this

```
message ControlExtensions {
  optional bool spread = 123456; // number to be defined
}

message RPC {
  repeated Message publish = 2;
  optional SpreadExtension spread = 123456; // number to be defined
}

message SpreadExtension {
  repeated uint32 publish_indices = 1;
}
```
with `publish_indices` identifying to which entries in `RPC.publish` does SPREAD apply.

johannamoran:
@gmelodie thanks for your feedback. 

@MatheusFranco99 - Excited to see this moving forward with your review of this feedback.

MatheusFranco99:
Hey all, really appreciate the feedback! Will review and address all comments very soon :)

pubsub/gossipsub/spread/README.md
MatheusFranco99: Thanks! Agree it's a duplicate, I just kept the "separate random-walk hops from wide-area hops" idea since it's weaker in the overview section and it's key to the design thesis.

pubsub/gossipsub/spread/README.md
MatheusFranco99: Exactly, added a discussion on it

MatheusFranco99: Indeed. Added it
Thank you!

pubsub/gossipsub/spread/README.md
MatheusFranco99: Yes, as extension. Added a discussion for it + the `publish_indices` nice idea

MatheusFranco99:
@gmelodie @richard-ramos @johannamoran 

Thanks for all the feedback!
Aligned to your suggestions and fixes.
Pls let me know what you think :)

pubsub/gossipsub/spread/README.md
```diff
@@ -0,0 +1,413 @@
...
+occasionally branch out instead of always forwarding to a single peer.
+
+Stretch - A performance metric defined as the ratio between the actual end-to-end delivery latency
+and the direct (usually optimal) communication latency between sender and receiver.
+
+Virtual Coordinates - Synthetic coordinates assigned to nodes in a geometric space such that
+distances between coordinates approximate pairwise network latency, allowing latency estimation
+without direct measurement.
+
+Voronoi Diagram (Dirichlet Tessellation) - A partition of a space into regions according to a set of
```
gmelodie: Nit: maybe use an enumeration or item list for these definitions
nit2: make it bold like 
```suggestion
**Voronoi Diagram (Dirichlet Tessellation)**: A partition of a space into regions according to a set of
```

gmelodie (approved):

pubsub/gossipsub/spread/README.md
MatheusFranco99: done

richard-ramos (approved):

Issue #723, opened 2026-06-14
Hybrid post-quantum Noise handshake (X25519 + ML-KEM-768)


royzah:
### Why
`/noise` is classical X25519, so recorded sessions are open to harvest-now-decrypt-later. A hybrid KEM closes that, and stays secure if either primitive holds.

### Proposal
A new, separately negotiated handshake (not a change to `/noise`), keeping the spec's "one id, one fixed suite" design:

- id: `/noise-mlkem768-hfs/0.1.0` (provisional)
- suite: `Noise_XXhfs_25519+ML-KEM-768_ChaChaPoly_SHA256`. XX with [Hybrid Forward Secrecy](https://github.com/noiseprotocol/noise_wiki/wiki/Hybrid-Forward-Secrecy); ML-KEM-768 (FIPS 203) mixed into the chain alongside the X25519 DHs.

Identity payload and signature are unchanged. Authentication stays classical (it cannot be broken retroactively); only confidentiality is HNDL-sensitive.

Implementations advertise the hybrid id ahead of `/noise`, so a mixed fleet falls back with no flag day.

### Reference impl
Working in Rust: libp2p/rust-libp2p#6481 over mcginty/snow#210. Revives #2168.

### To pin for interop
1. Protocol id string.
2. Exact HFS token placement and KDF mixing order, with KATs, so go/js/rust agree byte for byte.
3. ML-KEM-768 only, or the 512/768/1024 family?

royzah:
Bumping this for some outside eyes before any stack ships bytes, since the wire format needs to be agreed cross-impl.

Two things I'd love a read on:

1. **hfs vs full PQNoise-hybrid.** Since libp2p auth lives in the identity-key signature over the Noise static key (not the static key itself), the handshake only needs PQ for forward secrecy, which `Noise_XXhfs_25519+ML-KEM-768` covers. A full hybrid (PQ static-key auth too) seems redundant here. Does that hold, or is there a case for PQ on the static key?

2. **What to pin for interop:** the exact protocol id, HFS token placement + KDF mixing order (with KATs so go/js/rust match byte for byte), and ML-KEM-768-only vs the 512/768/1024 family.

Rust reference impl is up (libp2p/rust-libp2p#6481). Would like go/js folks in early so we don't converge on different names.

cc @thomaseizinger @mxinden @MarcoPolo @achingbrain - thoughts?

royzah:
[...]

PR #716, opened 2026-04-28
noise-pq: add Noise_XXhfs_25519+MLKEM768_ChaChaPoly_SHA256 spec (Stage 1 Working Draft)

paschal533:
## Summary

This PR adds a new protocol spec for `Noise_XXhfs_25519+MLKEM768_ChaChaPoly_SHA256`, a post-quantum hybrid extension of the classical Noise XX handshake used in libp2p, under the protocol id `/noise-mlkem768-hfs/0.2.0`.

**Update (2026-09-17):** The profile was renamed from `Noise_XXhfs_25519+ML-KEM-768_ChaChaPoly_SHA256`. Noise (revision 34, §8.2) requires each algorithm name to consist solely of alphanumerics and `/`, so the hyphenated name was never valid; thanks to @royzah for spotting it. The rename is wire-incompatible: the name is hashed into the handshake hash, so old-name and new-name peers fail when the initiator decrypts Message B. The libp2p identifier therefore moved from `/noise-mlkem768-hfs/0.1.0` to `/noise-mlkem768-hfs/0.2.0`. Message sizes are unchanged. See §2.1. [#727](https://github.com/libp2p/specs/pull/727) by @royzah uses the same `MLKEM768` name with the identifier `0.1.0`.

**Update (2026-06-24, history):** The spec was revised from an earlier X-Wing draft to use raw ML-KEM-768 (FIPS 203). The Noise XXhfs pattern already provides classical security through three independent DH operations (`ee`, `es`, `se`), so using X-Wing (which bundles an extra X25519 inside the KEM) would be redundant, adding 64 bytes of wire overhead with no security benefit. Raw ML-KEM-768 gives the same hybrid security guarantee with a smaller footprint. The Rust implementation, [libp2p/rust-libp2p#6481](https://github.com/libp2p/rust-libp2p/pull/6481) by @royzah, also uses raw ML-KEM-768.

The spec is filed at Stage 1 (Working Draft) per the libp2p spec lifecycle.

- Specifies the `e1` and `ekem1` HFS token extensions from the Noise HFS draft, applied to the existing Noise XX pattern
- Uses raw ML-KEM-768 (NIST FIPS 203) as the KEM primitive in the `ekem1` slot
- Covers wire format (Msg A 1,216 B, Msg B 1,200 B, Msg C 64 B), token ordering requirements, cipher state split, ML-KEM implicit rejection semantics, and interoperability requirements
- Includes a test vector schema matching the reference implementations

## Motivation

[...]

PR #727, opened 2026-07-06
noise: hybrid post-quantum handshake (X25519 + ML-KEM-768)

royzah:
Concrete draft for #723, so we have something byte-level to react to instead of an open question.

Additive, separately negotiated handshake that mixes ML-KEM-768 into the `XX` pattern via Noise HFS. Advertised ahead of `/noise`, falls back with no flag day. Auth stays classical (identity key signs the static key), only secrecy is hybridized.

Pins the protocol id and suite, defers token/KDF order to Noise HFS. Still open (called out in the doc):

1. the protocol id string
2. test vectors so go/js/rust match byte-for-byte
3. ML-KEM-768 only vs the 512/768/1024 family

Reference impl: libp2p/rust-libp2p#6481 (off-by-default `mlkem-hfs`). Would love go/js eyes early. cc @jxs @thomaseizinger @mxinden @MarcoPolo @achingbrain

### DIFF
--- a/noise/pq-mlkem768-hfs.md
+++ b/noise/pq-mlkem768-hfs.md
@@ -1,0 +1,89 @@
+# Hybrid post-quantum Noise (X25519 + ML-KEM-768)
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+| --------------- | ------------- | ------ | --------------- |
+| 1A              | Working Draft | Active | r0, 2026-07-06  |
+
+Authors: [@royzah]. Interest Group: [@jxs], [@thomaseizinger], [@mxinden],
+[@MarcoPolo], [@achingbrain].
+
+## Why
+
+[`/noise`][noise] is X25519 only, so recorded sessions are open to
+harvest-now-decrypt-later. This adds a separate handshake that mixes an
+ML-KEM-768 KEM into the same `XX` pattern via Noise [HFS][hfs]. It stays secure
+if either primitive holds.
+
+Auth is unchanged and stays classical: in libp2p the identity key signs the
+static key, so identity cannot be broken after the fact. Only secrecy is at risk,
+so only the ephemeral key exchange is hybridized.
+
+## Protocol ID
+
+```text
+/noise-mlkem768-hfs/0.1.0
+```
+
+Advertised ahead of `/noise`. Two capable peers pick it; otherwise negotiation
+falls back to `/noise` with no extra round trip.
+
+## Suite
+
+Mixed into the handshake hash, so it must match byte-for-byte:
+
+```text
+Noise_XXhfs_25519+ML-KEM-768_ChaChaPoly_SHA256
+```
+
+## Handshake
+
+`XX` with the `hfs` modifier. Token placement and KDF order are exactly Noise
+HFS; this spec adds nothing to them.
+
+```text
+-> e, e1
+<- e, ee, ekem1, s, es
+-> s, se
+```
+
+- `e1`: initiator sends the ML-KEM-768 encap key (1184 B).
+- `ekem1`: responder encapsulates, returns the ciphertext (1088 B), mixes the
+  shared secret.
+
+The libp2p identity payload (signature over the static key) is carried and
+verified as in [`/noise`][noise], unchanged.
+
+```mermaid
+sequenceDiagram
+    participant I as Initiator
+    participant R as Responder
+    I->>R: e, e1
+    R->>I: e, ee, ekem1, s, es  (+ identity)
+    I->>R: s, se  (+ identity)
+```
+
+## To pin for interop
+
+1. The protocol ID string.
+2. Test vectors: fixed keys and KEM randomness with per-step handshake hash and
+   transport keys, so go/js/rust match byte-for-byte.
+3. ML-KEM-768 only, or the 512/768/1024 family with 768 as the baseline?
+
+## Reference
+
+- rust-libp2p: [#6481][pr], off-by-default `mlkem-hfs` feature; classical path
+  untouched. Revives [#2168][old].
+- KEM: ML-KEM-768 in `snow` ([mcginty/snow#210][snow]). Until it ships in a
+  release, the impl pins `snow` via `[patch]`; KATs land with it.
+
+[noise]: https://github.com/libp2p/specs/blob/master/noise/README.md
+[hfs]: https://github.com/noiseprotocol/noise_wiki/wiki/Hybrid-Forward-Secrecy
+[pr]: https://github.com/libp2p/rust-libp2p/pull/6481
+[old]: https://github.com/libp2p/rust-libp2p/pull/2168
+[snow]: https://github.com/mcginty/snow/pull/210
+[@royzah]: https://github.com/royzah
+[@jxs]: https://github.com/jxs
+[@thomaseizinger]: https://github.com/thomaseizinger
+[@mxinden]: https://github.com/mxinden
+[@MarcoPolo]: https://github.com/MarcoPolo
+[@achingbrain]: https://github.com/achingbrain
--- a/noise/pq-mlkem768-hfs.md
+++ b/noise/pq-mlkem768-hfs.md
@@ -10,13 +10,13 @@ Authors: [@royzah]. Interest Group: [@jxs], [@thomaseizinger], [@mxinden],
 ## Why
 
 [`/noise`][noise] is X25519 only, so recorded sessions are open to
-harvest-now-decrypt-later. This adds a separate handshake that mixes an
-ML-KEM-768 KEM into the same `XX` pattern via Noise [HFS][hfs]. It stays secure
-if either primitive holds.
+harvest-now-decrypt-later. This specification adds a separate handshake that
+mixes an ML-KEM-768 KEM into the same `XX` pattern via Noise [HFS][hfs]. It
+stays secure if either primitive holds.
 
-Auth is unchanged and stays classical: in libp2p the identity key signs the
-static key, so identity cannot be broken after the fact. Only secrecy is at risk,
-so only the ephemeral key exchange is hybridized.
+Authentication is unchanged and stays classical: in libp2p the identity key
+signs the static key, so identity cannot be broken after the fact. Only secrecy
+is at risk, so only the ephemeral key exchange is hybridized.
 
 ## Protocol ID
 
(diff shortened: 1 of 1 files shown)


noise/pq-mlkem768-hfs.md
```diff
@@ -0,0 +1,89 @@
...
+
+Advertised ahead of `/noise`. Two capable peers pick it; otherwise negotiation
+falls back to `/noise` with no extra round trip.
+
+## Suite
+
+Mixed into the handshake hash, so it must match byte-for-byte:
+
+```text
+Noise_XXhfs_25519+ML-KEM-768_ChaChaPoly_SHA256
```
richard-ramos: The `-` must be dropped according to https://noiseprotocol.org/noise.html#cryptographic-algorithm-name-sections

noise/pq-mlkem768-hfs.md
```diff
@@ -0,0 +1,89 @@
...
+so only the ephemeral key exchange is hybridized.
+
+## Protocol ID
+
+```text
+/noise-mlkem768-hfs/0.1.0
+```
+
+Advertised ahead of `/noise`. Two capable peers pick it; otherwise negotiation
+falls back to `/noise` with no extra round trip.
```
richard-ramos: It seems to me that there's still a roundtrip if the receiver does not support `mlkem768`:

1. Emitter sends `/noise-mlkem768-hfs/0.1.0`
2. Receiver replies `na`
3. Emitter sends `/noise`
4. Receiver replies back with `/noise`
5. Handshake starts

richard-ramos: Idea for lines 27-28:

```
An initiator supporting both protocols SHOULD propose `/noise-mlkem768-hfs/0.1.0` before /noise. A responder supporting this protocol MUST accept that protocol ID. If the responder returns `na`, the initiator MAY subsequently propose /noise.
```

noise/pq-mlkem768-hfs.md
```diff
@@ -0,0 +1,89 @@
...
+sequenceDiagram
+    participant I as Initiator
+    participant R as Responder
+    I->>R: e, e1
+    R->>I: e, ee, ekem1, s, es  (+ identity)
+    I->>R: s, se  (+ identity)
+```
+
+## To pin for interop
+
```
richard-ramos: I'd rename this to `Open issues` or similar.

noise/pq-mlkem768-hfs.md
```diff
@@ -0,0 +1,89 @@
...
+`XX` with the `hfs` modifier. Token placement and KDF order are exactly Noise
+HFS; this spec adds nothing to them.
+
+```text
+-> e, e1
+<- e, ee, ekem1, s, es
+-> s, se
+```
+
+- `e1`: initiator sends the ML-KEM-768 encap key (1184 B).
```
richard-ramos:
```suggestion
- `e1`: initiator sends the ML-KEM-768 encapsulation key (1184 B).
```

noise/pq-mlkem768-hfs.md
```diff
@@ -0,0 +1,89 @@
...
+| --------------- | ------------- | ------ | --------------- |
+| 1A              | Working Draft | Active | r0, 2026-07-06  |
+
+Authors: [@royzah]. Interest Group: [@jxs], [@thomaseizinger], [@mxinden],
+[@MarcoPolo], [@achingbrain].
+
+## Why
+
+[`/noise`][noise] is X25519 only, so recorded sessions are open to
+harvest-now-decrypt-later. This adds a separate handshake that mixes an
```
richard-ramos:
```suggestion
harvest-now-decrypt-later. This specification adds a separate handshake that mixes an
```

noise/pq-mlkem768-hfs.md
```diff
@@ -0,0 +1,89 @@
...
+ML-KEM-768 KEM into the same `XX` pattern via Noise [HFS][hfs]. It stays secure
+if either primitive holds.
+
+Auth is unchanged and stays classical: in libp2p the identity key signs the
```
richard-ramos:
```suggestion
Authentication is unchanged and stays classical: in libp2p the identity key signs the
```

noise/pq-mlkem768-hfs.md
```diff
@@ -0,0 +1,89 @@
...
+
+Mixed into the handshake hash, so it must match byte-for-byte:
+
+```text
+Noise_XXhfs_25519+ML-KEM-768_ChaChaPoly_SHA256
+```
+
+## Handshake
+
+`XX` with the `hfs` modifier. Token placement and KDF order are exactly Noise
```
richard-ramos:
```suggestion
`XX` with the `hfs` modifier. Token placement and KDF order are exactly as specified by Noise
```

noise/pq-mlkem768-hfs.md
```diff
@@ -0,0 +1,89 @@
...
+2. Test vectors: fixed keys and KEM randomness with per-step handshake hash and
+   transport keys, so go/js/rust match byte-for-byte.
+3. ML-KEM-768 only, or the 512/768/1024 family with 768 as the baseline?
+
+## Reference
+
+- rust-libp2p: [#6481][pr], off-by-default `mlkem-hfs` feature; classical path
+  untouched. Revives [#2168][old].
+- KEM: ML-KEM-768 in `snow` ([mcginty/snow#210][snow]). Until it ships in a
+  release, the impl pins `snow` via `[patch]`; KATs land with it.
```
richard-ramos:
```suggestion
  release, the implementation pins `snow` via `[patch]`; KATs land with it.
```

noise/pq-mlkem768-hfs.md
```diff
@@ -0,0 +1,89 @@
...
+verified as in [`/noise`][noise], unchanged.
+
+```mermaid
+sequenceDiagram
+    participant I as Initiator
+    participant R as Responder
+    I->>R: e, e1
+    R->>I: e, ee, ekem1, s, es  (+ identity)
+    I->>R: s, se  (+ identity)
+```
```
richard-ramos: Is this diagram needed or would the  pattern described in line 43 be enough?

noise/pq-mlkem768-hfs.md
royzah: Right, 8.2 allows alphanumerics and `/` only. Suite is now `Noise_XXhfs_25519+MLKEM768_ChaChaPoly_SHA256`, and both implementations follow:

- [mcginty/snow#210](https://github.com/mcginty/snow/pull/210): [858dc27](https://github.com/royzah/snow/commit/858dc27) renames the parsed name and `Kem::name()`.
- [libp2p/rust-libp2p#6481](https://github.com/libp2p/rust-libp2p/pull/6481): [1ae21ce](https://github.com/royzah/rust-libp2p/commit/1ae21ce) renames `PARAMS_XX_HFS`.

Handshake tests pass on both. Thanks for catching it.

noise/pq-mlkem768-hfs.md
royzah: Correct, the `na` costs a round trip. Took your wording verbatim, so the claim is gone.

noise/pq-mlkem768-hfs.md
royzah: Renamed to "Open issues".

noise/pq-mlkem768-hfs.md
royzah: Applied.

noise/pq-mlkem768-hfs.md
royzah: Applied.

noise/pq-mlkem768-hfs.md
royzah: Applied.

noise/pq-mlkem768-hfs.md
royzah: Applied.

noise/pq-mlkem768-hfs.md
royzah: Applied.

noise/pq-mlkem768-hfs.md
royzah: Agreed, the pattern block says it. Diagram removed.

paschal533:
I'm closing libp2p/specs#716 in favour of this one, so there's a single spec for the suite.

Some of what #716 carried maps onto your open issues, and it's yours to take or leave:

- **Test vectors (your open issue 2).** Five deterministic vectors with fixed keys and seeded KEM randomness, carrying all three handshake messages, the final handshake hash and both `split()` cipher keys. They're reproduced byte for byte by two independent implementations, TypeScript and Python, from separate generators: [fixture](https://github.com/paschal533/pq-noise-artifacts/blob/main/vectors/pqc-test-vectors.json). I haven't replayed them against snow, because its ML-KEM key generation doesn't take an injected RNG.
- **Interop evidence.** Every ordered listener/dialer pairing of TypeScript, Python, Nim and rust-libp2p#6481, each against itself included, three runs each: 48 of 48 passed, one encrypted message each way so both cipher states are exercised, with a negative control that fails the same matrix when one side uses the old suite name. [Results and logs](https://github.com/paschal533/pq-noise-artifacts/tree/main/interop/results/20260917T213457Z).
- **Wire sizes.** Messages are 1,216, 1,200 and 64 bytes with an empty payload, and message B measured 1,306 bytes with an Ed25519 identity payload, so nothing crosses a 1,500 byte MTU.
- **Security considerations.** An on-path attacker can suppress the hybrid protocol id during multistream negotiation and leave both peers on `/noise`, since negotiation is unauthenticated. #716 described that and a strict mode that refuses the classical fallback.

Say which of these you want and I'll open a PR against your branch, or just lift the text.

One factual note, not an argument: the four implementations above currently ship `/noise-mlkem768-hfs/0.2.0`, because renaming the suite changed the handshake hash. I'll follow whatever this spec settles on for your open issue 1.

paschal533:
One thing I would like to propose for the draft, with a demonstration behind it rather than an argument.

**The downgrade is real and silent, and I can now show it rather than assert it.** Two js-libp2p nodes, both configured with the hybrid encrypter first and `/noise` second, negotiate over plaintext multistream-select. An on-path proxy that corrupts the hybrid protocol id makes the genuine listener answer `na`; both peers then complete a mutually authenticated classical `/noise` session and neither signals anything. Runnable harness and transcripts: [experiments/downgrade-demo](https://github.com/paschal533/pq-noise-artifacts/tree/main/experiments/downgrade-demo).

Nothing detects it afterwards, which is the part worth knowing. I checked four channels: `connection.encryption` is recorded locally on each side and never transmitted, so there is nothing to compare; the identify protobuf has no field for the connection encrypter, only the application-stream protocol set; the peer store records neither the encrypter nor the Noise static key; and the two encrypters generate separate static keys, but nothing ever compares them across connections.

**Suggested addition to the draft, in security considerations:** an implementation SHOULD provide a mode in which `/noise` is neither offered nor accepted, so that an operator who needs the post-quantum property can choose connection failure over silent downgrade. That mode is the only mitigation available today. It is worth stating normatively because it is a deployment choice the spec can make visible, and because implementations differ right now: it is reachable by configuration in the TypeScript, Python and Nim implementations, while `protocol_info()` in the Rust implementation returns `/noise` alongside the hybrid identifier unconditionally, so a Rust operator cannot opt out at all. That looks like a small addition if you think it is worth making.

Two things I am deliberately not proposing here. Binding the negotiation transcript into the Noise prologue does not work, because a successful downgrade runs classical `/noise` and never reaches the hybrid handshake code; I proposed that earlier and have withdrawn it. And the actual fix, having the identity signature cover the handshake hash plus the offered protocol set on the classical path too, is a change to `/noise` rather than to this suite, so it belongs in its own proposal rather than holding this one up.

Happy to write the security-considerations text if you want it.

royzah:
You're right, I checked. `protocol_info()` hands back `[hybrid, /noise]` and there's no way to drop the second one. Off gives classical only, on gives both, hybrid-only doesn't exist, and it's a build flag rather than something you pick at deploy. So a Rust operator genuinely can't opt out.

I'll put a runtime knob on `Config` in https://github.com/libp2p/rust-libp2p/pull/6481

Security considerations text: yes please, write it. The SHOULD reads right, and https://github.com/paschal533/pq-noise-artifacts/tree/main/experiments/downgrade-demo is why. Agree on dropping the prologue binding too, a downgrade never reaches that code.

On the id, I'm on 0.1.0 and you're on 0.2.0. I'll follow whatever the spec lands on.

PR #729, opened 2026-07-08
gossipsub: Topic Streams Extension

MarcoPolo:
Gossipsub v1.3 uses a single stream per direction for all RPCs. This introduces some problems: unnecessary head of line blocking between messages (especially problematic when a large message in one topic delays small messages in another topic) and topic name overhead on each message.

The Topic Streams Extension addresses these problems. It moves topic scoped application messages to separate long-lived streams.

### DIFF
--- a/pubsub/gossipsub/extensions/extensions.proto
+++ b/pubsub/gossipsub/extensions/extensions.proto
@@ -8,6 +8,7 @@ message ControlExtensions {
 
   optional bool testExtension = 6492434;
 
+  optional bool topicStreams = 6492435; // See topic-streams.md 
 }
 
 message ControlMessage {
@@ -58,4 +59,28 @@ message PartialMessagesExtension {
 
   // An encoded representation of the parts a peer has and wants.
   optional bytes partsMetadata = 4;
+}
+
+message TopicRPCHeader {
+  optional string topic = 1;
+}
+
+// TopicScopedMessage is identical to Message, except the topic field is unused
+message TopicScopedMessage {
+	optional string from = 1;
+	optional bytes data = 2;
+	optional bytes seqno = 3;
+
+  // Included for computing signatures, not used on the wire
+  optional string unset_topic_name = 4;
+
+	optional bytes signature = 5;
+	optional bytes key = 6;
+}
+
+message TopicRPC {
+  oneof payload {
+    TopicScopedMessage publish = 1;
+    PartialMessagesExtension partial = 2;
+  }
 }
--- a/pubsub/gossipsub/extensions/extensions.proto
+++ b/pubsub/gossipsub/extensions/extensions.proto
@@ -71,8 +71,12 @@ message TopicScopedMessage {
 	optional bytes data = 2;
 	optional bytes seqno = 3;
 
-  // Included for computing signatures, not used on the wire
-  optional string unset_topic_name = 4;
+  // Field 4 is `topic` on the canonical pubsub Message (see pubsub/README.md).
+  // It is reserved on the wire because the topic comes from TopicRPCHeader.
+  // Implementations MAY still populate field 4 locally when computing
+  // signatures so they can hash a protobuf object without copying the whole
+  // message. This is an implementation choice; field 4 is not sent on the wire.
+  reserved 4;
 
 	optional bytes signature = 5;
 	optional bytes key = 6;
--- a/pubsub/gossipsub/topic-streams.md
+++ b/pubsub/gossipsub/topic-streams.md
@@ -48,7 +48,7 @@ Violation](#aborting-the-connection-on-protocol-violation).
 
 ## Aborting the Connection on Protocol Violation
 
-If a receiver receives a message from a peer that violates a MUST condition,
+If a receiver receives a message that violates a MUST condition,
 the receiver MUST reset the connection to the peer and send error code
 `0xd52505` when the transport allows it. This code signifies a Topic Streams
 Protocol Violation error, and is derived from the first 3 bytes of the sha256
@@ -57,7 +57,7 @@ hashsum of the string `gossipsub-topic-streams`. (i.e. `echo -n
 
 ## Topic Streams
 
-A peer opens a bidirectional stream for each topic that it wishes to send
+A peer MUST open a bidirectional stream for each topic that it wishes to send
 application messages for. Despite being bidirectional streams, they are treated
 as unidirectional streams. If both sides wish to publish messages for a given
 topic, both sides MUST open a bidirectional stream.
@@ -93,12 +93,12 @@ applications.
 `TopicRPC` messages MUST NOT be empty. They MUST contain either a partial or
 publish message. The data length of the application message MUST be non zero.
 
-If there are multiple streams for a single topic, the receiver SHOULD process
-them in the order the streams were opened by the initiator. The receiver
-SHOULD limit the number of concurrent topic streams for the same topic to 3 by
-resetting future streams and downscore peers that open more. Initiators SHOULD
-limit the number of concurrent topic streams to 1 per topic. The initiator MUST
-close the old stream before writing on a new stream for a given topic.
+The initiator MUST close the existing topic stream for a topic before it opens
+a new one. Because the transport can reorder packets, the receiver can get a
+new stream before the close of the old one. This is not a violation. The
+receiver MUST accept up to 3 open streams for a topic and SHOULD process them
+in the sequence opened. If there are more than 3, the receiver SHOULD reset
+the excess streams and downscore the peer.
 
 If the receiver receives a topic stream for a topic it is not subscribed to and
 has not recently published partial messages to (via fanout), it SHOULD
--- a/pubsub/gossipsub/topic-streams.md
+++ b/pubsub/gossipsub/topic-streams.md
@@ -117,8 +117,8 @@ message to the application.
 ### Partial Messages
 
 If the partial message extension has been negotiated with this extension, peers
-MUST send each other Partial Messages on the topic stream, not the control
-stream.
+MUST send each other Partial Messages on the topic stream.
+Peers MUST NOT send Partial Messages on the control stream.
 
 `PartialMessagesExtension` protobuf messages MUST omit the `topicID` field when
 sent over the wire. Implementations MUST set the `topicID` field after
--- a/pubsub/gossipsub/topic-streams.md
+++ b/pubsub/gossipsub/topic-streams.md
@@ -76,7 +76,7 @@ SHOULD reset the stream. One second is a reasonable timeout for most
 applications.
 
 `TopicRPC` messages MUST NOT be empty. They MUST contain either a partial or
-publish message.
+publish message. The data length of the application message MUST be non zero.
 
 If there are multiple streams for a single topic, the receiver SHOULD process
 them in the order the streams were opened by the initiator. The receiver
--- a/pubsub/gossipsub/topic-streams.md
+++ b/pubsub/gossipsub/topic-streams.md
@@ -70,9 +70,15 @@ The protocol id for a topic stream is `/gsts/v0beta`.
 ### Control Stream
 
 When this extension is negotiated, the original gossipsub stream becomes the
-control stream. Application messages (such as the `Message` or
-`PartialMessagesExtension` messages) MUST NOT be published on the control
-stream.
+control stream. Subscriptions, `IHAVE`, `IWANT`, `GRAFT`, `PRUNE`, `IDONTWANT`,
+`ControlExtensions`, and any other non-application RPC or control traffic MUST
+remain on this stream.
+
+A stream opened with protocol id `/gsts/v0beta` and identified by a
+`TopicRPCHeader` is a topic stream. Application publish payloads (`Message`,
+represented on the wire by `TopicScopedMessage`, and
+`PartialMessagesExtension`) MUST be sent on a topic stream and MUST NOT be
+published on the control stream.
 
 ### Topic Stream Header
 
@@ -111,8 +117,16 @@ peers and use a short-lived stream when responding to `IWANTs`.
 
 ### Topic Scoped Messages
 
-When a peer wishes to publish a message, it MUST publish a `TopicScopedMessage`
-and it MUST NOT publish a message on the control stream.
+A stream opened with protocol id `/gsts/v0beta` and identified by a
+`TopicRPCHeader` is a topic stream. The original gossipsub stream is the
+[control stream](#control-stream) after negotiation.
+
+When a peer wishes to publish an application `Message`, it MUST send it as a
+`TopicScopedMessage` on the topic stream and MUST NOT publish it on the control
+stream. `PartialMessagesExtension` application payloads likewise MUST be sent
+on the topic stream and MUST NOT be published on the control stream. All
+subscriptions and other non-application RPC or control traffic MUST remain on
+the control stream.
 
 Implementations MUST NOT set the topic name when sending the message over the
 wire.
--- a/pubsub/gossipsub/topic-streams.md
+++ b/pubsub/gossipsub/topic-streams.md
@@ -80,10 +80,10 @@ publish message. The data length of the application message MUST be non zero.
 
 If there are multiple streams for a single topic, the receiver SHOULD process
 them in the order the streams were opened by the initiator. The receiver
-SHOULD limit the number of concurrent topic streams for the same topic to 3 and
-downscore peers that open more. Initiators SHOULD limit the number of
-concurrent topic streams to 1 per topic. The initiator MUST close the old
-stream before writing on a new stream for a given topic.
+SHOULD limit the number of concurrent topic streams for the same topic to 3 by
+resetting future streams and downscore peers that open more. Initiators SHOULD
+limit the number of concurrent topic streams to 1 per topic. The initiator MUST
+close the old stream before writing on a new stream for a given topic.
 
 If the receiver receives a topic stream for a topic it is not subscribed to and
 has not recently published partial messages to (via fanout), it SHOULD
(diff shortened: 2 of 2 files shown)


pubsub/gossipsub/topic-streams.md
```diff
@@ -0,0 +1,128 @@
...
+Upon opening the topic stream, the initiator MUST send a single length-prefixed
+`TopicRPCHeader` protobuf message. The `TopicRPCHeader` MUST contain a topic. The receiver uses this to identify the topic associated with the stream.
+Afterwards, only length-prefixed `TopicRPC` protobuf messages are allowed.
+
+If the receiver does not receive a topic header within a reasonable timeout, it
+SHOULD reset the stream. One second is a reasonable timeout for most
+applications.
+
+`TopicRPC` messages MUST NOT be empty. They MUST contain either a partial or
+publish message.
```
sukunrt: I think we should specify non 0 data length

pubsub/gossipsub/topic-streams.md
```diff
@@ -0,0 +1,128 @@
...
+protocol negotiation completes.
+
+The protocol id for a topic stream is `/gsts/v0beta`.
+
+### Control Stream
+
+When this extension is negotiated, the original gossipsub stream becomes the
+control stream. Application messages (such as the `Message` or
+`PartialMessagesExtension` messages) MUST NOT be published on the control
+stream.
```
sukunrt: What do you think about a design where we have per topic control streams too? 

So the control stream would share the topic relevant: IHave, IWant, IDontWant, and maybe Graft and  Prune

This allows: 
Easier implementation. Need to read a control scoped RPC. 
Heartbeat can now be per topic and be staggered by having different topics at different times, and also more frequent for some topics.
Less HOL Blocking when sharing IHave, IWant, IDontWant

And mostly the fact that there really isn't any reason that control messages for a topic should interleave with control messages of another topic. 

Similarly What do you think about? 
1 topic control stream as defined above
1 topic data stream **per message**. 
With this design we get nice cancellation properties. So we can cancel in progress sends easily. And if we add message preambles / ids before sending messages we can also reset on the receive side.

pubsub/gossipsub/topic-streams.md
```diff
@@ -0,0 +1,128 @@
...
+
+`TopicRPC` messages MUST NOT be empty. They MUST contain either a partial or
+publish message.
+
+If there are multiple streams for a single topic, the receiver SHOULD process
+them in the order the streams were opened by the initiator. The receiver
+SHOULD limit the number of concurrent topic streams for the same topic to 3 and
+downscore peers that open more. Initiators SHOULD limit the number of
+concurrent topic streams to 1 per topic. The initiator MUST close the old
+stream before writing on a new stream for a given topic.
```
sukunrt: it's a bit confusing, why would there be 3 streams? 

Is this because of subscribe, unsubscribe, subscribe, unsubscribe kind of pattern?

MarcoPolo: right, or if the implementation opens a new stream per message. Because there is a potential for packets to be reordered or delayed, the sender should avoid concurrent streams that could appear to the receiver as multiple concurrent streams open at the same time.

pubsub/gossipsub/topic-streams.md
MarcoPolo: 6b526f0c7a99

pubsub/gossipsub/topic-streams.md
MarcoPolo: Discussed synchronously. Resolved that theoretically 1 topic control stream would be better, the actual utility doesn't justify its implementation complexity. Decided to leave as is.

pubsub/gossipsub/topic-streams.md
```diff
@@ -0,0 +1,128 @@
...
+the same time or earlier than a peer's subscriptions. Since a publisher should
+not publish messages to a peer before learning of its subscriptions, there is
+no window when a publisher wishes to publish a message, but does not know if
+the peer supports Topic Streams.
+
+## Aborting the Connection on Protocol Violation
+
+If a receiver receives a message from a peer that violates a MUST condition,
+the receiver MUST reset the connection to the peer and send error code
+`0xd52505` when the transport allows it. This code signifies a Topic Streams
```
gmelodie, line 53: the message violates the condition, no? not the peer
```suggestion
If a peer receives a message that violates a MUST condition,
the receiver MUST reset the connection to the peer and send error code
`0xd52505` when the transport allows it. This code signifies a Topic Streams
```

pubsub/gossipsub/topic-streams.md
```diff
@@ -0,0 +1,128 @@
...
+Protocol Violation error, and is derived from the first 3 bytes of the sha256
+hashsum of the string `gossipsub-topic-streams`. (i.e. `echo -n
+"gossipsub-topic-streams" | sha256sum | head -c 6`)
```
gmelodie, line 56: Is this usual for specs to do? Otherwise, I see no point to explain where it comes from (maybe for curiosity in an appendix?

pubsub/gossipsub/topic-streams.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Topic Streams
+
+A peer opens a bidirectional stream for each topic that it wishes to send
```
gmelodie, line 60:
```suggestion
A peer MUST open a bidirectional stream for each topic that it wishes to send
```

pubsub/gossipsub/topic-streams.md
```diff
@@ -0,0 +1,128 @@
...
+application messages for. Despite being bidirectional streams, they are treated
+as unidirectional streams. If both sides wish to publish messages for a given
+topic, both sides MUST open a bidirectional stream.
+
+The responder of the bidirectional stream MUST NOT write on the stream after
+protocol negotiation completes.
```
gmelodie, line 66: So it is only bidirectional to accomodate the initial handshake? If so this should be stated explicitly

pubsub/gossipsub/topic-streams.md
```diff
@@ -0,0 +1,128 @@
...
+If the receiver does not receive a topic header within a reasonable timeout, it
+SHOULD reset the stream. One second is a reasonable timeout for most
+applications.
+
+`TopicRPC` messages MUST NOT be empty. They MUST contain either a partial or
+publish message. The data length of the application message MUST be non zero.
+
+If there are multiple streams for a single topic, the receiver SHOULD process
+them in the order the streams were opened by the initiator. The receiver
+SHOULD limit the number of concurrent topic streams for the same topic to 3 and
```
gmelodie: I wonder if it'd be better to use variables here for this, as well as the stream reset timeout or other constants, and then add a table below with recommended default values for these.

pubsub/gossipsub/topic-streams.md
```diff
@@ -0,0 +1,128 @@
...
+downscore peers that open more. Initiators SHOULD limit the number of
+concurrent topic streams to 1 per topic. The initiator MUST close the old
```
gmelodie: > Initiators SHOULD limit the number of concurrent topic streams to 1 per topic

Should we include an explanation about what to do when there are multiple concurrent topic streams for a single topic? Because I can't see how that would work exactly. If it makes no sense whatsoever, then maybe we can change that SHOULD to a MUST?

pubsub/gossipsub/topic-streams.md
```diff
@@ -0,0 +1,128 @@
...
+stream before writing on a new stream for a given topic.
+
+If the receiver receives a topic stream for a topic it is not subscribed to and
+has not recently published partial messages to (via fanout), it SHOULD
+downscore the peer. The receiver MUST NOT downscore a peer for opening a topic
```
gmelodie, line 105: > it SHOULD downscore the peer

Is this true for other spec violations here as well? Maybe we should include this sentence in other places as well

pubsub/gossipsub/topic-streams.md
```diff
@@ -0,0 +1,128 @@
...
+stream for a topic the receiver recently unsubscribed from, as the peer may not
+have received the unsubscribe message before opening the topic stream.
```
gmelodie, line 107: nit
```suggestion
downscore the peer. 

**Note:** The receiver MUST NOT downscore a peer for opening a topic
stream for a topic the receiver recently unsubscribed from, as the peer may not
have received the unsubscribe message before opening the topic stream.
```

pubsub/gossipsub/topic-streams.md
```diff
@@ -0,0 +1,128 @@
...
+closed when either the peer unsubscribes from the topic, or the publisher will
+no longer publish on the topic. Either side may close the stream.
+
+Implementations MAY choose to keep topic streams open only for mesh and fanout
+peers and use a short-lived stream when responding to `IWANTs`.
+
+### Topic Scoped Messages
+
+When a peer wishes to publish a message, it MUST publish a `TopicScopedMessage`
+and it MUST NOT publish a message on the control stream.
```
gmelodie: This got me thinking... how does a peer know that a message is not a control message? Apologies if I'm missing important gossipsub-related information about this.

pubsub/gossipsub/topic-streams.md
```diff
@@ -0,0 +1,128 @@
...
+implementations MUST reconstruct the full `Message` by setting `topic` to the
+value from the stream's `TopicRPCHeader`, and verify the signature against
+that. Implementations SHOULD also populate the `topic` before delivering the
+message to the application.
+
+### Partial Messages
+
+If the partial message extension has been negotiated with this extension, peers
+MUST send each other Partial Messages on the topic stream, not the control
+stream.
```
gmelodie: nit
```suggestion
If the partial message extension has been negotiated with this extension, peers
MUST send each other Partial Messages on the topic stream.
Peers MUST NOT send Partial Messages on the control stream.
```

pubsub/gossipsub/topic-streams.md
```diff
@@ -0,0 +1,128 @@
...
+Upon opening the topic stream, the initiator MUST send a single length-prefixed
+`TopicRPCHeader` protobuf message. The `TopicRPCHeader` MUST contain a topic. The receiver uses this to identify the topic associated with the stream.
+Afterwards, only length-prefixed `TopicRPC` protobuf messages are allowed.
+
+If the receiver does not receive a topic header within a reasonable timeout, it
+SHOULD reset the stream. One second is a reasonable timeout for most
+applications.
+
+`TopicRPC` messages MUST NOT be empty. They MUST contain either a partial or
+publish message. The data length of the application message MUST be non zero.
```
richard-ramos, line 94: > The data length of the application message MUST be non zero

Is this restriction mandatory? because pubsub does allow empty `data`.
https://github.com/libp2p/specs/blob/master/pubsub/README.md#the-message

pubsub/gossipsub/topic-streams.md
```diff
@@ -0,0 +1,128 @@
...
+## Negotiation
+
+Peers MUST use this extension with each other when both sides advertise support
+for this extension in the `ControlExtensions` message.
+
+The `ControlExtensions` message is the first message a peer sends and comes at
+the same time or earlier than a peer's subscriptions. Since a publisher should
+not publish messages to a peer before learning of its subscriptions, there is
+no window when a publisher wishes to publish a message, but does not know if
+the peer supports Topic Streams.
```
richard-ramos, line 38: What to do in case of races?: Assuming you have to peers (A and B) the following can happen (specially in quic):

1. `A` and `B` write `topicStreams=true` on their gossip streams
2. `A` receives `B`'s extension advertisement so `A` knows that both sides support the extensions
3. `A` opens `/gsts/v0beta`
4. But, because the streams are muxed independently, `B` can receive the topic stream before it read `A` extension advertisement,
5. Then in that moment `B` knows it support topic streams, however it still does not know that `A` supports it.

Should A open the stream ONLY after receiving the extension advertisement from B? or should `B` assume `A` opening a stream for `/gsts/v0beta` is proof enough that `A` supports topic streams?

pubsub/gossipsub/topic-streams.md
```diff
@@ -0,0 +1,128 @@
...
+has not recently published partial messages to (via fanout), it SHOULD
+downscore the peer. The receiver MUST NOT downscore a peer for opening a topic
+stream for a topic the receiver recently unsubscribed from, as the peer may not
+have received the unsubscribe message before opening the topic stream.
+
+### Lifecycle
+
+A topic stream is created when a node publishes a topic message to a peer. It is
+closed when either the peer unsubscribes from the topic, or the publisher will
+no longer publish on the topic. Either side may close the stream.
```
richard-ramos, line 113:
```suggestion
A topic stream is created when a node publishes a topic message to a peer. It is 
closed when the peer unsubscribes or the sender no longer intends to send 
messages on that topic. Either side may close the stream.
```

pubsub/gossipsub/extensions/extensions.proto
```diff
@@ -59,3 +60,27 @@ message PartialMessagesExtension {
...
+}
+
+// TopicScopedMessage is identical to Message, except the topic field is unused
+message TopicScopedMessage {
+	optional string from = 1;
+	optional bytes data = 2;
+	optional bytes seqno = 3;
+
+  // Included for computing signatures, not used on the wire
+  optional string unset_topic_name = 4;
```
richard-ramos: Should this use
https://protobuf.dev/programming-guides/proto3/#reserved ?

```suggestion
  // Included for computing signatures, not used on the wire
  reserved 4;
```

pubsub/gossipsub/topic-streams.md
```diff
@@ -0,0 +1,128 @@
...
+SHOULD reset the stream. One second is a reasonable timeout for most
+applications.
+
+`TopicRPC` messages MUST NOT be empty. They MUST contain either a partial or
+publish message. The data length of the application message MUST be non zero.
+
+If there are multiple streams for a single topic, the receiver SHOULD process
+them in the order the streams were opened by the initiator. The receiver
+SHOULD limit the number of concurrent topic streams for the same topic to 3 and
+downscore peers that open more. Initiators SHOULD limit the number of
```
richard-ramos: What happens with streams that are opened after the limit is reached? are they reset?

pubsub/gossipsub/topic-streams.md
MarcoPolo: Historically no, but that's partly because we didn't have great support for error codes. Now we have more support for error codes (at least in QUIC and Yamux transports), and it is helpful for debugging to know why a peer disconnected from you. Did you do something wrong? Are they saturated? Or are they going offline?

pubsub/gossipsub/topic-streams.md
MarcoPolo: The bidirectional nature is to be consistent with how Gossipsub uses streams. Unidirectional streams are only supported with QUIC. 

Also, as you point out, bidirectional streams are needed for the multistream select flow. but nothing else in this protocol requires it.

Any suggested text? Happy to add it!

pubsub/gossipsub/topic-streams.md
MarcoPolo: Because of packet reordering, the receiver cannot know if the sender purposefully or accidentally had two or more concurrent topic streams for a given topic. If we make this a MUST, implementations may incorrectly mark a peer as misbehaving just because of packet reordering.

The idea here is that senders should not have more than one concurrent topic stream, this should limit, in practice, the number of concurrent streams a receiver could view.

pubsub/gossipsub/topic-streams.md
MarcoPolo: Yes, I think it would be a good addition. Please open a separate PR and tag me for review!

pubsub/gossipsub/topic-streams.md
MarcoPolo: This is defined in the `Control Stream` section.

> When this extension is negotiated, the original gossipsub stream becomes the
control stream. Application messages (such as the `Message` or
`PartialMessagesExtension` messages) MUST NOT be published on the control
stream.

It could certainly be more clear though. I'll update

pubsub/gossipsub/topic-streams.md
MarcoPolo: It is when this extension is negotiated. This is intentional as it is likely incorrect or malicious behavior to have an empty data field.

pubsub/gossipsub/topic-streams.md
MarcoPolo: Added a recommendation to reset.

pubsub/gossipsub/topic-streams.md
MarcoPolo: If B receives an inbound topic stream from A and B supports topic streams, it should assume A supports topic streams. If B later receives the extension message that A doesn't support topic streams it should close the connection with a protocol violation. 

Thanks for the callout, I'll update the text.

pubsub/gossipsub/extensions/extensions.proto
MarcoPolo: Good point. I think it makes sense for this to be reserved in the specs protobuf. Implementations may still want to use the field to make it easier to compute the signature of a protobuf object without copying the whole message, but that's an implementation decision. I'll leave a comment to that effect.

pubsub/gossipsub/topic-streams.md
gmelodie: Maybe:
> The responder of the bidirectional stream MUST NOT write on the stream after protocol negotiation completes. The stream is only created bidirectional for the initial handshake.

pubsub/gossipsub/topic-streams.md
gmelodie: I see! This is tricky to word indeed. How about:

> The initiator MUST close the existing topic stream for a topic before it opens a new one. Because the transport can reorder packets, the receiver can get a new stream before the close of the old one. This is not a violation. The receiver MUST accept up to 3 open streams for a topic and SHOULD process them in the sequence opened. If there are more than 3, the receiver SHOULD reset the excess streams and downscore the peer.

gmelodie (approved):

richard-ramos (approved):

PR #730, opened 2026-07-08
tls: Add ECH

MarcoPolo:
This document specifies minor changes to the libp2p TLS Handshake to enable support for RFC 9849: TLS Encrypted Client Hello. The primary benefit is to make identifying libp2p connections harder to passive network observers by hiding the “libp2p” ALPN in the encrypted ClientHelloInner.

Related and prereq PRs:
- https://github.com/multiformats/multiaddr/pull/183
- https://github.com/multiformats/multicodec/pull/407

### DIFF
--- a/tls/ECH.md
+++ b/tls/ECH.md
@@ -1,0 +1,74 @@
+# ECH
+
+| Lifecycle Stage | Maturity      | Status | Latest Revision |
+| --------------- | ------------- | ------ | --------------- |
+| 1A              | Working Draft | Active | r0, 2026-07-01  |
+
+Authors: [@marcopolo]
+
+Interest Group: [@sukunrt], `todo`
+
+[@marcopolo]: https://github.com/marcopolo
+[@sukunrt]: https://github.com/sukunrt
+
+## Overview
+
+This document specifies minor changes to the [libp2p TLS Handshake](./tls.md) to
+enable support for [RFC 9849]: TLS Encrypted Client Hello. The primary benefit
+is to make identifying libp2p connections harder to passive network observers by
+hiding the "libp2p" ALPN in the encrypted ClientHelloInner.
+
+## Changes to the libp2p TLS Handshake
+
+### Outer Handshake
+
+The outer handshake MUST NOT be a libp2p TLS handshake.
+
+#### ALPN
+
+Unless an ALPN is explicitly specified by the application for the connection,
+client implementations SHOULD set it to `http/1.1` for TCP connections and `h3`
+for QUIC connections. The ALPN MUST NOT include `libp2p` or the muxer protocol
+IDs used by [inlined muxer negotiation].
+
+#### Authenticated Rejection
+
+In RFC 9849, clients can only use the server provided `retry_configs` if the
+outer handshake authenticates successfully with the given
+ECHConfig.contents.public_name. This means that if servers wish to support the
+`retry_configs` fallback they MUST use a valid domain name and hold the
+corresponding Server Certificate. This is the retry mechanism in RFC 9849;
+there is nothing libp2p specific about this.
+
+If the server does not have a valid public_name and certificate, the client can
+only fail the connection.
+
+### Inner Handshake
+
+The encrypted ClientHelloInner performs the standard libp2p TLS Handshake. The
+ALPN MUST conform to the libp2p TLS Handshake.
+
+## Server Key Rotation
+
+Servers SHOULD rotate their keys once a month, and keep the prior ECH Config
+keys around for 1 week to assist stale clients.
+
+## Client ECH Config Caching
+
+Clients SHOULD cache the ECHConfigList for no more than 48 hours. Note that a
+server can provide an updated ECHConfigList upon connection via an identify
+message.
+
+## `/ech` in the multiaddr
+
+The `/ech` component SHOULD appear after the `/tls` or `/quic-v1` component.
+Servers use this multiaddr to advertise ECH support and their ECHConfigList.
+
+## Caller Provided ECHConfigList
+
+Implementations SHOULD allow users to supply an ECHConfigList when dialing.
+
+<!-- References -->
+
+[RFC 9849]: https://www.rfc-editor.org/info/rfc9849/
+[inlined muxer negotiation]: ../connections/inlined-muxer-negotiation.md


tls/ECH.md
```diff
@@ -0,0 +1,74 @@
...
+IDs used by [inlined muxer negotiation].
+
+#### Authenticated Rejection
+
+In RFC 9849, clients can only use the server provided `retry_configs` if the
+outer handshake authenticates successfully with the given
+ECHConfig.contents.public_name. This means that if servers wish to support the
+`retry_configs` fallback they MUST use a valid domain name and hold the
+corresponding Server Certificate. This is the retry mechanism in RFC 9849;
+there is nothing libp2p specific about this.
```
gmelodie, line 41:
```suggestion
Servers MUST use a valid domain name and hold the corresponding Server Certificate to support the `retry_configs` fallback. This is the retry mechanism in RFC 9849; there is nothing libp2p specific about this.
```

tls/ECH.md
```diff
@@ -0,0 +1,74 @@
...
+
+If the server does not have a valid public_name and certificate, the client can
+only fail the connection.
```
gmelodie, line 44:
```suggestion
The client MUST fail the connection If the server does not have a valid public_name and certificate pair.
```

tls/ECH.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Inner Handshake
+
+The encrypted ClientHelloInner performs the standard libp2p TLS Handshake. The
+ALPN MUST conform to the libp2p TLS Handshake.
+
+## Server Key Rotation
+
+Servers SHOULD rotate their keys once a month, and keep the prior ECH Config
+keys around for 1 week to assist stale clients.
```
gmelodie, line 54: Maybe add the reasoning behind the time windows here? (i.e. once a month, 1 week)

tls/ECH.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Client ECH Config Caching
+
+Clients SHOULD cache the ECHConfigList for no more than 48 hours. Note that a
```
gmelodie, line 58: Reasoning on the 48h?

tls/ECH.md
```diff
@@ -0,0 +1,74 @@
...
+only fail the connection.
+
+### Inner Handshake
+
+The encrypted ClientHelloInner performs the standard libp2p TLS Handshake. The
+ALPN MUST conform to the libp2p TLS Handshake.
+
+## Server Key Rotation
+
+Servers SHOULD rotate their keys once a month, and keep the prior ECH Config
```
richard-ramos, line 53:
```suggestion
Servers SHOULD rotate their ECH key pairs once a month, and keep the prior ECH Config
```

tls/ECH.md
```diff
@@ -0,0 +1,74 @@
...
+
+Interest Group: [@sukunrt], `todo`
+
+[@marcopolo]: https://github.com/marcopolo
+[@sukunrt]: https://github.com/sukunrt
+
+## Overview
+
+This document specifies minor changes to the [libp2p TLS Handshake](./tls.md) to
+enable support for [RFC 9849]: TLS Encrypted Client Hello. The primary benefit
```
richard-ramos, line 17:
```suggestion
enable support for [RFC 9849]: TLS Encrypted ClientHello. The primary benefit
```

tls/ECH.md
```diff
@@ -0,0 +1,74 @@
...
+#### ALPN
+
+Unless an ALPN is explicitly specified by the application for the connection,
+client implementations SHOULD set it to `http/1.1` for TCP connections and `h3`
+for QUIC connections. The ALPN MUST NOT include `libp2p` or the muxer protocol
+IDs used by [inlined muxer negotiation].
+
+#### Authenticated Rejection
+
+In RFC 9849, clients can only use the server provided `retry_configs` if the
```
richard-ramos, line 36:
```suggestion
In RFC 9849, clients can only use the server-provided `retry_configs` if the
```

tls/ECH.md
```diff
@@ -0,0 +1,74 @@
...
+Clients SHOULD cache the ECHConfigList for no more than 48 hours. Note that a
+server can provide an updated ECHConfigList upon connection via an identify
+message.
+
+## `/ech` in the multiaddr
+
+The `/ech` component SHOULD appear after the `/tls` or `/quic-v1` component.
+Servers use this multiaddr to advertise ECH support and their ECHConfigList.
+
+## Caller Provided ECHConfigList
```
richard-ramos, line 67:
```suggestion
## Caller-Provided ECHConfigList
```

tls/ECH.md
```diff
@@ -0,0 +1,74 @@
...
+Authors: [@marcopolo]
+
+Interest Group: [@sukunrt], `todo`
+
+[@marcopolo]: https://github.com/marcopolo
+[@sukunrt]: https://github.com/sukunrt
+
+## Overview
+
+This document specifies minor changes to the [libp2p TLS Handshake](./tls.md) to
```
richard-ramos, line 16: For TCP, how would this work? because with TCP the flow is:

1 - TCP connection
2 - Multistream negotiation in plaintext sending `/multistream/1.0.0` and `/tls/1.0.0`
3 - Then the TLS handshake happen

meaning that an observer should be able to recognize libp2p being used. Would using `/tls/ech` mean that we gotta start TLS handshake directly, and skip the multistream select negotiation? if so, probably worth mentioning in the spec

lidel: +1. The multiaddr has the same gap: in today's specs `/tls` under `/tcp` only appears as `/tls/ws` or `/tls/http`, which run WebPKI TLS rather than the libp2p handshake, so `/tcp/<port>/tls/ech` has no defined meaning. Scoping this spec to QUIC until a TLS-first TCP mode exists may be simplest, but also defer solving problem. Better to, at very least, clearly state it in the spec.

tls/ECH.md
lidel: +1

```suggestion
Servers SHOULD rotate their ECH key pairs once a month, and keep the prior ECH Config
keys around for 1 week to assist stale clients.

Servers MUST persist ECH private keys across restarts for as long as a matching
ECHConfig may still be advertised or held by clients: an `/ech` value has no TTL and ends up where multiaddrs live for months:
bootstrap lists, static peering configs, delegated routing caches, and
caller-provided addresses. Without a certificate for `public_name` there is no
`retry_configs`, so dials with a config for a dropped key fail until the client
learns a fresh `/ech` address.
```

tls/ECH.md
```diff
@@ -0,0 +1,74 @@
...
+### Outer Handshake
+
+The outer handshake MUST NOT be a libp2p TLS handshake.
+
+#### ALPN
+
+Unless an ALPN is explicitly specified by the application for the connection,
+client implementations SHOULD set it to `http/1.1` for TCP connections and `h3`
+for QUIC connections. The ALPN MUST NOT include `libp2p` or the muxer protocol
+IDs used by [inlined muxer negotiation].
```
lidel: @MarcoPolo  FYSA stock Go `crypto/tls` (and quic-go on top of it) builds ClientHelloOuter by cloning the inner hello and replaces only SNI and random, so `libp2p` and the muxer IDs still go out in cleartext while ECH reports success (golang/go#71220). 

This is not "golang bad" type of thing, iiuc rustls has the same single ALPN list 🙃. So.. without a rule, the straightforward implementation ships a silent no-op 🙈 

Spec should clearly state this known and common problem:

```suggestion
for QUIC connections. The ALPN MUST NOT include `libp2p` or the muxer protocol
IDs used by [inlined muxer negotiation].

Some TLS stacks build ClientHelloOuter as a copy of ClientHelloInner, including
its ALPN. On such a stack the `libp2p` ALPN stays in cleartext even when ECH is
accepted. Implementations MUST NOT offer ECH unless their TLS stack can set the
ClientHelloOuter ALPN independently of the ClientHelloInner ALPN.
```

tls/ECH.md
lidel: Strong +1 to the MUST here.

Failing this connection does not end the dial, though: go-libp2p moves on to the peer's other addresses and starts TCP dials some ms after QUIC unless QUIC already connected. 

:point_right: [RFC 9848, section 5.1](https://www.rfc-editor.org/rfc/rfc9848.html#section-5.1) forbids this fallback for DNS-advertised ECH because it "would negate the privacy benefits of ECH". 

Building on the suggestion above:

```suggestion
If the server does not have a valid public_name and certificate, the client MUST
fail the connection.

Failing the connection does not end the dial: libp2p dialers try a peer's other
addresses, often in parallel. A non-ECH dial sends the `libp2p` ALPN in
cleartext, so an on-path observer only has to drop ECH handshakes to see it. A
client that holds an `/ech` address for a peer MUST NOT dial that peer's non-ECH
addresses, in parallel or after an ECH failure.
```

tls/ECH.md
lidel: +1, and it needs a rule on `public_name`: it goes out as cleartext SNI on every ECH connection ([RFC 9849, section 6.1](https://www.rfc-editor.org/rfc/rfc9849.html#section-6.1), step 5). Reusing a `/tls/ws` cert whose name embeds the peer ID or a libp2p-specific domain would leak what ECH is meant to hide.

List caveats explicitly: 

```suggestion
Servers MUST use a valid domain name and hold the corresponding Server Certificate
to support the `retry_configs` fallback. This is the retry mechanism in RFC 9849;
there is nothing libp2p specific about this.

`public_name` is sent in cleartext on every ECH connection. It MUST NOT contain
the peer ID or a libp2p-specific domain. It MUST be a host name with at least
two labels, even without a certificate, as some TLS stacks reject single-label
names.
```

tls/ECH.md
```diff
@@ -0,0 +1,74 @@
...
+
+#### Authenticated Rejection
+
+In RFC 9849, clients can only use the server provided `retry_configs` if the
+outer handshake authenticates successfully with the given
+ECHConfig.contents.public_name. This means that if servers wish to support the
+`retry_configs` fallback they MUST use a valid domain name and hold the
+corresponding Server Certificate. This is the retry mechanism in RFC 9849;
+there is nothing libp2p specific about this.
+
```
lidel: hm.. not sure if i got this right, but holding the cert here feels to be not enough: after rejecting ECH the server continues with ClientHelloOuter, and a server that keeps the [libp2p TLS ALPN rule](https://github.com/libp2p/specs/blob/78e75c6f4e5dd9de4bc1524c686c0f45df4b461a/tls/tls.md#alpn) aborts with `no_application_protocol` before `retry_configs` are sent.

Should below missing steps be added here?

```suggestion

On ECH rejection, the server MUST complete the outer handshake without the libp2p
TLS ALPN rules: accept the ClientHelloOuter ALPN, present the `public_name`
certificate, send `retry_configs`, then close the connection.

```

lidel (changes requested):
Thank you for kicking this off, it is worth doing. Honestly though, given where ipfs/libp2p maintenance is heading, and the TLS stacks are today, getting it right will take a while, so worth setting expectations low and focus on specs being realistic.

FWIW:

- **(Bad)** Sadly, in its current form this is a no-op in Go and Rust 🙈 IIUC stock Go `crypto/tls` (also used by quic-go) and rustls send the same ALPN list in the inner and outer ClientHello, so the `libp2p` ALPN still goes out in cleartext while ECH reports success ([details](https://github.com/libp2p/specs/pull/730#discussion_r4084058831)). Users would think `libp2p` is hidden when it is not ( :exclamation: )
  - Upstream: 
    - golang/go#71220 tracks the outer ALPN
    - rustls has no issue for it yet but also lacks server-side ECH it seems? (rustls/rustls#1980).
- **(Good)** Once TLS stacks allow a separate outer ALPN, ECH does beat deep packet inspection matching on the `libp2p` string.

See the inline comments, they mostly narrow what the spec promises and add rules where a quick LLM-assisted implementation would leak or break.

I'd suggest parking this spec until Go and Rust implementations exist and the upstream gaps above are closed, so it describes something real rather than security theater :)


ps. two caveats worth writing down for drive-by reader in case someone thinks ECH is a silver bullet for "privacy in libp2p":

- a censor that drops ECH handshakes gets a plain `libp2p` handshake via the peer's other addresses, unless the spec forbids that fallback ([suggestion](https://github.com/libp2p/specs/pull/730#discussion_r4084080110))
- peers announced via the DHT can be labeled by destination alone by anyone with a DHT crawl (prior art: https://github.com/dennis-tra/nebula); ECH cannot fix that, but a short section noting it would set honest expectations, e.g. that ECH inspection-resistance truly works only when peer addresses are not publicly discoverable

PR #732, opened 2026-08-05
waggle: initial version

jxs:
first draft to start the conversation, info is in the spec.


pubsub/waggle.md
```diff
@@ -0,0 +1,116 @@
...
+To unsubscribe from a topic, a peer closes the corresponding stream, indicating that it no longer wishes to participate in dissemination for that topic. Upon closure, connected peers update their subscription state by removing the association between that peer and the topic.
+The protocol does not prescribe how topics are created, named, or authorized. Topic management and access control are application-defined.
+
+## Publishing and dissemination
+
+ Objects are disseminated through incremental publication of their constituent pieces. A publisher may divide an object into multiple pieces and publish those pieces independently, including through repeated publication attempts. Pieces do not need to originate from a single publisher; any peer possessing a piece may participate in its propagation. Each published piece belongs to a topic and MUST be disseminated through the corresponding topic stream.
+
+To publish a piece, a node selects a random subset of its connected peers and transmits it to them. Upon receiving a previously unknown piece, a peer updates its local object view and begins propagating the piece through its own connected peers.
```
dknopik: I think this should be "subset of peers subscribed to the piece's topic". I'd also not explicitly require that this subset is random, the application should choose the subset. Of course, the application may choose that subset randomly.

pubsub/waggle.md
```diff
@@ -0,0 +1,116 @@
...
+
+The receiving peer uses the metadata it maintains about remote peers to identify peers that are known to be missing the received piece. The peer then transmits the piece to those peers, allowing dissemination to leverage existing knowledge of incomplete views. In addition, the peer advertises its updated metadata to peers that maintain a view of its state, allowing them to track the newly available piece.
+
+To discover additional dissemination opportunities, a peer also selects a random subset of connected peers for which it has no current metadata view and advertises its current object metadata to them. These exchanges allow peers to progressively discover missing pieces and establish additional synchronization relationships throughout the network.
```
dknopik: again, subset of subscribed peers.

Also, I am unsure if this should be codified here. At the very least, there should be a MAY somewhere in here IMO.

pubsub/waggle.md
```diff
@@ -0,0 +1,116 @@
...
+
+Topic identifiers, object identifiers, and piece metadata are opaque protocol fields whose semantics are defined by applications. The protocol does not prescribe how these fields are generated, assigned, encoded, or interpreted.
+
+Applications SHOULD use stable identifiers that allow peers to consistently refer to the same topic or object across independent connections and dissemination attempts. Applications MAY use content-derived identifiers, such as cryptographic hashes, when content addressing is appropriate.
+
+Applications are responsible for defining the representation and interpretation of piece metadata, including how pieces are tracked and how objects are reconstructed.
+
+## Object Cache
+
+The lifetime of cached object and piece state SHOULD be configurable by implementations and tunable by applications according to their resource constraints and dissemination requirements.
```
dknopik, line 77: There should not be a waggle maintained object cache. Waggle can not by itself utilize such a cache, as all fields are opaque, so the caching should just be responsibility of the application,

pubsub/waggle.md
```diff
@@ -0,0 +1,116 @@
...
+
+The protocol uses Protocol Buffers (protobuf) as its wire format. Messages are encoded using the following schema:
+
+```protobuf
+message Message {
+  oneof payload {
+    TopicSubscription subscription = 1;
+    ObjectPieces publication = 2;
+  }
+}
```
dknopik: As `TopicSubscription` is only sent exactly once, I'd omit this message definition

jxs: thanks Daniel, meanwhile updated the protobuf schema according to what we spoke

pubsub/waggle.md
jxs: the fields being opaque doesn't mean you can't store them as bytes as we do with partial messages cache, which compares and validate against new received pieces

pubsub/waggle.md
jxs: that's a great insight, agree and updated, thanks Daniel

pubsub/waggle.md
jxs: updated to remove the random reference. Why MAY? this is part of the dissemination, the protocol should guarantee that applications are able to in the limit flood the network so that a full dissemination is achieved.
If applications set the gossip factor to 0 this step is skipped

pubsub/waggle.md
dknopik: such a general purpose cache might be very inefficient, depending on the object being disseminated. I'd just let the application manage it.

In the end, maybe this boils down to being an API question: Either a highly configurable and modifiable cache that lives in the waggle implementation, or letting this be the responsibility of the application. I lean towards the latter.

cortze: From the application's perspective, I think you can always set a TTL or a timestamp deadline for each of the message types or topics.

If this is configurable per topic, I don't see the problem of the cache being directly at waggle.

Nashatyrev:
To build some intuition for the proposal, could we walk through the degenerate case of an object consisting of a single piece?

As I read it, after the initial publication fanout, a peer does not forward the piece to another peer until it has metadata indicating that peer is missing it. Instead, it first disseminates its current object metadata in a FloodSub-like manner to progressively establish those views. Once such metadata exists, the piece can be selectively pushed to peers known to be missing it.
Is that the intended model?

pubsub/waggle.md
```diff
@@ -0,0 +1,105 @@
...
+Peer scoring is an application-defined concern. The broadcast protocol should surface relevant information to the application and let it decide how to score and penalize peers, rather than baking scoring semantics into the network layer.
+There is limited utility in maintaining a stable mesh through GRAFT and PRUNE messages. Publishing to a subset of connected peers potentially selected using latency or other application-defined criteria provides the same robustness and coverage with significantly reduced implementation complexity.
+
+## Topic Membership
+
+Topics define independent dissemination scopes within the network. To subscribe to a topic, a peer opens a new dedicated outbound stream with its connected peers and MUST include the topic it wishes to participate in. Each topic is carried on its own dedicated stream. A peer subscribing to multiple topics opens a separate stream for each topic.
```
sukunrt, line 50: How about an outbound stream for each side of the connection. 

Say peers A and B want to subscribe to topic X, peer A will open a outbound waggle stream for topic X and peer B will open an outbound waggle stream for topic X. The first stream would be read only for peer B and the second one would be read only for peer A. This would avoid simultaneous connect problems.

PR #735, opened 2026-08-24
gossipsub: downscore peers on protocol violations

gmelodie:
Follow-up to https://github.com/libp2p/specs/pull/729#discussion_r3775156137, where @MarcoPolo asked for a separate PR.

The Topic Streams Extension says a receiver SHOULD downscore a peer for some spec violations. Other gossipsub documents state MUST and MUST NOT conditions without saying what a receiver does when a peer breaks them. This PR states the consequence once, in the extensions framework document, and applies it where an extension needs a specific rule.

`gossipsub-v1.3.md` gets a new "Peer Scoring on Protocol Violations" section:

- A peer that sends a message which violates a MUST or MUST NOT condition SHOULD be downscored through the `P₇` behavioural penalty from gossipsub v1.1.
- The rule covers only the conditions that a receiver can check from the messages it receives. It does not cover the process rules of this document, such as the requirement to update `extensions.proto`.
- An extension MAY define a stronger remedy, such as a connection reset. This keeps the rule compatible with #729, which resets the connection and sends an error code.
- A receiver MUST NOT downscore a peer for a message that the peer could have sent before it processed the receiver's latest subscription change. #729 carves out the same race for topic streams. Without it, a node that unsubscribes penalises every honest peer with a message already on the wire.

The Extensions control message rules now state their consequence: a receiver SHOULD downscore a peer that sends the message after the first message on the stream, or sends more than one.

`partial-messages.md` points at the v1.3 section, and states the one rule that needs the extension's own terms: a node that receives a `partialMessage` field for a topic it did not request partial messages on SHOULD downscore the sending peer, unless the peer could have sent it before processing the node's latest `SubOpts` for that topic.

Two documents are left alone on purpose. `gossipsub-v1.1.md` already owns the scoring machinery, and adding a blanket penalty rule to it is a semantic change rather than an editorial one. `extensions/experimental/test-extension.md` exists to test interop, so a penalty there is noise.

### DIFF
--- a/pubsub/gossipsub/gossipsub-v1.2.md
+++ b/pubsub/gossipsub/gossipsub-v1.2.md
@@ -2,7 +2,7 @@
 
 | Lifecycle Stage | Maturity                  | Status | Latest Revision |
 |-----------------|---------------------------|--------|-----------------|
-| 1A              | Working Draft             | Active | r1, 2023-07-14  |
+| 1A              | Working Draft             | Active | r2, 2026-08-30  |
 
 Authors: [@Nashatyrev], [@Menduist]
 
--- a/pubsub/gossipsub/gossipsub-v1.2.md
+++ b/pubsub/gossipsub/gossipsub-v1.2.md
@@ -76,6 +76,10 @@ on per message basis when the size is exceeded or just use `IDONTWANT` for all m
 
 To prevent DoS the number of `IDONTWANT` control messages is limited to `max_idontwant_messages` per heartbeat  
 
+The receiver SHOULD ignore the `IDONTWANT` messages beyond `max_idontwant_messages` within a heartbeat, and SHOULD
+downscore the peer that sent them, but only when `max_idontwant_messages` is agreed across the network. An
+implementation that disagrees with its peers on this value downscores honest peers.
+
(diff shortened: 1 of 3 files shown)


gmelodie:
@MarcoPolo I cannot tag you as a reviewer as you asked, but here you go :)

pubsub/gossipsub/gossipsub-v1.2.md
```diff
@@ -76,6 +76,10 @@ on per message basis when the size is exceeded or just use `IDONTWANT` for all m
 
 To prevent DoS the number of `IDONTWANT` control messages is limited to `max_idontwant_messages` per heartbeat  
 
+The receiver SHOULD ignore the `IDONTWANT` messages beyond `max_idontwant_messages` within a heartbeat, and SHOULD
```
MarcoPolo, line 79: please update the revision number at the top

MarcoPolo (approved):
looks good. Please update the revision number at the top

pubsub/gossipsub/gossipsub-v1.2.md
gmelodie: done!

